PR #67, opened 2016-10-26
Fade grid during transition

trvrb:
I'm quite happy with the result. @colinmegill: could you make sure I'm doing this safely however? I'm storing state within the grid component to keep drawing the old grid for 500ms before swapping in the new grid.


colinmegill:
I'll take a look!

PR #68, opened 2016-10-26
Tree transition

trvrb:
Switch from VictoryAnimation to CSS transitions for animating branches in the tree. This significantly improves animation performance. However, we're forced to use SVG lines for branches rather than SVG paths. Because of this, the radial layout is slightly less pretty.  However, other layouts (Path-O-Gen style and unrooted) should be fine as lines.

@rneher --- see what you think and merge if happy

### DIFF
--- a/src/components/tree/grid.js
+++ b/src/components/tree/grid.js
@@ -1,6 +1,5 @@
 import React from "react";
 import Radium from "radium";
-import {VictoryAnimation} from "victory";
 import GridLine from "./gridLine";
 
 /*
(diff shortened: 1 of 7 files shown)


rneher:
firefox doesn't show any lines...

trvrb:
Thanks @rneher. I'll fix this. Firefox was working just recently. I'll roll back a couple commits and see where I broke it.

trvrb:
Closing this for the moment, while we decide how to proceed with SVG paths vs SVG lines.

PR #69, opened 2016-10-26
Zoom v1

colinmegill:


trvrb:
I would propose doing a simpler set of controls here. More like Google Maps or Dark Sky:

https://darksky.net/47.6344,-122.3422

The default control is pan. This is done by dragging anywhere in the svg, just like when the current pan tool is active. Clicking the plus zooms in a level itself, rather than activating the plus tool. Same for minus.

Just need to be sure that this would play well with branch clicks. I think we can separate mouse click to select branch from drag to pan. Also, I think clicking branches to select nodes will be better than lassoing groups of nodes.

colinmegill:
Yes! Let's do it. I was headed in that direction, but ran into some documentation issues. Submitted an issue last night and thankfully got a quick response here so we're all set to do that now: https://github.com/chrvadala/react-svg-pan-zoom/issues/7

trvrb:
That's great.

colinmegill:
All set.

PR #70, opened 2016-10-26
Strip @Radium from most components

trvrb:
I think we're good here. Visual inspections says things are still working. A few components like those in `framework` and `map` suffered when removing, so I left the `@Radium` in these cases.

@colinmegill: please merge if this works for you.


trvrb:
Closing PR in favor of `pure_js_tree`. We should still go through an make a profiling pass on Radium to improve general React render performance. I'll make an issue.

PR #72, opened 2016-10-28
Better tree transition

trvrb:
Transition tips via CSS. Fade branches via CSS by temporally keeping a stored state of `distanceMeasure`, `layout`, etc...


rneher:
it this doesn't render in firefox, throwing error 
```
TypeError: CSS2Properties doesn't have an indexed property setter for '0'
```

trvrb:
Closing PR in favor of `pure_js_tree`.

PR #106, opened 2017-01-01
Fix legend zoom

### LINKED ISSUE #104: Tree goes to half width/height
Steps to reproduce:

Load up in single column layout, move to two column layout, hover on legend item.

Possible reasons: 

Wrong lifecycle method being used?
PhyloTree function method misfire?

rneher:
this implements a number of fixes:
 - no longer zooms into terminal node
 - colormaps are now specified in augur and exported along with the color options
 - formats legend numbers (could use some tweaking)
 - uses srqt counts for map circle radii
 - fixes the tree redrawing issue that caused the iffy behavior on legend mouse-over (closes #104)
 - adds confidence intervals to the tree

### DIFF
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -80,7 +80,8 @@ class App extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     const tmpQuery = queryString.parse(window.location.search);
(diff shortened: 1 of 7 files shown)


colinmegill:
LGTM

PR #110, opened 2017-01-24
upgraded to webpack 2 and babel 6

jameshadfield:
Ideally someone would check this works on their setup!


trvrb:
I'm going to close this without merging. Would be great to move to webpack v2 at some point, but easier to keep track of what's going on without PRs that are not actually PRs.

jameshadfield:
see #444

PR #148, opened 2017-01-27
Route to app using react router

trvrb:
This doesn't completely address the router issue, but I think this is a good direction. This uses `this.props.location.pathname` passed in from react-router rather than `window.location.pathname` and `this.props.location.query` rather than `queryString.parse(window.location.search)`. I've left `state.location` in App.


trvrb:
@colinmegill --- Can you review this? Nothing is breaking relative to `master`. I'm currently working on top on this branch to manipulate query params with react-router.

trvrb:
@nextstrain/auspice ---

Okay. I think I'm going to shelve this for a while and attend to other things. @colinmegill, could you take a look next week and merge if you deem appropriate? Briefly, this:

1. Cleans up state of params that had only lived in URL as query params. Things like `layout: radial`. I really didn't like how it was difficult to determine what is _truth_ and when params have updated when the only thing possible to do is `window.location.search`. I've tried to make this saner by moving these params to redux store in `state.controls`. Redux is now the central source of truth and the query params in URL location are updated to reflect this truth. On initial app load, [I initialize redux params from react-router's `props.location.query`](https://github.com/nextstrain/auspice/blob/124-react-router/src/components/app.js#L133). When a layout is changed, [I use react-router to update the query params in location and dispatch a Redux action](https://github.com/nextstrain/auspice/blob/124-react-router/src/components/controls/choose-layout.js#L43).

2. I have query params `colorBy`, `layout`, `distanceMeasure`, `dateMin`, `dateMax` moved to this reacter-router / redux pattern. Still need to move filters (like authors), but this can be a separate PR.

3. The dataset selection still needs attending to. This is still using `changeRoute` in App. But nothing is broken.

4. You'll see an immediate benefit here in that we no longer get the slot machine effect when loading http://localhost:4000/zika/. That URL sticks and query params only appear as they are selected. This looks cleaner and has the advantage of making it more obvious to a user that clicking layout=radial has had an effect in the URL.

5. The conflicts in `choose-layout.js` and `choose-metric.js` can be resolved to keep `124-react-router` over `master`.

PR #149, opened 2017-01-27
Remove unused npm packages

trvrb:
I'm pretty sure this is all safe to remove, but thought I'd do a PR anyway.


jameshadfield:
@trvrb check that you're not now using `history` with `react-router`, but I wouldn't think so

trvrb:
I just tested an `npm install` after merge. Everything seems to be working fine. Will merge unless I hear objections.

PR #156, opened 2017-02-02
115 changing dataset

jameshadfield:
This PR is ongoing

- A bunch of code has been removed that was never being used - namely some colors in `globals.js`, all of `colorScales.js`, `colors.js`, `ordinalColormaps.js` and `datasets.yaml`
- Color scales now work for ebola *despite* them not being provided in the metadata JSON. Note that "regions" (ebola) and "districts" (zika) look weird because those traits have not been inferred (by augur) for branches in the tree. So the colors on the branches are "unknown".
- I've changed how colorBy changes are propagated, so that now the router update happens together with the redux state change, hopefully avoiding things getting out of sync
- Confidence Intervals are no longer drawn - let's work on this feature in a separate branch for the time being
- if country (e.g.) colors are supplied in the metadata JSON, "new" countries (i.e. in the tree, not in metadata) are given grayscale colours

**known bugs to fix**
- sometimes reloads don't work correctly, but this will be taken care of when we move things from app state to redux


trvrb:
Sorry I just created the conflict here @jameshadfield. I really like moving the paired `location.query` / redux store updates to the reducer. I'll roll back 08f506942485ad63cd5b89fd7fb8475565812731 to prevent merge conflict.

trvrb:
Okay. Rolled back with fa49bda7583e2441fd204daaa38af7a3e6dd7874. Merge is now clean.

trvrb:
I think the grayscale colors for missing bits of metadata is perfect.

jameshadfield:
Much nicer colors! I've just fixed the case where there were more values than colors in globals (e.g. try zika -> district) by looping back over colors (not ideal). It looks better than it really is, because the legend only holds 26 items

trvrb:
Great! Thanks for catching the edge case. Looping is probably fine. I'm imagining the legend will just show the first 26 items. We're going to run into this soon with Zika as more territory is covered and we have more countries. It seems like the legend could be shrunk if it's populated with more items and still be legible. Maybe fit like 30-40 elements comfortably? But there will be an upper limit.

trvrb:
Looking at behavior now:

* Selecting "ebola" in the "Dataset" dropdown updates tree and entropy, but does not update map
* We lose query params in the URL when switching datasets, but they remain in redux and the tree is consistent. I thinking changing datasets should probably reset all these params. This is because a `colorBy` that exists for Zika might not exist for Ebola and a the date range might be non-overlapping.

trvrb:
Number 1 above is due to behavior of `changeRoute`. I recommend we solve this in a separate PR (issue #161).

PR #162, opened 2017-02-02
Phylo comments v2

trvrb:
Include doc string comments in `phyloTree` via @rneher. Plus a bit of refactoring. This seems to be fulling working except I'm getting the following error:

```
Error: <path> attribute d: Expected number, " M NaN NaN L 258.06…".
```

when loading radial (http://localhost:4000/zika/?l=radial). Other layouts are fine. I'll look but I suspect you might be faster @rneher.


jameshadfield:
@trvrb that error is not new

trvrb:
@jameshadfield: Ah. Thanks for the heads up. Okay in this case, can you take a quick look more generally? I think this can be merged.

jameshadfield:
looks good to me

PR #164, opened 2017-02-02
Fix color by genotype after entropy click

trvrb:
This is basically the same fix as pushed by @rneher in c660843e4fefe0cca01987f9fe727a469eac3227, but uses the new `changeColorBy` action. It also improves behavior in `color-by.js` to populate text box with genotype numbers.


trvrb:
I think this is safe to merge, but let me know if you see any issues @jameshadfield.

PR #170, opened 2017-02-06
Fix zoom for terminal nodes

trvrb:


trvrb:
Looks great. Thanks @rneher.

Issue #158, opened 2017-02-02
Move colorScale to redux


trvrb:
Currently, [`colorScale` is being recalculated on each App render](https://github.com/nextstrain/auspice/blob/master/src/components/app.js#L358) and then being passed down to subcomponents that refer to `colorScale`. This should get moved out of App and into redux. There should be a global `store.controls.colorScale` calling the action `changeColorBy` should update this shared `colorScale`. This is a bit complicated by the fact that setting the coloring scale requires knowledge of `store.tree` and `store.sequences`, but it looks like there are patterns for this:

https://thesabbir.com/redux-with-example-share-state-between-two-reducers/

with a call to `getState`.

Issue #161, opened 2017-02-02
Move dataset path to redux / react-router


trvrb:
Currently logic of dataset path (ie `/flu/h3n2/3y` vs `/zika/`) is handled through direct manipulation of `window` via [`changeRoute` in App](https://github.com/nextstrain/auspice/blob/master/src/components/app.js#L293). This should be moved to redux / react-router. We'll have stored state in redux and use `router.push` to update URL.

I'd propose to put this in `store.controls` alongside `colorBy`, `layout`, etc..., where `store.controls.dataset` still follows the scheme set up by @rneher in [`parseParams`](https://github.com/nextstrain/auspice/blob/master/src/util/parseParams.js).

jameshadfield:
Keeping this open for now, but moving to backlog, as we aim for launch.

For launch, RR will handle some paths, but `parseParams` will still be employed for the virus dataset logic.

trvrb:
We've just decided that the current setup with react-router loading App (https://github.com/nextstrain/auspice/blob/master/src/index.js#L41) and then App querying the URL (https://github.com/nextstrain/auspice/blob/master/src/components/app.js#L77) is the right way to go.

Issue #153, opened 2017-01-28
issues with changing dataset


jameshadfield:
All issues with changing from zika -> ebola (etc)

Issue #144, opened 2017-01-26
dead code removal


jameshadfield:
Currently (`0499371`), components such as react-router are being included in the bundle (as they are imported in the code, despite not being used). As we move to webpack2 these should be able to be removed automagically, and we can test this by comment out the include statement and comparing bundle size

At the moment, simply commenting out react-router reduces it from 1.64 to 1.58Mb.

PR #176, opened 2017-02-08
Move to redux

jameshadfield:
_This PR is ready to be reviewed -- run `npm i` first to get `react-router-dom`_
_master has been merged into this branch_

### Overview
* Lots of things have been moved out of App.js (which is now <150 lines 👍 )
  * colorScale is now in redux
  * redux initialisation (i.e. defaults) has been moved to the reducers
  * URL parsing has been offloaded to helper functions in `util/urlHelpers`
  * `changeRoute` no longer exists
  * App doesn't connect to any redux state (except for dispatch), and thus doesn't pass any of this down to children via props.
  * App no longer uses location in `state`, this is always taken from `this.context.router`
* Unused files have been placed in the folder "old_code", in order to be less confusing.
* React Router is now v4 💯  but we aren't really using it to it's capabilities with regards to pathnames. All the query parsing is done by us (this is required, RR doesn't help here). This is the biggest change from v3, `location.query` no longer exists. There are still some broken bits (see below).
  * There are a number of URL helper functions in `util/urlHelpers`
  * To keep things in sync between actions, when we dispatch an action that will change, e.g. `colorBy`, we also change the URL with a call to `modifyURLquery`. E.g. see [this commit](https://github.com/nextstrain/auspice/pull/176/commits/4ee75b67ce1fef4e627a57b15e5c50e0de4fcde5)
  * router is not passed down anywhere, it's always called as `this.context.router`
  * we still use `parseParams`, some of which may one day be moved to RR
* The filenames in `src/actions` have been renamed to be more intuitive as the number of actions grows, and we get more complex actions (thunks)
* `treeView` has been cleaned up, but there is still more to do here.
  * `shouldComponentUpdate` is used to prevent unnecessary rendering (i'd be interested to benchmark this)
  * tree-related info, namely `tipVisibility`, `nodeColors`, `branchThickness`, `tipRadii` are stored in `reduxState.tree.X`
    * in order to know when to re-render, i.e. when these things have changed, i've put in a version number to each, so you can do a quick check like `if (this.props.tree.tipRadiiVersion !== nextProps.tree.tipRadiiVersion)`
  * the tree now rescales when the sidebar closes
* The controls components are more self-contained, in that they don't rely on props being passed from parents via `{...this.props}`, but there's still more to do here.

### Bugs (i'm sure there are many more than this...)
* filter URLs are a bit of a mess - both modifying the URL and especially restoring state from it.
* filters in general are broken
* genotype colorBy doesn't work

### To Improve
* dataset changes are slow and many things are calculated too many times (e.g. colorScale)
* branchThickness is always updated, even tho it's only ever set when the tree loads.


jameshadfield:
This PR would close #158 #161 #107 #144 #99 #153

colinmegill:
A couple regressions I found:

* author filters no longer work
* location filters no longer work

Since the scope of this PR is large, I'm ok with making issues for the regressions and merging this in. Overall, this is a big improvement in code quality and standardization.

trvrb:
Great to see this merged. Thanks @jameshadfield!

Issue #167, opened 2017-02-03
SVG path vs polyline


trvrb:
In auspice v1, lines look like:

<img width="461" alt="rounded-corners-polyline" src="https://cloud.githubusercontent.com/assets/1176109/22577251/563b4370-e975-11e6-97cf-5e06bbfec9ea.png">

In auspice v2, line currently look like:

<img width="675" alt="rounded-corners-path" src="https://cloud.githubusercontent.com/assets/1176109/22577345/da7233a6-e975-11e6-8832-920e1b6ea733.png">

I think (but I'm not certain) this can be fixed with something like this modification to node x and ys:

https://github.com/blab/nextflu/blob/master/auspice/js/auspice/tree.js#L118

rneher:
we need path bc branches are sometimes arcs. the main difference is that in v1 branches are L shaped, while now they are T shaped. I think T is more sensible, since the base T-bar corresponds to the internal node. if branches are L, many branches overlap in polytomies and the internal nodes gets obscured. But I agree that this doesn't really play nicely with the frequencies.

trvrb:
I see. Thanks Richard. The best thing to do in this case is to modify the SVG paths directly to include little arcs connecting segment joins like so:

http://stackoverflow.com/questions/10177985/svg-rounded-corner

rneher:
I don't think this works bc there is no connection between the T-bar and the stem of the descendant clade. for each node, it draws a line to the parent and a vertical bar. rounded corners would work as expected if all branches had the same width.

trvrb:
Okay. This makes sense. Then I think the only fix is like mentioned above and to jitter the x's and y's to compensate for branch thickness.

rneher:
alternative would be to split the path and use the width of the dominant child as width for the T-bar/arc. either way would be messy.

trvrb:
It looked like `chop_branches` (#183) has resolved this. However, I don't know what's going on with `branchthickness`. @rneher: Is `branchthickness` part of this issue? If not, perhaps this can be closed?

rneher:
the branch `branchthickness` solves one lingering issue that the path aren't properly aligned on initial load bc the tree is drawn before branch thickness are set. whenever branch thickness are changed, we need to re-layout and redraw. there are funny async issues though and this is currently worked around by waiting a few hundred ms after changing the branch thickness and redraw then. also the versioning of  branch thickness does not yet work (@jameshadfield) this would be good to have in place if a full redraw happens after branch thickness change

PR #183, opened 2017-02-17
Chop branches

rneher:
improvements to phylotree. should not have changed the API


rneher:
fixes #167

jameshadfield:
@rneher this code looks good to me

jameshadfield:
image
It's still not 100% yet, but it's a massive improvement and should be much easier to tweak now that .branches holds both the stem and the T

rneher:
should be fixed now, issue was that tree is drawn before branch thickness is set. I put a rather hacky delayed redraw to fix it.

PR #186, opened 2017-02-18
Branchlabels

rneher:


trvrb:
I just tested this branch. There might be some small fiddling with placement, font size and amount of text to clip, but this seems generally great. There does seem to be an issue however, with how to interact these labels with +/– tree zoom. Right now these just size up, like branches. I don't think this should stop a merge however, as this is a general issue with tree zoom and will have to be addressed regardless of branch labels.

rneher:
if this callback in treeview

https://github.com/nextstrain/auspice/blob/branchlabels/src/components/tree/treeView.js#L330

knows about the zoom level, this would be addressed.

rneher:
padding etc is here:

https://github.com/nextstrain/auspice/blob/branchlabels/src/util/phyloTree.js#L154

can be also be set in the options dict on instantiation which overwrites defaults.

trvrb:
Thanks @rneher. Great to put this in `setDefaults`. Would you agree on merging this now?

rneher:
sure.

PR #199, opened 2017-02-23
branch & tip hover functionality

jameshadfield:
## regressions
none as far as I know

## code changes
* branch/tip hover tree manipulations are now within callbacks in treeView
* branch/tip click events change nothing on the tree - this has been done in a seperate branch (not yet public)
* phyloTree id names for branches changed to `branch_S_###` and `branch_T_###`

## style changes
* hovering on branches highlights the branch
* hovering on tips changes the radius


rneher:
looks good to me!

trvrb:
These changes look great! Thanks @jameshadfield. The only possible issue I see is when hovering on a vertical part of a branch. Often won't highlight the whole thing due to overlapping branches. One suggest would be to just highlight `branch_S_###`. This would result in exactly the same behavior for unrooted and clock.

Issue #200, opened 2017-02-23
draw T branch to parent, not child


jameshadfield:
If we define a branch as the stem (the horizontal bit in rectangular view), we currently draw the T (vertical bit in rectangular mode) to the children at the same step. This should be changed so that the T is drawn from a branch to it's parent (similar to nextstrain.org). This causes issues with T thicknesses and inference of country and is especially apparent on branch `dynamic-branch-thickness`.

nextstrain.org:
![image](https://cloud.githubusercontent.com/assets/8350992/23280128/db9473ae-f9cc-11e6-9fac-55fbf229bf2f.png)

new nextstrain:
![image](https://cloud.githubusercontent.com/assets/8350992/23280163/fd56ee0e-f9cc-11e6-9242-dac669c2e535.png)

@rneher thoughts?

rneher:
Both have advantages and disadvantages. I think the T belongs to the parent and should have the state of the parent as in new nextstrain. Especially in multifurcations, you otherwise have a large number of branches covering each other up. this also messes with the branch tool-tips: they would show mutations of the branch on top even though these mutations are not shared by other children descending from the same multifurcation.

jameshadfield:
Yes I see. Branch `hover-stem-only` restricts the hover state to the stem, which would solve some of those issues. 

The issue @trvrb and I found was in views like this: 
![image](https://cloud.githubusercontent.com/assets/8350992/23280630/d45e4c48-f9ce-11e6-9d62-20c0645bce5e.png)

rneher:
I agree this looks a little odd. but from a logical point of view I like the T way of plotting it.

trvrb:
@rneher: The situation I really didn't like is shown in the above figure from @jameshadfield. We have the virus moving from purple (French Polynesia) to green (Brazil) and then to red (Haiti). I don't think the bottom part of this branch should be green. It should be red. Like it is in old auspice. Like if the branches were straight lines rather than 90º joins.

![d45e4c48-f9ce-11e6-9d62-20c0645bce5e](https://cloud.githubusercontent.com/assets/1176109/23281065/89c3e6c8-f9d0-11e6-9972-9647a862f130.png)

rneher:
I don't agree. the ancestral virus in your circle was in Brazil. It moved to Haiti at some point along the horizontal part.

trvrb:
[...]

PR #201, opened 2017-02-23
branch highlighting on stem only

jameshadfield:
Do people prefer this behaviour over what's currently in master?

You can now only hover / click on the stem of a branch (so only affects rectancular & radial views), and highlighting is only applied to the stem.


trvrb:
I think this is better than the previous behavior in `master`. Hovering on T bars is currently strange as I think it's unclear that the mutations listed refer to mutations on the parent branch, rather than mutations on the child branch. This fixes the confusion. Would be nice to get functionality back however per #200.

rneher:
I don't understand the confusion. mutation etc happened on the S part and are present in the T part. child mutations happen later but the children inherent the state at the T.

PR #204, opened 2017-02-25
Minify titlebar and improve responsiveness

### LINKED ISSUE #34: Help page

### LINKED ISSUE #122: make titlebar component responsive
(after #121)

### LINKED ISSUE #191: Static divs on screen resolutions close to 780px
Perhaps due to bootstrap media queries, with the width is slightly less than 780px the divs still have 780px width, so text is wider than the screen.

This is not related to padding of divs.

This affects iPads

### LINKED ISSUE #195: Make whole "tile" clickable on splash page

trvrb:
This PR moves the titlebar in App into the sidebar. This fixes the Escher-esque laying that had continually bothered me. It also gives a nice place to attach the sidebar toggle to, so it's no longer floating. The toggle is now just an extension of the titlebar.

This also cleans up a few broken things, including:

* Made title on splash page responsive
* Remove help page
* Improve tile breaks and window breaks
* Made whole tiles clickable

Fixes #34 
Fixes #122 
Fixes #191 
Fixes #195


jameshadfield:
looks good!

on my laptop the tree & map are never side-by-side - is this the plan?

I think the title-bar should be fixed (i.e. always visible), both in the sidebar an in static pages, to prevent things like this:

image

trvrb:
>I think the title-bar should be fixed (i.e. always visible), both in the sidebar an in static pages, to prevent things like this...

I agree this is almost certainly better. I tried to do this but got stuck with making it work with sidebar. I can return to this latter this weekend.

>on my laptop the tree & map are never side-by-side - is this the plan?

They should end up side-by-side on a big enough screen. Laptop size felt squished however and I purposely set the break to side-by-side past my laptop size.

side-by-side

PR #208, opened 2017-02-25
Zoom clade grid tooltip

rneher:
combined PR. tooltips and grid_zoom had some conflicts but works now fine for me. also merged in the title bar. @trvrb and @jameshadfield please check.


trvrb:
I'm getting a bug when zooming to clade and then zooming out with "-" a couple times:

zoom-bug

rneher:
yes, can confirm. This is the usual different updates cancelling each other issue. looks quite artsy though. I presume this happens when you click - mutliple times in rapid succession?

trvrb:
Yes. Multiple clicks within like 1 second of each other.

rneher:
after this fix, I didn't manage to break it again....

trvrb:
Nice! I can't reproduce the bug anymore. This is great. Another small bug however. After zooming to a clade, when you zoom all the way out, you end up with the root node selected, like so:

selection-remains

Would be better (but not urgent) zooming all the way to the root, cleared the info panel.

rneher:
it now dismisses the info box by setting the selected branch to null of we are at the root. However, the behavior on resize is odd. It resets the tree without currently resetting visibility. There is a comment in the code that reset is desired...
 
https://github.com/nextstrain/auspice/blob/zoom_clade_grid_tooltip/src/components/tree/treeView.js#L91

not sure I like this on a small window size change

PR #212, opened 2017-02-26
Info panel styling

### LINKED ISSUE #175: Style and position info panel and tooltips
The text in the tree info panel needs some simple styling applied.

### LINKED ISSUE #180: Link back to original data
Feedback is to make "more obvious links through to the original datasets". We can most easily accomplish this by having a link out in the Info panel for each tip. We currently maintain a URL associated with each sequence in fauna. Can just pass this through Augur to expose it to Auspice. This would go to Genbank in some cases, GitHub in others, etc...

This would go in Item page as well when completed.

trvrb:
Style info panel / tooltip and create data link. @jameshadfield: You might want to do more with the styling, but I'm pretty happy at the moment. There's some positioning difficulty but I'll leave this for a separate issue.

Fixes #175 
Fixes #180

### DIFF
--- a/src/components/tree/branchSelectedPanel.js
+++ b/src/components/tree/branchSelectedPanel.js
@@ -8,34 +8,37 @@ const BranchSelectedPanel = ({branch, viewEntireTreeCallback, responsive}) => {
       // width: responsive.width - 280 /*legend*/ - 35 /*padding*/,
       position: "absolute",
       // left: 280 /* legend width */ + 20, // RESPOND TO LEGEND COLLAPSE!
-      left: 5,
+      right: 5,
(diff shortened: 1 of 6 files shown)


rneher:
we the prev version had a fixed height. that allows positioning of the panel respective to all four corners. is there a way to figure out the height before it is rendered?

rneher:
my prev comment if fixed by a06e82832f256f5c5aca43b7188aa546d7f91872

jameshadfield:
@rneher my understanding is that it's not possible to calculate text widths, div heights, etc until it's rendered by the DOM - which is why I took the easy option and specified width and height ;)

PR #233, opened 2017-02-27
minor static page improvements

jameshadfield:


jameshadfield:
going to merge this as @rneher may not be online before release

PR #242, opened 2017-02-28
compute responsive

colinmegill:


colinmegill:
This fixes map not responding to sidebar events

PR #250, opened 2017-03-02
Single column responsive tweets

colinmegill:


colinmegill:
@trvrb it seems like the cleanest option is to switch off the two layout feature - it was interfering with Twitter's UI. It was a little bit more visually busy, and certainly people are used to reading tweets single file anyway :)

jameshadfield:
much cleaner! 
I'm just going to add in a couple more tweets - ok to push to this PR?

colinmegill:
@jameshadfield Absolutely! I'm all set

PR #253, opened 2017-03-13
Map responds to controls

colinmegill:


trvrb:
This super cool. One possible bug though. Many filters end up with disconnected demes on the map. Like so:

filter-zika-to-caribbean

I can't think of a scenario where this should occur given the trace-subtending-branches algorithm we're theoretically using. Actually... I'm seeing missing transmission edges before any filtering has occurred. Here's Zika in South America on e06c630828a2cc338a9f6a8ccc8d4da0e7e093a2:

missing-connections

Notice that Peru is disconnected. This is not an issue on current `master`:

correct-connections

jameshadfield:
Filtering is now applied from the "valid tips" to the root and time is a slice through the tree. Here "valid tips" mean tips that match the filters, are in-view (this occurs when a branch has been zoomed to), and are within the time parameters. This means that you will not get the images seen above if no time bounds are applied, but you will start to get a disconnected graph as you tighten your time bounds. 

South America (no time constraints):
image

South America (2016-2017):
image

PR #286, opened 2017-03-19
Update date range from metadata on load

### LINKED ISSUE #258: dynamically set time-min & max
- [ ] hardcode both in augur
- [ ] redux init sets this from JSON

trvrb:
This finds `date_range` in `meta.json` with `date_min` and `date_max`:

```
 "date_range": {
  "date_min": "2013-06-29", 
  "date_max": "2016-11-21"
 }, 
```

And after `meta.json` is loaded it fires an action to update `controls` in redux store. This seems to be working just fine. If `date_range` is not in `meta.json` then the defaults in `globals.js` are kept, just as before. @jameshadfield: could you look to see if this done as cleanly as possible in terms of redux? You're more familiar with this than I am now. I've pushed new JSONs to `data.nextstrain.org`, so this should work with remote data.

Fixes #258

### DIFF
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -2,8 +2,40 @@ import { updateColorScale, updateNodeColors } from "./colors";
 import { dataURLStem } from "../util/globals";
 import * as types from "./types";
 
-/* request metadata */
+const updateDateRange = function () {
+  return function (dispatch, getState) {
+    const { controls, metadata } = getState();
+    /* bail if all required params aren't (yet) available! */
+    if (!(metadata.loadStatus === 2)) {
+      return null;
+    }
+    if (metadata.metadata.date_range) {
+      if (metadata.metadata.date_range.date_min) {
+        dispatch({
+          type: types.CHANGE_ABSOLUTE_DATE_MIN,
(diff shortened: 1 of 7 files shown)


trvrb:
Hold of merging this. I just noticed that this overwrites URL params from whatever is in `meta.json`. URL should take precedence.

jameshadfield:
@trvrb working on this now

trvrb:
@jameshadfield --- I noticed that in the live app there's currently a bug where for Ebola, URL params are being overwritten on load. For Zika it's working fine. I assume this has something to do with the default routes being picked up.

jameshadfield:
this is a bit bigger - currently on nextstrain.org/ebola if there are dates in the URL and you refresh then they are clobbered :(

jameshadfield:
@trvrb the dev version now correctly handles URL dates (and layout etc) - if this is working ok for you then i think this PR is good to go in

trvrb:
This looks great @jameshadfield. Though I just noticed an issue with numeric to calendar date conversion that I'm working on now.

trvrb:
I believe this is ready to go. I'm going to merge.

PR #287, opened 2017-03-20
Improve app performance

trvrb:
This PR does three main things:

* Snap to rather than transition on attr / style tree updates. I think this works _much_ better for things like date slider and legend hover. Makes the app feel much more responsive.

* Profiling revealed that react-sidebar was thrashing the app when it would attempt to save it's width after component update. Because sidebar contains tree, map, etc... it would be constantly updating. This would causing a ~50ms hit after almost every action. I've forked react-sidebar to [nextstrain/react-sidebar](https://github.com/nextstrain/react-sidebar) and removed the code in question on branch `remove-savewidth`. This PR swaps us to using the forked repo here: https://github.com/nextstrain/auspice/compare/snappy?expand=1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R63. You'll need to `npm install` to see the improvement.

* Remove radium from non-framework components. Profiling was revealing decent overhead to having radium on DOM heavy components like tree. I've stripped it out of a few places.

Overall, I think this is a definite performance improvement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "react-router-dom": "^4.0.0-beta.5",
     "react-select": "^1.0.0-rc.2",
     "react-sidebar": "^2.2.1",
+    "react-sidebar": "git://github.com/nextstrain/react-sidebar.git#remove-savewidth",
     "react-svg-pan-zoom": "2.0.0",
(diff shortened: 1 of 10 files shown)


trvrb:
Only good feedback here. Going to merge.

PR #291, opened 2017-03-28
Entropy improvements

jameshadfield:
Entropy panel changes:
* d3 (v4) used (victory chart removed)
* zoomable (on key press - hopefully works cross browser)
* zebra style colours to avoid confusion with genotype colours in tree / map
* AA / NT toggle

This PR does not address:
* Stacked color bars according to mutations (requires dynamic tree traversal)
* Entropy reflective of zoomed in tree / filters
* AA / NT toggle does not change info box in tree

Minor issues:
* clipping isn't working - so when you zoom you'll get bars outside the x-axis bounds. I don't think this is a dealbreaker


trvrb:
Thanks James. This is awesome! Issues:

* Entropy panel does not resize on window resize.
* Seems perhaps reasonable to have the option / resize work while moused over the genome view.
* Might look more aesthetically pleasing to have gray bounding box behind genome annotations. Not sure though.
* The zoom is really nice. Though would be more intuitive to have handles on the edges of the gray box, the way that [CrossFilter](http://square.github.io/crossfilter/) behaves. I now see that you can grab the edges of the gray box to resize. This is great. Perhaps just adding a small visual cue?
* Zooming currently pushes the entropy bars to the left of the y-axis. This should be hidden.
* It's possible to zoom to negative values. Shouldn't be possible.
* I think it would help to have the a bar colored as highlighting if this site is used to color by. This will give the user some feedback that clicking on a bar actually did something.
* The `move` cursor should probably be an open hand cursor instead, like map. But open for disagreement.
* The "Note: Holding option / alt allows you to pan & zoom" randomly disappears. Though I do find the text somewhat distracting as well. Don't know the best thing to do here.
* On initial load, I'm not able to grab the right hand side of gray box with mouse.

jameshadfield:
@trvrb I think all of those have now been addressed. If you're happy it's ready to go in.

trvrb:
This is looking so good. Thanks for the efforts here @jameshadfield. Just a couple very small remaining issues:

* There's an off-by-one error when clicking on a site. For example, for Ebola clicking on "Codon 409 in protein GP" results in coloring by `GP_410`. There's the same off-by-one for nucleotides as well. I think the tooltip is the issue, but not entirely sure.
* Should probably default to "AA" rather than "NT". Seems the more likely place of initial interest.
* Overlapping genes are an issue. Our only example is "sGP" vs "GP" in Ebola. Will be worse with overlapping reading frames. I'm okay not worrying about this for the moment and we can return when we have a virus that demands it.

jameshadfield:
@trvrb first 2 points fixed - i'll file an issue for the 3rd now

trvrb:
@jameshadfield --- Awesome! Merging now.

Issue #293, opened 2017-04-08
Allow overwriting of default colorBy


trvrb:
`colorBy` currently defaults to `country`. If `country` doesn't exist, `colorBy` should default to first element in `color_options` in `meta.json`.

This is currently on the branch `set-default-colorBy`. Might be basically there.

trvrb:
This is resolved as originally stated.

Issue #292, opened 2017-04-08
Fix initial setting of date range in redux


trvrb:
Currently, when `date_range` is loaded through `meta.json`. It correctly registers in the redux store on initial load. However, I believe the tree starts its loading before this redux event finishes.

trvrb:
Commit 7843e59fb3eab16aa919546ae3556a1aab2e11b9 on `master` fixes the redux half of the bug.

trvrb:
I believe this is fixed in 164ccba4efd214e6d582e402406e3bf42200387a.

PR #294, opened 2017-04-08
Include dengue

trvrb:
Getting dengue up and running. Will require:

- [x] splash page tile
- [x] fixing date range load issue #292 
- [x] fixing default colorBy issue #293


trvrb:
I think we're set now. Thanks @sidneymbell!

PR #301, opened 2017-04-13
 apply fitToViewer before zooming into clade to make sure that zoomed…

wdingx:
After panning the tree, clicking on branch executes function zoomIntoClade but the zoomed area can extend beyond viewer. 
Adding fitToViewer before zoomIntoClade will ensure that the zoomed region always fits the viewer.


trvrb:
Thanks @wdingx. Definite improvement. Merging.

PR #302, opened 2017-04-14
Fix default georesolution map bug

sidneymbell:
Appears to also fix the default coloring issue.


trvrb:
The desired default state of `country` shouldn't be buried in the logic of `loadData`:

https://github.com/nextstrain/auspice/pull/302/files#diff-4dc7fd61bd86087afe91768c79f27934R64

The ability to set default from `globals` should be maintained:

https://github.com/nextstrain/auspice/pull/302/files#diff-1789c8dd3ea02256186b9024c4238fbdR56

`globals` should consolidate these decisions as much as possible.

Also, if a tree has `country` and `region` and `division` you need to be able to specify default resolution in `meta.json`.

jameshadfield:
I vote to make it essential to specify in meta.json

trvrb:
My thinking was generally to have defaults in `globals`, but allow overwriting via `meta.json`. If we make it mandatory to specify geo resolution in `meta.json`, where do we stop with mandatory defaults? Would we need to specify tree layout and tree branch length metric as well?

trvrb:
My proposed behavior would be:

1. Start `geoResolution` out with whatever is specified in `globals.defaultGeoResolution`. Country seems like a useful baseline.

2. If `defaultGeoResolution` is specified in `meta.json`, then overwrite `geoResolution` with whatever is in `meta.json`.

3. Check `tree.json` to see if current `geoResolution` exists as an attr. If it does not exist, move up the resolution stack until a matching attr is found, ie. if `division` is specified, but tree lacks `division`, then move up to `country` and if then if tree lacks `country` move up to `region`.

sidneymbell:
@trvrb - Thanks for looking at this, sorry for the slow response (just saw this).

I think the proposed behavior above makes sense. (3) is already implemented here. I'll see if I can tweak things today to get (1) and (2) in order.

And yes, I can switch to a `master` PR at that point with issue linkage (just was still a bit wary about `master` given my newbie-ness, and it's currently a dengue-specific fix). Sorry about that / thanks for the suggestion.

sidneymbell:
Now fixes the dengue map. Will make a separate branch to fix the default color by and date range.

colinmegill:
This brought in James' changes, so we can wait to merge until that branch is also ready. This now renders map and includes a default.

sidneymbell:
Now also includes James' fix to the default colorby. Works for dengue. Only remaining issue with dengue is the default date range.

trvrb:
Closing this in favor of #314.

PR #310, opened 2017-04-27
analysis slider (if specified in meta.json)

jameshadfield:
This PR adds an "analysis date slider" when the option is specified in the meta.json file. 

Analysis date is already used as a `colorBy` in H7N9 - http://www.nextstrain.org/flu/h7n9/ha?c=fauna_date&r=division - and the new slider can be seen here - http://auspice-dev.herokuapp.com/flu/h7n9/ha?c=fauna_date&r=division

I probably won't merge until zika / ebola have been updated to have this as a `colorBy`


trvrb:
@jameshadfield --- Nice work! There were just a couple things that I think need updating here before the analysis slider can be live.

* Dates in the analysis date slider should be proper calendar dates. Like in the date slider.
* I'd suggest moving analysis date slider to the filters section. This is because it acts as a filter (pruning tips) rather than acting like the date slider, which also prunes branches. I also think it would be good to have this less prominent. It's a really nice feature, but should not be above "Color By".

I'm going to merge this because we need your modifications to `loadData` for downstream branches by @sidneymbell.

Issue #315, opened 2017-04-29
Provide defaults for colorBy, distanceMeasure and layout


trvrb:
The improvements to `colorBy` to get dengue working are definitely helpful. Currently, we default `colorBy` to the first element in `color_options` in `meta.json` as originally specced. This is safe and will result in a valid initial coloring. However, there is no real control in augur for what order `color_options` are read into the JSON. We need to provide a stated default like what's provided for `geoResolution` for `colorBy` as well. I'd suggest that this piece of `meta.json` could expand for other defaults as well, like so:

```
defaults: {
  geoResolution: "region",
  colorBy: "ep",
  distanceMeasure: "div",
  layout: "rect"
},
```

If a default is present here then this would override the current selection of the first element in `color_options`:

https://github.com/nextstrain/auspice/blob/master/src/reducers/controls.js#L153

If elements are missing from `defaults` in `meta.json` (or if `defaults` is missing entirely) then just keep what's specified as default in `globals`.

trvrb:
Oh. This was the issue that PR #316 fixed. Closing now.

PR #313, opened 2017-04-28
Sidebar errors

jameshadfield:
Fixes #307 #308 #312

PR #316, opened 2017-04-29
Add meta.JSON defaults for layout, colorBy, and distanceMeasure.

sidneymbell:
Fixes issue #316


trvrb:
Awesome. Thanks @sidneymbell!

Small etiquette note. You don't want to say "fixes #316" when #316 is the PR itself. The "fixes" keyword will tie the PR to an existing issue. You can have a PR that fixes multiple issues like #313. A little weird that PRs and issues share the same numbering system, but whatever...

trvrb:
For posterity, this fixed #315.

PR #319, opened 2017-05-02
fix-295

### LINKED ISSUE #295: Grid is not updating properly when changing dataset
On current `master` (05038e75694aedb35efc61b399a7871e5613f05f) there is a bug when changing between datasets. Loading the app and toggling between Ebola and Zika will result in the tree changing, but the grid remaining the same as the previous dataset. Mousing over anything (causing a rerender) updates the grid to what it should be.

jameshadfield:
fixes #295


trvrb:
This looks great. Thanks @jameshadfield!

Issue #306, opened 2017-04-25
display notifications / warnings / errors


jameshadfield:
I'd like to implement a system similar to https://github.com/atom/notifications

Namely, a `notifications` component and reducer plus some actions

thoughts?

jameshadfield:
current progress on branch `notifications`

![notification](https://cloud.githubusercontent.com/assets/8350992/25644491/659add90-2f5b-11e7-8066-49c4434f679d.gif)

jameshadfield:
to do:
- [ ] persist "error" notifications

jameshadfield:
See branch `notifications`

trvrb:
@jameshadfield: I see the `awaiting_merge` flag. Are you suggesting that `notifications` be PRed? Or that you're waiting on a different branch to be merged before you can proceed here?

trvrb:
@jameshadfield: Can you PR `notifications`? Seems to be the proper way to request feedback.

jameshadfield:
now merged

PR #324, opened 2017-05-06
Notifications

jameshadfield:
I think a PR is more appropriate than issue #306 

Note that no notifications are currently triggered from proper errors, and of course the buttons on the splash page should be removed.


trvrb:
This is a fantastic direction @jameshadfield. I'm wondering however, if this isn't better discussed as the original issue #306 and then go with the PR when it's ready for code review. But perhaps the functionality is ready to go in even if no notifications are triggered as of yet?

jameshadfield:
agreed!

jameshadfield:
reopening as per @trvrb's request

trvrb:
This is a good general direction. The notifications behave well, though I have to admit, I'm somewhat at a loss for when to exactly use these. Seems perhaps distracting to pop up a notification when switching datasets for example. I'm entirely sure though. The functionality seems good, just have to figure out the most appropriate use case.

PR #326, opened 2017-05-16
Mugration

### LINKED ISSUE #256: Tree confidence inferred state fade
- [ ] mouse over shows probability
- [ ] opacity E [0,1] - [Lemey et al 2014](http://bedford.io/papers/lemey-air-travel/)
- [ ] mouse over potentially shows probability info box (time as well?)

jameshadfield:
This PR aims to communicate likelihood values of geo-inferred states ("mugration") via branch opacity and a table in the info panel. It uses a (not-yet-pushed) version of augur to export the values

to-do list:
- [x] what colour should the branch be on hover?
- [x] store the data more efficiently in the JSON. Update: gzipped tree files increased from 81kb to 96kb (likelihoods for two geo traits, zika dataset). This is OK I think.
- [x] improve styling of table in info panel
- [x] radio button controls

Would close #256

thoughts @rneher @trvrb ?


jameshadfield:
to run you'll need updated JSONs. SOmething like this should work:
```
cd <augur>
git fetch
git checkout mugration_confidence
cd zika
python zika.prepare.py
python zika.process.py
cp auspice/*json ../../auspice/data/ #maybe different for you
```
then, in the auspice directory
```
node dev-server.js local
```

jameshadfield:
image

Issue #257, opened 2017-03-15
temporal confidence intervals


jameshadfield:
- [ ] show one figtree / BEAST like bar on hover
- [ ] display info in tooltip / panel

trvrb:
Resolved in #329.

PR #329, opened 2017-06-09
Confidences

jameshadfield:
@colinmegill I'd like to start code-reviews for each PR in order to improve programming style and technique. Could you please review this PR. 

**aesthetics**
In conjunction with `augur` this PR incorporates confidences for both inferred traits (*mugration*) and TimeTree. These are indicated by changes in branch colouring (for the former) and shaded bars at the nodes for the latter.
Salient points:
* non-temporal confidences are always displayed if available, and cannot be toggled
* temporal confidences by default show when you hover over the branch, and there is a toggle that turns them all on. They are only available for rectangular trees.

These have been discussed with @trvrb and all looks good - @rneher are you ok with this?


rneher:
This looks good to me -- nice work!

jameshadfield:
will close #256 #257

trvrb:
This appears to be working great. I'm going to merge into `master`. Maybe we can aim to rebuild Zika and Ebola to take advantage of this alongside a release.

Issue #288, opened 2017-03-20
Map redraws fresh each render


trvrb:
@colinmegill ---

In profiling app performance, I noticed that (as I think you suspected), the current map code has some decent overhead attached to it. For example:

<img width="1369" alt="map-call-stack" src="https://cloud.githubusercontent.com/assets/1176109/24086703/cdcf9f42-0cd1-11e7-9d81-c12b2744f685.png">

Each render, `geodesic` needs to be computed fresh. That's 25ms here. Drawing markers are another 25ms. These should just be computed once on load and then attrs changed to show/hide, change transmission width, change deme radii, etc... as state changes. This should improve performance, but will also make it possible to transition map states to move toward animation.

trvrb:
I believe that #330 fixes this.

trvrb:
Specific issue closed by #330. I'm going to do a separate issue for remaining map perf issues.

Issue #279, opened 2017-03-16
map transmission curves


jameshadfield:
The geodesics look good for intercontinental stuff but are essentially straight lines for smaller distances. I think we should use custom curves where (similar to gytis' work) curvature is inversely proportional to distance.

Additionally, transmissions from A-B and B-A currently overlap, whereas they should be flipped / curved in different directions.

trvrb:
Closed by #330.

Issue #269, opened 2017-03-15
Intelligently choose a centroid to minimize over dateline transmissions given latlong box


colinmegill:

trvrb:
@colinmegill ---

I've just spent a couple hours with `Leaflet.Geodesic` plugin. I'm fairly convinced now that this will not be a simple fix. My suggestion is to shelve this. There is a bigger reason why however. If you look at a zoomed map near the South American equator for Zika, you'll see all the lines are nearly straight. This should be resolved per #279. We should go to Bezier curves or the like. In doing so, we can address this current issue. What do you think?

trvrb:
This is (indirectly) resolved by #330. Being able to pan left or right fixes this.

trvrb:
Closed by #330.

Issue #238, opened 2017-02-28
Map bounds not quite right


trvrb:
On `master`, I'm getting transition lines for Zika that go right to the border of the map, but look like they stop before they should:

<img width="1087" alt="map-bounds" src="https://cloud.githubusercontent.com/assets/1176109/23388596/128cf482-fd31-11e6-8cdb-d3daa62edb34.png">

jameshadfield:
the (just pushed) dev.nextstrain.org version looks a bit different

![image](https://cloud.githubusercontent.com/assets/8350992/23388835/6b4aa51e-fd19-11e6-8c34-d884f826eca3.png)

trvrb:
I'm getting the image above (in my post) on `dev.nextstrain.org` from my laptop when doing a shift-refresh.

jameshadfield:
strange my image is from dev.nextstrain version 0.3 (the latest) so perhaps different browsers?!?

trvrb:
My guess is just different screen widths.

trvrb:
Fixed in #330.

trvrb:
Closed by #330.

PR #330, opened 2017-06-16
Map animation v1

### LINKED ISSUE #66: User clicks on a line, highlight both instances of great circle to show incoming & outgoing from side of map

sidneymbell:
Implements animations for map + tree as well as several controls. Responds to user-input animation length in seconds, user `minDate` from date slider, and whether animated transmissions should be cumulative.

Closes #66, #238, #269, #279, #288.


trvrb:
Let me know when this is ready for review. I'll take a closer look then, including merge issues.

sidneymbell:
@trvrb - I believe this is ready for review. The "threads" feature for spacing out multiple transmissions that we discussed has not been implemented yet, but all of the bugs and UI tweaks we discussed on Friday are fixed/implemented. @colinmegill - am I forgetting anything? :)

colinmegill:
I think we're there with this last push on Friday

trvrb:
Again, this is really cool. Thanks @sidneymbell, @evogytis and @colinmegill! Some remaining issues / bugs before this can go live:

* If you click play and start the animation and then try to grab the date slider it results in a tug-of-war. Clicking play and then manipulating the date slider should send the animation into a "pause" state.
* Small, but the "Play" and "Pause" buttons should be identical dimensions.
* Repeatedly playing and pausing the animation shrinks the animation window further and further from the desired width.
* Rename "Animation is Cumulative" to "Animate cumulative history" or "Accumulate animation lines" (toggles should be verbs). Make this a toggle like [James made for confidence intervals](https://github.com/nextstrain/auspice/blob/master/src/components/controls/choose-metric.js#L73). 
* I feel like a free form text box for "Animation length" is not the right UX. Deleting back leads to "0" and you can't delete the 0. Feels weird. This really needs to be just slow (45 seconds), medium (30 seconds) and fast (15 seconds). Like standard media controls. Call this "Animation speed".
* Dengue is still causing the windowpane bug at timepoints where no tips are are visible.

colinmegill:
Will wrap these up this week.

colinmegill:
@sidneymbell have you already re-added the triplicate map where it is needed?

sidneymbell:
@colinmegill - Yup. Just road-tested it with dengue, too. Toggling it in the metadata JSON here:

``` "defaults": {
  "geoResolution": "region",
  "mapTriplicate": "true"
 }```

Appropriately changes the map display and animation behavior.

**Comment by colinmegill** [comment]
@trvrb We're not repro-ing " Repeatedly playing and pausing the animation shrinks the animation window further and further from the desired width." Could you confirm / add steps to repro

**Comment by trvrb** [comment]
@colinmegill: to reproduce the shrinking time window just:
1. Load `map-animation-v1` and go to http://localhost:4000/zika
2. Click "Play" and immediately pause.
3. I see 2011-07-24 to 2011-11-11. This is 110 days (via Wolfram Alpha, http://www.wolframalpha.com/input/?i=2011-07-24+to+2011-11-11)
4. Click play, click pause, click play, click pause, etc... 
5. I see 2012-09-28 to 2012-12-25. This is 88 days.
6. Doing this more I get 2014-05-30 to 2014-07-28. This is 59 days.

**Comment by trvrb** [comment]
I found two additional bugs:

- [ ] There appears to be a reversion where tree and map are no longer appearing side-by-side with a large window. Getting this on `map-animation-v1`.

- [ ] There should be map bounds. Currently can spin the map right or left far from display area effectively losing the display. There should either be a hard bound a couple maps distant east or west (at the appropriate bound).

**Comment by jameshadfield** [comment]
About to merge `master` into `map-animation-v1` (commit 19d479e1c73a7e9ff6c7e5c4828ca86553348844). It looks impressive.

Here are some issues I noticed:
- [ ] (on the non-cumulative mode) at the end of the animation the date slider should return to the full date range. It currently shows a small slice at the end and nothing is visible.
- [ ] Clicking on the date slider during animation pauses the animation (good) however the "play/pause" button still shows Pause and requires another click before Play is shown.
- [ ] Changing animation speed during an animation does not change the speed - although the highlighted speed label does change. 
- [ ] Changing Tree layout during an animation causes quite a few tree display issues. I suggest disabling these actions while an animation is in progress.
- [ ] Every now and again I get an error (trace below) where the animation freezes.

```
VM3764:62 Uncaught TypeError: Cannot read property '0' of undefined
    at Bezier (eval at <anonymous> (bundle.js:5244), <anonymous>:62:49)
    at extractLineSegmentForAnimationEffect (eval at <anonymous> (bundle.js:5244), <anonymous>:205:16)
    at SVGPathElement.eval (eval at <anonymous> (bundle.js:5244), <anonymous>:221:32)
    at SVGPathElement.attrFunction (eval at <anonymous> (bundle.js:3066), <anonymous>:669:21)
    at eval (eval at <anonymous> (bundle.js:3066), <anonymous>:962:16)
    at d3_selection_each (eval at <anonymous> (bundle.js:3066), <anonymous>:968:30)
    at Array.d3_selectionPrototype.each (eval at <anonymous> (bundle.js:3066), <anonymous>:961:12)
    at Array.d3_selectionPrototype.attr (eval at <anonymous> (bundle.js:3066), <anonymous>:652:17)
    at updateVisibility (eval at <anonymous> (bundle.js:5244), <anonymous>:220:4)
    at Map.maybeUpdateDemesAndTransmissions (eval at <anonymous> (bundle.js:5220), <anonymous>:313:46)
```

P.S. I couldn't reproduce trevor's error about the animation window shrinking as you press pause/play repeatedly.

**Comment by jameshadfield** [comment]
`master` has now been merged in, however this has resulted in some extra issues (or ones I didn't notice before). We can hopefully work together on these next week. 

- [ ] There are issues changing datasets - I believe because the map is running before the tree is ready. I tried to address these here https://github.com/nextstrain/auspice/blob/map-animation-v1/src/util/mapHelpersLatLong.js#L229 but without much success
- [ ] related to point 1, #297 is still there - is this going to be an easy fix? 
- [ ] I had to add a few bits of code to `mapHelpersLatLong` to replicate 660c1f064c69163d2d8a6cb9281527f6dd19bb79 - please check this conditional and what side effects it may have https://github.com/nextstrain/auspice/blob/map-animation-v1/src/util/mapHelpersLatLong.js#L229
- [ ] The following error is now displayed on page load, despite `getInitialState` only being called in `slider.js` which I don't think has been changed?
```
Warning: getInitialState was defined on t, a plain JavaScript class. This is only supported for classes created using React.createClass. Did you mean to define a state property instead?
```
- [ ] enabling the `mapTriplicate` boolean correctly duplicates zika transmissions, however when you pan the map it snaps back to the main view after mouse release, so you can't reposition the world... Is this intentional?
- [ ] The flu dataset, which has no transmissions (see 2nd point, above), still animates (great!) with the demes changing size. It is, however, really laggy as it gets towards the tips. Not an issue for this PR but something to consider for the future. 
- [ ] The ebola dataset is not animating, with the `Bezier` error occurring when I press play (stack trace printed in previous comment)

**Comment by trvrb** [comment]
@jameshadfield:

I'm getting 

```
Warning: getInitialState was defined on t, a plain JavaScript class. This is only supported for classes created using React.createClass. Did you mean to define a state property instead?
```

Only after running `npm install` from `map-animation-v1`. Running `npm install` from `master` and then switching back to `map-animation-v1` removes warning.

----------------------------------

I'm still reproducing the shrinking date window issue.

**Comment by colinmegill** [comment]
@jameshadfield I see the stacktrace you shared above as an error every time Zika animation is run now, but it's completely new to me. @sidneymbell, did you see this before the merge? Is it possible it's a data issue?

**Comment by sidneymbell** [comment]
@colinmegill - I haven't seen it before either.

**Comment by colinmegill** [comment]
@jameshadfield This is new as of you seeing it - if you didn't touch code before you saw it, that leaves a new augur build and data error?

**Comment by sidneymbell** [comment]
Re: 
>The ebola dataset is not animating, with the Bezier error occurring when I press play (stack trace printed in previous comment)

Just confirmed this is a dataset-specific bug. It's the same problem that was previously in the dengue build. I suspect this is due to the recent changes to geo metadata definitions in fauna.

I.e., this happens when two locations have a distance between them of `0`, but it still tries to draw a transmission because they have different geo metadata attached to them. For example, Ebola is failing when it tries to draw the transmission between these two points:
image

This then causes the value for `end` passed to `Bezier` [here](https://github.com/nextstrain/auspice/blob/map-animation-v1/src/util/mapHelpers.js#L209) to be `-Infinity` because the denominator is `0`. 

This was the same bug that was previously messing with dengue. Fixed with a dataset rebuild. We should add some sanity checking to auspice, but more importantly fix this on the backend.

**Comment by trvrb** [comment]
If it helps your debugging, I've noted that eddac025bd087f31c1c8933b64351a5cb952083d works fine with live Ebola JSONs. Broken after Bezier goes in.

**Comment by trvrb** [comment]
In live http://data.nextstrain.org/ebola_meta.json, I have:

```
coyah: { latitude: 9.7090404, longitude: -13.3889559},
kerouane: { latitude: 11.4231137, longitude: -9.5938615}
```

so I don't think it's a distance = 0 issue. @sidneymbell: can you identify the issue more specifically?

sidneymbell:
Yes, this is a Bezier issue. This needs to be sanity checked in auspice.

It's also a data issue: a rerun of dengue fixes this exact bug. Likewise, ebola animation worked fine before the most recent data rerun.

I can look more specifically at exactly what causes it in the morning, but I can say for sure that it arises when the denominator of `end` [here](https://github.com/nextstrain/auspice/blob/map-animation-v1/src/util/mapHelpers.js#L208) is `0`, causing `end` to evaluate to `-Infinity`. Which, looking again, I totally spoke too soon/was mistaken -- the primary problem is clearly a date issue that can be fixed on the auspice side. 

Worth noting, however, that there may still be a secondary geo data issue. In dengue the transmission that caused the bug was between two locations that were actually from the same region and shouldn't have been drawn as a transmission, but had been mislabeled by a recent `--update_locations`. Rolling back to the old version of regions and repopulating vdb/dengue fixed it.  (The geo metadata thing is how I caught and fixed this initially / why I assumed it was a geo problem)

trvrb:
Ah. I see. Thank you. Denominator of `end` is 0 when `destinationDate-originDate`. This is about parent / child nodes that have different locations but the same date. This arose in your dengue example due to mismatched region naming. However, I believe this may exist in the Ebola tree naturally. As in there are zero length branches with different locations. Notice `num_date` in your example above is the same for parent and child nodes. I just found this. It's `V1331`.

ebola

This is a "real" zero length branch. This needs to be fixed in auspice. I'd suggest just adding a small number to the denominator.

sidneymbell:
Yup, exactly. My brain's a bit too dengue centric :)

The auspice side will be a one-line fix tomorrow. The fauna geo thing should also be addressed at some point, but less urgently.

colinmegill:
Great find!

trvrb:
Almost all issues have been resolved. There are a couple very minor things that I'll carve out into separate issues and don't prevent merge. There is one major issue outstanding:

- [ ] Switching from "Ebola" to "Avian" causes the app to crash. Strangely Zika to Avian works and Ebola to Zika works, etc... Specific to this dataset switch.

ebola_to_avian


@jameshadfield: could you take a look at this tomorrow?

jameshadfield:
I'll take a look - FYI despite looking like a `loadMeta...` error, it's actually from `getLatLongs`->`aggregated`

trvrb:
2aaafea1aa97b310cc36ba14d33ee0c6c0ce6e5a resolves this. Issue of `this.props.visibility` not being available when `getLatLongs` is called.

@jameshadfield: I merged `simplify` branch and this didn't resolve. Still, there seemed to be no issues merging `simplify` on top of `map-animation-v1`.

@colinmegill: Can you look at 2aaafea1aa97b310cc36ba14d33ee0c6c0ce6e5a and see if `latLongs` returning `null` if data isn't available is the best pattern or if you would do this check some other way.

colinmegill:
@trvrb that's maybe a little downstream of where the proper check should be, but if it fixes the edge case, I would move to resolve that in a subsequent issue, this works.

trvrb:
@colinmegill: I'll carve out the clean up into a separate issue.

However, I now see that switching between Ebola and Avian bizarrely updates tree but only updates map color, not lat long placement. @colinmegill, would you be able to look into this? It can end up as a separate issue as well.

jameshadfield:
@trvrb i'll merge simplify into master once this PR is in, if that's ok.

trvrb:
This has some remaining small problems. However, I'm merging now and will carve these out as separate issues.

PR #331, opened 2017-06-18
Electronapp

zachcp:
Hi Team Nextstain,

I was playing with the possibility of packaging up Nextstrain as a Desktop Application, and, as it turns out, its relatively straightforward. This PR provides the code and working instructions to package Nextstrain inside of the [Electron Application Shell](https://electron.atom.io/). There are a few outstanding issues required  to make it as polished as the web experience (see below) that basically consist of correcting some file-paths and reducing the total package size (~168Mb on OSX). The app builds nicely and this could be an excellent way to provide versioned data to your end users.  If this of interest, I'd be happy to help solve these issues.

Issues:
- favicon is not passed into the bundle
- splash page images  (ebola/zika, etc) are not visible/the url is wrong
- customize doc ICON needs to be specified.
-  large size of Bundle - aggressively exclude unused NPM modules.
- specify the electron build parameters for a standardized build

best regards,
zach cp


zachcp:
Screenshot of Nextstrain as a local App

screen shot 2017-06-17 at 10 52 12 pm

screen shot 2017-06-17 at 10 51 09 pm

zachcp:
I will play with electron packaging configuration in this repo:here: https://github.com/zachcp/nextstrain-electron until I get a hang of how to package/minimize the builds. The ~168Mb includes all the Node modules which can be excluded to get down to  ~100Mb - about the size of the electron runtime.

I also  thought the raw data was included in this bundle but its not - electron is still reaching our via REST calls to the Nextstrain server.  My interest in playing with this (beyond learning some interesting tech), is to learn to develop a self-contained data application that can distributed to end users. Versioning at the app level would include data versioning.  So once the build is figured out I would be interested in replacing the REST calls with local file-system calls to downloaded data.  

zach cp

rneher:
thanks! this is super interesting. To use local data, you need to chance dataURLStem defined here:
https://github.com/nextstrain/auspice/blob/master/src/util/globals.js#L67

(to get the data, do smth like 
```
curl http://data.nextstrain.org/ebola_tree.json --compressed -o data/ebola_tree.json
curl http://data.nextstrain.org/ebola_sequences.json --compressed -o data/ebola_sequences.json
curl http://data.nextstrain.org/ebola_meta.json --compressed -o data/ebola_meta.json
curl http://data.nextstrain.org/ebola_entropy.json --compressed -o data/ebola_entropy.json
```
)

trvrb:
Hi Zach,

Thanks for PR! This is a great direction (and something I hadn't really considered until now). In this case of working with versioned / local data it makes much more sense to have an installed app. This would be really cool to include on a MinION mobile sequencing laptop.

Ideally, we should be able to work from a single codebase here, right? And just have different npm builds for browser vs electron?

zachcp:
I think the single codebase is the way to go as well.  I am still unfamiliar with how electron packaging works - i.e. making sure files are placed in correct directories (hence the separate repo where you I could control the directory structure).

It also looks like you guys already have the data side of things worked out - very nice. However, I  run into an issue when trying to use local data (see attached). Local data access may require using some node-specific commands to load. 


screen shot 2017-06-18 at 11 03 32 pm

zachcp:
superceded by #345

PR #334, opened 2017-06-20
Seasonal flu

jameshadfield:
This PR:
* adds seasonal flu to auspice
* moves the definitions of datasets to a separate file (will be JSON eventually)
* fixes map errors when nodes are missing the `geoResolution` attr
* the (absolute) min date is taken from the root of the tree

It is currently on http://auspice-dev.herokuapp.com and the S3 files are up to date.

I'd like to merge to master and push live today. Please let me know if there are any objections.
cc @sidneymbell @colinmegill @rneher @trvrb


rneher:
thanks James, this is looking really good. There are still a few outliers. I just to have a function in augur that would look for the first child of the root and if these were few strains and had a branch length >0.03 I would clip this branch and add the child nodes to an tentative outlier list automatically. 

regarding the transmissions between regions, I am still not convinced displaying them for flu is useful. I see the value for the color on the tree, but the map looks to messy for my taste. Also, are there flu specific reason to include Caribbean as a separate region?

Do we have a plan re titer and frequencies? for flu both of them are quite central features and before this replaces the old site, I think we should have titers and frequencies...

rneher:
The Vic/Yam trees with their small evolutionary rates look odd as a time tree for 2y and 3y. The reason is obviously lack of evolution and identical sequences circulating for several years. but then for these less rapidly evolving lineages we don't really need 2y resolution.

jameshadfield:
@rneher I've uploaded (to S3) `vic` &  `h3n2` builds without region inference

the tree pruning function would be really helpful - it seems every build I run there's another outlier! Removing 2y for `vic` & `yam` is super easy if desired

I believe the plan is to add in frequencies (no plan), timeline (sidney) and titers (me, pretty close) as we build them - all the while maintaining nextflu. Do you think this is OK?

rneher:
If we maintain nextflu as is, that would do it (we actually have time tree nextflu working for about a year now but it just never hit the live site). frequencies should be pretty straightforward (just a line graph that talks to the tree...

trvrb:
This is awesome. Thanks James. Some responses:

* Richard, the regions thing (Caribbean, etc...) is my doing. I attempted to streamline regions in fauna to be something that could be used for Zika, dengue, flu and other things on the drawing board. It felt too complicated to try to have separate region maps per virus. Here's the regions as they stand: https://github.com/nextstrain/fauna/blob/master/source-data/geo_regions.tsv. I looked at a bunch WHO material, etc... to try to come up with a sane assignment, but there could be adjustments. I don't see an issue in having Caribbean here for flu as a region to be colored. When we make frequency plots for reports we would just leave this out of the region list to draw from.

* Let's just drop 2y across the board for now. I don't like having it for H3N2 and not for Vic and Yam and I don't think it adds that much here for H3N2 anyway.

* I'm anticipating keeping nextflu.org running for a while, at least until this reaches feature parity.

* James, let's move colors to mirror current nextstrain dengue with blue tones for Asia, orange tones for Americas and green tones for Africa, Europe and Western Asia.

* I actually think the region-level geo inference is a good thing and likely to be decently accurate. The map should look better with Gytis's Bezier curves which should go live shortly. I completely agree about not doing geo inference on countries and leaving branches as gray when doing "country" coloring and in this case not showing transmission lines on the map.

-------------------------------

Important things missing:

* vaccine strains
* clade annotations

I don't think either block this from going live, but they're low-hanging in terms of useful features.

rneher:
re regions: I am not we will be able to have consistent regions across all pathogens, but I don't see a reason why we should. There is a good reason to have south pacific for zika, but not for flu. If I was iceland, i'd be asking for north atlantic. Totally ok with north/subsaharan africa though (many earlier virus are still annotated as "africa")

The only reason for having the 2y resolution was to show more recent viruses, but maybe a better way to do this would be to have region specific builds.

I agree that geo inference on region level is good, at least it gives some indication were the backbone might have been. but the map looks a little too much "everything connected to everything" for my taste. don't feel strongly though.

jameshadfield:
Merging now - the final augur builds are running overnight and will both push the JSONs -> S3 and deploy this version in the morning.

jameshadfield:
P.S. @colinmegill @sidneymbell there are 2 commits:
660c1f064c69163d2d8a6cb9281527f6dd19bb79
521707280ab854124ca213f6f62110c10e1576f5
in the map code which are needed for seasonal & avian flu. Hopefully these merge ok.

Issue #340, opened 2017-06-23
Coloring of AA states from entropy plot is buggy


trvrb:
To reproduce go to:

http://www.nextstrain.org/flu/h3n2/ha/3y?c=region&r=country

This will color the flu tree by region. Next click on column corresponding to HA1 site 159 in the entropy plot. This should recolor according to `gt-HA1_159`. However, only the tips get recolored. Branches stick with the original region coloring, like so:

![coloring](https://user-images.githubusercontent.com/1176109/27503766-7b18ec78-5834-11e7-844a-31cbbc2bfe4d.png)

Additionally, trying to load this directly doesn't work as well:

http://www.nextstrain.org/flu/h3n2/ha/3y?c=gt-HA1_159&r=country

Shows region colors.

I'm getting the same buggy behavior with Zika.

@jameshadfield Do you have an obvious idea here?

jameshadfield:
Nothing obvious - I thought I'd removed the URL genotype bit, but clearly not.

I'll look into this next week.

jameshadfield:
Hovering over the nodes restores them to the correct colour - so I think the redux state is correct but phloTree isn't displaying it correctly.

Issue #300, opened 2017-04-12
general entropy sync issues


jameshadfield:
currently the map restores colorBy for genotype correctly, but the entropy panel doesn't reflect this 

*URLs*
- [X] genotype URLs are not being set by the entropy panel
- [x] ensure genotype URLs are being correctly processed

*ColorBy*
- [X] if colorBy is no longer a genotype, bars must be de-selected
- [x] genotype colorBy must be processed by the entropy panel 

*Panel interaction bugs*
- [x] selected entropy bar disappears when zoom bounds change
- [x] selected entropy bar disappears when browserDimensions change

*Potential improvements not considered*
* maybe zoom in to the gene encompassing the selected base, ideally with a clean transition

PR #342, opened 2017-06-26
Tree bugs

jameshadfield:
This PR fixes 2 bugs:
* #340, where genotype colours were not correctly applied
* a bug where if a branch (i.e. clade) was selected and then the sidebar collapsed or the browser size changed, the tree would zoom out.

Issues remaining for future PRs:
* genotypes in the URL are broken and have been temporarily removed - see #300

In general, reducing the number of dispatched actions (i.e. by making fewer actions with more effects) seems to solve a lot of bugs. This should be our general aim going forward.

@trvrb if this works for you please merge.


trvrb:
This is awesome. Merging now. Thanks James!

PR #343, opened 2017-06-27
Display clade names for nodes when defined by augur inputs

huddlej:


jameshadfield:
Nice work!

* The labels aren't being removed on dataset changes - add something like `this.svg.selectAll('.branchLabel').remove();` to the `clearSVG` method 

* I think line 943 should be   `selectAll('.branchLabel')` (not `tip`)

* There are some related (but not currently used) functions, such as `drawBranchLabels` - could you comment these out for the meantime?

* Similarly, line 1085 refers to svg elements with class `.branchLabels` (not `.branchLabel`) - should this be changed or removed?

huddlej:
Thanks, @jameshadfield ! I've addressed most of the issues you noticed. I left in the reference to `.branchLabels` and a couple of the unused branch label functions for now. These were added by @rneher back in February. His branch label implementation is nicely flexible (using a callback function to determine what the labels are).

Maybe my approach here could be replaced by Richard's? It would be worth discussing offline.

src/util/phyloTree.js
```diff
@@ -925,16 +927,28 @@ PhyloTree.prototype.drawBranches = function() {
...
+//     .data(this.nodes) //.filter(function (d){return bLFunc(d)!=="";}))
+//     .enter()
+//     .append("text")
+//     .text(function (d){return bLFunc(d);})
+//     .attr("class", "branchLabel")
+//     .style("text-anchor","end");
+// }
+
+
+PhyloTree.prototype.drawCladeLabels = function() {
```
jameshadfield: looks good

src/util/phyloTree.js
```diff
@@ -925,16 +927,28 @@ PhyloTree.prototype.drawBranches = function() {
...
-    .data(this.nodes) //.filter(function (d){return bLFunc(d)!=="";}))
+    .data(this.nodes.filter(function (d) { return typeof d.n.attr.clade_name !== 'undefined'; }))
     .enter()
     .append("text")
-    .text(function (d){return bLFunc(d);})
+    .style("visibility", "visible")
+    .text(function (d) { return d.n.attr.clade_name; })
     .attr("class", "branchLabel")
-    .style("text-anchor","end");
+    .style("text-anchor", "end");
```
jameshadfield: tmp // review

PR #345, opened 2017-06-28
Electron App Take 2

zachcp:
Hi Team Nextstrain,

This pull request supersedes  #331 and I believe it addresses almost all of the initial concerns. Here is brief summary of the changes:

-  Updated: this repo use an npm-based solution that uses webpack/node for all build steps.
-  Changed: all configuration is now contained within the `package.json` file and a new file `webpack.config.electron.json`. 
- Added: the `index.js` in the root directory is the script needed to run electron and launch nextstrain.
- Added: the `build` directory which contains the icon files
- Added: the output will end up in the `releases` folder
- Changed: the electron build will download data to the `data` directory and copy this into the app.
- Updated: instructions on how to build the app are  in the `Readme`.  

There are currently two outstanding issues that can be relatively easily addressed:

1. I am only downloading the ebola/zika data correctly. The other json files need to be correctly specified.
2. the iconset was generated with the background.png file.  The resolution would be improved by improved by having a proper icon set with a full set of sizes.

There are also a few config/repo issues:
1. the `webpack.config.electron.json` file is really only being used to pass in the local data parameter.  probably can be done better.
2. express inside electron is not recommended but I think its best for now. i don't know if there are any issues with ports/firewalls etc. In any event everything is working now.
3. I haven't checked network calls for css/js/google analytics. Does it make sense to pack all of this up during the build step (no internet required?)
4. directory structure - make the `build` dir as part of the build step? is output to `releases` okay?

screen shot 2017-06-27 at 10 49 38 pm

Please let me know if there are any changes I can make.

Best Regards,
zach cp


trvrb:
@zachcp ---

Sorry this has sat for two weeks. Again, I think this is awesome and I really appreciate the efforts. We've been trying to push a new feature, but I'm hoping I can dig into this later this week.

zachcp:
@trvrb ok great; sounds good

trvrb:
This really cool @zachcp. Feels nicely like a real application. I built this super easily and the existing web app does not seem to be affected. Directory naming scheme seems fine to me. And okay to add `express` and `asar ` to our dependencies. My only real question is what's going on here:

https://github.com/zachcp/auspice/blob/app2/webpack.config.prod.js#L19

I would have thought that `config.prod.js` would be fully separate from `config.electron.js`. Seems slightly amiss.

zachcp:
@trvrb fantastic!   I believe those lines are vestigial and can be removed; I'll check. I'll also rebase against master. 2 other quick Qs:

1. Do you have a vector version of the Nextstrian logo? If so, I can create an icon set.
2. Do you have a list of all of the JSON data files that are used in Nextstrain? I believe, for example, that FLU has a different pattern than the other viruses.

zcp

trvrb:
@zachcp --- Fantastic. Thank you. I don't think you need to rebase. Conflicts are extremely light.

1. Here is PDF line art for the Nextstrain logo: 

[nextstrain_logo.pdf](https://github.com/nextstrain/auspice/files/1154669/nextstrain_logo.pdf)

2. I was planning to make a simple `datasets.json` that lists these patterns. We need a central file for this for other purposes anyway. The could be used to do a programatic `fetch_data` script. I'll plan to commit this to `master` rather than here.

zachcp:
@trvrb  I removed the unused config lines. The app still builds fine. if you want to merge this the other issues - data, icons can come later.

trvrb:
Thanks @zachcp! Two things:

1. Could you actually rebase onto current `master`? I just merged locally and found that the resulting electron app built just fine and ran successfully, except that data didn't seem to be getting loaded properly. Had blank panels for tree and map. I couldn't figure out how to fix this. Hopefully you can. Very sorry to make you do more work here. Hopefully with this merged we can keep checking that electron app is working.

2. Could you drop `package-lock.json` from the merge? I'm still running npm 4. I think this is a good addition, but I'd like to separate out the functionality into a separate PR.

zachcp:
@trvrb 

I rebased/squashed then merged onto master. I, too, couldn't build but it was due to several packages which for some reason needed to be explicitly installed:

```
npm install binomial
npm install linspace
npm install outer-product
npm install react-addons-css-transition-group
npm run dist:electron
```

this works/builds for me now.

trvrb:
The rebased version just worked for me after a simple `npm install` (without the explicit installs above). Again, this is fantastic! Thanks so much for the contribution @zachcp.

zachcp:
No problem; i learned a lot and it was  a chance to poke around in this cool code base.

One other thought about packaging: if you can replace GET HTTP calls with something like `d3.json` for loading the local files, you could avoid using express locally and you wouldn't need electron (100Mb + bigruntime ) at all. You could use MacGap (~1mb, small runtime but no bells/whistles).

Issue #338, opened 2017-06-21
Branch tooltip should display nucleotide mutations


trvrb:
When "NT" is selected in the diversity panel, the branch tooltip should display nucleotide mutations instead of AA mutations.

![nucleotide-tooltip](https://user-images.githubusercontent.com/1176109/27405552-d190f71c-5686-11e7-827e-3bd1bcd1aceb.png)

Issue #335, opened 2017-06-21
Info box text


trvrb:
Current info box says "Grubaugh_et_al" rather than "Grubaugh et al":

![info-box](https://user-images.githubusercontent.com/1176109/27399822-e9062d36-5672-11e7-97de-2235737d0f69.png)

jameshadfield:
Also fixed for the filter menus. 
I'm going to fix some other issues with tree panels on the same branch.

PR #346, opened 2017-06-28
Tree info panels

### LINKED ISSUE #332: tree tip info box improvements
- [ ] date CI if applicable

jameshadfield:
This PR cleans up the tree info panels for hovering over a branch & tip as well as clicking on a tip.
Closes #332 #335 #338

BRANCH HOVER:
* Nucleotide mutations shown when selected in entropy panel (no screenshots)
image
image

TIP HOVER:
* shows less information
image

TIP SELECT:
* information hidden if not available.
* if both URL + accession, the accession is now a link
* only applicable attrs are shown (no division for flu etc)
image
image

@trvrb if this looks good please merge

### DIFF
--- a/src/components/charts/entropyInfoPanel.js
+++ b/src/components/charts/entropyInfoPanel.js
@@ -1,8 +1,8 @@
 /*eslint-env browser*/
 import React from "react";
-import * as globals from "../../util/globals";
-import {dataFont, infoPanelStyles} from "../../globalStyles";
-import { prettyString } from "../tree/tipSelectedPanel";
+// import * as globals from "../../util/globals";
+import { infoPanelStyles} from "../../globalStyles";
+// import { prettyString } from "../tree/treeViewFunctions";
 
 const InfoPanel = ({hovered, mutType}) => {
 
--- a/src/components/controls/all-filter.js
+++ b/src/components/controls/all-filter.js
@@ -1,6 +1,5 @@
 import React from "react";
 import Radium from "radium";
-import { filterAbbrFwd } from "../../util/globals";
 import ChooseFilter from "./choose-filter";
 import { connect } from "react-redux";
 
--- a/src/components/controls/recursive_filter.js
+++ b/src/components/controls/recursive_filter.js
@@ -1,11 +1,11 @@
 import React from "react";
 import Radium from "radium";
 import Select from "react-select";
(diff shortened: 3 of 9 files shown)


trvrb:
@jameshadfield --- This is fantastic. I really like the behavior where tooltip reflects current coloring. This will extend nicely when we have lots of `colorBy` options (epitope mutations, etc...).

Issues:
* Not huge, but I'm still seeing "division" in info panels for seasonal flu. However, I think it's a good thing to list "country" and "division" for flu even if we only allow "region" `colorBy` and "region" geo reconstruction.
* I'm not getting an accession link for seasonal flu. This may be on the augur end.
* I like the behavior of switching between "NT" view and "AA" view. My only concern is that this functionality is quite hidden. It's not obvious that changing Entropy affects Phylogeny display. This isn't blocking and I don't have a great solution, but worth thinking about.
* I do think that the tip tooltip should reflect both `colorBy` and `distanceMeasure ` the same way branch tooltips work. If `distanceMeasure` is date, then show date. If `distanceMeasure` is divergence, then show divergence.

rneher:
I like those tool tips as well, but the darker font is very difficult to read in my browsers (ubuntu firefox/chrome).
image

jameshadfield:
yeah... that's quite different!

rneher:
looks like its single pixel width...
image

rneher:
this is font weight 500/300 on my screen

image

jameshadfield:
@trvrb thanks! 

* point (1) - division in flu - this is in the tree file coming from augur so we should remove it from there. This PR will not display `region` / `country` / `division` in the tooltip _if_ they're not in the `attr` of the node.
* point (2) - seasonal flu missing accession - This is not exported augur. It looks sparser because this PR removes fields such as accession, URL when they're not available. Zika has everything so looks much better.
* point (3) - toggling AA/NT - I agree but let's defer.
* point (4) - tip hover now shows date (and confidence interval if applicable) for time trees and divergence otherwise

@rneher 
I've changed the fonts to 500 & 300

trvrb:
@jameshadfield ---

This is excellent. I understand better now for points (1) and (2). I think only showing what're in attrs makes complete sense and I like the spec. Thanks for adjusting (4). I just noticed that this now reports inferred dates for tips with date uncertainty (`2017-XX-XX`, etc...). This is a great feature. I'm considering this good. Thanks James!

PR #352, opened 2017-06-29
define defaultColorBy on data load. Closes #347

### LINKED ISSUE #347: color goes black when switching AA<->NT
when you color by genotype (via click on diversity panel) and then change NT <-> AA (or vice versa), everything goes black. IMO color shouldn't update in this case, and it doesn't if color is region or similar.

jameshadfield:
@rneher / @trvrb if this works for you please merge
Closes #347

### DIFF
--- a/src/actions/treeProperties.js
+++ b/src/actions/treeProperties.js
@@ -6,7 +6,6 @@ import { calcVisibility,
 import { determineColorByGenotypeType } from "../util/urlHelpers";
 import { changeColorBy } from "./colors";
 import * as types from "./types";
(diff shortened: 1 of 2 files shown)


trvrb:
This looks good to me. Behavior is good and code makes sense. Eventually I think the better spec would be switching between AA and NT would go from `colorBy` = `NS1_349` to `colorBy` = `nuc_3517` (for example). Hmm... actually. I don't see an obvious way to implement this however. And there will be codons with multiple corresponding nuc changes and nuc changes that correspond to no AA change. So, I don't have a great suggestion. I think switching back to the default makes sense.

PR #368, opened 2017-07-07
363 date window

### LINKED ISSUE #363: Date window shrinks on map play/pause
1. Load `master` and go to http://localhost:4000/zika
2. Click "Play" and immediately pause.
3. I see 2011-07-24 to 2011-11-11. This is 110 days (via Wolfram Alpha, http://www.wolframalpha.com/input/?i=2011-07-24+to+2011-11-11)
4. Click play, click pause, click play, click pause, etc... 
5. I see 2012-09-28 to 2012-12-25. This is 88 days.
6. Doing this more I get 2014-05-30 to 2014-07-28. This is 59 days.

trvrb:
Fixes #363 

This moves from days in moments to years in d3. I think much cleaner. @colinmegill, you might want to take a quick look, but seems to work well. I'll plan to merge unless there's something major wrong.

### DIFF
--- a/src/components/controls/date-range-inputs.js
+++ b/src/components/controls/date-range-inputs.js
@@ -6,6 +6,7 @@ import Slider from "./slider";
 import { connect } from "react-redux";
 import { controlsWidth } from "../../util/globals";
 import { modifyURLquery } from "../../util/urlHelpers";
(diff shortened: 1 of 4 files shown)


trvrb:
Never mind. This doesn't actually fix #363. Working more now. Hold off on review.

Issue #367, opened 2017-07-07
Animation performance


trvrb:
Check if there's low hanging fruit to improve animation performance. The 100ms tick can probably be improved for Zika. @colinmegill: How was 100ms tick chosen? Ideally, we want this to tick as fast as render works but no faster.

colinmegill:
@trvrb am going to do a branch for requestAnimationFrame now that we know we want animations with fixed times, I would expect this to be an improvement.

trvrb:
Excellent.

jameshadfield:
Cool. For what it's worth, I think the fixed times should be faster. `flu` with `slow` takes 4min to animate on my laptop!

jameshadfield:
Flu animation render stack time (i.e. the code executed during each `setInterval`) is around 560ms for me (about 1/3 of the way through the tree). Of this, `getLatLongs` is running twice and taking 260ms + 175ms. `updateStylesAndAttrs` (tree) is taking 80ms. Those 3 calls are 92% of the stack time.

trvrb:
I'm getting 220ms frames (4 FPS) for default Ebola animation on production. This looks to be about 100 ms for tree and 100 ms for map. `getLatLongs` is indeed running twice (taking 20ms each time). `updateVisibility` is a separate call taking 70ms.

![profile](https://user-images.githubusercontent.com/1176109/27980979-6d2d9cf2-6339-11e7-8915-2fe1713d4512.png)

This more what I'm expecting. 3y flu is about as big as Ebola and I'm getting 410ms frames on production. `getLatLongs` is taking 200 ms (twice). I thought that this might be because this was "country" geo resolution, but we have the same issue with "region" geo resolution when flu has 10 regions and Ebola has 50. Or I guess it has 30 given it's tripled. Okay. This makes sense. If flu is not tripled then performance is similar to Ebola with `getLatLongs` taking 20ms instead of 200 ms. Tripling demes results in 9x more lines. So 20ms vs 200ms makes sense.

Moving forward I see two obvious ideas:

* Don't fully triple map. We need a "west" copy of Japan, but not a "west" copy of Europe. Need a half copy on the left and a half copy on the right giving 2x demes. I don't know how difficult this would be to implement.
* Avoid the duplicate `geLatLongs` call.

There may be internal logic that can get cleaned up as well.

trvrb:
[...]

PR #377, opened 2017-07-08
Animation refactor

trvrb:
Progress on #367 

Before this branch (current `master`) if getting the following performance for Ebola:

* ~230ms animation frames
* ~20ms `getLatLongs`

With this branch in place I'm getting:

* ~190ms animation frames
* ~10ms `getLatLongs`
* ~50ms `updateVisibility`

Before this branch (current `master`) if getting the following performance for 3y flu region resolution:

* ~400ms animation frames
* ~45ms `getLatLongs`

With this branch in place I'm getting:

* ~270ms animation frames
* ~15ms `getLatLongs`
* ~150ms `updateVisibility`

_Going to keep updating these numbers as I work on the PR. Serving to help me remember if there's performance increases._


trvrb:
Okay. I think I'm done with this for the moment. @colinmegill: I tried to clean up code where possible, things like:

* replacing things like `long0` with `longOrig`, `long1` with `longDest` and `winningPair` with `closestPair` to make self documenting
* don't pass around high-level redux object `controls` to things like `updateVisibility` that have to be reached into and instead pass in low-level `numDates`.
* carve out functions for `leafletLatLongToLayerPoint` and `maybeGetTransmissionPair` to reduce code duplication

Biggest functional change is duplicating map with adding a half map to the west and a half map to the east, rather than the full triplicating. This improves performance and provides a more natural set of bounds.

This is just FYI for @colinmegill, but @jameshadfield could you do a quick review on Monday and merge if it looks good.

jameshadfield:
Looks good to me - i've made some comments in issue #367

Will let @colinmegill take a look, but I plan to merge later this afternoon (since animation is currently disabled in master anyway)

colinmegill:
Ok, will take a look now!

trvrb:
Great. Thanks both.

src/components/map/map.js
```diff
@@ -488,38 +497,6 @@ class Map extends React.Component {
       }
     }, animationTick);
 
-    // controls: state.controls,
```
colinmegill: This comment can stay, I will clean up on the requestAnimationFrame pass

src/components/map/map.js
```diff
@@ -25,18 +24,16 @@ import {
     // datasetGuid: state.tree.datasetGuid,
     treeVersion: state.tree.version,
     treeLoaded: state.tree.loaded,
-    controls: state.controls,
+    splitTreeAndMap: state.controls.splitTreeAndMap,
```
colinmegill: This is way cleaner

src/util/mapHelpersLatLong.js
```diff
@@ -94,134 +122,65 @@ export const getLatLongs = (nodes, visibility, metadata, map, colorBy, geoResolu
...
-        )
-      });
+      let lat = geo[geoResolution][key].latitude;
+      let long = geo[geoResolution][key].longitude + OFFSET;
+      if (long > westBound && long < eastBound) {
+        demesAndTransmissions.demes.push({
+          location: key, // Thailand:
+          total: value.length, // 20, this is an array of all demes of a certain type
+          color: averageColors(value),
+          coords: leafletLatLongToLayerPoint(lat, long, map)
```
colinmegill: This (apparently, but I could be wrong here) no longer relies on `new L.LatLng`, is that intentional?

colinmegill:
All set - this is looking great

trvrb:
These changes look good James.

src/util/mapHelpersLatLong.js
trvrb: I did this to reduce code duplication. `leafletLatLongToLayerPoint` should be handing back a `new L.LatLng(lat, long)` in the exact same fashion as before, just shelved out into a function.

PR #379, opened 2017-07-09
Include Bezier comments

trvrb:


trvrb:
Thanks for this @evogytis!

PR #394, opened 2017-07-11
animation changes URL date on pause / complete / reset. Closes #372

### LINKED ISSUE #372: Date URLs don't work with animation
Currently, running animation updates redux `minDate` and `maxDate` but does not push these updates to the URL. I think this is the correct thing to do (would look string to have the URL rapidly ticking). However, this levels a discordance when the animation is paused. Like so:

url

I think the proper spec is to push redux state to URL on animation stop.

jameshadfield:


trvrb:
This is excellent. Thanks James!

PR #395, opened 2017-07-11
Filters Header

jameshadfield:
This PR contains the prototype for a filters header. 

Styling and functionality needs improving, and there are lots of edge cases regarding filters!

~Hopefully someone can take this over while i'm gone.~ 

image


trvrb:
This is a great start. I actually didn't even think to borrow the styling from the styling we're using in the tree panel. Seeing it this way made me think that the proper place to put filter info is above the tree panel like you have here for author.

filters

After all the clade filter affects the map as well. I think we need to sit down and mock this up together. I'd really like to figure out a systematic design for these sorts of elements. This affects citations as well. Should these be the same card format? The white text on gray background achieves the effect of a temporary element that can be dismissed. However, I could also imagine a permanent info box at the top initially listing total sequence count that gets elaborated upon by filters.

jameshadfield:
Agree that this requires some thinking & sketches. Getting author attribution right needs to be done perfectly - which involves some fauna / augur work also.

jameshadfield:
Closing and removing branch. Will apply changes onto branch `data-title`

PR #396, opened 2017-07-12
upgrate react packages

jameshadfield:
This PR upgrades react from `15.3.1` to `15.6.1`. This allows performance tools to be used, which will allow us to quantitatively measure our efforts in #367 (and elsewhere).

https://github.com/facebook/react/releases


trvrb:
Everything seems to check out. Tested locally and tested on dev.

PR #398, opened 2017-07-13
Set absoluteDateMax from tree; closes #364

### LINKED ISSUE #364: Ebola date range
Ebola has absolute date range of 2014/03/04 to today. This makes it so half the date slider is non-active. Should fix to have absolute max date equal to the final sample. I don't know how much of this is augur vs how much is auspice.

This impacts animation in that a bunch of animation happens with a blank display.

sidneymbell:


trvrb:
This is great. Thanks Sidney.

PR #399, opened 2017-07-13
Fix control display for "divergence". Closes #382.

### LINKED ISSUE #382: Dengue is silently defaulting to divergence
Switching to any of the dengue datasets displays `m=div` instead of `m=num_date`. This is an okay design decision. (I think it's pretty equivalent for these datasets) However, on initial load the "DIVERGENCE" button is not active. It should be blue.

sidneymbell:


trvrb:
Thanks for this Sidney. However, running from `divergenceDefaultDisplay` I'm still getting issue #382. "TIME" starts out highlighted and then when the JSONs load, it deselects, but "DIVERGENCE" is not highlighted.

dengue

sidneymbell:
Thanks, Trevor. Sorry, I totally missed this somehow. It's fixed now.

trvrb:
This was superseded by PR #416.

Issue #381, opened 2017-07-10
improve bezier speed


jameshadfield:
Related to #367 but better served as its own issue

**Current Data Flow**

each tick calls `updateVisibility`, which has `D3` data binds to the demes and transmissions. This data bind is for all transmissions, whether visible or not (at time of writing, n=147 for zika). The bezier calculation (see below) is calculated for all transmissions, and SVG elements created (and modified at each tick) regardless of whether visible or not. The stroke-width attr effectively hides those which are not visible, but they're always calculated and SVG elements exist.

**Current Bezier calculation**

The Bernstein polynomial is calculated for each transmission, for each tick. 15 points are then calculated for the fraction of the path present in the current time slice (negative fractions are mathematically sound and are currently passed through for curves outside of the time slice). 

**Steps to make things better**

* For each transmission`A-B`, pre-calculate a single Bernstein polynomial (or `n` if there are `n` separate transmission events from `A-B` that we don't want over top of each other)
* Only run D3 data binds for visible transmissions (requires using a key to keep data binds in sync)
* Either
  * take the pre-computed polynomial and make a curve of 15 points representing the current time slice (15 shouldn't be hardcoded, but related to the length of the visible path)
  * (faster, my preferred option) pre-compute (when we make the polynomial) a curve of (say) 5-150 points (depending on length), then at each tick, slice this array of points so that it's relevant to the time window and draw this.

**Potential Increases**

Vast. E.g. current `zika` @ medium speed takes around 300 ticks (flu will take way more, so will zika as we move to animation frames)
* Bernstein polynomial: 45000 calculations can be reduced to 150.
* Curve interpolation: 45000 calculations can be reduced to 150 precomputed arrays, with (computationally cheap) array slices at each tick
* SVG elements: number in the DOM will be number visible, not all 150
* SVG updates only on visible elements, not all 150

jameshadfield:
cc @colinmegill @trvrb

colinmegill:
Fully support, though no opinion on which direction. Seems like this will be a dramatic speedup for the bezier side.

jameshadfield:
I'm strongly voting for precomputing the curves array and taking slices during each tick. We have to optimise a lot in order to get map animation ticks fast enough and this is an easy gain.

trvrb:
I agree about precomputing curves and slicing.

[...]

PR #402, opened 2017-07-18
Precalculate deme and transmission data

### LINKED ISSUE #392: precalculate LatLongs
(as mentioned in https://github.com/nextstrain/auspice/issues/367#issuecomment-313839395)

The thing that takes up all the time is this loop:

https://github.com/nextstrain/auspice/blob/master/src/util/mapHelpersLatLong.js#L130

For each of 3 offsets x each of ~500 transmission events x 3 offsets = ~4500 calculations each cycle for flu. These are basically the same calculations being done each cycle. The output is just for each transmission (that has an origin and a destination) compute the lat/long of the origin and the lat/long of the destination. offsets (due to map tripling) result in duplicating transmissions at different lat/longs.

However, this information only needs to be computed on load. We have a list of all transmissions in the tree (regardless of visibility) for the specified geoResolution. Can pre-compute this transmission list (with lat/longs) on load from the full tree. Store this in map state. Then each render cycle, just find the visible transmissions in this list. Will be immensely faster.

trvrb:
Resolves #392 

This creates `demeData` and `transmissionData` in map state. These are flat objects listing every deme (there may be duplicates due to west and east triplication) and every transmission event. These are constructed on first load and then only updated on update operations (change color, change date slider, etc...).

Performance is roughly comparable at the moment. I'm getting 8ms for `updateDemeAndTransmissionData` for Ebola compared to 18ms on `master`. Large additional speed improvements could be gained from switching from arrays to dictionaries. However, this will complicate things I think best left until after Bezier (#381).

This is mostly working, but there's more to do here.

- [x] Change dataset needs to be fixed
- [x] Change geo resolution needs to be fixed


trvrb:
With commit 515e78a8cf94ba8fac6f29caf86f5de19e512a21, Ebola is now running `updateDemeAndTransmissionData` in 3ms.

trvrb:
I think this is now good to go. @colinmegill could I request a review here? Please merge if you're satisfied or go ahead and make fixes. I'm not going to have much more a chance to work on this in the next few days.

colinmegill:
Read through all of this, this looks really great. Merging now. Exciting to have smarter updates here.

PR #406, opened 2017-07-22
Precompute bezier

### LINKED ISSUE #381: improve bezier speed

colinmegill:
Closes #381


colinmegill:
Moving discussion over here - @jameshadfield @trvrb if we do 5-150 points (based on length), the idea is then to not compute the actual position of the Bezier but just find the closest point, is that correct? It seems like `computeResponsive` might also be an input to this - the animation will be jumpy the larger the window, because the pixel space will be wider (could be wrong).

trvrb:
@colinmegill --- This is a really great start. I went through and I like how this is being implemented. In terms of how to slice the Bezier curve with date slider, I bet we could error on the side of having lots of points along the line. Just have a function for how many to make. It shouldn't be expensive to store these. The most expensive bit will be the SVG manipulation if we have extra points that are not really necessary for resolution.

For performance, I would recommend storing these points as something like

`BezierPoints`: `[{x1,y1}, {x2,y2}, .... , {xn, yn}]` and `BezierDates`: `[d1, d2, ...., dn]`

You essentially want a fast way to look up the points to use. In this case, you could compare `minDate` and `maxDate` to `BezierDates` and get a `minIndex` and a `maxIndex` you can then slice `BezierPoints` appropriately. Or rather update visibility of line segments appropriately.

With this in place, we can get creative how the length of `BezierPoints`.

colinmegill:
Wonderful! Was thinking along similar lines.

trvrb:
@colinmegill ---

There are some remaining issues here that I'm going to leave to you:

- [ ] Filtering currently doesn't work. Beziers stick around when they shouldn't. (Go to Zika, filter to "Jamaica")
- [ ] Flu map is broken. (Go to flu, doesn't load)
- [ ] "Inchworm" effect on missiles. (Go to Zika, click animation play, watch the initial transmission from Vietnam to French Polynesia, it contracts and expands oddly --- this may be due to Bezier curve length or something else)

jameshadfield:
Just a quick note on the flu thing (i haven't looked at why it's broken) - I foresee plenty of cases where a map is present with demes and no transmissions, so it'd be great if the `map` component handled these cases really well.

trvrb:
>Just a quick note on the flu thing (i haven't looked at why it's broken) - I foresee plenty of cases where a map is present with demes and no transmissions, so it'd be great if the map component handled these cases really well.

Agreed. Working for H7N9 to just show demes (circles). Don't know what's going on with seasonal flu.

>"Inchworm" effect on missiles.

I thought of (I believe) a really good way of handling this via a weighted linear interpolation of x,y coordinates. I'll try to implement.

trvrb:
db2d7e6837b64c444e63983230f7f016b062f3e8 fixes the flu issue. This was caused by flu lacking lat/longs for certain demes in `meta.json`. This _shouldn't_ have happened in augur export, but we still need to have auspice robust to this sort of thing.

trvrb:
332668bb9250135e1a3d18bc7e61c0bb58ef5c59 fixes the inchworm issue. Interpolated start and end points give a fully continuous animation

trvrb:
faf9fe0ef23df8bc2c1db967375a5044194dedc5 fixes the filtering issue. Filtering hadn't been activated. Now only draws "visible" transmissions.

trvrb:
@jameshadfield: Could you review this please? I believe that everything is working as intended and that there are no regressions, but it would be good to have a second set of eyes.

cc @colinmegill

trvrb:
f578936e6f60152017b1bfebffe6df396645e764 removes `minTransmissionDate`. This originally had been necessary when bending Beziers according to date. This is no longer planned for and `minTransmissionDate` is completely extraneous. When implemented, "threads" of multiple transmissions between A and B won't need to refer to this either.

src/util/mapHelpers.js
```diff
@@ -1,81 +1,107 @@
...
+  const startIndex = _.findIndex(bezierDates, function (d) { return d > numDateMin });
+
+  // find end
+  const endIndex = _.findLastIndex(bezierDates, function (d) { return d < numDateMax });
+
+  // startIndex and endIndex is -1 if not found
+  // this indicates a slice of time that lies outside the bounds of BCurve
+  // return empty array
+  if (startIndex == -1 || endIndex == -1) {
+    return [];
```
jameshadfield: Is this right? What if part of the transmission were in the time slice, yet the max date is after the transmission? (or vice versa)

jameshadfield:
Also, how are lodash methods (`_.something`) and `d3` running in `mapHelpersLatLong.js` functions when the libraries aren't imported in that file?!?

colinmegill:
@jameshadfield good question. It looks like we have them on window, though that does not appear to be coming from index.html script tags:

image

src/util/mapHelpers.js
trvrb: Hmm... my console logs when constructing this seemed correct but now I'm not sure. Need to diagram this out.

trvrb:
@jameshadfield: 3c00b465ae90a35f59ef32885a2ce352db2faf97 explains logic for the different scenarios of `numDateMin` and `numDateMax` relative to transmission bounds. You'll see that we get `startIndex` _or_ `endIndex` of `-1` only when the entire bounds don't overlap. Good to document though.

Going to merge now.

PR #413, opened 2017-07-28
use react performance tools for animation ticks (dev only)

jameshadfield:
This adds code which will console log the time spend in different components for one animation frame and cumulative for 10 once the play button is clicked. Resets when the animation resets.

Only loaded on the dev version.

@colinmegill please check you are happy with this code being in master.


trvrb:
@jameshadfield ---

I think this is an excellent addition. However, I'm not sure I agree with this always displaying on map animation when in dev mode. There will be times when we're debugging map etc... and don't want this. Why not just add a simple `enableAnimationPerfTesting` flag to `global.js`? I like how this can never be on when `NODE_ENV == "production"`.

jameshadfield:
@trvrb think this is good to go now

trvrb:
Thanks James!

PR #422, opened 2017-08-10
fix CI bug. Closes #337

### LINKED ISSUE #337: Switching tree layouts triggers confidences
On live site if I go to Zika:

http://www.nextstrain.org/zika

then click "Radial" then click "Rectangular", I get confidence intervals on all nodes despite "Show confidence intervals" being toggled off.

jameshadfield:
See #337 

I'm happy with the code & i've tested by changing the layouts with/without CIs, changing to divergence etc etc and I believe it to be working correctly, but could do with @trvrb confirming this


trvrb:
Fantastic to have this fixed. Thanks so much for taking of this @jameshadfield.

Examined code and behavior. Wasn't able to break it. Nice work.

PR #424, opened 2017-08-16
move to air-bnb eslint config + some custom (sane) rules

### LINKED ISSUE #419: update eslint rules
I'd like to update these and remove a bunch

While we're at it - simple poll - who is using eslint and do you have strong opinions on it? (you can edit this post)
@huddlej 
@sidneymbell 
@colinmegill 
@trvrb --- I'm not using it, but would like to. I think I need a tutorial.
@rneher

jameshadfield:
This PR moves to an updated set of eslint configs, which should help improve react code.

I've added some rules to `.eslintrc`, mainly to disable stylistic linting errors which are not impeding code understanding - e.g. forcing `{ a: 1 }` over `{a: 1}`. Please disable more rules here as you see fit - I believe the aim of linting should be to highlight errors rather than enforce unnecessary stylistic consistency. 

Note that this PR introduces further linting errors to the code, but it's better to change this now. 

This PR would close #419


trvrb:
@jameshadfield: Shouldn't these have been added to `devDependencies` and not `dependencies`? We now have duplication of `eslint` between these.

jameshadfield:
moved to `devDependencies` in #426

PR #426, opened 2017-08-18
Update linting rules and move to devDependencies

trvrb:
All the linting modules should be in `devDependencies` and not in `dependencies`. This was unnecessarily increasing size of webpack.

I just pushed to dev to make sure this didn't break anything and it looks okay.


trvrb:
I added `padded-blocks: off` because this was failing and I thought a space made things much more readable:

```
const extractLineSegmentForAnimationEffect = (
  numDateMin,
  numDateMax,
  originCoords,
  destinationCoords,
  originNumDate,
  destinationNumDate,
  visible,
  bezierCurve,
  bezierDates
) => {

  if (visible === "hidden") {
    return [];
  }

  // want to slice out all points that lie between numDateMin and numDateMax
  // and append interpolated start and end points
```

I added `arrow-body-style: off` because I this was failing and I thought it a natural phrasing:

```
export const pathStringGenerator = d3.svg.line()
  .x((d) => { return d.x; })
  .y((d) => { return d.y; })
  .interpolate("basis");
```

Lines with `.x` and `.y` were causing the issue. This seemed like a natural way to use D3 method chaining.

trvrb:
The use of `one-var: off` and `one-var-declaration-per-line: off` exist to allow

```
  let latOrig, longOrig, latDest, longDest;
  let transmission;
```

This seemed like the most readable way to declare these variables.

The use of `no-console: [1, { "allow": ["warn", "error"] }]` allows `console.warn` and `console.error` but not `console.log`. I think there are important warnings we'd like to include in production code like:

```
console.warn("Warning: Lat/long missing from metadata for", key);
```

trvrb:
The use of `no-param-reassign: [1, { "props": false }]` was to handle another common pattern:

```
_.map(demeData, (d) => {
    d.coords = leafletLatLongToLayerPoint(d.latitude, d.longitude, map);
  });
```

trvrb:
Also, I'm getting a lot of things like `visibilityVersion is missing in props validation`. This goes to https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md.

Do we want to enforce props validation or ignore the linting?

trvrb:
This completes my linting pass on Map code. The only major linting rules I'm not sure about are:

```
react/prop-types
react/sort-comp
```

James, please review and merge. I figure we can keep iterating on linting rules in small increments.

trvrb:
Actually, the majority of our React components are flagrantly breaking the `react/prop-types` and
`react/sort-comp` rules. Decided that for now, at least, better to remove these from the linting spec.

jameshadfield:
this all looks great - hopefully everyone feels free to just add to `.eslintrc` as we need to

PR #427, opened 2017-08-22
Download & Upload Data Framework, Authors in Footers are Filters

### LINKED ISSUE #296: export tree svg and newick/nexus
It would be nice to be able to export the tree for example as newick or nexus prepared by augur, or as svg with the current coloring. A derivative of the old auspice did this as 
```
	function exportTreeSVG(){
		var tmp = document.getElementById("treeplot-container");
		var svg_tmp = tmp.getElementsByTagName("svg")[0];
		// Extract the data as SVG text string
		var svg_xml = (new XMLSerializer).serializeToString(svg_tmp);
		var blob = new Blob([svg_xml.replace("cursor: pointer;", "")], {type: "text/plain;charset=utf-8"});
		saveAs(blob,'tree.svg');
	}
	d3.select("#svgexport")
		.on("click", exportTreeSVG);
```
this worked pretty well but for fonts

### LINKED ISSUE #421: drag & drop CSV colorBy information
This is a feature proposal to allow (client side) addition of `colorBy` data to a dataset by means of dragging and dropping a CSV file with a column of `strain` (matching the taxa in the tree) and additional columns processed as `colorBy` variables. Ancestral state reconstruction is not performed. 

This feature would work well with the ability to download line lists. It is a mechanism by which "private" data can be added to public analyses.

jameshadfield:
This PR has 2 main effects: 

(1) it merges in the machinery for file download / upload, and disables the former via the flag `enableDownloadModal` in `util/globals.js`. This allows remaining bugs to be fixed as separate issues and reduces future merge conflicts. 

(2) The authors in the footer now act like filters - and this is in-sync with the authors drop down in the sidebar. 

Salient points:
* I've removed the display of authors in the footer for seasonal flu, as they are not exported from augur (check the live site!)
* Dragging CSV files onto the browser is not disabled by `enableDownloadModal`

This PR closes #421 
This PR closes #296


package.json
```diff
@@ -98,13 +98,21 @@
     "d3-shape": "^1.2.0",
     "d3-time-format": "^2.0.5",
     "d3-zoom": "^1.1.3",
+    "eslint": "^4.4.1",
```
trvrb: Careful with this. `eslint` should only be in `devDependencies`.

trvrb:
@jameshadfield --- This is great! Thank you. I didn't review UX of the not enabled changes. Just looked for bugs relative to `master`. Will merge now after cleaning merge conflicts.

PR #428, opened 2017-08-24
Move to d3 v4

trvrb:
This PR completes a move to D3 v4. Rather than importing the entirety of D3, this is importing individual modules like so:

```
import { event } from "d3-selection";
import { min, max, sum } from "d3-array";
import { scaleLinear } from "d3-scale";
```

This will keep bundle size down. Almost all the changes were incredibly small. More major refactors were:

1. The new `d3.tree` that replaces `d3.layout.tree` in v4 no longer provides a `.nodes()` function. We were using this in several places to flatten a tree into a node list. I replaced this with a `flattenTree()` function in `treeHelpers.js`. This should have the added benefit of speeding up code. `d3.layout.tree` was doing a bunch of under-the-hood layout calculations that we weren't using at all. The new `flattenTree()` is much more streamlined.

2. The new `d3-time-format` from v4 no longer provides a `.parse` function on the `d3.time.format` object and instead uses a separate `timeParse` object. This necessitated splitting the previous `state.controls.dateFormat` into `state.controls.dateFormatter` and `state.controls.dateParser`.

3. The d3 exit / enter / append methods used in `phyloTree` to attach and update grid lines were not to spec (but working somehow in v3). I modified this to work in v4.

Also, removes some unused `dependencies` in `package.json`. You'll obviously need to `npm install` after merge. With this PR `bundle.js` is 1.4M and `bundle.js.gz` is 383K. Prior to this PR `bundle.js` was 1.6M and `bundle.js.gz` was 434K.


rneher:
I am not fully sure what the `merge` does, but if that is what it takes, fine.

trvrb:
I got this bit of code from here: https://github.com/d3/d3-selection#selection_merge. Actually a bit hazy on why this was truly necessary, but it fit their mold.

trvrb:
Thanks Richard and thanks James. I'm going to merge this now.

PR #433, opened 2017-08-29
Improve date conversion code and remove moment.js

trvrb:
This PR mainly cleans up date conversion code. It does make one user facing change.

1. Replace all logic related to moment.js with d3 date logic. There was never any reason to have both. I much favor D3 here, as it is actually accurate in its conversion between numDates and calDates (tested and D3 can go back and forth repeatedly, while moment does not). With moment.js removed from codebase, bundle size drops from 351kb to 308kb.

2. I realized I was being stupid with my initial implementation of `numericToCalendar` and `calendarToNumeric` in `util/dateHelpers`. These need a `dateScale` and either a `dateFormatter` or a `dateParser`. I had been keeping these in the redux store. However, they never change and it's so much nicer to just declare them in `dateHelpers`: https://github.com/nextstrain/auspice/compare/replace-moment?expand=1#diff-d7e8181afaf0f5f903ec1d7d26f9dbd7R4. This makes calls to `numericToCalendar` and `calendarToNumeric` much cleaner: Before: `calendarToNumeric(nextProps.dateParser, nextProps.dateScale, nextProps.dateMin)` and after: `calendarToNumeric(nextProps.dateMin)`.

3. User facing change is the removal of [react-datepicker](https://github.com/Hacker0x01/react-datepicker). This had relied on a moment date being supplied to it / returned from it. This component was used to click on a start or end date below the date slider and adjust date using a calendar UI. I never liked this as it was tedious, requiring 36 clicks to adjust a date by 3 years. There is the ability to type, but it behaves funnily and doesn't allow right arrow for example. I would vastly favor a simple input box that is validated. The current PR just replaces this with text, which I believe is still an improvement for the time being.

@jameshadfield --- Could you take a look at this?


trvrb:
Thanks for reviewing this James.

PR #437, opened 2017-08-31
300 entropy

jameshadfield:
This PR tidies up a number of issues with the entropy panel, listed in issue #300 

CUrrently at http://auspice-dev.herokuapp.com/

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,7 +22,6 @@ rules:
   no-param-reassign: [1, { "props": false }]
   no-unused-expressions: ['error', {"allowTernary": true }]
   no-restricted-syntax: ['error', 'ForInStatement', 'LabeledStatement', 'WithStatement'] # allow ForOfStatement
-  array-callback-return: off # we commonly mutate in place in array map statements
   react/jsx-tag-spacing: ["error", { "closingSlash": "never", "beforeSelfClosing": "never", "afterOpening": "never" }]
   react/jsx-filename-extension: [1, { "extensions": [".js", ".jsx"] }]
   react/jsx-first-prop-new-line: off
--- a/.eslintrc
+++ b/.eslintrc
@@ -20,6 +20,7 @@ rules:
   one-var-declaration-per-line: off
   no-console: [1, { "allow": ["warn", "error"] }]
   no-param-reassign: [1, { "props": false }]
+  no-unused-expressions: ['error', {"allowTernary": true }]
   no-restricted-syntax: ['error', 'ForInStatement', 'LabeledStatement', 'WithStatement'] # allow ForOfStatement
   array-callback-return: off # we commonly mutate in place in array map statements
   react/jsx-tag-spacing: ["error", { "closingSlash": "never", "beforeSelfClosing": "never", "afterOpening": "never" }]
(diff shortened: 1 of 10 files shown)


.eslintrc
```diff
@@ -20,6 +20,9 @@ rules:
   one-var-declaration-per-line: off
   no-console: [1, { "allow": ["warn", "error"] }]
   no-param-reassign: [1, { "props": false }]
+  no-unused-expressions: ['error', {"allowTernary": true }]
+  no-restricted-syntax: ['error', 'ForInStatement', 'LabeledStatement', 'WithStatement'] # allow ForOfStatement
+  array-callback-return: off # we commonly mutate in place in array map statements
```
trvrb: I really liked this actually as this is a simple fix of `map` -> `forEach` for proper JS. I'd recommend to keep this rule.

src/components/charts/entropy.js
```diff
@@ -8,30 +8,33 @@ import { changeColorBy } from "../../actions/colors";
...
 import EntropyChart from "./entropyD3";
 import InfoPanel from "./entropyInfoPanel";
-import { changeMutType } from "../../actions/treeProperties";
+import { TOGGLE_MUT_TYPE } from "../../actions/types";
+import { analyticsControlsEvent } from "../../util/googleAnalytics";
+import { modifyURLquery } from "../../util/urlHelpers";
 import "../../css/entropy.css";
 
-const calcEntropy = function (entropy) {
+const calcEntropy = function calcEntropy(entropy) {
```
trvrb: This can also be linted as:

```
const calcEntropy = (entropy) => {
```

which I think is cleaner.

trvrb: Although I see now that below you used

```
export const determineColorByGenotypeType = (colorBy) => {
```

So maybe you had a reason for the `const funcName = function (funcName) {` style?

trvrb:
@jameshadfield ---

Looks great overall. I was able to find a further edge case:

* Go to Zika: http://localhost:4000/zika
* Click "NT" and click on a bar. Tree, etc... now is colored and bar now is colored.
* URL is something appropriate like http://localhost:4000/zika?c=gt-nuc_7500
* Try to load from this URL and you get the tree and map properly colored but the entropy bar is not colored.
* Entropy bar is properly colored when loading from AA, like so: http://localhost:4000/zika?c=gt-NS3_572

.eslintrc
jameshadfield: fine with me - we can explicitly return `undefined` in these cases

src/components/charts/entropy.js
jameshadfield: not here I didn't - it could be a fat arrow function.
There are cases, however, when you can't use it, namely if the function body uses `this` as the binding is different between `function () {}` and `() => {}`

jameshadfield:
^^ nice find

src/components/charts/entropy.js
trvrb: Cool. I didn't know about the `this` difference.

.eslintrc
trvrb: I was meaning that if you're not returning something to build a new array you shouldn't be using `map`. You should be using `forEach` instead. This is just a simple word swap.

jameshadfield:
All bugs fixed - i'll fix the merge conflicts now and merge :)

trvrb:
Perfect. Thanks James.

Issue #209, opened 2017-02-25
Keep titlebar visible


trvrb:
We should experiment with keeping the titlebar always visible at the top of the window. Should help to keep oriented in the app.

trvrb:
This was fixed with the move to the minified titlebar.

trvrb:
As discussed in #441 ----

Commit bf7d040d60ed8b2c1c7525204abac2e0c92422cf experiments with a fixed titlebar per issue #209, like so:

<img width="713" alt="fixed-titlebar" src="https://user-images.githubusercontent.com/1176109/30067799-b0cdd78c-9210-11e7-9b6d-6a0f19fd43e1.png">

I decided that I very much preferred the white background titlebar and just making the sidebar toggle fixed. But please speak up if you'd prefer the fixed titlebar. (I tried fixed titlebar and white and it really didn't work):

<img width="808" alt="white-background" src="https://user-images.githubusercontent.com/1176109/30067742-7ab86626-9210-11e7-8f74-b4b4c41b0f05.png">

This makes me close to convinced that we don't want a fixed visible sidebar. I'm going to close this issue for now. Can be reopened if necessary.

PR #441, opened 2017-09-05
Styles update

### LINKED ISSUE #274: Sidebar collapse arrow gets smooth transition rather than snapto

trvrb:
This PR addresses several smaller styles issues. Final look is very similar to what we have now:

final-look

1. It darkens the background very slightly and adds a more even drop shadow to cards. I definitely prefer this. 

*Before*:

before-1

*After*:

after-1

2. It fixes the persistent issue with cards being slightly too short on their right-hand sides. Gutters are now equal to left and right of cards and stay the same width after opening and closing sidebar. **Ideally, this needs to be tested on Windows and Linux, Max scrollbars are weird.**

3. The sidebar toggle is now properly affixed to top of screen, so that when sidebar is scrolled it stays put.

4. The sidebar toggle is now place _over_ the sidebar scrollbar. There had been a persistent issue where you couldn't click the toggle after scrolling.

5. The sidebar toggle now properly animates with the rest of the sidebar when opening and closing.

6. Sidebar behavior is improved in mobile so that cards slide left and right rather than compressing to a thin strip. It feels much better in practice.

----------------------------

Note that commit bf7d040d60ed8b2c1c7525204abac2e0c92422cf experiments with a fixed titlebar per issue #209, like so:

fixed-titlebar

I decided that I very much preferred the white background titlebar and just making the sidebar toggle fixed. But please speak up if you'd prefer the fixed titlebar. (I tried fixed titlebar and white and it really didn't work):

white-background

----------------------------

Resolves #274

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -31,6 +31,7 @@ rules:
   react/jsx-first-prop-new-line: off
   react/prop-types: off # possibly reinstate
   react/sort-comp: off # possibly reinstate
+  jsx-a11y/no-static-element-interactions: off
 parserOptions:
   ecmaVersion: 6
   sourceType: module
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -72,27 +72,22 @@ class App extends React.Component {
     return (
       <g>
         <DownloadModal/>
+        {this.state.sidebarOpen || this.state.sidebarDocked ?
+          <TitleBar minified/> :
+          <TitleBar minfied absent/>}
+        <ToggleSidebarTab
+          open={this.state.sidebarDocked}
+          handler={() => {this.setState({sidebarDocked: !this.state.sidebarDocked});}}
+        />
         <Sidebar
           sidebar={
-            <div>
-              <TitleBar minified={true}/>
-              <Controls/>
-              <ToggleSidebarTab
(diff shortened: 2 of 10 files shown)


colinmegill:
💯 - pronounced cards are better, and also prefer white background at the
top.

trvrb:
I'm getting the same scrollbar behavior on Mac with "Show scrollbars" set to "Always" on Mac as I am with PC using both Chrome and Edge browsers. This is slightly non-ideal (in that the right hand edge of card is slightly cut-off). Attempting a fix.

rneher:
this looks good on ubuntu firefox. Definitely improved!

trvrb:
So, I finally figured out how to dynamically determine scrollbar widths:

https://github.com/nextstrain/auspice/commit/6e2176491877ce4cada64f9d2774469b027a1a1a#diff-c351b41538f669c947418a391312c738R43

I tested and this works beautifully (giving consistent gutters) in Mac OS Chrome, Firefox and Safari (with and without always visible scrollbars). It also works perfectly on PC Chrome and Edge browsers. Also works on iOS mobile. I'm going to declare this resolved.

Merging now.

PR #442, opened 2017-09-06
Move leaflet to npm and fix mobile panning

trvrb:
This PR does two things:

1. Moves leaflet to npm (it had been in `util/leaflet.js`) and removes the unused `util/leaflet-plugins`.

2. Fixes a persistent issue where on mobile you would get trapped with the map taking up the screen and no way to scroll past it. This makes it so that a single finger drag scrolls the entire page, while two fingers will pan / zoom. This is the same behavior as the tree.

I don't think this needs much of a review, but if you want to take a look @colinmegill, go ahead.


trvrb:
Pushed to dev and confirmed that scrolling past map now works properly on iOS. Assuming this is the case with Android, but let me know if it's broken.

PR #444, opened 2017-09-06
Upgrade webpack to v3

trvrb:
This PR upgrades webpack to v3.5.5. Final `bundle.js.gz` sizes are almost identical those under v1. There were theoretical tree shaking improvements, etc..., but we may not be taking advantage of them with the current config. There are also theoretical performance improvements. I could believe this, but not hugely dramatic.

Still, I see no reason not to stay current. I believe everything is still working just fine, including the electron build.

@jameshadfield --- could you take a look here? Especially at the `webpack.config` files to double-check I'm not missing anything?


jameshadfield:
Will look this afternoon. 
Similar to https://github.com/nextstrain/auspice/commit/f74db1125b8bc72ac4ccc971d5bb0756decf6770 - what are your thoughts on babel 6 as well?

trvrb:
I think it would be great if you want to add the Babel upgrade to this branch.

jameshadfield:
All working for me. Adding Babel 6 now.

jameshadfield:
working and tested via dev-server and build+server. Merging. Will require `npm i`.

trvrb:
Nice! `redbox-react` v1.5.0 wasn't giving you problems? It was for me here: https://github.com/nextstrain/auspice/pull/445

jameshadfield:
No problems - what were you getting? It builds and runs fine for me. FYI I also upgraded things like `react-transform-catch-errors`

trvrb:
I was getting: `Uncaught Error: imports[1] for react-transform-catch-errors does not look like a React component.` after upgrading `redbox-react`.

https://stackoverflow.com/questions/37895944/react-transform-catch-errors-does-not-look-like-a-react-component

jameshadfield:
Do you still get this on `master` after the merge?

trvrb:
Just checked, and no. Not getting this anymore. That's great.

Issue #98, opened 2016-12-30
Optionally disable switch to 2 columns 


colinmegill:

trvrb:
Here is the spec we just came up for this:

![panel](https://user-images.githubusercontent.com/1176109/28950532-7b89b326-7879-11e7-95b8-4cfe0619e96a.jpg)

colinmegill:
v1

![image](https://user-images.githubusercontent.com/1770265/28954698-ca902aa6-7895-11e7-9245-451b881c6fe2.png)

trvrb:
I just made some tweaks to this in eeada1fb12d9cc20a37f3a9a0810405db2289347. I really like the basic idea. Works well even on laptop screen.

<img width="1366" alt="layout" src="https://user-images.githubusercontent.com/1176109/28956217-60ff629a-78a0-11e7-9b4e-4de989630868.png">

colinmegill:
Yes, that's much better immediately. Having everything on screen makes entropy colors clash with the other panels? Different palette needed?

jameshadfield:
I like the reduced-height entropy panel a lot (I can tidy up the y-axis stuff)

Unfortunately it's broken for me at eeada1fb12d9cc20a37f3a9a0810405db2289347

![aug-04-2017 09-49-39](https://user-images.githubusercontent.com/8350992/28978328-656244c4-78fa-11e7-9254-aacd135ca1e9.gif)

PR #447, opened 2017-09-07
Split / Full panels

colinmegill:


trvrb:
This needs a few things before merging:

1. A rebase on top of current `master`. I tried merging `master` into `#98-panel-layout` and it was quite ugly. Would suggest to work through and graft changes line-by-line onto current `master`.

2. Sensible defaults for `full` vs `thirds`. Defaulting to `thirds` looks really bad on smaller viewports.

3. Renaming `full` / `thirds` to something more sensible.

PR #448, opened 2017-09-07
Charon

jameshadfield:
This PR is not ready to be merged, however it is at a stage where it needs feedback & testing.

This PR adds API functionality to the server (`server.js`). When the client loads Auspice, a API request is made for a "manifest", which provides available datasets and splash page tiles (these used to be hardcoded in the Auspice code). Similarly, dataset JSONs are now attained via a API call to the server, rather than a direct `fetch` to S3. Note that the data (including the manifests and splash image tiles) is all stored on S3, but the server is making the calls (in response to an API request), not the client. 

Salient points: 
* `dev-server.js` has been removed. Command line arguments to `server.js` allow the dev functionality (see the comments in `server.js`). 
* `src/util/datasets.js` has been removed. This data is now in a manifest JSON on S3. As such, new datasets need to modify this file as well. 
* ~I have not tested this on heroku dev (will do tomorrow)~ this now works both on `localhost` and `http://auspice-dev.herokuapp.com/`
* mumps datasets are accessible via `localhost:4000?user=mumps` or `http://auspice-dev.herokuapp.com?user=mumps`
* the speed of loading http://auspice-dev.herokuapp.com is almost identical to nextstrain.org (both in the c. 800ms range)
* the speed of loading zika (data displayed, without map tiles loaded) is also the same
* Cloudflare should still be giving a benefit, as the same fetch-type calls are being done, just py the server not the client. 

How to run locally
```
cd auspice
npm install
sh scripts/get_data.sh
npm run start:local # equivalent to "node server.js dev localAPI localData"
chrome -> localhost:4000
```

@trvrb @colinmegill @huddlej it'd be great if you could both take a look at the way the server is laid out (it's extremely simple) and test this on your machine.

To do list:
- [x] Explore caching when getting files from S3 (might already be happening) UPDATE: speed is comparable to direct S3 fetch. Can thoroughly explore caching in the future. 
- [x] Ensure JSONs are coming through gzipped (and confirm that they were)


jameshadfield:
Results of performance testing (using chrome's dev tools):

All times are `charon` (API) vs `master`:

Splash page:
"DOMContentLoaded" (note that splash images are not loaded by this point): 1.63 vs 0.93s
# requests: 118 vs 119
transferred: both 1.9MB
manifest request: 200ms vs NA

Zika:
"DOMContentLoaded" 1.1 vs 3s
tree rendered times: 2.3 vs 4.0s
# requests: 17 vs 26
transferred: 711kB vs 662kB (not sure why this is different)
manifest request: 140ms vs NA

So in general, splash is a little slower, zika is quite a bit faster.

jameshadfield:
Results of stress testing on heroku dev using `loader.io`

10 request / second for 1minute. Holds up well.

Splash page:

[image:D4494AC0-E663-4DE6-BB2B-0A286EE11FCB-5201-0000065CCC144379/885ADF72-C622-416B-AD7B-6699A915A366.png]

charon?request=json&path=dengue_denv3_tree.json (2.0MB JSON file, via S3)


[image:3F354966-A700-49D9-BFDD-010C7828AB84-5201-0000069CECA8FD48/CF996C60-807E-450F-B1AF-D2BA474CB19A.png]

jameshadfield:
Thanks for the review @huddlej - really appreciated. I've renamed a few things as per your advice. Performance seems comparable to `master` and holds up well under stress.

I plan to merge this PR today.

trvrb:
This a great direction and well implemented. Thanks @jameshadfield. Some brief performance testing matches what you've found.

* I'm getting ~1400ms to finish loading JSONs on nextstrain.org/ebola (before Charon).
* I'm getting nearly identical ~1400ms to finish loading JSONs on http://auspice-dev.herokuapp.com/ebola (with Charon).

>Console error loading Ebola site: Error: <path> attribute d: Expected number, "MNaN,NaNLNaN,NaNC…".
This is a persistent error.

I read through the code. Looks great. I like the structure. Generally impressed here and I think it will be a good path forward to build on.

PR #449, opened 2017-09-08
Finish migration from React.propTypes to propTypes

### LINKED ISSUE #436: update how PropTypes are checked
this will break in React v16
see https://github.com/jameshadfield/phandango/commit/faa450792084bf8ff6a0e74b5c3e4d5384b0b99e

trvrb:
Resolves #436.

Motivated by @jameshadfield's getting `redbox-react` upgraded. I went through and upgraded remaining packages that were throwing warnings from `React.propTypes` per issue #436.

This will require an `npm install`. Console is now beautiful clear of warnings.

@jameshadfield --- Could you confirm this works for you quickly? Code doesn't need reviewing.


jameshadfield:
Works for me

Issue #451, opened 2017-09-12
async script loading


jameshadfield:
`leaflet.css` and `react-select.css` are loaded from unpkg.com. These are requested before the bundle and are blocking. Sometimes these take 10s - so the page load is around 11s! They should be made async.

jameshadfield:
closed by #452

PR #452, opened 2017-09-12
async loading of external CSS & react bundle

jameshadfield:
This PR moves external CSS calls from `<head>` to `<body>` to load asyncronously. The react bundle is also loaded asyncronously. This should fix the blocking cases detailed in #451 

Chrome dev-tools shows the loading of CSS & React bundle simultaneously:
image

@colinmegill as you wrote `index.html` can you please code review.


trvrb:
Tested. Looks great. Thanks James.

PR #453, opened 2017-09-14
server-side rendering of markdown reports

### LINKED ISSUE #35: Reports page

### LINKED ISSUE #438: server rendered static reports
Aim: to display reports (initially a straight port of http://nextflu.org/reports/...) as static pages generated from markdown (see https://github.com/blab/nextflu/tree/master/auspice/reports). These should not be part of the bundle, rather they should be server rendered and the HTML sent to the client.

* `ReactDOMServer.renderToStaticMarkup(element)` should be able to send HTML to the client - https://facebook.github.io/react/docs/react-dom-server.html
* https://www.npmjs.com/package/markdown-to-react-components purports to turn markdown files into react components. This should allow for future use cases of interactive elements within the reports.

jameshadfield:
### What this PR does
This PR adds (seasonal flu) reports to nextstrain. These are not included in the bundle but are rendered on the server and sent to the client via an API request. The data lives on S3 - the available reports are listed in the manifest and each report is saved as pure markdown. 

### How to run / test
Remotely: live at http://auspice-dev.herokuapp.com/reports
Locally: 
```
git fetch
git checkout 438-static
npm install
# option 1 - sources data locally
npm run getdata
npm run start:local
# option 2 - sources data from S3
npm run start
# option 3 - build & run production version
npm run build
npm run server
```

### What’s needed to merge this PR
This PR is quite involved and will require code review by @colinmegill (regarding the server rendering and webpack-ifying of the server) and aesthetic improvements by @trvrb (especially for the presentation of the list of available reports). Error testing (esp 404s) is ongoing - please let me know if you find one.

### What’s changed
* The server is more complicated and requires transpiling. For development (`npm run start`) we can use `babel-node`  instead of `node` to avoid this, however for production we must run `server.js` through webpack (see `webpack.config.server.js`) producing the (gitignored) `server.dist.js`. This must be in the root directory. `npm run build` and `npm run server` commands have been updated.
* react-hot-loader is now used (a good move anyways). The old setup was incompatible with the more complex server.
* The (S3 hosted) reports are written in markdown, with images also on S3 and accessed via the API, e.g. `H3N2 frequencies`. These image filenames must all be lowercase.
* Server rendering of React HTML which is then injected into the page by the client is new. There's actually not that much out there on this - it seems to be only really used for the initial `<App>` render, not for individual pages like this. The implementation of this needs to be reviewed.

### Issues closed:
closes #438 
closes #35


jameshadfield:
re: layout of the "table of contents" "list of reports" - while there's only 4 there now, I envisage having this containing not just flu reports, but other virus blog posts / reports, blog posts about new features (e.g. "new: download - upload" functionality). Perhaps we should change the word from "reports" to "posts"?

trvrb:
>Perhaps we should change the word from "reports" to "posts"?
Yes. I think calling these "posts" make a lot of sense. In this case, perhaps some tagging would be good. Taking a look at the PR now. Thanks James.

trvrb:
@jameshadfield ---

Can you test these latest changes?

You can run `npm run start` to serve static files from rawgit sourced from the themis repo. If you run `npm run start:local` it will source static files from `static/`.

I decided to go with raw markdown and converting live on the server as this gives a bunch of options that don't exist in Jekyll. It also saves a compile when developing.

My main question at the moment is where tile images should live. I've moved them to live in static content. This seemed to fit the standard use of a separate static assets location. Felt better than having them alongside datasets where they can be accidentally deleted. But would appreciate your thoughts.

I'd like to:
1. Clean up themis directory structure
2. Move from rawgit to S3/Cloudfront to allow compression etc...
3. Add YAML front matter to posts (requires switching to Remark from react-markdown)
4. Fix Markdown image links to not need the funny charon request syntax

But none of these things are urgent and I think can be done in smaller individual PRs.

jameshadfield:
Thanks @trvrb. I'm making a few alterations which should allow this to be merged tomorrow - i see a very clear distinction between data and posts:
* Each should have their own manifest file
* Tile images should live in /data/ as they are not related to "posts"

PR #455, opened 2017-09-15
98: "grid vs full" panel layout architecture & toggle

### LINKED ISSUE #98: Optionally disable switch to 2 columns

jameshadfield:
This PR replaces #447 

This PR closes #98 

This is cherry-picked to make things cleaner.

In addition to @colinmegill's sidebar toggle:
* "thirds" is now called "grid"
* the default setting is based on `twoColumnBreakpoint` (controls reducer)
* browser dimension changes cause the redux state of `panelLayout` to change where necessary. This is much cleaner than calculating this in each component.


trvrb:
@jameshadfield (and @colinmegill) --- Thanks for putting this together. I made the changes in cf2503fc4ba16f72caab49f0e511872bf820eed0 to get a perfect grid rather than an approximate grid. I think this is a big improvement. I tested on big monitor, on laptop and on tablet and phone (via Chrome) and I'm quite happy with the behavior. Gridding the app on phone looks a bit funky, but not awful. Still perhaps eventually graying out grid when screen is too narrow would be good. But I don't want this to stop the PR. Merging now.

Issue #450, opened 2017-09-11
Map SVG demes & transmissions need translating


jameshadfield:
The map SVG comprises two parts - the saving of the background map panes as a base64 PNG inside the SVG, and the demes & transmissions (SVG paths etc). This works well for the default view, and zooming appears to be ok. Hovever, panning brakes both of these parts.

This issue:
Map demes & transmissions are there but need to be translated (i.e. moved across)

kairstenfay:
Can give more specific instructions about how to reproduce this issue? (and #429, too)

Issue #431, opened 2017-08-28
Data attribution needs to be written


jameshadfield:
Will be displayed in the Download Data Modal

file: `src/components/controls/downloadModal.js`

Issue #429, opened 2017-08-25
map SVG is broken after panning


jameshadfield:
This issue depends on the merge of PR #427

The map SVG comprises two parts - the saving of the background map panes as a base64 PNG inside the SVG, and the demes & transmissions (SVG paths etc). This works well for the default view, and zooming appears to be ok. Hovever, panning brakes both of these parts:

- [ ] Map panes do not save and there are `404`s in the console
- [ ] Map demes & transmissions need to be translated in the SVG

CameronDevine:
I am not able to reproduce this issue. It appears this should have been closed by #587.

jameshadfield:
![auspice429](https://user-images.githubusercontent.com/8350992/76175799-17b72280-6213-11ea-9670-6f874199e7b6.png)
@CameronDevine are you able to reproduce this?

CameronDevine:
@jameshadfield Yes, I am able to reproduce this. It appears there is an issue with the [mapbox/leaflet-image](https://github.com/mapbox/leaflet-image) library (See mapbox/leaflet-image#89 and mapbox/leaflet-image#117). When a map image is requested that spans the "anti meridian" (The line of 180 degree longitude) the result is incorrect. Areas of the map with a longitude less than -180 degrees are blank, while areas greater than +180 degrees seem to have a portion of the map incorrectly placed in the remaining area.

rleir:
@CameronDevine But Japan is about 138E longitude, and the line of cutoff in the example seems to go through Japan. Roughly the anti-meridian of Boston (funny coincidence, location of MIT). I had a similar problem with the anti-meridian of Ottawa, Canada in this leaflet map: https://leirtech.com/example/conn/

and solved it by making sure the longitudes values were within the range [Ottawa+180 .. Ottawa-180] with the 3 lines of code at or about 
https://github.com/rleir/connection-map/blob/3821a0aad443b2f63b8099af00000abe3d0e0988/src/info2geojson.py#L327
@jameshadfield Where in the Auspice code do you create the SVG map? cheers -- Rick

CameronDevine:
I believe this issue should have been closed when #925 was merged in. However, I think the changes in that pull request, specifically the change to `dom-to-image`, caused #1066. The image generation happens in the `window.L.getMapTiles` function of `src/components/map/map.js`.

**Note, this is only an issue when downloading an SVG of the data displayed.**

rleir:
[...]

PR #462, opened 2017-09-27
Dataset title, better filter display & download modal

### LINKED ISSUE #354: Display author and title when filtering by author
It would be good to provide a method for individuals to share their own datasets and give prominent credit to contributors. I feel like this is currently overly hidden. My suggestion is when filtering by author, just show author and title at the top. Something like:

faria

when filtering to "Faria et al". Additionally, we need to fix the URL handling of filters, as this is only really useful if a URL to this view can be shared. This should look something like: http://www.nextstrain.org/zika?f:authors=Faria%20et%20al. This might be able to be streamlined further.

jameshadfield:
## This PR introduces
* a title ("info") card above the map/tree, which contains the dataset title and a list of active filters (see screenshots). This card is collapsable via a toggle.
* an expanded footer to contain all available filters (see screenshots)
* the removal of filters from the sidebar
* the download modal is now running.

## Blocking:
* The data attribution needs to be written - @trvrb - #431

## Notes:
* This is live and working for zika at http://auspice-dev.herokuapp.com/zika
* Augur should be rerun for most datasets - this is happening this week. After this, the `controls` can be removed from the metadata JSON, but this can be left in for now.
* This still works relatively gracefully on "older" augur builds - see http://auspice-dev.herokuapp.com/flu, http://auspice-dev.herokuapp.com/dengue
* Bugs #429 and #450 remain and will be fixed eventually.
* closes #354  

## Screenshots
image

image

image

image

image


jameshadfield:
@trvrb I think this is ready to be merged now

trvrb:
Thanks @jameshadfield. There were a couple small bits of logic that I was wanting to adjust. I'll do this now and then we can merge if you're happy with them.

trvrb:
@jameshadfield --- I just removed date filters from the info box. I didn't like how the numbers flipped through while dragging date slider. I you feel strongly I'm happy to discuss / undo.

I also disable the data download button so that this can be merged now and not depend on #431 (I'll do this, I promise).

Going to merge now.

Issue #458, opened 2017-09-19
invalid URLs shouldn't load main app page


jameshadfield:

jameshadfield:
closed in #463

PR #463, opened 2017-09-30
panels defined by meta.json

### LINKED ISSUE #356: panel display via (augur defined) attributes
(re-)implement augur control over the display of panels, via a `meta.json` attribute named `panels`. This will allow custom control over which panels are displayed.

On the auspice side, JSON loading will be:
* (synchronously) fetch meta, tree & sequences
* as the `panels` attribute dictates, load additional JSONs

Depends on: `simplify` branch merged.
Things depending on this: Frequencies & Timeline
Timeframe: around a month (I'll be in england for a while)

jameshadfield:
## This PR
* adds a spinner which replaces the (empty) map & tree until they load
* only loads components specified in `meta.json` (all builds currently specify entropy, map & tree)
* closes #356 

Tested on all live data on the dev server. No problems.

cc @sidneymbell this will affect the timeline.

### DIFF
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -4,6 +4,7 @@ import { updateColors } from "./colors";
 import { updateVisibleTipsAndBranchThicknesses } from "./treeProperties";
 import { turnURLtoDataPath } from "../util/urlHelpers";
 import { charonAPIAddress } from "../util/globals";
+import { errorNotification } from "./notifications";
 
 // /* if the metadata specifies an analysis slider, this is where we process it */
 // const addAnalysisSlider = (dispatch, tree, controls) => {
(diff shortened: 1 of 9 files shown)


jameshadfield:
@trvrb this is good to go I believe.

The entropy panel loads a bit later than before, as it is only fetched _after_ the meta.json is processed. But everything else is improved - spinners and panel specifications.

This also allows us to upload data (say `beast_X.json`) to S3 then access them at `nextstrain.org/beast` without modifying the manifest. If the datafiles don't exist a notification pops up and the app goes back to the splash page.

jameshadfield:
closes #458

jameshadfield:
now up on dev server - look at http://auspice-dev.herokuapp.com/mers 🎉

trvrb:
This is great. Tested behavior and seems good. Code looks good too. I especially like the ability to load arbitrary datasets, not specified in manifest and bouncing back to splash if they don't exist.

PR #464, opened 2017-10-01
Styles update, dark sidebar, minimal cards

### LINKED ISSUE #328: genotype text input box bugs
* only seems to allow numeric input (despite "NS1_348" being a valid input)
* cannot delete first character
* should be debounced to allow entering of "1234" without changing the coloring to "1", "12", "123"

trvrb:
Fixes #328 (genotype input bugs)

After the PR the styling looks like:

dark-sidebar

This is currently up on http://auspice-dev.herokuapp.com/.

This PR does several things:
1. First and foremost, it darkens the sidebar. This was a design decision cascading from the implementation of the info panel. I really didn't like the negative-space gutter interaction of all the cards that we were using. I didn't like having text on a card and I didn't like text on the background, then cards, and then text, like seen in 948dec228bf22f35a88aadfad4180e117c815c1c.

cards-interface

In moving to a dark sidebar the visual weight was mitigated to a substantial degree by thinning down the sidebar from 280 pixels to 220 pixels (enabled by moving dataset "title" from sidebar to info).

@amcrisan, @evogytis, @jameshadfield, @huddlej and I thought the dark sidebar worked better. @colinmegill and @sidneymbell thought the light sidebar worked better. You can do a head-to-head comparison with branch `styles-v3` vs `styles-light`.

I think this gives clean separation of controls and viz. To get to (2) I needed to go with (1).

2. Cards still exist, but the drop shadow and beige background are gone. This allows a clean viz area with white-on-white that allows text and graphics to be flowed together.

3. Select boxes have been moved to react-select and input boxes have been styled. This should result in a more consistent appearance in Firefox, Safari and IE.

4. Genotype colorBy select has been improved to give a "gene" dropdown and fix the annoying input bugs per #328.

5. Panel glyphs have been added.

6. Sidebar toggle is now a chevron consistent with legend toggle.

@rneher --- I hope the new styling works for you. Let me know if it doesn't.


colinmegill:
One path to avoid the saturation of the dark sidebar would be to: 

1. Get rid of the grey and cards on the main panel, as has been done
2. Recede the sidebar by flipping the drop shadow to the left side
3. Put the grey background there (so it's a tray, rather than an overlay)

This is one option to accomplish much of the same thing without the heaviness of the dark sidebar. 

I like the line between the content areas. I would make it lighter grey, and balance the white space around it.

colinmegill:
The filters are not *currently* so many that they could not be summarized at the top. It might look something like: 

> Showing [inference based on] 200 sequences, from 35 authors, in 68 countries across 3 regions, between May 2015 and June 2016. 

This would add context when it's most needed - on load - and prevent things from expanding and contracting too much (which looks glitchy), because the bulk of it would always be there.

trvrb:
@colinmegill ---

Thanks for the feedback. I did your 1, 2 and 3 in branch `styles-light-v2`. Please take a look. I actually really like this. I'm about 50/50 now on light vs dark sidebar. Happy with both options.

With dark sidebar, when first landing, the first thing you look at is the sidebar and then the tree. With light side, the first thing you look at is the tree, and then the sidebar.

I tried "balancing the white space" around the card demarcations. Really did not like this. Cards lost their anchoring. Did lighten them up though.

rneher:
I like the new design, but I can't quite be decided whether I like the dark light version better. The big dark blob generates too strong contrast (I use a lot of dark themes, but its either all dark or all light). The light design might benefit from a grey background shading for the controls.

jameshadfield:
_better is the enemy of good_

(1) my vote is for `styles-v3`
(2) Sure, I can add more context on load. But context upon filtering is equally important.

colinmegill:
@rneher agree that it's usually all dark and all light, and @trvrb agree
that light doesn't pull focus away from the tree. If you do decide to stick
with light, and keep the shadow, make the sidebar, as the shaded item, the
'lower' one, or remove shadow:

[image: image.png]

trvrb:
All, thanks for the feedback. I'm going to go with the light sidebar for the time being, but save the dark sidebar to the branch `stale-dark-sidebar` so we can return to this if we change our minds.

@rneher ---

I greyed down the light sidebar slightly.

I agree about this issue with unrooted display. I don't think the fix is restoring the card design however. We could do a lattice grid for unrooted, or experiment with matte backing (like the oceans in the map).

@colinmegill and @jameshadfield ---

I like the idea of context on load (extending the "Showing 379 of 379 genomes." sentence). Not because of the expanding contracting info box, but just because it seems useful.

trvrb:
> If you do decide to stick with light, and keep the shadow, make the sidebar, as the shaded item, the 'lower' one, or remove shadow:

It is now the "lower" one. The drop shadow is inset.

trvrb:
I'm going to merge this now. I think it's a solid foundation. There are some more small tweaks that can happen, but by

rneher:
very good. It's a big improvement either way.

PR #467, opened 2017-10-05
metadata reducer flattened

jameshadfield:
This PR flattens the metadata reducer, which should improve the development effort and reduce bugs.

Before: `reduxState.metadata.metadata` held most of the information
After: `reduxState.metadata` holds everything, and `reduxState.metadata.loaded` can be relied upon.

### DIFF
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -115,26 +115,26 @@ class App extends React.Component {
           }}
         >
           {
-            (!this.props.treeLoaded || !this.props.metadata.metadata) ? (
+            (!this.props.treeLoaded || !this.props.metadata.loaded) ? (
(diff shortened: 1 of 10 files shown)


trvrb:
@jameshadfield --- This is nice. One immediate thing: I'd strongly prefer that this spec be consistent across `tree`, `metadata`, `entropy`, `sequences` and `frequencies`. Looks like we currently have `state.tree`, ie this PR is bringing `metadata` into consistency with `tree`. However, we do use `state.entropy.entropy`.

jameshadfield:
yes. `state.entropy` is wrong and should be fixed.

trvrb:
Great. I'll let `entropy` be a separate PR. Tested and merging now.

PR #468, opened 2017-10-05
don't calculate color scale before data is loaded

jameshadfield:
`getColorScale` now only runs once upon dataset load, instead of 3+ times


trvrb:
This is great. Confirmed new desired behavior. Thanks @jameshadfield!

PR #471, opened 2017-10-12
S3 staging

jameshadfield:
This PR adds a toggle to the sidebar which changes the source of data from `data.nextstrain.org` to `staging.nextstrain.org`. A sentence is added to the info box when the staging server is enabled. If desired I can add a notification pop up (instead?). 

Charon API has been extended to take an extra query parameter `s3` to the manifest & dataset calls. Splash images are currently always sourced from `data.nextstrain.org`, this can be revisited in the future.


trvrb:
Nice! This is great.

PR #472, opened 2017-10-17
Enable downloads

### LINKED ISSUE #431: Data attribution needs to be written

jameshadfield:
This PR enables the download modal popup and contains a number of stylistic improvements as well as bug fixes. 

It is tested and working with all live datasets.

This PR is needed ASAP, as it is referred to in the submitted manuscript.

Closes #431

image


trvrb:
Looks great overall, some small suggested changes:

* The current ordering of the metadata csv is strange. For Zika I have: `Strain,Division,Num Date,Num Date Confidence,Region Confidence,Title,Url,Country,Region,Country Entropy,Db,Accession,Raw Date,Authors,Date,Paper Url,Div,Country Confidence,Journal,Region Entropy`. Can this be fixed to be something like `Strain,Num Date,Region,Country, etc...` Don't care about the exact ordering, would just something that is more sane.
* Avoid tabs in your csvs. For instance, strains in authors csv are separated by tabs. Would suggest spaces instead  (or pipes or whatever).

PR #476, opened 2017-11-23
Narrative machinery

jameshadfield:
This PR incorporates the machinery needed for narratives. It does not address styling, UX, narrative content or expansions to state stored in the URL.

The markdown syntax used for the narratives is normal markdown with the addition of lines such as `[NEXTSTRAIN_URL](ebola?dmax=2015-04-09&dmin=2015-02-07)` or `
[NEXTSTRAIN_NO_ACTION]`. I propose finalising this in a subsequent PR. 

The general data flow is:
* after JSONs are loaded, the narrative markdown is requested
* the server fetches this from https://github.com/nextstrain/themis/tree/master/files/narratives
* the server parses this into blocks of rendered HTML with associated URLs if applicable
* this state is sent (in JSON format) to the client
* this becomes redux state in the client
* the client renders a right-hand-sidebar with `<Narrative>`
* `<Narrative>` creates `div`s with listeners as defined in the markdown, which cause the URL to be parsed when the `div` is hovered over and the state of the app to change accordingly. The content of these `div`s is the server rendered HTML.

NB: the `enableNarratives` global flag must be set to `false` before merging into master!


trvrb:
The narrative work here is really excellent @jameshadfield. I like it a lot. And I really appreciate merging and setting an `enableNarratives` flag.

PR #480, opened 2017-12-12
separate state from URLs

jameshadfield:
### This PR is complete but needs further bug testing.

This PR greatly simplifies how the app handles URL state by removing react router and introducing middleware. Before we had a complicated 2-way relationship between the URL and the app state. Now the entire app works off (redux) state and the URLs are kept in sync by middleware.

* Any change to the app "navigation" state (e.g. reloading a page, clicking a link, browser back-foward) is handled in one place: the `changePage` action. This modifies the `datasets` reducer handles what "page" nextstrain is on (app vs about vs splash...) as well as the dataset being displayed (zika, dengue/all...). ( The "state" of the app (e.g. colorBy) is still handled by the controls reducer.)
 
* This (and other) actions are intercepted by middleware which changes the URL content when necessary. That is, if we remove this layer, the app still works but the URL is never updated. 

## Questions:
* Should some actions (e.g. changing colorBy) change the URL in such a way that you can "go back" via the browser back button? If so, which actions should and shouldn't do this?

## Still to do:
* ~bugs when refreshing on pages such as `/dengue/all`~
* ~remove `history` npm package~
* ~use `pushState` where necessary and ensure back button works as expected~
* ~posts (reports) need updating~
* ~minimise / describe exactly which components access URL state (only middleware should ever change the URL). Ideally URL state is only ever read in one place.~
* ~query params remain when going to splash page (and others)~
* ~fix the choose dataset selector~

## FAQ (move to wiki)
* How are query params (`?c=country...`) read on page load? 
They remain untouched in the URL and `loadJSONs` reads them and passes them to the controls reducer along with the JSON data

* How are query params read during interaction with the app?
They're not - actions are dispatched to change state and middleware keeps the URL up to date

* How is the desired dataset (e.g. zika, flu/...) loaded?
The `changePage` action (perhaps via `popstate`) sets this as redux state (`datasets.datapath`), which is used by the `loadJSONs` action (including triggering the narrative markdown fetch)

* What is the order of actions on initial page load?
    * The initial page is set in the default state of the `datasets` reducer by examining the URL (so no `popstate` action on load). If this is `<App>` then the app loads with a spinning nextstrain logo as we don't yet have any data. Similarly, the `<splash>` page will load with a spinning logo where the tiles should be, as this information requires the manifest JSON.
    * The manifest JSON is fetched (`<Monitor>` -> `componentDidMount` -> `getManifest`)
    * The arrival of this JSON triggers `MANIFEST_RECEIVED` and `changePage` actions
    * The `changePage` does nothing here (the reducer state is already correct) unless `page === "app"` where it sets `datasets.datapath` (previously `undefined`)
    * If we're on `<App>`, the lifecycle methods detect a change in `datasets.datapath` and trigger the `loadJSONs` action

* NB: current URL state _is_ stored in the `datasets` reducer, but this is only used upon back/forward navigation logic - this is desired because I want to be able to use the back/forward buttons and _not_ reload the entire app (like we did before). 

* Where is URL state _read_?
    * googleAnalytics
    * `loadJSONs` - query state is passed to the reducers at the same time as the JSONs, which allows the correct state to be set.
    * footer - the acknowledgements are hard coded and are selected via the URL (could be via `datasets.datapath` or set in the JSONs)
    * the `browserBackForward` action, triggered from a `popstate` listener in `<Monitor>`. Used to `changePage` or `changePageQuery` when the URL changes via back/forward buttons. Note that this is the only place that uses the `state.datasets.urlX` values (necessary for comparison of old URL vs new URL).
    * initial state of `state.datasets` (i.e. does the splash page load, or the app, or...). This could be achieved via an initial `changePage` action, but we aim to minimize actions...
    * `getManifest` uses the pagename to load the appropriate data (see above) - this could potentially be improved.


trvrb:
@jameshadfield: This is a fantastic direction. I'm reviewing the PR now. Generally I totally see how this should add substantial sanity to the app. Having a one-way flow of information from redux to URL makes a lot of sense.

--------------------

>Should some actions (e.g. changing colorBy) change the URL in such a way that you can "go back" via the browser back button? If so, which actions should and shouldn't do this?

My gut feeling is that anything that modifies query string in the URL (to the right of `?`) should not register as a state of the app in its history, while every update to the path string updates the app's history. I don't feel hugely strongly about this however.

--------------------

I'll dig into code, but first a bugs pass:

- [ ] If I go to http://localhost:4000/zika and then select "ebola" in the dataset dropdown to get to http://localhost:4000/ebola, I can then click back to get back to the Zika page. This works. But if I then click forward, the url changes to http://localhost:4000/ebola, but the app state does not update. Strangely, this is not entirely consistent, sometimes a forward action will register and sometimes it will not. Clicking around, I can't figure out an obvious logic for when it works and when it does not.

Actually... I can't find any other bugs besides this forward / back button issue.

--------------------

I don't quite understand the relationship between this `url-middleware` branch the `entropy`  branch (PR #478). It looks like some, but not all of `entropy` additions are in `url-middleware`. I see that changing visibility updates the entropy plot. However, it looks like the `url-middleware` branch does not require an update to the JSON format? Given JSON requirements of `entropy` it would seem to make sense to merge `url-middleware` first and then attempt `entropy`. Would you agree?

trvrb:
Re: merging. Sounds like a good strategy.

jameshadfield:
This branch now merged into `entropy` and the sporadic forwards/backwards bug is now an issue

Issue #502, opened 2018-02-20
Missing nextflu features


trvrb:
## Overview

This issue is meant to document features that currently exist on nextflu.org (via https://github.com/blab/nextflu/tree/master/auspice) that should be ported into nextstrain/auspice. Fortunately, we currently use the same flu JSONs for nextflu.org and for nextstrain.org, so that the builds are the same, although nextstrain.org may need some additional info in `meta.json` to load features.

This looking at migration of nextflu.org to nextstrain.org. Migration of who.nextflu.org is a separate matter and will need attending to after migration of nextflu.org.

I've organized this into:
* Features necessary to migrate live nextflu.org to nextstrain.org
* Features that are nice to have but not required
* HI (titers) features

## Necessary

1. Strain name search. I use this all the time when interacting with nextflu. I'd think this would work like any other filter. Would be amazing to live filter tips as search string is honed in on...
2. ~Frequencies. PR #497 is a fantastic direction for this.~
3. Annotation of AA mutations to branches. This is way nicer to work with than having to mouseover everything and is critical when digging into flu trees. I would default this to off the way it is on live nextflu.
4. ~local branching index colorBy (I'm okay with this being computed in augur rather than live updated in auspice --- might even prefer)~, ~clade colorBy~

## Nice to have

5. Legend bounds need fixing to be properly rounded like they are on nextflu. The four decimal places with random cuts looks strange.
    * _EDIT_ #506 
6. Nextflu populates genotype colors in a more pleasing fashion so that majority is the pleasing blue-green and mutates are brighter colors. Nextstrain often has the reverse.
    * _EDIT_ #503
7. Fix colorBy ordering. I think this might be random at the moment (due to dictionaries not being sorted), when there should be a clear ordering, like so: https://github.com/blab/nextflu/blob/master/auspice/_data/color_options.json.
    * _EDIT_ fixed
8. Allow multiple `colorBy`s
    * _EDIT_ #509
9. keep controls state if possible - i.e. switching 3y -> 6y should maintain colorBy, tree display...
    * _EDIT_ #510

## HI titers (EDIT: will revisit)
10. Password protection
11. JSONs (HI, tree, subs) contain egg, cell, hi, fri, center information (i.e. there's 3 JSONs in total). @trvrb @rneher @huddlej to build these / design this schema.
12. HI distance, subs model, tree model
13. Titer matrix panel?

_Edited to remove LBI_

_Edited after meeting_

jameshadfield:
[...]

Issue #503, opened 2018-02-21
fix legend ordering


jameshadfield:
fix legend ordering for genotypes to reflect when particular genotypes were first observed

#497

trvrb:
There are a couple situations where this comes up:

With genotypes, where the ancestral state is not the first thing in the legend.

http://www.nextstrain.org/flu/h3n2/ha/3y?c=gt-HA1_135

<img width="1056" alt="freq-1" src="https://user-images.githubusercontent.com/1176109/36743250-aa1022ca-1b9e-11e8-97dc-c22ef94512ba.png">

But also with clades where unassigned is not the first entry in the legend:

http://www.nextstrain.org/flu/vic/ha/12y

<img width="1062" alt="freq-2" src="https://user-images.githubusercontent.com/1176109/36743365-fe42b222-1b9e-11e8-8d80-96efab27a918.png">

Generally, I don't think this is a huge issue, but I could imagine eventually separating colorBys into continuous, ordinal and categorical. Examples of continuous are `ep` and `num_date`. Examples of ordinal are `country` (where there's a reason two countries are placed next to each other on the color ramp). Examples of categorical are `clade_membership` and `genotype`.

I think just categorical variables would want to be reordered according to tree appearance, while continuous and ordinal should stick to their natural ordering.

PR #504, opened 2018-02-26
Nextflu features

jameshadfield:
This PR addresses points 1-4 of #502. As such, once merged, nextflu.org can redirect to nextstrain.org. The "Nice to have" features will be tackled next (separate PRs). The titer features will come after that (one day). 

### Points in depth:
1. Strain name search - partial matches (i.e. "auck" shows c. 4 genomes in the dropdown) don't result in a change in the tree as it wasn't working smoothly -- i.e. you had to wait for the tree to update while you were typing. Can revisit this at some point - maybe we can change the tip radius on the fly.  Selected strain appears in the info panel (top of screen) as well as the URL (i.e. you can restore the state to a single strain via the URL). 
2. Frequencies - live already.
3. AA mutations - working for every dataset - see the branch label dropdown menu in the sidebar. Only labels are shown where the descendent visible tips account for more than 5% of the total visible tips (same as nextflu). Clade labels are always displayed.
4. ColorBys: LBI & clade are live already (although the latter requires updated augur JSONs).

### Somewhat unrelated points
* The JSON processing on initial load has been shifted from the reducers to a single action - fewer dispatches, fewer potential bugs, faster code.


rneher:
Thanks, James. Being able to focus on new auspice will be a big step forward. 
a few comments (in random order): 
 * branch labels would be better right justified on end of branch
 * label for frequencies are pretty small. 
 * could it be that legend colors are ascending rather than descending in frequency? Would be nice to have the most relevant colors first. 
 * shouldn't `Normalized data` for frequencies by default?
 * the time tree view, the vaccine cross shows up at its `first chosen` date. Maybe we should implement some sort of background shading or a horizontal bar when a particular vaccine was chosen and used:
  |---------XXX NH 2017/18 XXX
 * don't show the gender.
 * legend item numbering and rounding could be prettier. 
 * I thought we had the joining of parent T-bars and child node sorted out, but is there are kinks. Also, we should probably draw children first, parents later. 
 * when zooming into a clade, the truncated frequencies look weird. Not sure how this can be avoided, though. 
 * I think the flu folks would prefer `HA1`was default genotype coloring rather than `nucleotide`

rneher:
Zooming into a clade and looking at normalized `HA1:92` frequencies gives me this;
image

rneher:
nextflu show a clade averaged age instead of the age of the individual tips. makes for a visually more pleasing experience. when plotting actual age, probably good to make non-existent values grey rather than black.

rneher:
nextflu's tooltips for branches can be quite handy (to look for mutations, titer drops, average ages, etc). currently nextstrain doesn't have this "on the fly" info. thoughts?

rneher:
I was thinking smth like this for vaccine strains (in the background obviously, not covering up the tree)
image

trvrb:
@rneher: Interesting take. I like the direction. I also wondered about making the dashed lines go down to the bottom of the page and then over so as not to occlude so much.

rneher:
some visual that links the strain to the period it is used as a vaccine would be great...

trvrb:
@jameshadfield: This is fantastic.

* Strain search UI is great. I can understand why the iterative filtering of the tree presented a challenge. I really like how the search only applies to visible tips. Though could we resize tip circle to be bigger when only a single strain is selected? It's being hard to pick out once selected.

* I agree with @rneher that right aligned AA branch labels would be better. This should also apply to clade labels.

* AA branch labels are proper size for flu, but appear overly large for Zika and mumps.

* The coloring issues are all in augur. 

@rneher:

>nextflu's tooltips for branches can be quite handy (to look for mutations, titer drops, average ages, etc). currently nextstrain doesn't have this "on the fly" info. thoughts?

The Nextstrain spec was only showing colorBy attributes for the active colorBy. I think this the best way to do this as otherwise there are too many things in the tooltip.

>shouldn't Normalized data for frequencies by default?

I'm torn. Zooming to a clade with genotype with raw data looks like:

freq

which I think fits nicely with my own expectations when subsetting the data.

_Actually... I think perhaps better to thread frequencies issues separately as they are not part of this PR, but deserve attention._

>could it be that legend colors are ascending rather than descending in frequency? Would be nice to have the most relevant colors first.

I believe this is https://github.com/nextstrain/auspice/issues/503

jameshadfield:
Everything fixed or broken out into seperate issues.

trvrb:
Excellent! Thanks so much James.

PR #517, opened 2018-02-27
Style tip stroke separately from branch stroke

trvrb:
Nextflu has tip stroke set to full color and branch stroke set to a grayer version. In the code here:

https://github.com/blab/nextflu/blob/master/auspice/js/auspice/colors.js#L224

I believe this makes for a significantly more distinguishable / poppy tree. Here's `master` for flu:

master

And here's `tip-styles`:

tip-styles

The tips are not so blurred together. Also makes Zika, etc... look better.

I think this is a definite improvement. And I thought that my approach with separating existing `stroke` arrays into separate `tipStroke` and `branchStroke` arrays was the right approach. However, I got stuck when trying to modify PhyloTree.change to allow for this. I had tried using `setStyleViaFunction` without success.

@jameshadfield: Would you take a look at this at some point?

### DIFF
--- a/src/components/tree/index.js
+++ b/src/components/tree/index.js
@@ -12,7 +12,7 @@ import HoverInfoPanel from "./infoPanels/hover";
 import TipClickedPanel from "./infoPanels/click";
 import { changePhyloTreeViaPropsComparison } from "./reactD3Interface";
 import * as callbacks from "./reactD3Interface/callbacks";
-import { calcStrokeCols } from "./treeHelpers";
+import { calcBranchStrokeCols } from "./treeHelpers";
 import { darkGrey, dataFont } from "../../globalStyles";
 
 const resetTreeButtonStyle = {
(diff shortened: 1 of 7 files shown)


jameshadfield:
Massive improvement.

1. Lowering the branch opacity lower-bound looks great for flu but perhaps a bit washed out for zika?
0.4:
image

0.6:
image

2. The `mouseleave` event of branches makes them disappear :( - try mousing over a bunch of branches to see!

trvrb:
I see what you mean about Zika being washed out. This resulted from different logic for branches with confidences and branches without. d27d4019d106742821a8ea4b25f26ea349af2200 is a small fix for this.

trvrb:
Thanks for finishing this up @jameshadfield. Nice to see how actually easy the addition was to PhyloTree.change. I like the interface.

I can confirm that this is working for me.

PR #518, opened 2018-02-28
Dynamic y-axis for frequencies

### LINKED ISSUE #505: frequencies questions
cc @huddlej @trvrb @rneher 

### Some relevant questions (add more as needed):
* Should normalisation be the default?
* How should residual error be dealt with? Currently if the raw frequency for a pivot is `<0.01` I collapse it to zero -- the alternative being to expand what is probably rounding / smoothing errors to what looks like 100% frequency. But this approach leads to sawtooth like displays:
image

jameshadfield:
* Removes normalisation toggle for frequencies
* y-axis scale adjusts to fit the data in the panel
* closes #505

image

The A2 clade:
image


trvrb:
Nice! I really like this. My only suggestion would be to eventually move to transitions. Should help to show what's going on. But I would leave this for a separate PR.

Issue #333, opened 2017-06-19
transmissions without colorBy


jameshadfield:
For traits where the internal nodes have an undefined color by (e.g. `colorBy="country"`, for seasonal flu we have chosen not to infer ancestral country states) these show up as grey transmission lines. If I understand how `map` works, for the screenshot, for all the `region` transmissions (these exist, and the geo resolution is region) the country `colorBy` (which is all grey for internal nodes) is being averaged - is this right?

![image](https://user-images.githubusercontent.com/8350992/27304634-8bd7ce56-54f3-11e7-8528-baf8093fcc2a.png)

Would the best solution be to drop transmission lines in this case? 
This would mean we could reinstate `country` as a `geoResolution` which would always show only demes (since there will be no `country` transmissions in the tree). I think this would be valuable.
cc @sidneymbell @trvrb @colinmegill @rneher

rneher:
I would prefer not to show the transmissions for flu. I don't think that we have any concrete evidence for the majority of the transmission events and explicitly drawing generates a false impression of confidence.

jameshadfield:
@rneher remove the transmissions for region as well?
I still think we should show demes - otherwise the map should be removed completely!

rneher:
its probably a little less bad for regions, but I'd rather have only demes (which report data) than a hairball of unreliable transmissions.

jameshadfield:
demes only on branch `map-with-no-tranmissions` - only works with the seasonal flu datasets (these are on S3) with `geoResolution`->`country`
![image](https://user-images.githubusercontent.com/8350992/27305469-9a93bf9c-54f6-11e7-9b73-95888988698a.png)

trvrb:
I think in general principal having some attrs that color tips only and then don't have an ancestral reconstruction is great. I agree these should have gray branches in the tree and show no transmission lines on the map. For seasonal flu I would do country as tip only, but still keep the region transmission inference / coloring. This looked like it was working fairly well on the test data.

colinmegill:
To respond to James' original message, the colors are no longer averaged on transmissions (they are on demes) as we treat each transmission individually now for the animation.

jameshadfield:
OK. In an effort to get seasonal flu working today I'm going to use the demes only code (the second screenshot) for situations where the geoResolution has no associated transmissions. 

Once the map animation branch is merged I will remove the transmissions that appear in the first screenshot.

sidneymbell:
Nice work, @jameshadfield. 

[...]

PR #529, opened 2018-03-21
Adjust branch label and vaccine cross styling

trvrb:
@jameshadfield ---

The changes to branch label styling shouldn't be controversial. However, I've decided to remove the dashed lines from the vaccine crosses. I believe it was adding confusion rather than removing it. If we go in this direction I think we need the full description of choice and deployment. But even then I really feel like it will be overloading the tree. Perhaps the best thing to think toward is a separate vaccine panel.

Also, I think the symbols (vaccine crosses, serum gears) look much better with a bit of white space surrounding them. This was accomplished previously by using a character as the cross and giving a stroke of `#fff` and a fill of `#333`. Something to think about.

old

new

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,7 +21,7 @@
     https://codepen.io/tigt/post/async-css-without-javascript
     https://stackoverflow.com/questions/9271276/is-the-recommendation-to-include-css-before-javascript-invalid
     -->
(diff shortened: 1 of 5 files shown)


jameshadfield:
@trvrb the reason they are SVG shapes not crosses with the nice styling is to do with attaching handlers to them - see https://github.com/nextstrain/auspice/pull/498. If they are always going to be directly over the tip then I guess it's not necessary. However I think there is some work in the future on this where we could show info regarding deployment time etc

PR #534, opened 2018-03-26
Second tree & tanglegram

jameshadfield:
Similar to PR #533, this adds functionality via a url only, e.g. `http://localhost:4000/flu/h3n2/ha/3y?tt=na`. The sidebar dropdown (for the second tree) is disabled. There are still some bugs in the second tree but it's pretty close. 

There are (as far as I can tell) no bugs when only one tree is in view, but will test again after it's merged into master.


rneher:
This works great for me! Would be nice to calculate LBI on both trees, maybe use the average as color. A better untangle heuristic might improve interpretability.

rneher:
that's because the additions in augur were done a long while back. ;)

trvrb:
Ah. Yes. Thank you @rneher. 😁

rneher:
mostly the prepare/filter sequences step to ensure the same (or at least similar) set of sequences and the clade assignment derived from HA needed to be assigned to NA. no big deal.

neavemj:
Hi James, thanks for the great feature! 

I was just wondering how to enable the second tree option in the sidebar? I have a segmented virus and I can get the tanglegram via the url but the sidebar doesn't have the 'second tree' option. Perhaps I need to change something in the config file?

I'm running Auspice v1.34.0 and Augur 3.0.2.dev1.

Thanks again,

Matt.

neavemj:
Thanks James,

I'd love to help with testing! I'm currently working on a few different builds - I'll send you a message.

This fixed worked :). I just had to make sure that the strings were in capital letters, even though it is lower case in the files. In this case, the segments are simply named segment1, segment2, ... segment10. Might be worth adding these as I guess there are a few viruses that are annotated this way.

Thanks again and I'll be in touch.

Matt.

PR #539, opened 2018-04-16
Linting pass

trvrb:
I don't think anything should be objectionable here. The only thing to really note, is that currently we use both:

```
const filterAbbrRev = {
  "all": "all"
}; 
```

```
const filterAbbrRev = {
  all: "all"
}; 
```

Would be better to pick one and be consistent through out. This is the `quote-props` option. I might prefer quoting attributes (first of the two above).


jameshadfield:
Can we not use both? I use the second one pretty consistently, but don't have any problem with the codebase using both.

trvrb:
I changed the linter to want to be `consistent`: https://github.com/nextstrain/auspice/pull/539/files#diff-1dc6ee56b778cd91e0327b52aaeaa1b9R45

This enforces the same style within a block. I'm very happy to allow both.

PR #544, opened 2018-04-23
modify temporal color scale & legend to focus on tip dates

jameshadfield:
Previously, the color range for temporal trees (i.e. the blue -> red) was based upon the root date -> the latest tip date. For deeply diverged trees (e.g. dengue, lassa) this meant that essentially all the tips were red and represented by the final box in the legend.

This PR modifies the legend to span the actual tip ranges. The colour ranges are skewed to mostly focus on this range as well. I think this results in much nicer and more useful displays, although for lassa all the recent samples are still red! cc @trvrb @sidneymbell 

image

image

image

image

### DIFF
--- a/src/util/colorScale.js
+++ b/src/util/colorScale.js
@@ -10,9 +10,8 @@ import { setGenotype, orderOfGenotypeAppearance } from "./setGenotype";
 
 const unknownColor = "#DDDDDD";
 
-const getMinMaxFromTree = (nodes, nodesToo, attr, tipsOnly) => {
-  let arr = nodesToo ? nodes.concat(nodesToo) : nodes.slice();
-  if (tipsOnly) arr = arr.filter((n) => !n.hasChildren);
+const getMinMaxFromTree = (nodes, nodesToo, attr) => {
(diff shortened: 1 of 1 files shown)


sidneymbell:
This is lovely, far more informative. Thanks, @jameshadfield.

trvrb:
@jameshadfield --- I really like the direction here. Date coloring was completely uninformative for some builds. A couple thoughts:

1. I completely see why you set the color ramp to reflect tip sampling. Otherwise, http://localhost:4000/lassa/s?c=num_date will still give little discrimination in recent isolates due to the presence of the single isolate from 1969. Still, seems somewhat bespoke relative to the rest of the app to have a non-linear colorBy.

2. Generally, it seems like this issue should be addressed in the colorBy export. Ie should colorBy reflect all nodes? Or should it be clamped to tips? Should it be a linear mapping of domain to range? Or should it be weighted by sampling?

I like our defaults of all nodes and linear. But giving the option for the others seems like the more general way to address this.

I'm happy to merge this now and address point (2) with time.

PR #545, opened 2018-05-01
Fix toggle position and remove extraneous whitespace

trvrb:
This is a small fix to placement of sidebar toggle and title.

Before (current `master`):

before

After (`fix-toggle`):

after

_Fixes #526_


trvrb:
Merging as this is a small bug fix.

PR #546, opened 2018-05-02
Fix frequencies x range

trvrb:
This fixes the issue where frequencies panel didn't fill the axis. It was caused by basing xScale off of `ticks` rather than `pivots`. A simple replacement here solved the issue.

Before:

freq-before

After:

freq-after

_Fixes #542_


trvrb:
Merging as this is a small bug fix.

Issue #262, opened 2017-03-15
QA pass & issue creation


colinmegill:

jameshadfield:
What is this issue? @colinmegill @trvrb

trvrb:
This is the beginning of the "mobile" epic #263 where the aim is to actually make this good on mobile. First step is just identifying issues.

PR #550, opened 2018-05-06
Styling and UI improvements

trvrb:
@jameshadfield ---

This PR represents a fairly major pass at styling and UI. In doing so I had two goals in mind: first, to improve consistency and clarity of UI targets and second to improve mobile behavior. All changes are cosmetic. No changes to function. Specific additions:

1. The biggest change is to sidebar behavior. I realized that the tiny chevron toggle was the root of the issue. I changed general behavior to always show sidebar when display is large and to highlight toggle when display is small. I borrowed the basic metaphor from [React docs](https://reactjs.org/docs/hello-world.html), where on mobile a circular button appears in corner to toggle between nav and text. I did the same thing here to provide an always on toggle to swap between content and sidebar on mobile.

Here is mobile before the PR:

mobile-before

The toggle in the top left was nearly impossible to hit and when first encountering the UI it wasn't clear that you should even bother with the toggle.

Here is mobile after the PR:

mobile-after

Tapping big green button brings up controls (with obvious iconography). Hitting again or swiping over hides sidebar.

I don't think it will ever be a great experience to manipulate auspice on mobile, but getting basic nav down is important as this should make narratives possible.

The only major mobile issue left is tree panning and I'm working on this now.

2. Make all "buttons" uppercase text to improve consistency of what can be clicked and what can't.

3. Make "tab" elements in panels to highlight manipulable objects. Before it wasn't at all obvious that "AA" and "NT" in entropy were meant to be clicked on. Before this looked like:

div-before

After the PR, this looks like:

div-after

I'm using this same UI for "RESET LAYOUT" in tree as well.

4. Keep titlebar inline so that there are only two scrollable surfaces (sidebar + content) rather than three (sidebar + content + titlebar) as there are currently.

5. Minor changes to improve controls spacing, slightly embolden title text, thin sidebar drop shadow, adjust card margins, properly animate sidebar slide behavior, improve map animation buttons and disable keyboard search in Select boxes.

_This PR is up at [dev.nextstrain.org](http://dev.nextstrain.org/) to allow mobile behavior to be tested. This is a simple alias to `auspice-dev` Heroku app._

_Resolves #262._


trvrb:
* I mainly tested on my iPhone 6 and did some of the chrome emulation.  I do a more thorough investigation here.

* Removing toggle from desktop was a purposeful decision. When screen space is available I don't think there's ever a need to hide sidebar. Question is whether these two UIs are enough to cover all screen sizes. I thought so, but open for discussion.

jameshadfield:
Ahh ok - let's discuss in person tomorrow. I wouldn't worry about the mobile fixes for now.

trvrb:
Some examples of mobile layouts:

### iPhone 6 Plus

iphone-6-plus-vertical

iphone-6-plus-horizontal

### iPhone X

iphone-x-vertical

iphone-x-horizontal

### iPad

ipad-vertical

ipad-horizontal

trvrb:
My thinking with regards to always on sidebar on desktop and sidebar toggle on mobile was based largely on [React Docs](https://reactjs.org/docs/hello-world.html). With them (and many similar sites) you always have a sidebar with nav when the display is large enough and then when the display gets to a certain smallness you get a toggle between the two. Trying to shoehorn in sidebar toggle on desktop felt like it was being counter-productive, in that getting a UI that works for toggle on desktop and toggle on mobile was being super hard. Just aiming for toggle on mobile made things much easier and fit the established web pattern.

trvrb:
Merging this now.

PR #552, opened 2018-05-07
Separate Nucleotide and Gap/N Mutations in Hover infoPanel

emmahodcroft:
This separates nucleotide mutations into two groups, so that 'N's and '-'s do not prevent the user from seeing base-to-base changes. 

Example with both types of mutations:
infopanel_n_4

If you only have Gaps/Ns, this is all that's shown:
infopanel_n_2

Similarly, if only nucleotide (base-to-base) mutations, that's only shown:
infopanel_n_3

This is my first foray into javascript so there may be better/more standard ways to do some of this, please edit if so :)

### DIFF
--- a/src/components/tree/infoPanels/hover.js
+++ b/src/components/tree/infoPanels/hover.js
@@ -108,10 +108,36 @@ const getMutationsJSX = (d, mutType) => {
   if (mutType === "nuc") {
     if (typeof d.muts !== "undefined" && d.muts.length) {
       const nDisplay = 9; // max number of mutations to display
-      const n = d.muts.length; // number of mutations that exist
-      let m = d.muts.slice(0, Math.min(nDisplay, n)).join(", ");
-      m += n > nDisplay ? " + " + (n - nDisplay) + " more" : "";
-      return infoLineJSX("Nucleotide mutations:", m);
+	  const nGapDisp = 4; // max number of gaps/Ns to display
+
+	  //gather muts with N/-
+	  let ngaps = d.muts.filter(mut => {return mut.slice(-1) == "N" || mut.slice(-1) == "-"
+										|| mut.slice(0) == "N" || mut.slice(0) == "-";})
+	  const gapLen = ngaps.length; // number of mutations that exist with N/-
+	  //gather muts without N/-
+	  let nucs = d.muts.filter(mut => {return mut.slice(-1) != "N" && mut.slice(-1) != "-"
+										&& mut.slice(0) != "N" && mut.slice(0) != "-";})
+	  const nucLen = nucs.length; //number of mutations that exist without N/-
+	  
+	  let m = nucs.slice(0, Math.min(nDisplay, nucLen)).join(", ");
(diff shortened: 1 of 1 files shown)


src/components/tree/infoPanels/hover.js
```diff
@@ -110,31 +110,36 @@ const getMutationsJSX = (d, mutType) => {
       const nDisplay = 9; // max number of mutations to display
       const nGapDisp = 4; // max number of gaps/Ns to display
 
-      //gather muts with N/-
-      let ngaps = d.muts.filter(mut => {return mut.slice(-1) == "N" || mut.slice(-1) == "-"
-                                        || mut.slice(0) == "N" || mut.slice(0) == "-";})
+      // gather muts with N/-
+      const ngaps = d.muts.filter((mut) => {
```
trvrb: Arrow functions are clearer with parenthesis surrounding argument.

src/components/tree/infoPanels/hover.js
```diff
@@ -110,31 +110,36 @@ const getMutationsJSX = (d, mutType) => {
...
+        return mut.slice(-1) === "N" || mut.slice(-1) === "-"
```
trvrb: String comparisons should always be done with `===` or `!==`: https://stackoverflow.com/questions/3586775/what-is-the-correct-way-to-check-for-string-equality-in-javascript

src/components/tree/infoPanels/hover.js
trvrb: If variable is never reassigned, use `const` instead of `let`.

src/components/tree/infoPanels/hover.js
```diff
@@ -110,31 +110,36 @@ const getMutationsJSX = (d, mutType) => {
...
+      if (nucLen === 0) {
         return infoLineJSX("Gap/N mutations:", mGap);
       }
       return (
         <p>
-        {infoLineJSX("Nucleotide mutations:", m)}
-        <br/><br/>
-        {infoLineJSX("Gap/N mutations:", mGap)}
+          {infoLineJSX("Nucleotide mutations:", m)}
+          <div height="5"/>
```
trvrb: A `div` will give you more control over spacing.

trvrb:
This is really great @emmahodcroft. I appreciate the screenshots as well. I just made a few small stylistic tweaks.

I use a "linter" to identify non-standard JavaScript and would not have noticed most of these otherwise. The linter behavior is specified here: https://github.com/nextstrain/auspice/blob/master/.eslintrc. To get it to work for me I just needed to enable a plugin in Atom (https://atom.io/packages/linter). The linter really helps me when coding JavaScript, notices typos, etc...

I think this is good to go however. Thanks so much for the contribution.

emmahodcroft:
Thank you, Trevor, for the comments! Javascript is definitely different from the strict world of Java, so it's good to learn. 
James did mention linters when we spoke about auspice last week, but I had completely forgotten about it until now. I will get one!

jameshadfield:
Thanks @emmahodcroft 
👍 for the linter - it's super useful

PR #556, opened 2018-05-08
Clade zoom behavior

trvrb:
Improvements to clade zoom and tree reset.

1. Tree reset via `redrawTree` is much simpler. It just calls `updateVisibleTipsAndBranchThicknesses`. Nuking the tree was having bad side effects for me. Mainly it introduced a bug on branch lengths so that after zooming to a clade and clicking "reset tree" you'd get the following (from live):

zoom-clade-bug

This PR resolves this common issue. By not nuking the tree you also retain the animation. I'm able to break the tree rarely enough that I don't think we need to be nuking it here.

2. Remove "view entire tree" filter button. I thought that having "reset layout" and "view entire tree" were overly synonymous and that "reset layout" was the more obvious of the two.

3. Slightly pad y so that small clades don't look weird. Before this change we had:

before

After this change we have:

after

4. Pad x when tip labels are present so that tip labels do not overflow. I hacked this a bit by just using a multiple of character count. I don't think it's worth it investing in a `getBBox` style solution.

Before this change:

before-clade

After this change:

after-clade

_Fixes #555_

### DIFF
--- a/src/components/info/info.js
+++ b/src/components/info/info.js
@@ -7,18 +7,6 @@ import { prettyString } from "../../util/stringHelpers";
 import { months } from "../../util/globals";
 import { displayFilterValueAsButton } from "../framework/footer";
 
-const resetTreeButton = (dispatch) => {
-  return (
-    <div
-      className={`boxed-item active-clickable`}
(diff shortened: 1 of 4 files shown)


trvrb:
Merging this based on our discussion today.

Issue #538, opened 2018-04-16
App errors out when requesting "multi-part" dataset not in manifest


trvrb:
Current, if you request a dataset that has multiple "parts", ie `flu` / `h3n2` with this dataset being in manifest, the server fails at parsing the URL. For example, requesting:

https://nextstrain.org/flu/h1n1pdm/ha/pandemic?user=lassa

errors out although

https://nextstrain.org/flu/h1n1pdm/ha/pandemic

works as `pandemic` was added to `manifest_guest.json`, but not `manifest_lassa.json`. This is because the server isn't properly parsing these requests. Calling this locally shows the error:

```
API request: /charon?request=json&path=flu_h1n1pdm/ha/pandemic_meta.json&s3=live
API request: /charon?request=json&path=flu_h1n1pdm/ha/pandemic_tree.json&s3=live
Error: ENOENT: no such file or directory, stat '/Users/trvrb/Documents/src/auspice/data/flu_h1n1pdm/ha/pandemic_meta.json'
API request: /charon?request=json&path=flu_h1n1pdm/ha/pandemic_tip-frequencies.json&s3=live
Error: ENOENT: no such file or directory, stat '/Users/trvrb/Documents/src/auspice/data/flu_h1n1pdm/ha/pandemic_tree.json'
Error: ENOENT: no such file or directory, stat '/Users/trvrb/Documents/src/auspice/data/flu_h1n1pdm/ha/pandemic_tip-frequencies.json'
```

With the manifest in place this becomes:

```
API request: /charon?request=json&path=flu_h1n1pdm_ha_pandemic_meta.json&s3=live
API request: /charon?request=json&path=flu_h1n1pdm_ha_pandemic_tree.json&s3=live
```

barneypotter24:
I encounter the same error running locally with non-HA or NA flu segments.

After updating `manifest_guest.json` I can navigate to http://localhost:4000/flu/seasonal/h3n2/mp/2y using drop-down menus, however I cannot navigate there by URL request.

```
Charon API request: /charon?request=json&path=flu_seasonal_h3n2_ha_2y_meta.json&s3=live
Charon API request: /charon?request=json&path=flu_seasonal_h3n2_ha_2y_tree.json&s3=live
Charon API request: /charon?request=json&path=flu_seasonal_h3n2_ha_2y_tip-frequencies.json&s3=live
Charon API request: /charon?request=json&path=flu_seasonal/h3n2/mp/2y_meta.json&s3=live
[...]

PR #576, opened 2018-06-05
Replace all slashes with underscores when getting dataset path

barneypotter24:
Replaces all instances of forward slashes with underscores when parsing a URL to determine the path to search for JSONs. Previously only replaced the first occurrence with an underscore, causing files with multiple underscore-delimited fields to not be found except when added to `manifest.json`.

_Fixes #538._


trvrb:
This fixes issue #538.

trvrb:
Tested and working for multipart flu datasets. Merging.

PR #581, opened 2018-06-12
fix narrative twitter errors

jameshadfield:
This removes fatal errors when using offline (as `window.twttr` was `undefined`), but not sure if it fixes your errors @tsibley.

Unfortunately we can't use `react-twitter-embed` (which we use on the static site) as the markdown is being turned into HTML on the server. I'm sure it'd be possible to run the twitter widget on the server side, but I'm hoping there's a simpler fix.


tsibley:
Thanks, James.  This indeed fixes the issue I was running into with Firefox's tracking protection.

PR #583, opened 2018-06-22
Improvements for running sans manifest

tsibley:
See the individual commit messages for each structured change and the rationale.

### DIFF
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -10,6 +10,13 @@ const makeDataPathFromParsedParams = (parsedParams) => {
   return tmp_levels.map((d) => d[1]).join("_");
 };
 
+export const makeDataPathFromPathname = (pathname) => {
+  return pathname
+    .replace(/^\//, '')    // strip leading
+    .replace(/\/$/, '')    //   and trailing slashes
+    .replace(/\//g, '_');  // replacing all internal ones with underscores
+};
+
 /* match URL pathname to datasets (from manifest) */
 export const getDatapath = (pathname, availableDatasets) => {
   if (!availableDatasets) {return undefined;}
@@ -17,7 +24,7 @@ export const getDatapath = (pathname, availableDatasets) => {
   if (parsedParams.valid) {
     return makeDataPathFromParsedParams(parsedParams);
   }
-  return pathname.replace(/^\//, '').replace(/\/$/, '').replace(/\//g, '_');
+  return makeDataPathFromUrl(pathname);
 };
 
 export const chooseDisplayComponentFromPathname = (pathname) => {
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -12,9 +12,9 @@ const makeDataPathFromParsedParams = (parsedParams) => {
 
 export const makeDataPathFromPathname = (pathname) => {
   return pathname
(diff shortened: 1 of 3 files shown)


huddlej:
This works for me!

tsibley:
Thanks for testing, @huddlej!

src/actions/navigation.js
```diff
@@ -10,14 +10,21 @@ const makeDataPathFromParsedParams = (parsedParams) => {
...
+
 /* match URL pathname to datasets (from manifest) */
 export const getDatapath = (pathname, availableDatasets) => {
   if (!availableDatasets) {return undefined;}
   const parsedParams = parseParams(pathname, availableDatasets);
   if (parsedParams.valid) {
     return makeDataPathFromParsedParams(parsedParams);
   }
-  return pathname.replace(/^\//, '').replace(/\/$/, '').replace(/\//g, '_');
+  return makeDataPathFromUrl(pathname);
```
jameshadfield: This is undefined - do you mean `makeDataPathFromPathname`?

tsibley: Yes, I did!  I thought I had tested this, but I now realize I was actually testing the `return` two lines up on line 25. My test cases were:

1. `/zika` and `/flu/seasonal/h3n2/ha/3y` (and variants) *without* a manifest present
2. `/zika` and `/flu/seasonal/h3n2/ha/3y` (and variants) *with* a manifest present

but it should have also included:

3. Some dataset _not_ in the manifest but loaded *with* a manifest present

I'll fix this up.

tsibley: (The reason for the name difference is that I had originally chosen `makeDataPathFromUrl`, but then adjusted it to parallel `makeDataPathFromParsedParams`.)

tsibley:
> dataset name in the sidebar is nice (we can tweak the layout / replace _ later)

I chose explicitly not to replace `_` as then what's displayed matches the filename, which is useful for folks operating with custom files sans-manifest.

tsibley:
Repushed with adjusted changes addressing the misspelled function name and accounting for viewing a dataset that's not in the loaded manifest.

Re: linting: Ah ha, I didn't realize a linter was in use.  When running `npm run lint` I get a _lot_ of errors/warnings, though, including in files I didn't touch:

```
✖ 477 problems (407 errors, 70 warnings)
```

jameshadfield:
Yup, there are plenty of linting errors. But that doesn't mean we should introduce more! A bit of history: linting was introduced quite late in the piece, so over time i've been cleaning things up file by file.

jameshadfield:
https://eslint.org/docs/rules/no-multi-spaces is a pretty ridiculous rule tho, we shouldn't be following that

tsibley:
> Yup, there are plenty of linting errors. But that doesn't mean we should introduce more!

I agree in theory, but in practice how am I supposed to know what's new?  How are you doing this?

jameshadfield:
image
I can see them in atom.

tsibley:
Ah. I'm not using Atom. I'll revisit the linting stuff later today.

tsibley:
Thanks for adding that and merging.

trvrb:
@tsibley and @jameshadfield --- 

This is great. I like the display of dataset path when lacking manifest. Everything seems to be working for me.

PR #584, opened 2018-06-26
Upgrade compression-webpack-plugin, electron, mocha

trvrb:
@jameshadfield ---

I upgraded these three packages to fix GitHub's security warnings. Didn't seem to cause any problems. If this works for you, could you merge?


jameshadfield:
:+1:

PR #585, opened 2018-06-26
Expand dataset pathnames appropriately. Closes #575

### LINKED ISSUE #575: URL path is no longer redirecting appropriately
@jameshadfield ---

If I go to https://nextstrain.org/flu it properly loads the `flu/h3n2/ha/3y` dataset. However, the URL stays at https://nextstrain.org/flu. It should instead load as https://nextstrain.org/flu/h3n2/ha/3y as it did previously. I think this bug was introduced quite recently.

jameshadfield:


trvrb:
This is working for me and resolves the issue where https://nextstrain.org/flu doesn't automatically expand to https://nextstrain.org/flu/seasonal/h3n2/ha/3y.

PR #586, opened 2018-06-28
Meta safety checks

trvrb:
@jameshadfield ---

I went through and systematically tested removal of each field in `meta.json` and corrected errors associated with removal of each. I even got it to the point where the app works if `meta.json` is completely absent. Here is Zika from modular augur without a `meta.json` file:

metaless

It defaults to just the tree as entropy is impossible without `meta.annotations` and map is impossible without `meta.geo`. But I think this is pretty great. I tested this with other tree JSONs and it's working for other pathogens as well.

@rneher --- This is my proof of principle for intermediate tree JSON files in augur (https://github.com/nextstrain/augur/issues/134). They will be easily viewable.

I would recommend making this drag-and-drop, but I'll need help from @jameshadfield to make this happen.

### DIFF
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -103,9 +103,8 @@ const fetchDataAndDispatch = (dispatch, datasets, query, s3bucket, narrativeJSON
       values.forEach((v, i) => {
         if (v) data.JSONs[promisesOrder[i]] = v; // if statement removes undefinds
       });
-      // console.log(data);
-      if (!(data.JSONs.meta && data.JSONs.tree)) {
-        console.error("Tree & Meta JSONs could not be loaded.");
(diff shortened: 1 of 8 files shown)


jameshadfield:
Looks good. Will make sure things work today -- could you modify https://github.com/nextstrain/static/blob/master/content/docs/03-bioinformatics-pipeline/05-exported-JSON-files.md please? 

re: drag & drop, technically it's not hard, but I remember discussing this with you in the past and deciding not to take auspice in this direction. Happy to revisit.

trvrb:
Metadata description updated: https://github.com/nextstrain/static/commit/6a52873dc5a93eb93d70bb2349965775791c98ba

PR #587, opened 2018-06-28
Reinstate (improved) download ability

### LINKED ISSUE #562: reinstate & improve download data button / page
* the UI could be improved
* it's currently broken and has been removed from the page

jameshadfield:
This PR re-enables downloading of data, and closes #562.

### Improvements
* The SVG is now saved as a single file, essentially mirroring the positioning of the panels as they appear in the browser.
* Information text is written to the SVG
* Pressing the magic key `d` is a shortcut for downloading the SVG
* The Nextstrain paper citation has been added to the auspice footer.
* Map panning is correctly rendered in the SVG _(most of the time)_

### Known bugs
* changing datasets updates the map in the DOM, but upon SVG download the old map is written. Unsure how to fix this (I've tried tearing down `window.L` and/or `this.state.map.remove()` within `<Map>`, however this causes fatal leaflet errors. I remember @colinmegill having troubles here so don't think the fix is simple.)
* sometimes map panning causes SVG elements outside the map-panel to be written to the SVG. I tried applying a `clipPath`, however this resulted in internal elements which crossed the boundary also being removed.

In general, the `mapDemesAndTransmissions` SVG is rather complicated and @colinmegill would have the most knowledge here. On the flipside, when the map contains errors they are rather obvious ;)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,9 +16,9 @@
       "integrity": "sha512-sW77BFwJ48YvQp3Gzz5xtAUiXuYOL2aMJKDwiaY3OcvdqBFurtYfOpSa4QrNyDxmOGRFSYzUpabU2m9QrlWE7w==",
       "dev": true,
       "requires": {
-        "chalk": "^2.0.0",
-        "esutils": "^2.0.2",
-        "js-tokens": "^3.0.0"
+        "chalk": "2.3.0",
+        "esutils": "2.0.2",
+        "js-tokens": "3.0.1"
       },
       "dependencies": {
         "ansi-styles": {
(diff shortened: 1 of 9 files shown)


trvrb:
@jameshadfield --- This is great. Some small things that I noticed in review:

1. The download modal should be larger given all the text that's squeezed into it. This is especially apparent when window is small:

download_modal

I wonder about instead making it take up the entire display and give it a big x in the top right?

This is also because it's not visually consistent with the other modals. Other modals are dark grey background with rounded corners and white text. If you're going for a modal where you click outside, make it consistent with other modals in the app.

2. Newick should be downloaded as `.nwk` to match augur format.

3. I'd much prefer metadata as `.tsv` to `.csv`. This is my general preference across the board. Was csv an active decision?

jameshadfield:
@trvrb points 1-3 have been addressed now and I think this is good to go.

jameshadfield:
This PR closes #450 and closes #429

trvrb:
One small comment that's worth thinking about even if I don't have an immediate solution: The download modal states

>Showing 6 of 502 genomes, from 1 author, from 1 region, from 5 countries, dated Jul 2011 to Jan 2018.

But then clicking download "strain metadata" or "author metadata" downloads all 502 strains and 82 authors.

trvrb:
Go ahead and merge whenever you'd like @jameshadfield.

PR #589, opened 2018-07-02
Local data sans dev server

tsibley:
This restores the ability to run the non-dev server with local data. See the commit message for more details.


tsibley:
Ah, thanks for that catch, @jameshadfield. I've repushed with the fix.

Question for you: What's the status of the Electron build?

> Q: which npm script are you using to build the bundles to source local data? I can't see the appropriate one in package.json

I'm building with `npm run build` and running the server as `npm run server localData`. The build isn't parameterized on local/remote data; just the running server. (This is as it was, I just made it work again.)

trvrb:
>Question for you: What's the status of the Electron build?

Almost certainly broken.

tsibley:
In that case I won't try to determine if `path.join` in `index.js` (used by Electron) needs to be `path.resolve`.  (Thought I don't _think_ it does.)

tsibley:
@jameshadfield Ping. This is ready for re-review.

Issue #598, opened 2018-07-19
Electron app


jameshadfield:
The electron app was once functional but needs debugging / improvements. Commit https://github.com/nextstrain/auspice/commit/0ed5acf7539aed818bd59550d9ed75e8a7f3ce96 has/will remove it from the codebase, but this should be used to fix this issue.

PR #599, opened 2018-07-19
Improve README & remove electron

jameshadfield:
Various README improvements and `npm` script improvements.

Re: electron, see #598. As it stands it doesn't work and is a potential source of confusion.


trvrb:
Looks good.

PR #603, opened 2018-07-20
simplify server

jameshadfield:
This PR removes the need for `server.js` to be transpiled. Running the server is now simply `node server.js [dev] [localData]` (or `npm run server`, `npm run dev` etc). This results in speed increases in performance mode, removes the build step, and makes development easier. I've got ideas to improve things even further but they will be on top of this PR.

Once this is in `release`, https://github.com/nextstrain/nextstrain.org can be similarly simplified.

@tsibley can you double check things work as expected for you?


tsibley:
This appears to work for me. `npm run dev` (and `dev:local`) work without a `dist/` directory, while `npm run server` (and `server:local`) first requires `npm run build`. I think this is the expected behaviour, right?

jameshadfield:
> npm run dev (and dev:local) work without a dist/ directory, while npm run server (and server:local) first requires npm run build

Correct -- dev is building a bundle behind the scenes, but you can't access it I don't think. This allows changes to the source code to update the bundle on the fly without needing to rebuild.

trvrb:
This all looks great James. Nice to see package.json simplified as well.

PR #604, opened 2018-07-25
Move logic to server

### LINKED ISSUE #588: Source JSONs from github repos
URLs such as nextstrain.org/community/X/Y should attempt to source JSONs from the github organisation `X` and repository `Y`. This requires that augur-produced JSONs are committed, e.g. to `repoName/data/auspice` or `repoName/auspiceJSONs`. 

If it's multi-part dataset (e.g. nextstrain.org/community/X/flu/seasonal/h3n2/3y), then the repository should be called `flu` and the JSONs named `flu_seasonal_h3n2_3y_tree.json` etc.

Considerations:
* use rawgit including compression
* allow the repo `Y` to be named `nextstrain-Y` (e.g. you can imaging that `myLabGitHub/flu` is already taken, and it isn't obvious that it's a nextstrain related repo) -- @trvrb do we want to enforce this?

### LINKED ISSUE #601: Narratives won't load without a manifest file
While trying to get a narrative markdown file to load without a manifest, auspice tries to load an `undefined` pathogen. Adding a proper `manifest_guest.json` file to the data directory corrects the issue.

### LINKED ISSUE #602: Narrative URLs need to have a query string even if it is empty
Trying to define a narrative with the following Markdown url fails:

```
`nextstrain url=/ebola/`
## Some header
```

with the following javascript traceback:

```
Error in fetchNarrativesAndDispatch TypeError: Cannot read property 'length' of undefined
    at parseParams (parseParams.js:16)
    at getDatapath (navigation.js:23)
    at loadData.js:132
```

However, changing the markdown to the following with an empty query string fixes the problem.

```
`nextstrain url=/ebola/?`
## Some header
```

This is a surprise especially for users who don't understand query strings.

jameshadfield:
This PR moves relevant functionality to the (auspice) server. There were lots of changes, so there's bound to be bugs -- some testing would be much appreciated!

### Outward facing changes
* The data source (e.g. staging server? local data?) is now specified by prepending the source to the URL
    * To access a local copy of zika, the URL is `localhost:4000/local/zika`
    * A list of local datasets is available at `localhost:4000/local` (same for staging)
    * Github builds (i.e. JSONs committed to the repo in the `auspice` directory) are available at `localhost:4000/community/ORGNAME/REPONAME/EXTRA/PATHS`. For instance `localhost:4000/community/nextstrain/WNV/NA` loads the WNV build from https://github.com/nextstrain/WNV/. resolves #588
    * Appending `:local` to npm commands to change the server functionality is now deprecated.
    * In the future, the list of "sources" can be expanded, e.g. `s3::...` could specify an arbitrary S3 bucket.
* The local data folder can be specified on the command line via `npm run server -- local:PATH`, `node server.js local:PATH`, `npm run dev -- local:PATH` etc. The `PATH` must currently be relative (to the auspice directory), but this will be improved as we make `auspice` globally installable. The default path is still `data/` in the auspice directory.
* Tanglegrams are now specified like `/flu/seasonal/h3n2/3y/ha-na` instead of `/flu/seasonal/h3n2/3y/ha?tt=na`. A warning notification is given if you use the old syntax.
* More descriptive errors are displayed on the splash page (when they arise)
* A bunch of narrative bugs have been fixed (but plenty remain)
    * resolves #602 resolves #601 

### Internal changes
* The server combines the tree & meta JSONs into a structure more similar to schema v2. This means the initial load only requires a single fetch from the client. This will make a transition to v2 much easier, and allow patching of the v1 schemas before they arrive at the client.
* The manifest JSON is loaded by the server & turned into a list of available datasets, which is given to the client when needed. This is also build for local data by scanning the directory. The structure of the underlying manifest is only really used for ordering these paths. Changes to the manifest format are welcome, but beyond the scope of the PR.
* Frequencies JSONs are only fetched when the metadata specifies this panel
* Page changes & data loading logic is much simplified

UPDATE: once this is in `master`, the the nextstrain.org server will break. https://github.com/nextstrain/nextstrain.org/pull/2 will fix this.


tsibley:
I'll take the time to go through this.  Some initial thoughts based on your notes above:

Are community builds obviously denoted as such in the UI? I realize the URL contains the word, but most people never look at URLs. We probably want to make clear that community builds are just that in the UI.

> Tanglegrams are now specified like `/flu/seasonal/h3n2/3y/ha-na` instead of `/flu/seasonal/h3n2/3y/ha?tt=na`.

Does this mean that we're forbidding filenames from containing hyphens then?

trvrb:
I think this is a really great direction James. Declaring sources in URL rather than npm flags is much more convenient. I have similar immediate response as @tsibley on 

>Tanglegrams are now specified like /flu/seasonal/h3n2/3y/ha-na instead of /flu/seasonal/h3n2/3y/ha?tt=na.

I would do this one of two ways.

1. Keep the original URL of `/flu/seasonal/h3n2/ha/3y?tt=na` (note that live is `ha/3y` rather than `3y/ha`). This fits with the idea that we are source the `/flu/seasonal/h3n2/ha/3y` but with "view" parameter `tt=na`.

2. If moving into the path, please keep ordering consistent. I would prefer `/flu/seasonal/h3n2/ha:na/3y`. Don't want hyphen as this may be used in segment name. Colon seems pretty safe and semantically clear. This fits with the idea that we are "sourcing" both segments.

We could declare that segment / gene has to be the final token in the path, but this may be difficult to enforce.

I don't know if I prefer option (1) or option (2).

trvrb:
>Are community builds obviously denoted as such in the UI? I realize the URL contains the word, but most people never look at URLs. We probably want to make clear that community builds are just that in the UI.

Definitely want this. And make it very clear for our builds and community builds that build is housed at github.com/nextstrain/zika or github.com/trvrb/flu or wherever.

jameshadfield:
Thanks both
* i'll change `ha-na` to `ha:na`
* The "source" of the datasets is known to the client, so a separate PR can modify the header to clearly display this. 

All the other builds are served using lowercase (i.e. you can go to `/fLu/SeaSONal/` and it _should_ be just fine), but the github stuff needs to match the case of the repo/org.

tsibley:
Instead of displaying a warning notification about `?tt=…` being "deprecated", can we just correct the URL for them automatically?  It's nice to leave automatic redirects behind at old URLs so that they keep working. People may have emailed around links with `?tt=` or embedded them in write-ups, and we don't want those to suddenly stop working.

tsibley:
The `/status` and `/status/staging` pages no longer seem to work. The server spits out messages like:

```
Charon API request: /charon?request=json&want=/staging/ebola&type=tree
Server doesn't know how to handle the request "undefined"
```

tsibley:
@jameshadfield Can you explain what your plan is for `treeToo` vs. `treeTwo`?

I know that on master the second tree is referred to as `treeToo` everywhere. This branch introduces some instances of `treeTwo` limited to the JSON the server returns and client code handling it.

src/actions/recomputeReduxState.js
```diff
@@ -458,26 +461,26 @@ export const createStateFromQueryOrJSONs = ({
...
+  //   frequencies.version++;
+  //   frequencies.matrix = computeMatrixFromRawData(
+  //     frequencies.data,
+  //     frequencies.pivots,
+  //     tree.nodes,
+  //     tree.visibility,
+  //     controls.colorScale,
+  //     controls.colorBy
+  //   );
+  // }
```
tsibley: I'd recommend just removing this block of unused code rather than commenting it out. Chunks of commented out code are a bit of a "code smell", especially when using a version control system. They're rarely necessary and tend to just accumulate as clutter.

tsibley:
Going to `/local/WNV/NA` doesn't work for me, even though `data/WNV_NA_{meta,tree}.json` exist. If I rename the files to lowercase, then going to `/local/WNV/NA` works but changes the URL to `/local/wnv/na`. Is forbidding uppercase letters in the path the expected behaviour?

src/server/sourceSelect.js
```diff
@@ -49,7 +49,11 @@ const constructPathToGet = (source, providedUrl) => {
   let datasetFields; // this _does not_ take into account the 2nd tree
   let treeName;
 
-  const parts = providedUrl.replace(/^\//, '').replace(/\/$/, '').split("/");
+  const parts = providedUrl
+    .replace(/^\//, '')
+    .replace(/\/$/, '')
+    .replace("live/", "")
```
tsibley: This should probably be anchored to the start of the string, otherwise no dataset can include a path segment ending in `live`.

jameshadfield:
Thanks @tsibley 

> The /status and /status/staging pages no longer seem to work. 

Fixed now.

> I'd recommend just removing this block of unused code...

Fixed (actually reinstated it, it shouldn't have been commented out). For the record, I'm all for not leaving commented code around (i've put in issues in auspice about exactly this), but it's a view not universally shared.

> Is forbidding uppercase letters in the path the expected behaviour?

Perhaps this could be a separate PR. I'm not sure of the best thing to do here. Currently paths have to be lowercase, except for Github. I want the URL to be uppercase if it looks better, but it `/wnv/na` shouldn't fail. We should make an issue with a bunch of situations and what behavior we want. 

> Instead of displaying a warning notification about ?tt=… being "deprecated", can we just correct the URL for them automatically?

Probably, but I don't know how big of an issue this will be. Certainly not for the narratives where this has never worked in the first place. 

> Can you explain what your plan is for treeToo vs. treeTwo?

`treeToo` will probably be replaced with `treeTwo` over time.

src/server/sourceSelect.js
jameshadfield: yup, done

src/components/splash/index.js
```diff
@@ -27,19 +106,28 @@ class Splash extends React.Component {
...
+              <div>
+                <p style={{color: "rgb(222, 60, 38)", fontWeight: 600, fontSize: "24px"}}>
+                  {"😱 404, or an error has occured 😱"}
+                </p>
+                <p style={{color: "rgb(222, 60, 38)", fontWeight: 400, fontSize: "18px"}}>
+                  {`Details: ${this.props.errorMessage || this.state.errorMessage}`}
+                </p>
+                <p style={{fontSize: "16px"}}>
+                  {"If this keeps happening, or you believe this is a bug, please "}
+                  <a href={"mailto:hello@nextstrain.org"}>{"get in contact with us."}</a>
```
tsibley: I think @trvrb wanted to avoid real `mailto:` links for spam reasons?

tsibley:
> Currently paths have to be lowercase (same as currently)

https://nextstrain.org/WNV/NA works currently, whereas it doesn't on this branch.

I think we should be allowing, respecting, and preserving whatever case is provided. Paths should be case-sensitive, like nearly all URLs. That's the case currently on master, but the behaviour is changed on this branch, and I don't understand why.

> `treeToo` will probably be replaced with `treeTwo` over time.

Any objection to a commit which does this rename all at once instead of bit by bit?  It seems straightforward.  (Famous last words…?)  I'd prepare it.

>> Instead of displaying a warning notification about ?tt=… being "deprecated", can we just correct the URL for them automatically?

> Probably, but I don't know how big of an issue this will be. Certainly not for the narratives where this has never worked in the first place.

Well yes, we can't know the usage very well (without detailed server access logs), but it's polite to leave behind redirects when moving around public URLs. I'm not worried about the narratives. Why break [links from 3 months ago](https://twitter.com/hamesjadfield/status/986322812301275136) when we don't have to?  :-)

tsibley:
All my notes from this review are above, either as general comments or line-specific comments.

jameshadfield:
> Any objection to a commit which does this rename all at once instead of bit by bit? It seems straightforward. (Famous last words…?) I'd prepare it.

Go for it.

I'll make the URLs all case sensitive, redirect `tt=...` and then merge if there are no further objections

src/server/sourceSelect.js
```diff
@@ -0,0 +1,121 @@
...
+    return {available: global.LIVE_MANIFEST, source};
+  } else if (source === "staging" && global.STAGING_MANIFEST) {
+    return {available: global.STAGING_MANIFEST, source};
+  }
+  return {available: undefined, source};
+};
+
+
+const guessTreeName = (parts) => {
+  const guesses = ["HA", "NA", "PB1", "PB2", "PA", "NP", "NS", "MP", "L", "S"];
```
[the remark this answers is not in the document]
jameshadfield: Tree name specified in the JSON (& required in the schema).

src/components/splash/index.js
jameshadfield: What's your understanding of the potential for spam? If you can see a way of keeping the link active without `mailto:` then please go for it. I don't like how on nextstrain.org they aren't links.

jameshadfield:
* The URLs & Paths are now case sensitive. Tested on a few different sets without any issues.
* `tt=...` URLs are changed to the new syntax and a warning notification is shown

I'll merge soonish if there are no objections. Thanks @tsibley for looking over this.

PR #606, opened 2018-07-30
align branch labels according to tree orientation:

rneher:
this orients the amino acid labels properly on the left-facing tree. 
old:
image
new:
image


jameshadfield:
Thanks @rneher 👍

Issue #547, opened 2018-05-02
Map transmissionIndices bug


jameshadfield:
Sometimes this error occurs:
```
mapHelpersLatLong.js?a8f2:402 Uncaught TypeError: Cannot read
   property 'forEach' of undefined
       at eval (mapHelpersLatLong.js?a8f2:402)
       at Array.forEach (<anonymous>)
       at eval (mapHelpersLatLong.js?a8f2:389)
       at Array.forEach (<anonymous>)
       at updateTransmissionDataColAndVis (mapHelpersLatLong.js?a8f2:387)
       at updateDemeAndTransmissionDataColAndVis
   (mapHelpersLatLong.js?a8f2:428)
       at Map.maybeUpdateDemesAndTransmissions (map.js?e5a9:312)
       at Map.componentWillReceiveProps (map.js?e5a9:105)
       at callComponentWillReceiveProps
   (react-dom.development.js?cada:6389)
       at updateClassInstance (react-dom.development.js?cada:6575)
   ----
```
Caused by `transmissionIndices[id]` being undefined.

A try-catch fix was implemented in 1.18.1 (https://github.com/nextstrain/auspice/commit/ee4bca6a58a8175f7fdd1cfaf60dd1796fc818b4) however the underlying issue persists.

jameshadfield:
@trvrb are you familiar with colin's code here?

trvrb:
Can you tell if this was introduced recently? Or is persistent? And is there a reliable way to reproduce?

jameshadfield:
JSONs in the slack channel. May be a JSON thing, i've never seen it before.

rneher:
fundamentally, the issue here is to decide what to do/how to fail when the meta.json doesn't contain coordinates. in the case above, this happened because an ancestral state was inferred that wasn't even in the country list (I think we pinned the ancestral state to vietnam, but vietnam wasn't sampled and hence excluded from the `geo` fields in the meta json).

jameshadfield:
Closed in #608

PR #608, opened 2018-08-02
this fixes (papers over) issue #547. 

rneher:
The issue was missing coordinates which result in missing transmissions from the index, which later results in 'type undefined' errors. This PR just sets lat/long to 0.0 instead. This will result in silly locations, but at least it doesn't break. This is a more general issue as to what should happen with incomplete geo-info.

### DIFF
--- a/src/components/map/map.js
+++ b/src/components/map/map.js
@@ -18,6 +18,7 @@ import { MAP_ANIMATION_PLAY_PAUSE_BUTTON } from "../../actions/types";
 // import { incommingMapPNG } from "../download/helperFunctions";
 import { timerStart, timerEnd } from "../../util/perf";
 import { lightGrey, goColor, pauseColor } from "../../globalStyles";
+import { errorNotification } from "../../actions/notifications";
 
 /* global L */
 // L is global in scope and placed by leaflet()
@@ -181,7 +182,8 @@ class Map extends React.Component {
         demeData,
         transmissionData,
         demeIndices,
-        transmissionIndices
+        transmissionIndices,
+        demesMissingLatLongs
       } = createDemeAndTransmissionData(
         this.props.nodes,
         this.props.visibility,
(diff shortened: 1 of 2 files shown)


jameshadfield:
Hey @rneher -- could you run these JSONs through `augur validate <JSONs>` -- i'm interested if we pick up that they are "invalid". 

I'm happy enough with a temporary solution to stop things crashing, but this is a deeper seated issue in the design of the map component which assumes that every value has a corresponding lat-long. I guess we should (a) not show links to and from "unknown" demes (unknown in the sense that their lat/long is unknown) and (b) display an error to the user.

jameshadfield:
Thanks... i'll fix `export` ASAP as this should definitely be an error. The issue of how Auspice should deal with "invalid" JSONs is a thorny one...

rneher:
note that these were the ancient jsons that caused this issue back in May. Don't even know where they came from, you posted them on slack. My main motivation was to shorten the long issue list of auspice.

trvrb:
I'm in favor of merging this. Having lots of redundancy in auspice seems like the way to go. Even if these JSONs won't validate, still better to have auspice not crash.

trvrb:
I see. I'm agnostic in this case.

rneher:
I tend to agree that not displaying them is better than drawing to some place in the Atlantic. But it is difficult for the user to know that smth is amiss if only a few lines are missing. My main reason to go back to this was to figure out where exactly this goes wrong for what reason. We had previously just caught it right before drawing. The root cause is info in the meta_json and that previously resulted in a corrupted the transmission data structure.

jameshadfield:
Augur PR modifies `augur validate` to show the following when trying to validate these JSONs:

```
Validating that /Users/james/nextstrain/auspice/data/pater_meta.json and /Users/james/nextstrain/auspice/data/pater_tree.json are internally consistent...
	ERROR:  "vietnam", a value of the geographic resolution "country", appears in the tree but not in the metadata.
	ERROR:  	This will cause transmissions & demes involving this location not to be displayed in Auspice
	WARNING:  Color option "country", which contains a color_map, is missing "vietnam"
```

jameshadfield:
Here's a possible fix (feel free to trash a1c66ac if it's not desired):

Demes/transmissions involving undefined lat/longs are not rendered & this error notification is shown on page load:

image

jameshadfield:
@rneher @trvrb shall we merge this into master as currently stands?

rneher:
My main reason to get back to this was to understand why this error happened and to reduce the number of `high priority bug` tags in the issue list. I think this is good to merge.

jameshadfield:
Great!
Nice work on working out the root cause of this issue, and it's now well communicated in both auspice (via error notification box) and augur (via `augur validate`)

Issue #483, opened 2018-01-09
y-axis in clock tree display


rneher:
The clock tree display could use an y-axis with divergence.

jameshadfield:
Added in #610 -- thanks @rneher

PR #610, opened 2018-08-03
allow negative dates, y labels, better spacing

rneher:
This fixes the long-standing issue #483 and other non-documented problems with the tree grid. It now allows
 * negative dates (yes, there were pathogens before Christ)
 * adds ylabels indicating divergence in the clock display. This could use some tweaking
 * improves the spacing of grid lines 
 * removes the decimal points from dates like 2015.0 (unless major grid resolution is <1y).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -41,6 +41,7 @@ rules:
   import/prefer-default-export: off
   no-multi-spaces: ["error", { ignoreEOLComments: true }]
   no-labels: off
+  no-unused-labels: off
   no-continue: off
   no-unneeded-ternary: ["error", { "defaultAssignment": true }]
   quote-props: ["error", "as-needed"]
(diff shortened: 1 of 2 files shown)


jameshadfield:
Looks great -- thanks @rneher 

I added some comments and cleaned up the code a bit in c7393ea

And then I added horizontal grid lines for clock-layouts in 289add4.  If this isn't a good direction to go, i'll remove the commit before merging into master.

screen shot 2018-08-04 at 13 52 49

rneher:
great. I think the horizontal grid lines are useful.

trvrb:
I really like the horizontal grid lines for clock display. This labeling makes clock display much less opaque.

trvrb:
Actually... I'm getting a bug where if I toggle back and forth between "rectangular" and "clock" the horizontal grid lines will appear in front of the tree.

jameshadfield:
Ahhh merged without seeing that message! I'll fix the bug now

PR #612, opened 2018-08-04
Tree render order

jameshadfield:
This PR is in response to https://github.com/nextstrain/auspice/pull/610#issuecomment-410477963, which it fixes.

The main change is that tree elements are now collected into named groups, and these groups persist so that later renders (e.g. upon update) preserve the correct layering / ordering.
As a bonus, it's now much clearer to debug the elements as they are collected & named:

image

Additionally
* Vaccine dotted lines are removed as they were zero-length but being drawn in the DOM
* T-branch elements are not drawn into the DOM for clock/unrooted layouts

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -46,6 +46,7 @@ rules:
   no-unneeded-ternary: ["error", { "defaultAssignment": true }]
   quote-props: ["error", "as-needed"]
   prefer-const: ["error", {"destructuring": "all"}]
+  indent: ["error", 2, {"MemberExpression": "off"}]
 parserOptions:
   ecmaVersion: 6
   sourceType: module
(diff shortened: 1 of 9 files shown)


trvrb:
@jameshadfield ---

I really like collecting groups here. Makes a more inspectable SVG and will make an SVG that is easier to edit in Illustrator etc...

Two issues:

* There are two `id="majorGrid"` group elements. If two should exist, they should have different ids. Ah. One should be `minorGrid`.

* I was able to break confidence bands by just mousing around the tree resulting in confidence bands that refuse to show up. I couldn't reproduce this on live.

jameshadfield:
Yup, one `majorGrid` should have been `minorGrid`. That's fixed now.

trvrb:
This is great @jameshadfield. Everything working as far as I can tell. Please go ahead an merge.

Issue #510, opened 2018-02-27
keep state between dataset changes where possible


jameshadfield:
i.e. switching 3y -> 6y should maintain colorBy, tree display...

PR #613, opened 2018-08-04
Fix bugs when changing datasets with URL queries defined

### LINKED ISSUE #611: Changing datasets crashes with genotype colorby
@jameshadfield ---

I generally like the new behavior where visualization state is preserved when switching between datasets via the dropdown. Ie, if I'm at https://nextstrain.org/zika?l=clock and switch to ebola I arrive at https://nextstrain.org/ebola?l=clock. However, if the original visualization state doesn't exist in the target dataset then the app crashes out to 404. For example, if you start at https://nextstrain.org/zika?c=gt-NS1_349 and switch to ebola then the get the crash as ebola doesn't have `NS1` to color by. 

Actually... for non-genotype color bys you don't get a 404, but it's still unexpected behavior. For example, go from https://nextstrain.org/WNV/NA?c=wnv_strain and select zika and you end up at https://nextstrain.org/zika?c=wnv_strain, which then shows coloring by date.

This issue exists for geo resolutions in addition to color bys. For example, https://nextstrain.org/zika?r=region switches to https://nextstrain.org/ebola?r=region, which doesn't exist.

I don't know the best treatment for this. One option would be to return to the previous behavior of nuking state, but this is bad for things like panel grid. Fixing the crash on genotype color by seems like an obvious first step.

jameshadfield:
Closes #611

Fixed for the colorBy & geoResolution cases. 

This maintains the expected behavior when going from http://localhost:4000/local/flu/seasonal/h3n2/ha/6y?c=ep to the 3y dataset.

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -269,19 +269,41 @@ const checkAndCorrectErrorsInState = (state, metadata, query) => {
   if (!metadata.colorOptions) {
     metadata.colorOptions = {};
   }
-  if (Object.keys(metadata.colorOptions).indexOf(state.colorBy) === -1 && !state["colorBy"].startsWith("gt-")) {
+  const fallBackToDefaultColorBy = () => {
     const availableNonGenotypeColorBys = Object.keys(metadata.colorOptions);
     if (availableNonGenotypeColorBys.indexOf("gt") > -1) {
       availableNonGenotypeColorBys.splice(availableNonGenotypeColorBys.indexOf("gt"), 1);
     }
-    console.error("Error detected trying to set colorBy to", state.colorBy, "(valid options are", Object.keys(metadata.colorOptions).join(", "), "). Setting to", availableNonGenotypeColorBys[0]);
-    if (Object.keys(metadata.colorOptions).length > 0) {
+    if (availableNonGenotypeColorBys.length) {
+      console.error("Error detected trying to set colorBy to", state.colorBy, "falling back to", availableNonGenotypeColorBys[0]);
       state.colorBy = availableNonGenotypeColorBys[0];
       state.defaults.colorBy = availableNonGenotypeColorBys[0];
     } else {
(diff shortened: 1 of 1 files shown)


trvrb:
@jameshadfield ---

Nice work here. I did catch a small issue:

* Starting from http://localhost:4000/flu/seasonal/h3n2/ha/3y?c=ep if you go to `na` or go to `vic` you properly drop the `?c=ep` from the URL, but the resulting colorby is wrong in both cases. `na` gives an LBI coloring and `vic` gives a region coloring. Both should go to the default coloring by clade.

jameshadfield:
> Starting from http://localhost:4000/flu/seasonal/h3n2/ha/3y?c=ep if you go to na or go to vic you properly drop the ?c=ep from the URL, but the resulting colorby is wrong in both cases. na gives an LBI coloring and vic gives a region coloring. Both should go to the default coloring by clade.

This is now fixed in be70819

jameshadfield:
P.S. this PR resolves #510

trvrb:
Thanks for fixing the flu issue. One more bug here @jameshadfield. If I start at http://localhost:4000/flu/seasonal/h3n2/ha/3y?c=lbi and go to "zika" I get an empty URL as expected, however Zika then has date coloring rather than its proper default of country.

trvrb:
Oh... and another related bug (didn't think about filters until now). If I start at http://localhost:4000/ebola?f_authors=Kugelman_et_al and then go to "zika" I get a full app crash with `Uncaught TypeError: Cannot read property 'title' of undefined`.

jameshadfield:
> If I start at http://localhost:4000/ebola?f_authors=Kugelman_et_al and then go to "zika" I get a full app crash with Uncaught TypeError: Cannot read property 'title' of undefined.

Should be fixed in 79a3799 but needs testing

trvrb:
@jameshadfield ---

This is great. I can't find any more bugs. I think this new behavior of preserving state where possible definitely makes the app feel more unified. Please feel free to merge.

PR #614, opened 2018-08-05
fix bug in dataset dropdown options

### LINKED ISSUE #607: Dataset options not getting parsed correctly
@jameshadfield ---

It looks like there is a bug in how parsing of available datasets is occurring. There is no `pandemic` resolution for H3N2, Vic and Yam, but it shows up as an option in the dropdown:

dataset

This should only exist for H1N1pdm. This is how it was specified in `manifest.json`. When selected this dataset can't be found and the app throws a 404.

jameshadfield:
dataset matching now considers all previous entries in the hierarchy, not just the parent

This closes #607

### DIFF
--- a/src/components/controls/choose-dataset.js
+++ b/src/components/controls/choose-dataset.js
@@ -10,6 +10,10 @@ const renderBareDataPath = (source, fields) => (
   </div>
 );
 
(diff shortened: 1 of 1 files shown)


trvrb:
This is great. The fix makes sense and data selection is now working as intended. Please go ahead and merge.

PR #615, opened 2018-08-05
Show github links as dataset (in sidebar) for community URLs

jameshadfield:
For community URLs, this would change the dataset text in the sidebar to:

image


trvrb:
I think this is really great @jameshadfield. I made some small styling adjustments. If you find a better icon in Font Awesome (https://fontawesome.com/v4.7.0/icons/) feel free to adjust. `fa-code-fork` was a close contender.

I don't think I'd add anything else to the sidebar. We may consider a "Community Build" badge in the top right of main section (to the right of title) or something like this. Or may consider revising title block to include author information, but this is a larger UX decision.

If you're happy, please merge.

PR #617, opened 2018-08-06
Upgrade deps with security vulnerabilities

tsibley:
As prompted by GitHub's security scans: https://github.com/nextstrain/auspice/network/dependencies

See commit messages for more info about the process.


trvrb:
Nice to have these updated.

Issue #619, opened 2018-08-09
nucleotide genotype should shift zoom window in certain cases


jameshadfield:
Very minor bug here -- if the nucleotide base is changed in the dropdown to one outside the zoom window, the brush should move to be around that newly selected base. Currently, the brush doesn't move.

In this screenshot, the zoom window is ~ (0, 500), but the nuc position is 3000

![image](https://user-images.githubusercontent.com/8350992/43928756-e8a192d0-9be6-11e8-9964-d66e11b0787c.png)

jameshadfield:
Closed in #620

PR #620, opened 2018-08-10
Fix issue 619

emmahodcroft:
This fixes (hopefully) the issue [619](https://github.com/nextstrain/auspice/issues/619) where sometimes typing in a new nuc location would not zoom to that location. A test was added to check whether the nuc selected is within the visible region - if not, it re-zooms. 

I also improved the behaviour for when multiple nuc sites are selected and then the user switches to a gene whether none of those sites are inside the gene. Previously, it defaulted to "[gene] positions..." and didn't re-zoom. Now it sets position to 1 to trigger a colorBy event and thus a zoom to that gene. 

@jameshadfield can you check this seems to fix things?


jameshadfield:
:+1:

PR #621, opened 2018-08-10
Gap/N Excluded from Count/Entropy

emmahodcroft:
Addresses issue [199](https://github.com/nextstrain/augur/issues/199) (in augur). 

This change means mutations to/from 'N' or '-' (gap) aren't included in entropy and count calculations in the entropy graph. (Note that for the Lee TB dataset this means all nucleotide counts go to 1 - that's not a bug, it's real.) 

I also fixed a bug I noticed yesterday where 'N' or '-' at the start of a mutation wasn't being filtered correctly in the branch hover:
image

This now shows correctly:
image


jameshadfield:
:+1: this is great

Issue #622, opened 2018-08-10
flu NA gene disappears upon entropy panel zoom


jameshadfield:
For `h1n1pdm` the NA gene disappears when the zoom brush is dragged inwards from the right. These screenshots show this for https://nextstrain.org/flu/seasonal/h1n1pdm/na/12y

![image](https://user-images.githubusercontent.com/8350992/43976080-5c48688c-9c94-11e8-9427-81bff3aaefb5.png)
![image](https://user-images.githubusercontent.com/8350992/43976088-62ee1f92-9c94-11e8-914a-3280fcc1b4c6.png)

Interestingly, this does not happen with https://nextstrain.org/flu/seasonal/h3n2/na/12y !

jameshadfield:
Closed in #624

PR #624, opened 2018-08-12
fix issue 622, disappearing NA

emmahodcroft:
Should hopefully fix issue [622](https://github.com/nextstrain/auspice/issues/622), where NA gene disappeared when moving brush in from right.


jameshadfield:
👍 thanks @emmahodcroft

Issue #623, opened 2018-08-10
add entropy zoom bounds to URL


jameshadfield:
This will allow the zoom to be changed by paragraphs in the narrative 👍

PR #626, opened 2018-08-13
Entropy Zoom in URL

emmahodcroft:
This PR allows zooming to change the URL and also allows zoom coordinates to be read from the URL. (Issue [623](https://github.com/nextstrain/auspice/issues/623))

To allow the existing zoom functionality to work without cluttering the URL too much, zoom coordinates (`zmin` & `zmax`) should only show in the URL when the zoom is set manually rather than automatically. For example, selecting a gene from the drop down (or typing a nuc/AA) or clicking on an entropy site should mean that no zoom coordinates appear in the URL - these are set by the `colorBy` (`c` in the URL). 

However, moving the handles of the brush, the sides of the brush, or click-and-drag to set the brushes does cause zoom coordinates to appear in the URL. 

If just a `colorBy`/`c` is present in the URL, zoom will be set by this, as previously. If both a `colorBy`/`c` and zoom coordinates are present, colouring is set by the colorBy and the zoom coordinates override whatever 'default' zoom might apply to the selected AA/nuc, and are used for the zoom instead. 

If min or max zoom are out of bounds, they are set to 0 or absoluteZoomMax, respectively. If min zoom is larger than max, they are reversed. 

I've tested on Zika and TB, but would appreciate testing on other datasets & in general :)

### DIFF
--- a/src/actions/entropy.js
+++ b/src/actions/entropy.js
@@ -23,3 +23,8 @@ export const showCountsNotEntropy = (showCounts) => (dispatch, getState) => {
   dispatch({type: types.ENTROPY_COUNTS_TOGGLE, showCounts});
   updateEntropyVisibility(dispatch, getState);
 };
+
+export const changeZoom = (zoomc) => (dispatch, getState) => {
+  dispatch({type: types.CHANGE_ZOOM, zoomc});
+  updateEntropyVisibility(dispatch, getState);
+};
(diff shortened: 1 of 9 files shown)


emmahodcroft:
The latest commit should make the zoom-by-url work in narratives. (It already worked if you specified a gene/nuc, but now you can provide `zmin` and `zmax`.) 

One thing that is perhaps not so intuitive is that if zoomed in, it will *not* zoom out if you scroll back up the narrative (or to the next narrative point which maybe doesn't seem like it should be zoomed) unless you've specified this in the URL. 

I played with this a bit, but this is intrinsically linked to non-narrative behaviour where doing things like changing colorBy to 'country' or switching from AA to nuc shouldn't change the zoom. If we want to maintain that behaviour (which I think is probably good), then doing such things behind the scenes in a narrative has the same effect. But, it can easily be corrected if desired by zooming out explicitly!

jameshadfield:
This looks great @emmahodcroft -- i'm just going to wait until we have a consensus on the URL query fields to use (`zmin` / `xmin` / `pmin` / `gmin` !?!?! ) and then i'll merge 👍

PR #627, opened 2018-08-15
make calendarToNumeric parse negative dates

rneher:
this prevents negative dates to defaulting to 1970. It also no longer pretends to have month and day for BC dates.


jameshadfield:
👍 thanks @rneher

Issue #520, opened 2018-03-02
add selected named clade to URL state


jameshadfield:

jameshadfield:
Closed in 633 👍

PR #633, opened 2018-08-20
specify zoom on tree by clade in URL

emmahodcroft:
This PR should allow specification of the 'zoom' of the tree by specifying `clade=` in the URL. Clade is specified by `clade_annotation` in the tree.json and is what currently draws branch labels onto the tree (if specified). 

If the user zooms in on a clade this should change the URL, and if specified in the URL, should load accordingly. Could probably do with a bit more testing, but I think it plays nicely with everything else. 

This was a little tricky as `applyInViewNodesToTree` in `actions/tree.js` (sets visibility of nodes) modifies node.shell which does not yet exist at the point the URL queries are being 'read.' I've gotten around this by creating `node.inView` which is then used to set `shell` when it's created later (in `components/tree/phyloTree/phyloTree.js` constructor). I don't think this messes anything up, but perhaps there's a neater way. 

Should resolve issue [520](https://github.com/nextstrain/auspice/issues/520)

Allowing clades to be specified in modular augur by AA should hopefully be possible via the branch [clade ](https://github.com/nextstrain/augur/tree/clade) in augur

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -2,7 +2,7 @@ import queryString from "query-string";
 import { numericToCalendar, calendarToNumeric } from "../util/dateHelpers";
 import { reallySmallNumber, twoColumnBreakpoint, defaultColorBy, defaultGeoResolution } from "../util/globals";
 import { calcBrowserDimensionsInitialState } from "../reducers/browserDimensions";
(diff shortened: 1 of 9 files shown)


jameshadfield:
This looks great! Code looks good and couldn't find any obvious bugs.

PR #636, opened 2018-08-23
Release Narrative

jameshadfield:
This PR implements the narrative with styles & functionality as discussed in person. It is currently up on the dev-heroku server.

Main points:
* Markdown narrative files are parsed and link content to app state via URLs defined in the markdown. See https://github.com/nextstrain/narratives for example markdown files.
* Local narratives are sourced from a `local_narratives` directory, and corresponding `npm run ...` scripts have been added. The URL is localhost:4000/local/narratives
* auspice-dev.herokuapp.com/narratives loads a splash page with all available narratives.
* Community narratives can be loaded in a similar style to datasets, except that the markdown files are in a `narratives` directory (instead of `auspice`). For instance, the narrative files in https://github.com/nextstrain/narratives/tree/master/narratives can be accessed via auspice-dev.herokuapp.com/community/narratives/nextstrain/narratives/ebola (etc)
* Scrolling has been implemented using a modified version of https://github.com/VikLiegostaiev/react-page-scroller. This works well in general, except for paragraphs with text which overflows the available space, where I've had to hide the overflow. This should be fixed in a separate issue and will require further modifications to the scrolling script.
* Tested on Firefox, Safari & Chrome on MacOS with no issues. Tested on iPhone 7 and generally works well, although I expect some tweaks will be necessary as people begin to use it.

I would like to merge into master and put docs up, but not publicize until we have played with it for a while and have at least a couple of well written narratives online.


trvrb:
This is super slick. I think everything is well in order. Thank you for humoring me with the scrolling and nav changes. I couldn't find any bugs. Please go ahead and merge.

Issue #473, opened 2017-10-24
tree branch visibility is incorrect


jameshadfield:
As it stands, any branch that is present within the selected time slice is "visible". E.g. the branch joining a tip from 2016 with a parent in 2013 will be selected if the max date is 2014.

Furthermore, as both tips & branches use the same visibility array, the 2016 tip (in the above example) is shown as visible. 

This screenshot indicates zika filtered to a max date of 2014:
![image](https://user-images.githubusercontent.com/8350992/31964052-a0232acc-b8b7-11e7-839a-a681f35bea85.png)

trvrb:
I've been annoyed by this as well. My only thought for a solution has been to break up branches into multiple subbranches, each with 1 parent and 1 child node. However, I was worried about this for performance reasons as it would significantly increase the number of SVG objects, which has already shown to be a performance bottleneck.

Alternatively, there could be some way to change x-values for nodes dynamically based on time interval selected. This would give more of a growing tree effect than the current approach.

jameshadfield:
You could do it with one extra SVG object (branch) and using `stroke-dasharray` or changing the `d` attribute. With the way phylotree works, this isn't easy.

trvrb:
Yes. I'm afraid of digging too deep into Phylotree and would instead be thinking of solutions that change the "data" supplied to phylotree.

jameshadfield:
@evogytis and I have a possible solution to this:
* On the initial rendering of a tree (or after layout changes, zoom into clade etc etc) draw the skeleton tree - the entire tree with thin branches - what's currently the "not visible" state.
* When rendering "true" branches, only render those which are visible.
* If a branch is visible, but intersects the time slice, only draw the portion of the branch which is in view (i.e. modify the x co-ordinates)

This should actually make animation, filtering etc much faster, as the skeleton tree is left unchanged and so we are rendering far fewer branches at each time point.

jameshadfield:
Closed in #639

PR #638, opened 2018-08-24
Time slicing

jameshadfield:
Watching people interact with auspice has made it apparent that branches which are not in the selection (time slice, filters etc) are confusing to people. We currently draw them thinner, but they are still there, with on-hover info boxes etc. This PR improves the visual distinction between in-selection and out-of-selection branches / tips, especially for time slices.

* on-hover info boxes no longer appear for out-of-selection branches / tips
* There is no longer the ability to click on out-of-selection branches (and the cursor no longer implies this ability)
* out-of-selection branches are now thinner
* for rectangular layouts & temporal trees there is a grey background indicating the selected time slice. This looks especially good for animations and generally improves clarity (in my opinion). 
    * Note that a bunch of tips & parts-of-branches are rendered as "in-selection" but out of this grey-time-slice... this is issue #473, those tips are actually out of the current selection.

### DIFF
--- a/src/components/tree/index.js
+++ b/src/components/tree/index.js
@@ -4,6 +4,8 @@ import UnconnectedTree from "./tree";
 const Tree = connect((state) => ({
   tree: state.tree,
   treeToo: state.treeToo,
+  dateMinNumeric: state.controls.dateMinNumeric,
+  dateMaxNumeric: state.controls.dateMaxNumeric,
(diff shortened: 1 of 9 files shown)


trvrb:
I think this is a significant improvement. Really nice idea.

I couldn't break anything. Please go ahead and merge.

Issue #643, opened 2018-08-27
Clade zoom in narratives doesn't zoom out again


emmahodcroft:
If you specify to zoom to a clade via the URL in narratives, it zooms in, but there's no way to zoom out again...

jameshadfield:
Closed in #644

PR #644, opened 2018-08-27
Force default tree zoom to root

emmahodcroft:
This is hopefully to resolve issue [643](https://github.com/nextstrain/auspice/issues/643).

If a clade zoom is not explicitly set in the URL, it forces the zoom to the root.

This shouldn't change now non-narrative builds work, as if you load via URL it will by default zoom to the root (if a clade isn't specified).

@jameshadfield would you mind checking this is ok, and if you think there's a better way?


jameshadfield:
Perfect -- thanks!

PR #645, opened 2018-08-27
Improve Info text

### LINKED ISSUE #640: Dates in info text should refer to tips and not to nodes
@jameshadfield ---

Working with Lassa made me notice this just now...

Currently, we have some very weird language like 

>Showing 78 of 78 genomes, from 0 countries, from 11 divisions, dated Jul 1662 to Mar 2018.

for Lassa and 

> Showing 1470 of 1470 genomes, from 208 authors, from 10 regions, from 84 countries, dated Nov 1633 to Jul 2017.

for dengue. The text here is referring to "genomes". These are tips in the tree. Dates should be collected from tips rather than nodes.

The date filter doesn't have to change. Still makes sense to write

> Filtered to Jul 1662 to Dec 1942

jameshadfield:
Closes #640

### DIFF
--- a/src/components/download/downloadModal.js
+++ b/src/components/download/downloadModal.js
@@ -7,7 +7,7 @@ import { stopProp } from "../tree/infoPanels/click";
 import { authorString } from "../../util/stringHelpers";
 import * as helpers from "./helperFunctions";
 import * as icons from "../framework/svg-icons";
-import { getAcknowledgments, footerStyles} from "../framework/footer";
+import { getAcknowledgments} from "../framework/footer";
 import { createSummary } from "../info/info";
 
 const dataUsage = [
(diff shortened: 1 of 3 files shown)


trvrb:
@jameshadfield ---

This current implemention is an interesting solution:

lassa

Ie, add a sentence like "Nodes dated between Jan 2016 and Oct 2017. " This definitely now is technically correct. However, I'm not sure it's very useful. I was suggesting in issue #640 to change the text to:

> Showing 358 of 358 genomes comprising 6 countries, 8 authors and 2 host species dated Jan 1969 to Mar 2018.

This seems actually useful to me, while the current implementation is just repeating what we already know from looking at the date slider.

And then if filtered, the text would be:

> Showing 357 of 358 genomes comprising 6 countries, 8 authors and 2 host species dated Jan 2003 to Mar 2018. Filtered to Jun 1985 to Mar 2018.

Would you agree? Or do you prefer the "nodes dated" text?

jameshadfield:
@trvrb is this better?

image

jameshadfield:
Merging this as i'm confident that it's much better than the current `master` -- @trvrb if you want to change the wording (e.g. i've gone for "sampled from") then please just commit to master 👍

trvrb:
Perfect! I like this a lot. Thank you @jameshadfield.

PR #648, opened 2018-08-31
Add more visible toggle between narrative & interactive modes

jameshadfield:
This PR changes the UI for the toggle between narrative mode & interactive mode, as user testing indicated that it was too hidden.

Before:
image

After:
image


trvrb:
I think this is a definite improvement. I just made a small stylistic change. Feel free to merge.

PR #650, opened 2018-09-01
Tweak tree legend placement

trvrb:
The legend wasn't fully obscuring the gray timeslice bar. This commit fixes this.

Before:

before

After:

after


jameshadfield:
Looks much better. Thanks!

PR #654, opened 2018-09-04
Allow npm global install

### LINKED ISSUE #652: Include bin with auspice install
@tsibley and @jameshadfield ---

Related to https://github.com/nextstrain/augur/issues/147.

I was doing some work on https://github.com/artic-network/artic-ebov/blob/master/pages/ebov-phylogenetics-sop.md / http://artic.network/ebov/ebov-phylogenetics-sop.html and wanted an easy way to bring up an auspice view. The current ARTIC install relies completely on conda (http://artic.network/ebov/ebov-it-setup.html) and so I didn't want to ask for docker to be installed if we can avoid it. I got `augur` / `nextstrain build --native` installed easily as part of the conda install and was then trying to do the same for `auspice`.

Take a look at the `bin` branch. This adds
```
"bin": {
  "auspice": "./server.js"
},
```
to `package.json`. I also pushed this to npm at https://www.npmjs.com/package/auspice with the idea that you can then run `npm install -g auspice` to get a globally callable version of auspice.

However, calling `auspice` results in
```
fs.js:119
    throw err;
    ^

Error: ENOENT: no such file or directory, scandir 'dist'
    at Object.readdirSync (fs.js:759:3)
    at findAllCompressionFiles (/usr/local/lib/node_modules/auspice/node_modules/express-static-gzip/index.js:127:24)
```
due to `server.js` looking in the wrong place for files in `dist/` I believe.

You can reproduce this by checking out the `bin` branch and running `npm uninstall` followed by `npm install -g` from within the `auspice/` directory.

If you think it's a good idea, could one of you work on this?

jameshadfield:
Closes #652

This allows auspice to be globally installed via npm (currently working via `npm install -g auspice`).

Default local datasets directory for a globally installed auspice is `<cwd>/auspice` (or `<cwd>` if that doesn't exist). As before, the command line argument `data:<path>` overwrites this. Narratives operate similarly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ s3/
 ### npm ###
 node_modules/
 npm-debug.log
+*tgz
--- a/.npmignore
+++ b/.npmignore
@@ -8,6 +8,7 @@ s3/
 
 src/*
 !src/server/*
+!src/version.js
 
 ### OSX ###
 .DS_Store
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,16 @@
+### App specific ###
+big_flu
+/data/
+/releases/
+/static/
+s3/
+/local_narratives/
+
+src/*
+!src/server/*
+
+### OSX ###
+.DS_Store
+
+### npm ###
+npm-debug.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,23 @@
 language: node_js
 sudo: false
 node_js:
-  - "9"
+- '9'
 install:
-  - npm install
-  - pip install --user awscli
+- npm install
+- pip install --user awscli
 script:
-  - npm run build
+- npm run build
 after_success:
-  - if [[ "$TRAVIS_PULL_REQUEST" == "false" && "$TRAVIS_BRANCH" == "release" ]]; then
-      npm run gzip-and-upload;
-      npm run redeploy-site;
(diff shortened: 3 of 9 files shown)


trvrb:
This is really fantastic @jameshadfield! I just have three small suggestions:

1. Could you add some basic help to `auspice -h` and `auspice --help`?

2. Could you add `auspice -v` and `auspice --version`? Just a couple basic things I'd expect in a command line interface.

3. Could you add dev instructions for how to push to NPM? For example, I wasn't sure if `dist/` should exist or not? And maybe include a script in `devel/`? Or `scripts/`? And then include in the `.travis.yml` file. Otherwise, the NPM version will rapidly fall behind.

@tsibley separated dev scripts for `augur` and `cli` into a `devel` directory that I thought was a good way to do things.

jameshadfield:
I see. Thanks @trvrb. Are there any other points you'd like addressed for this PR?

trvrb:
Nope. I believe that should be it. Thanks James.

jameshadfield:
Suggestions (1) & (2) now done.
Instructions for deploying now in the README. Travis CI deploy to npm may or may not be working. This is #658.
I don't get as `package-lock.json` when installing globally, so i'm going to leave this for a separate issue.

trvrb:
The command line options are great! Thank you @jameshadfield.

PR #662, opened 2018-09-07
Remove .npmrc file

### LINKED ISSUE #661: NPM_TOKEN not defined
Bug that I just introduced, when running `npm start`, or anything really the value of `NPM_TOKEN` that's specified here https://github.com/nextstrain/auspice/blob/master/.npmrc is not defined. This causes the following error:
```
Leda:auspice trvrb$ npm start
Error: Failed to replace env in config: ${NPM_TOKEN}
    at /usr/local/lib/node_modules/npm/lib/config/core.js:415:13
```
Maybe there's a simple way to define a fall back in the `.npmrc` file? Sorry about this.

tsibley:
It broke local use of npm if NPM_TOKEN was missing from the environment.
Since the file was only meant for Travis CI, define the config
dynamically in .travis.yml instead.

Resolves #661.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-//registry.npmjs.org/:_authToken=${NPM_TOKEN}
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,5 +12,6 @@ after_success:
     npm run gzip-and-upload;
     npm run redeploy-site;
     npm run rebuild-docker-image;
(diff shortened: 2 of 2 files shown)


trvrb:
I think this will break NPM deploy on Travis. It needs this npmrc file. Probably best to remove "npm publish" from Travis script as well while we sort this out.

trvrb:
Ah! Nice. Sorry. Was on phone and didn't look at more than the commit message. My fault. This is great. Please merge and release.

PR #665, opened 2018-09-19
colours are consistent between two trees. Closes #663

### LINKED ISSUE #663: Second tree colors are slightly different than the first tree
Each color in the first tree is linked to colors that are slightly lighter in the second tree. It's pretty subtle, but a real behaviour:

image

Note that it only happens when toggling on the second tree after viewing the first. It doesn't happen if you load <https://nextstrain.org/flu/seasonal/h3n2/ha:na/3y> directly.

The bug appears to be an off-by-one error of some kind related to [`createDiscreteScale()`](https://github.com/nextstrain/auspice/blob/58b2723/src/util/colorScale.js#L53-L67) or the domain it is passed: the scale is computed with N values for the first tree and with N+1 when recomputed for the second. Note that [the first tree](https://nextstrain.org/flu/seasonal/h3n2/ha/3y) on its own has no "unassigned" color, but [the second does](https://nextstrain.org/flu/seasonal/h3n2/na/3y). So perhaps the first tree's colors aren't updated for the new scale. I was unable, in a reasonable amount of time, to make a fix that didn't have other unintended consequences and so moved on to other things.

@jameshadfield Seems like you might have more insight into this bit of context. Any thoughts?

jameshadfield:
Adding a second tree (as opposed to loading 2 at the same time) causes the node colours to recompute, however previously the left tree did not update to reflect these changes.

### DIFF
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -82,7 +82,7 @@ export const loadTreeToo = (name, fields) => (dispatch, getState) => {
   fetchJSON(`${charonAPIAddress}request=additionalJSON&source=${oldState.controls.source}&url=${fields.join("/")}&type=tree`)
     .then((json) => {
       const newState = createTreeTooState({treeTooJSON: json.tree, oldState, segment: name});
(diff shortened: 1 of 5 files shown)


tsibley:
Looks good to me. Thanks!

Issue #649, opened 2018-09-01
Allow trees with just branch_lengths or just clock_lengths


trvrb:
@jameshadfield ---

Currently, tree JSONs are required to have both `clock_length` and `branch_length`. However, if we do a light pass to directly visualize a tree from RAxML or from BEAST it will only have one or the other.

I encountered this when constructing a BEAST tutorial here: https://github.com/nextstrain/mers-beast-tutorial. I could add additional augur steps to have both `clock_length` and `branch_length`, but the simplest pass does not include both. I believe that auspice should be able to work with this sort of data.

You can see the resulting JSONs here: https://github.com/nextstrain/mers-beast-tutorial/tree/master/auspice and notice that they currently crash auspice: https://nextstrain.org/community/nextstrain/mers-beast-tutorial

This has just `div` and `branch_length`. I need to fix `augur refine` to properly call these `num_date` and `clock_length`, but the issue still stands.

I would propose greying out the option TIME or DIVERGENCE that doesn't apply to a particular dataset.

jameshadfield:
Agreed. This is part of the JSON 2.0 plan (the schemas for 2.0 do not require num_date or div) 👍

jameshadfield:
I believe @emmahodcroft is playing around with the implementation of this 👍

emmahodcroft:
@jameshadfield  and @trvrb - 
I've created a branch (`no_date`) which should now handle cases where there is `branch_length` and `div` but no date information. I also created a branch in `augur` (not pushed yet) which modifies `refine` so that if date information on the tips is present (but `--treetime` option not run, for example, so no dates on internal nodes) this is included in the `branch_lengths.json` file. (In theory this could then be visible in auspice in some manner - for example on mouse-over.)

`no_date` works by looking at the root node to see if it has `num_date`, and if not, setting a new variable, `displayDates`, to false. This keeps the default dates behind-the-scenes (seemed the easiest work-around), ensures all nodes will be visible regardless of 'date' (or lack thereof), forces the default view to be `div`, removes the date-slider, 'Time' option (in 'Branch Length'), and animation options from the controls, and removes the 'play/reset' buttons from the map. 

This seems to work with the files you provided Trevor, and some I mocked up myself. I still need to ensure this carries over to 'treeToo' if visible (next step).

If there are any example files that have date only (inverse of the above) this would be useful for me to start tackling the next bit - I can mock them up of course but if they've come from something 'real' it helps ensure I'm making the right assumptions.

[...]

PR #670, opened 2018-09-27
Handle trees without date information

emmahodcroft:
Addresses half of issue [649](https://github.com/nextstrain/auspice/issues/649). 

This PR should allow auspice to handle cases where there is `branch_length` and `div` but no date information (or, date info only at the tips).

It works by looking at the root node to see if it has `num_date,` and if not, setting a new variable, `branchLengthsToDisplay,` to `divOnly` if so (default is `divAndDate`). 

This keeps the default dates behind-the-scenes (seemed the easiest work-around), ensures all nodes will be visible regardless of 'date' (or lack thereof), forces the default view to be `div`, removes the date-slider, 'Time' option (in 'Branch Length'), 'clock' option (in 'Layout'), and animation options from the controls, removes the 'play/reset' buttons from the map, and doesn't show the 'sampled between' in the info text.

This can be expanded to have a `dateOnly` option for trees with dates only.

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -20,7 +20,7 @@ export const checkColorByConfidence = (attrs, colorBy) => {
 };
 
 export const getMinCalDateViaTree = (nodes, state) => {
-  // const minNumDate = nodes[0].attr.num_date - 0.01; /* slider should be earlier than actual day */
+  /* slider should be earlier than actual day */
   /* if no date, use some default dates - slider will not be visible */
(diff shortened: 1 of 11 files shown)


jameshadfield:
Nice work @emmahodcroft - I don't have a divergence-only tree to test on, but it works great for a time-only tree 👍 

Two minor things:
* I'd remove the "branch length" bit from the sidebar altogether (since you can't change it).
* Asking for `div` in the URL causes bad things to happen 😱 -- e.g. http://localhost:4000/local/mers?m=div

jameshadfield:
Emma i'm sorry this has been here for nearly a month! I'll test & merge this today.

jameshadfield:
This is great, and there are so many exciting possibilities opening up based off of this. Thanks!

PR #676, opened 2018-11-01
Allow tree nodes to be hidden

jameshadfield:
This allows `node.hidden` to decide whether the node is rendered or not.
Try `localhost:4000/ZEBOV` for an example of what this can do.

Allowed values: (Will be added to docs when they exist)
not set or `0` -- always shown
`1` -- always hidden
`2` -- hidden for time trees only
`3` -- hidden for div trees only

### DIFF
--- a/src/components/tree/phyloTree/change.js
+++ b/src/components/tree/phyloTree/change.js
@@ -3,6 +3,7 @@ import { calcConfidenceWidth } from "./confidence";
 import { applyToChildren } from "./helpers";
 import { timerStart, timerEnd } from "../../../util/perf";
 import { NODE_VISIBLE } from "../../../util/globals";
+import { getBranchVisibility } from "./renderers";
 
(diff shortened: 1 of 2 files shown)


tsibley:
How does the `d.n.hidden` property interact with the `d.visibility` property?

tsibley:
Could we use `never`, `always`, `timetree`, `divtree` instead of 0–3?

trvrb:
The `ZEBOV` example is really fantastic. Also see the use for exploded trees in general.

I don't think there should be much of a performance hit if any having an array of strings instead of an array of ints. Nice to have things be self documenting.

jameshadfield:
> Could we use never, always, timetree, divtree instead of 0–3?
Sure. 

Allowed values:
* property not set or any string that's not below -- always shown
* `"always"` -- always hidden
* `"timetree"` -- hidden for time trees only
* `"divtree"` -- hidden for div trees only

PR #677, opened 2018-11-01
Fix hot reloading.

### LINKED ISSUE #496: sort out react hot loader
* hasn't been working in a long time
* babel used to have `"plugins": ["react-hot-loader/babel"]` but it wasn't actually running and when it did run it caused an error to be thrown!

jameshadfield:
Requires running `npm install` as dependencies have changed

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,18 +3,18 @@
   "plugins": ["transform-decorators-legacy"],
   "env": {
     "timing": {
-      "plugins": [
-      ],
+      "plugins": []
     },
     "development": {
       "plugins": [
+        "react-hot-loader/babel",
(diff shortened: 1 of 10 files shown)


jameshadfield:
Can someone (@tsibley @emmahodcroft @trvrb) test that this still works for them?

`npm run build && npm run server` and `npm run dev` (You'll need to do `npm install` first)

tsibley:
The commands work for me as expected (prod server doesn't hot-reload, dev server does).

The dev server does issue this warning:

    (node:10421) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead

which doesn't happen on master.

jameshadfield:
Great. Thanks.

PR #682, opened 2018-11-19
Swap rawgit CDN to githubusercontent CDN

### LINKED ISSUE #672: change community builds CDN
Community URLs currently source the JSONs from github via the rawgit CDN.
Rawgit is shutting down -- see https://rawgit.com/
We should migrate the CDN -- https://www.jsdelivr.com/rawgit looks simple.

trvrb:
Currently, "community" builds from nextstrain.org/community are being sourced from rawgit. However, rawgit is shutting down.

This moves the CDN source to githubusercontent, which is the official source for "raw" files from GitHub.

This PR would resolve #672.


tsibley:
This looks good to me! The forcing content-type thing I was thinking of no longer applies because charon merges two upstream responses into one and thus sets the content-type itself now. It used to pass through the response minimally changed.

trvrb:
Great. I'm going to go ahead and merge.

PR #684, opened 2018-11-29
second tree option available by segment number

neavemj:
Small change to make the 'Second Tree' option appear when virus segments are named according to segment number. Must be "segment1", "segment2", ... "segment10".

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "auspice",
-  "version": "1.31.0",
+  "version": "1.34.0",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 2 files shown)


jameshadfield:
Nice, thanks! This is now in master version v1.34.1. See [this issue](https://github.com/nextstrain/auspice/pull/534#issuecomment-442642120) for further background.

P.S. I ammended the commit to remove the changes in `package-lock.json` which means this PR didn't become merged. But your commit is here: https://github.com/nextstrain/auspice/commit/f89fb86a72b826a5d0b7bd431e8f6dbb4b91ec4b

Thanks again 👍

Issue #687, opened 2019-01-02
future server API


jameshadfield:
This issue is to discuss the design of the server API employed by auspice. To briefly recap, auspice needs to know which datasets / narratives are available as well as obtaining the dataset or narrative to view. 

Here are @tsibley's thoughts, taken from https://github.com/nextstrain/auspice/pull/683#issuecomment-450027666

### Current

The Charon API, [as described in the white-labelling docs][api doc], relies on a dynamic server able to respond to the following endpoints:

    /charon/getAvailable
    /charon/getAvailable?prefix=flu_seasonal

    /charon/getDataset?prefix=flu_seasonal_h3n2_ha_2y
    /charon/getDataset?prefix=flu_seasonal_h3n2_ha_2y&type=tip-frequencies

    /charon/getNarrative?prefix=flu-report-2018-Q3.md

Pros:

* Server allows for custom logic to dynamically respond with different datasets for different people (e.g. logged in users).

Cons:

* Requires a dynamic server.

* Statically-hosted builds (e.g. for GitHub Pages) and S3-served builds (e.g. for Nextstrain) must be special-cased since no dynamic server is possible.

* Paths in the user-facing URL require transformation back and forth, which historically has been a source of bugs.

[api doc]: https://github.com/nextstrain/auspice/blob/extend/docs/charonAPI.md

### Proposed

My proposed "data API" is to use standard HTTP access for what it's good at by changing Auspice to request these URLs instead:

    /datasets.json
    /narratives.json

    /dataset/${datasetPath}/${type}.json
    /dataset/flu/seasonal/h3n2/ha/2y/{tree,meta,tip-frequencies}.json

    /narrative/${name}.json
    /narrative/flu-report/2018/Q3.md

Pros:

* Endpoints can be provided _either_

  - by a set of files on a static host (e.g. GitHub Pages, S3), _or_
  - by a dynamic server in order to customize responses for different people (e.g.  logged in users).

* Statically-hosted builds (e.g. for GitHub Pages) and S3-served builds (e.g. for Nextstrain) just work with no special-casing or additional server logic in Auspice.

* Paths in the user-facing URL match those behind the scenes without any transformation.

* Standard HTTP caching strategies work without a custom server complicating it.

Cons:

_I don't see any downsides to this approach, but maybe you do?  Looking for feedback!_

jameshadfield:
[...]

PR #688, opened 2019-01-02
auspice extension architecture

jameshadfield:
Prior to this PR auspice was a repository which generated
* (a) the interactive visualisation app used in nextstrain.org
* (b) the server code used in nextstrain.org to handle requests from the client
* (c) a npm-installable program to visualise local datasets, with nextstrain branding

This PR changes auspice into a stand-alone program separate from nextstrain. Alone, it can be used to display local datasets (similar to before, but with new branding). It may also be extended by injecting code at build time.  In this fashion, the viewer used in nextstrain.org will be one particular extended-version of auspice. 

This PR also introduces docs for auspice, which are the best place to understand and try out the functionality introduced by this PR. They are located in `./docs-src` and build into a webpage to be run off github pages from `./docs`. [This page](https://nextstrain.github.io/sandbox) is a working clone of the docs in this branch and is what nextstrain.github.io/auspice will look like when this is merged into master.

---

* Server code is now the responsibility of the servers running extended versions of auspice. E.g. https://github.com/nextstrain/nextstrain.org/tree/whitelabel/auspice/server
* The "whitelabel" branch of github.com/nextstrain/nextstrain.org is configured to build auspice from this PR. There are still a few bits missing (community builds, narratives). A PR will arrive there shortly. Upon merging of both PRs the live nextstrain.org site will then be built using whitelabelled/extendable auspice. 
* The client-server API is relatively unchanged from current master. Please see https://github.com/nextstrain/auspice/issues/687 for discussion of this.
* Likewise, the dataset JSON format is relatively unchanged from current master. Migrating to v2.0 is a priority but outside the scope of this PR.
* The extension interfaces are preliminary, but working. Nextstrain can be built without aesthetic regressions. They should be improved in future PRs.
* Similarly the github pages splash & docs pages should be thought of as a first pass. They do, however, provide a centralised place for auspice documentation and marketing which we currently lack.


jameshadfield:
Merged following 2019-01-14 nextstrain meeting

PR #692, opened 2019-01-22
Add back `npm run {server,dev}` for the benefit of folks upgrading

tsibley:
These were the primary ways to run Auspice previously, and it's nice to
make them keep working even with the introduction of the new Auspice
CLI.  People upgrading will be less surprised.


jameshadfield:
No problem -- we can add these back for versions `1.x.x`. Thanks!

Note that in the future auspice should be interacted with by using `auspice` (or `npx auspice` if not installed globally), but I understand that this is better done with a v2 release.

tsibley:
Oh yeah, having a CLI for Auspice is great and these additions aren't meant to impede that forward progress, just offer a bridge for folks upgrading. It might make sense to add warnings to them (along with `view`, `start`, etc) so people can start getting used to the idea of `auspice` or `npx auspice`. Thanks for merging.

PR #701, opened 2019-02-04
Improve app load times using code splitting

jameshadfield:
Currently loading auspice (e.g. nextstrain.org/zika) involves fetching & parsing a JS bundle containing all the relevant app code (324kb minimised), then fetching the dataset JSON(s). 

Code-splitting is a feature where the app code (bundle) can be split into separate chunks. This allows a smaller initial fetch+parse and can improve load times for apps. This PR implements code splitting for a number of the components (see below) and improves load time. The time until nextstrain.org/flu/seasonal/h3n2/ha/3y is fully displayed has reduced from 7.9s to 6.3s (average from 4 runs on my machine using the dev server vs live server) and it "feels" faster than this because the map/tree displays and can be interacted with before the entropy/frequency panels).

This image shows the current live version (top) vs this PR (bottom):

lazy_load

---

The components split into separate bundles:
* main app (tree, sidebar etc)
* entropy panel (if you load nextstrain.org/flu/seasonal/h3n2/ha/3y you'll see that the tree appears slightly before the entropy)
* frequency (same as entropy -- and if you don't require frequencies this code is never fetched)
* splash page (the auspice one)
* notifications
* status page (i.e. the JS for this is never fetched until you actually go to the status page)

I tried splitting out the Map component, which sped up loading even more but looked worse (in my opinion) as there was a white panel next to the tree for ~0.5s. 

This PR is available for testing here: https://nextstrain-dev.herokuapp.com

Furthermore, we can now analyse the composition of each bundle using `auspice build --analyzeBundle` which results in an interactive display (screenshot below). I'll make a separate issue for the improvements that this has flagged up.

image


trvrb:
Thanks for the all the detailed work here @jameshadfield. I'm always a fan of performance improvements. Generally in favor of page loads. I looked through code changes and did a fair amount of testing. There does seem to be an improvement, though I seemed to be getting more like ~1s rather than ~2s, but I wasn't being entirely rigorous with Chrome's dev tools.

I don't like the idea of a blank map panel during load.

This is a more a rhetorical question for you: how much additional logic overhead is this bringing? Looking at the code it doesn't look bad at all.

Please feel free to merge.

PR #703, opened 2019-02-04
404 page preserves URL

### LINKED ISSUE #700: 404 page should preserve URL
Currently, if auspice 404s it punts back to the base URL. For example, http://nextstrain.org/zika1 404s, but the URL is changed to https://nextstrain.org/. I don't think this is good behavior. Often, there's just a slight misspelling and this also breaks the use of the back button. I believe 404s should resolve as they do on bedford.io for example: https://bedford.io/zika.

jameshadfield:
Closes #700 -- see that issue for more information.

### DIFF
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -76,10 +76,11 @@ export const changePage = ({
   }
 };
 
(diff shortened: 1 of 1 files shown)


trvrb:
This looks great! Thanks James.

Issue #708, opened 2019-02-27
Code splitting has broken d3 transitions


jameshadfield:
This bug was introduced between v1.35.5 and v.1.36.0 by PR #701 (code splitting). Nextstrain.org is still on v1.35.5 and it's recommended to continue to use that version until this bug is fixed.

## Description:
For some datasets (!) clicking on branches / tips in the tree causes the app to crash. Zika, flu/seasonal/h3n2/ha/3y, and flu/seasonal/h3n2/na/3y appear to work fine, however http://localhost:4000/flu/seasonal/h3n2/ha:na/3y is broken. To reproduce, click on (any?) branch / tip in the tree(s).

## Cause:
Error message given is `svg.selectAll(...).filter(...).transition is not a function`. The selection is not empty, but rather `transition` is not being added as a prototype to the `Selection()` object.

trvrb:
If this proves difficult to immediately fix, we may want to push a hot fix to 1.36.1 that reverts the code splitting PR. Even though live site is working. We won't be able to make small fixes to footer, etc... and anyone installing auspice from npm will have this issue.

PR #709, opened 2019-02-27
Check mutation type against tree

jameshadfield:
This PR sets `mutType` intelligently according to what mutations are available on the tree. Furthermore, it allows nuc-genotype colorBy values for datasets lacking gene annotations in the metadata JSON. 

Tested against a number of datasets with no regressions seen, but @emmahodcroft please try on a few more! ~Please note that any d3-transition errors encountered are unrelated to this -- see #708.~ Update: Rebased onto v1.36.1 so that there should be no install / d3-related errors.

@tsibley may care to examine d34c075.


tsibley:
@jameshadfield Thanks for these fixes! I'm not sure I see where the default `geneLength` passed to `decodeColorByGenotype` is `{}`. Can you point that out to me?

It seems like at the very least there should be a `nuc` key in any `geneLength` passed to `decodeColorByGenotype`?

jameshadfield:
>I'm not sure I see where the default `geneLength` passed to `decodeColorByGenotype` is `{}`. Can you point that out to me?
https://github.com/nextstrain/auspice/blob/master/src/reducers/controls.js#L38

>It seems like at the very least there should be a `nuc` key in any `geneLength` passed to `decodeColorByGenotype`?
Not if annotations aren't defined in the metadata json.

tsibley:
> Not if annotations aren't defined in the metadata json.

Ah, I see. I was thinking auspice could/did compute the nucleotide length itself.

trvrb:
I'm going to go ahead and merge this. Going on Emma's specific feature testing and I can't find any regressions.

Issue #707, opened 2019-02-26
Allow footer text to derive from meta.json


trvrb:
Currently text in footer (https://github.com/nextstrain/auspice/blob/master/src/components/framework/footer.js#L49) is the only remaining place where auspice source code has to get modified in a dataset specific manner. Notably, this currently causes issues with community builds, where for example Paul's work here: https://nextstrain.org/community/pauloluniyi/lassa/s, gets a footer borrowed from here: https://nextstrain.org/lassa/s, which is inaccurate and doesn't appropriately highlight Paul's work.

I recommend adding a `footer` field to `meta.json` that auspice uses to specify this text. Probably best to have this be HTML and so people can embed links. If `footer` is present auspice uses this. If not auspice falls back on current strategy of `window.location.pathname.includes`.

tsibley:
HTML might be ok in the short-term to get this going now, but pretty soon we will need to address the security implications of allowing that HTML injection. It might be easiest to not allow it from the beginning, particularly since nextstrain.org already allows rendering of untrusted inputs via community URLs.

We might address the security issues by filtering to a minimal allowed set of tags or by switching to some other, more restricted format which we render to safe HTML.

trvrb:
Ah! Good point @tsibley. Would markdown be a safe alternative?

tsibley:
Yes, if whatever we use to render Markdown → HTML has the appropriate restrictions, since most Markdown parsers allow inlining raw HTML and some allow arbitrary element attributes.

jameshadfield:
Currently the auspice server (and nextstrain.org server) include a node markdown library which is used to parse the narrative markdown files, sending (sanitized) HTML to the client to render. With the proposed server API, the parsing of the footer markdown (and also the narratives) would need to happen on the client.

jameshadfield:
I think best to leave this until after v2 release. We can add in a footer field in the v2 JSONs at a later date and maintain backwards compatibility.

jameshadfield:
Update: This "feature" is really important as we (and others) start to integrate auspice into other projects. The dataset JSON should be able to define this. In parallel, this should be made part of the augur auspice-export JSON, and the footers currently defined in auspice moved into their respective augur builds.

[...]

PR #710, opened 2019-02-27
Don't display custom ebola footer for community builds

jameshadfield:
As a temporary fix until #707 is addressed, this PR removes the custom ebola footer from all community-sourced ebola builds.


trvrb:
Thanks James. I've replaced this functionality in https://github.com/nextstrain/auspice/commit/7d7778b68b0fa756a4cefca7e994745ec46cc0d0.

Issue #280, opened 2017-03-16
dynamically set map deme radii


jameshadfield:
would prevent cases such as this:
![image](https://cloud.githubusercontent.com/assets/8350992/24022652/155a93ea-0a65-11e7-9188-314f05e8ed99.png)

trvrb:
Possibly addressed by changes in #330? We should revisit and see if working as intended.

jameshadfield:
Doing this for all the different use cases seems difficult to do perfectly, but relatively easy to make vast improvements over the current situation (right now it's hardcoded so only works for a small range of tip values). 

We could also expose a user selector to tweak this as desired, e.g:
![image](https://user-images.githubusercontent.com/8350992/47743743-fae10700-dc3c-11e8-8494-47f771450853.png)

tsibley:
What if we define a scaled set of maximum deme radii based on the current map zoom level, assign the largest demes to the zoom level max radii, and scale the rest of the smaller demes based on that?

jameshadfield:
This would be a relatively easy first step and would probably be a big improvement.

trvrb:
I think that PR #712 + a dynamically set max zoom (to be ~2 levels past initial zoom set by `fitBounds`) will fully resolve this. Initial sizing all works well, only looks bad when zoomed too far out. I think reasonable to add in a `maxZoom` and `maxBounds` to appropriately bound these more narrow datasets.

jcw349:
> Doing this for all the different use cases seems difficult to do perfectly, but relatively easy to make vast improvements over the current situation (right now it's hardcoded so only works for a small range of tip values).
> 
> We could also expose a user selector to tweak this as desired, e.g:
> ![image](https://user-images.githubusercontent.com/8350992/47743743-fae10700-dc3c-11e8-8494-47f771450853.png)

Are there plans to implement this still? I think it will help visualize local cases.  As of now, when we hover our mouse over small demes, the mouse misses the circle and nothing happens.

trvrb:
I'll leave this open, but it hasn't seemed a major issue to me. If others disagree, please chime in.

emmahodcroft:
[...]

PR #712, opened 2019-03-11
Map improvements including transmission ribbons

trvrb:
This PR includes a few separate map improvements.

1. Slightly jitter each separate transmission line so that if there are multiple events they appear as a ribbon rather than a narrow line. I think this is a significant improvement. And it has the nice effect where the transmission ribbon thickness properly interacts with visibility adjustments made through date slider or tree zoom.

Here's current Ebola:

ebola

Here's an animation of flu spread:

flu

This just adds an additional `extend` parameter that is propagated down through the map code to the underlying `bezier` math that pushes the arc out by `extend` number of pixels. `extend` for a particular pair of demes starts at 0 and increments with each transmission event.

This was mainly motivated by the ugliness of current regional flu view, but I think made a big difference overall. Compare before:

Screen Shot 2019-03-10 at 5 13 41 PM

To after:

Screen Shot 2019-03-10 at 5 14 52 PM

2. Improve deme circle sizing. Previously, deme circle area was proportional to the number of tips belonging to this deme. This resulted in large datasets like flu having circles that were too big and small datasets like Ebola in North Kivu having circles that were too small. This PR adds a normalization by the square root of the total number of nodes in the tree. I kept this as square root so that smaller datasets still have slightly smaller circles than larger datasets. Within a single auspice view, we maintain circle area as proportional to tip count (as I believe is most expected behavior).

This mostly addresses #280. Initial deme radii look good. Zooming out too far on a narrow set of regions (like Ebola in West Africa or H7N9 in China) doesn't look great, but it's really not bad.

3. Improve initial map bounds logic. This is a small fix to make initial zoom bounds better fit to the lat/longs that are present. Essentially we just needed to set Leaflet's `zoomSnap` to something more granular than the default 1.

4. Don't show error for missing lat/longs if deme is labeled "Unknown" or "unknown". This was apparent in https://nextstrain.org/community/inrb-drc/ebola-nord-kivu, where the samples that are labeled as "unknown" really should not have an associated lat/long. 

6. Add very slight border to deme circles. I think this looks stronger and more grounded.

_Resolves #425_.
_Resolves #369_.


rneher:
I like it. looks much better IMO!

trvrb:
I'm going to go ahead and merge this as well.

rneher:
There is one aspect that looks a little strange: 
image
The circles around the disk have sometimes odd colors. This is from
https://nextstrain.org/flu/seasonal/h3n2/ha/6m?c=gt-HA1_91&p=grid

trvrb:
Thanks for catching this Richard. I know what this is (stroke color not updating on colorBy change). I'll fix.

jameshadfield:
Looks great!

lixingguang:
cool！

PR #713, opened 2019-04-16
Fix initial slider render

trvrb:
There had been a persistent bug in which handles on the date slider don't appear in their proper locations on initial page load, ie:

Screen Shot 2019-04-16 at 8 53 40 AM

I tracked this issue down to a side effect of their being two initial renders of the Slider component. The second caused by Slider updating its internal state when it profiles how wide its container is. The first render had `offset` (which determines handle placement) set to `[0,0]` while the second render has `offset` set properly. This _should_ have been okay, however, for some reason, React's reconciliation wasn't updating the DOM properly upon second render.

I first tried to fix the reconciliation angle, but couldn't sort it out. This PR removes `_handleResize` method and instead fixes pixel dimensions on initial load. This should always be okay for us, as Slider is never resized responsively. This has the side effect of removing unnecessary rerenders of Slider during app use.

I also took this opportunity to lint `Slider.js` to modernize it to ES6 standards.

### DIFF
--- a/src/components/controls/slider.js
+++ b/src/components/controls/slider.js
@@ -4,6 +4,12 @@ import createReactClass from "create-react-class";
 import _assign from "lodash/assign";
 import _isArray from "lodash/isArray";
 
+// sizing parameters
+const upperBound = 208;
+const sliderLength = 220;
+const handleSize = 12;
+const sliderStart = 20;
+
 /**
  * To prevent text selection while dragging.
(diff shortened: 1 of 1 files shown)


jameshadfield:
Looks great and works well for me - really nice to get rid of this bug too, once you notice it it seems to always appear!

In the future one could convert this to a react class (or functional component with hooks) and get rid of `create-react-class` (about 7kb gzipped), but as you point out it may take a bit of debugging to get the `bind`s working correctly.

PR #720, opened 2019-04-22
Auspice consumes v2 JSONs

jameshadfield:
Tl;DR - It’d be fantastic if you could test this version of auspice if you have any augur-produced v2 JSONs 🎉 

```
npm install --global auspice@alpha
auspice view --datasetDir <dir with jsons>
```
Any v2 JSON (without a `_tree.json` or `_meta.json`) should be picked up (it should also work fine for v1 JSONs). If you need an example JSON:
```
curl https://raw.githubusercontent.com/nextstrain/auspice/json-v2/examples/minimal.json > minimal.json
```

### Some background
The “v1” JSONs (current schemas [here](https://github.com/nextstrain/augur/blob/master/augur/data/schema_meta.json) and [here](https://github.com/nextstrain/augur/blob/master/augur/data/schema_tree.json)) have changed over time and there has never been any guarantees that a JSON from one time would work with a later/earlier version of auspice (in fact this has been the source of many headaches!). 

Furthermore, about a year ago the auspice (and nextstrain.org) server began transforming these JSONs before sending to the client. Thus the data that auspice (the client) received was a different format altogether, somewhere half way between schema v1 & v2.

The [v2 schema](https://github.com/nextstrain/augur/blob/master/augur/data/schema.json) contains a number of improvements over the previous schema (and probably most noticeably, it’s a single JSON instead of two). Since there are no v1 features not in v2, it’s possible to transform v1 JSONs into v2 JSONs (see below).

Furthermore, as we want to expand the capabilities of auspice to build “static” sites (e.g. GitHub pages), we lose the ability to rely on the server to transform data. This PR is a huge step in that direction, as now v2 JSONs can be handed directly to the client with no modifications. 

This PR shouldn’t result in any aesthetic changes.

### Server changes

The server must now respond with v2 JSONs. There is a conversion function which the auspice server (“charon”) uses to convert metadata + tree JSONs into a v2 JSON. For release, this function (and others) will be made available so that other servers (e.g. nextstain.org) can use it along the lines of:
```js
const convertJsons = require(“auspice”).convert;
```

In terms of the API requests there are very few changes, and they all revolve around requests for two trees. Previously a server was required to handle dataset requests with the `?type=tree` query as well as `?prefix=".../tree1:tree2/...` encoding. Both of these are no longer required, instead auspice makes two independent requests and expects back two (v2) JSONs. The requests for the (v2) JSONs, as well as `tip-frequencies` and all other requests remain unchanged. The one modification a server can make — and it’s completely optional — is to set a property which instructs auspice to display a different URL.

### Producing v2 JSONs

Augur has been producing v2 JSONs for quite some time, but these were (as far as I know) a dead end with no use. It’s therefore to be expected that they may be in short supply for testing! 
Don’t forget that the auspice server — and shortly the nextstrain.org server — will convert v1 JSONs to v2, so v1 JSONs will still work without issues.

* A minimal example may be found [here](https://github.com/nextstrain/auspice/blob/json-v2/examples/minimal.json)
* Auspice can now convert v1 -> v2 using the same conversion function that the server uses to deliver v1 JSONs to auspice:
```
auspice convert --meta-json <json> --tree-json <json> --output <json>
```

### Proposed v2 schema changes

During development it became clear that we’re missing a few fields. Thus, the v2 JSONs here don’t _exactly_ match the current v2 schema. See [this augur PR for more info](https://github.com/nextstrain/augur/pull/277)

### To-do

* I anticipate there will be bugs (which is why it’s an alpha release!) 
* Expose (server) helper functions, such as `convertFromV1`. This should make building custom servers much easier.
* Documentation
* nextstrain.org PR using this version
* augur PR with proposed schema changes
* Test on IE / Edge @emmahodcroft 🙏 
* Reinstate the ability to drag & drop a CSV containing new metadata (colorings) for the tips.
* Fully test static sites / different customizations etc
* Remove the logging from the server / cli (but which is useful during alpha testing).


rneher:
some more general thoughts: Our author field is very similar to bibtex. should we try to align completely? there are node bibtex parsers....

rneher:
similarly, genome annotations have accepted formats such as GFF3:
https://www.ensembl.org/info/website/upload/gff3.html

emmahodcroft:
`npm install --global auspice@alpha`

@jameshadfield Unfortunately this doesn't work for me... Running with `sudo` (otherwise I immediately get errors) about access, I get errors about 'gifsicle'.  I can copy them to you somewhere?

I don't usually install with `--global` - unsure if that's part of the problem. But it won't let me install without that, as it complains I'm stalling a package a package under a package of the same name...

emmahodcroft:
My apologies! After more Googling, this seems to work for me... `sudo npm install --global auspice@alpha --unsafe-perm=true --allow-root`

jameshadfield:
This is now in the `v2` branch. Smaller PRs will follow with fixes as we approach the v2 release :)

PR #723, opened 2019-05-13
Don't Count X for AA Diversity Chart

emmahodcroft:
With the update to allow recovery of ambiguous sites for Fasta-input in `augur`, there's now the possibility for many AA sites be 'X' at the tips. At the moment, these are counted as 'mutations' for the diversity chart:

Before the change:
counts-before

But these should probably be disregarded from 'entropy' and 'count', as 'N' is in nucleotide data. I introduced a similar check as is already present for nucleotides in 'counts'. However, I decided that we may be interested in gaps ('-') in AA alignments (as this is a little more stringent than gaps in nucleotides, since they must be in-frame), so I allowed this to be counted both in 'entropy' and 'counts'. We can change this in future if it doesn't seem to actually be informative. 

After the change:
counts-after

### DIFF
--- a/src/util/entropy.js
+++ b/src/util/entropy.js
@@ -21,7 +21,11 @@ const calcMutationCounts = (nodes, visibility, geneMap, isAA) => {
         for (const prot in n.aa_muts) { // eslint-disable-line
           n.aa_muts[prot].forEach((m) => {
             const pos = parseInt(m.slice(1, m.length - 1), 10);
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @emmahodcroft this seems very sensible. I'll try and take a look at this over the next ~4 days, but if you don't hear from me feel free to merge 👍

PR #725, opened 2019-05-16
src/components/tree/phyloTree/grid.js:

rneher:
If the major grid is 5 or 10y use 5 minor ticks instead of 4. This avoids the odd case when our grid lines in the tree are:
2010, 2011.25, 2012.5, 2013.75 and 2015
and replaces this in favor of
2010, 2011, 2012, 2013, 2014 and 2015
(analogously for 10y intervals).


trvrb:
This is great @rneher. Thanks for the fix. I tested and this works well. Merging now.

Issue #731, opened 2019-07-02
Auspice narratives scroll-up bug


kairstenfay:
Aupice narratives has a known bug where scrolling down then scrolling up sometimes causes the page to load a blank white screen and then redirect from `/narratives` to `/na`.

The surest way to recreate this is to go to [an example narrative](https://wnv-paper-narrative.herokuapp.com/narratives/map-bug) and scroll all the way to the bottom. Then click the link to go back to the top or manually scroll up. 

This issue happens sporadically and frequently.

joverlee521:
Console error:
```
mapHelpers.js:218 Uncaught TypeError: Cannot read property 'demes' of null
    at updateVisibility (mapHelpers.js:218)
    at Map.maybeUpdateDemesAndTransmissions (map.js:354)
    at Map.maybeUpdateDemesAndTransmissions (react-hot-loader.development.js:661)
    at Map.componentWillReceiveProps (map.js:119)
    at callComponentWillReceiveProps (react-dom.development.js:11441)
    at updateClassInstance (react-dom.development.js:11651)
    at updateClassComponent (react-dom.development.js:14694)
    at beginWork (react-dom.development.js:15644)
    at performUnitOfWork (react-dom.development.js:19312)
    at workLoop (react-dom.development.js:19352)
updateVisibility @ mapHelpers.js:218
maybeUpdateDemesAndTransmissions @ map.js:354
maybeUpdateDemesAndTransmissions @ react-hot-loader.development.js:661
componentWillReceiveProps @ map.js:119
callComponentWillReceiveProps @ react-dom.development.js:11441
updateClassInstance @ react-dom.development.js:11651
updateClassComponent @ react-dom.development.js:14694
beginWork @ react-dom.development.js:15644
performUnitOfWork @ react-dom.development.js:19312
workLoop @ react-dom.development.js:19352
callCallback @ react-dom.development.js:149
invokeGuardedCallbackDev @ react-dom.development.js:199
invokeGuardedCallback @ react-dom.development.js:256
replayUnitOfWork @ react-dom.development.js:18578
renderRoot @ react-dom.development.js:19468
performWorkOnRoot @ react-dom.development.js:20342
performWork @ react-dom.development.js:20254
performSyncWork @ react-dom.development.js:20228
interactiveUpdates$1 @ react-dom.development.js:20495
interactiveUpdates @ react-dom.development.js:2170
dispatchInteractiveEvent @ react-dom.development.js:4882
[...]

Issue #722, opened 2019-04-23
Map cannot change geographic resolution in narrative mode


jameshadfield:
When in narrative mode, changing the geographic resolution causes the map component to break (no demes / transmissions rendered). If there are transmissions to be drawn, you get errors such as `Error trying to access x-y in transmissionIndices. Map transmissions may be wrong.`, otherwise there are no console errors.

* Since we can change the geo resolution via the sidebar dropdown, this shouldn't be too hard to fix.
* We should ensure we can change both color + geo resolution at the same time

[This narrative](https://wnv-paper-narrative.herokuapp.com/narratives/map-bug) recreates this error.

PR #736, opened 2019-07-06
Narratives bug

kairstenfay:
This fixes the bugs described in #722 and #731 (and maybe others?)

The last couple of commits are a WIP that need refactoring, but I won't have time to get back to it for a few weeks. 

It represents a joint effort between @joverlee521, @jameshadfield and me.


jameshadfield:
Thanks @kairstenfay and @joverlee521 🎉 

I've cherry-picked the commits onto `master` so that we can release this as a v1 release -- the only thing that really changed was the way traits are accessed on nodes (v1: `n.attr[geoResolution]`, v2: `getTraitFromNode(n, geoResolution)`).

PR #739, opened 2019-07-16
Use preset colours for (single) AA / NUC genotypes

jameshadfield:
Genotypes for a single aa/nuc will now always use the same colour scale. This helps with interpretation when comparing different genotypes.

Nucleotide:
image

Amino Acid colours via Taylor WR (1997) [Residual colours: a proposal for aminochromography](https://watermark.silverchair.com/100743.pdf). Protein Engineering 10(7) 743-746
image

### DIFF
--- a/src/util/colorScale.js
+++ b/src/util/colorScale.js
@@ -28,6 +28,45 @@ const createListOfColors = (n, range) => {
   return d3Range(0, n).map(scale);
 };
 
+/* AA colours taken from
+ * Taylor WR (1997) Residual colours: a proposal for aminochromography. Protein Engineering 10(7) 743-746
+ * https://watermark.silverchair.com/100743.pdf
+ */
+const colorsAA = {
+  D: "#FF0000", /* Asp Aspartate */
+  S: "#FF3300", /* Ser Serine */
+  T: "#FF6600", /* Thr Threonine */
+  G: "#FF9900", /* Gly Glycine */
+  P: "#FFCC00", /* Pro Proline */
+  C: "#FFFF00", /* Cys Cystine */
+  A: "#CCFF00", /* Ala Alanine */
+  V: "#99FF00", /* Val Valine */
+  I: "#66FF00", /* Ile Isoleucine */
+  L: "#33FF00", /* Leu Leucine */
+  M: "#00FF00", /* Met Methionine */
+  F: "#00FF66", /* Phe Phenylalanine */
+  Y: "#00FFCC", /* Tyr Tyrosine */
+  W: "#00CCFF", /* Trp Tryptophan */
+  H: "#0066FF", /* His Histidine */
+  R: "#0000FF", /* Arg Argnine */
+  K: "#6600FF", /* Lys Lysine */
+  N: "#CC00FF", /* Asn Asparagine */
+  Q: "#FF00CC", /* Gln Glutamine */
+  E: "#FF0066" /* Glu Glutamate */
+};
+
+const colorsNuc = { /* taken from https://www.umass.edu/molvis/drums/codes.html */
+  A: "#5050ff",
+  T: "#e6e600",
+  U: "#cc9900", /* not yet in augur but is a valid code */
+  G: "#00c000",
+  C: "#e00000",
+  K: "#ae00fe", /* Keto - G or T - not taken from DRuMS */
+  M: "#fd0162", /* aMino - A or C - not taken from DRuMS */
+  R: "#2e8b57", /* puRine - A or G */
+  Y: "#ff8c00" /* pYramidine - C or T */
+};
+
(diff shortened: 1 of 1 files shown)


trvrb:
@jameshadfield ---

I like where you're coming from here, but I'd like to request that you stick with colors drawn from https://github.com/nextstrain/auspice/blob/master/src/util/globals.js#L112. This has 20 colors. I find colors in the PR garish.

Screen Shot 2019-07-15 at 9 12 34 PM

Screen Shot 2019-07-15 at 9 15 01 PM

And I see now a separate issue as well. No matter what with 20 colors there will be AA combinations that are quite close like the above. I would think better to have individual AA colorings be clearly distinct (compare to the below) that to have full consistency across colorings.

Screen Shot 2019-07-15 at 9 16 34 PM

emmahodcroft:
Could we set this to be an option somewhere? For me, if I'm switching back and forth between a couple of sites, the fact the colours indicate different things makes things unnecessarily confusing, and I can see the same problem if I wanted to screen-grab these to use in a presentation or paper - it seems really random that the legends change. 

I see that this still has a problem of similar colours being selected, and I'm not sure I have the answer to that (I'll have a think). But certainly for me the priority would be consistency so that I don't have to preface slide/position changes with 'note the legend has changed...' every time :/

rneher:
Thanks James. This is great. While I think that colors can be improved, I feel quite strongly that
 1) the same nucleotide and amino acid should show up in the same color at different sites. 
 2) missing information (N/X) should be grey. 
Single site colorings are 99% of all color by genotype requests and there are good reasons to handle them separately. I also think that it is a useful feature to have similar amino acids have similar colors. An Isoleucine to valine change, for example, is not expected to do much in most cases and it seems sensible to reflect this in color distance. This would be particularly valuable for more diverged samples with substantial amino acid variation.

trvrb:
@emmahodcroft @rneher ---

Thanks for the thoughts. To start, I absolutely agree that "missing information (N/X) should be grey". We obviously diverge about consistent nuc/AA coloring across sites. I think this depends on whether the visual encoding / decoding task is meant to be primarily to make _evolutionary observations_ or to _identify genotypes_. I believe that the current approach is a good one for marking changes on the tree. When I row through mutations on the flu trees, I like this in keep major clade green and the minor clade yellow. This reduces cognitive load for me.

I see where you're coming from however in basically wanting to embed an alignment onto a series of colored trees.

I'd like to confront my preference with data. I'll play around with the two approaches and see how they feel for my common task of walking through flu variants and report back.

rneher:
testing this a bit is a good idea. My take is
 - I am sure it is possible to find colors that don't hurt the eye but also code for aa properties
 - for nucleotides, keeping the full legend with ACTG-N might reduce things jumping around
 - one aspect that differentiates nextstrain from most other phylo-viz projects is the tight integration with molecular evolution and phenotypes. This would be further accentuated when using aa colors that relate to aa properties. in the flu case, for example, the fact that 131 and 135 both pick up a charged residue might be important. So I would argue the moment you care about the genotype at a position, consistent coloring is preferable. If only the mutation pattern on the tree is of interest, coloring by abundance is sufficient.

rleir:
Hi all,
Who in the team has experience with design for Accessibility?  My limited knowledge of this is:
- Contrast is important. You need to choose colours with strong contrast (e.g. light greys are not OK). There are accessibility checker tools that rate the contrast on a page. The Leaflet map would fail this test badly. The tree is better, but also has contrast issues.
- You cannot appreciate how important this is to people with limited vision. For example, my father was a scientist and had failing eyesight, and I was trying to help him make good (or any) use of his computer.
- Zoom features are important. The Leaflet map is good at this (with caveats) but the tree does not have a convenient zoom.
- Some colours are more of a problem with common types of colour blindness (let's not worry about this at the moment).
I just want to mention Accessibility in passing, because it is just one of many issues. Let's keep it in mind while doing the more critical work.
Thanks
Rick

emmahodcroft:
Thank you for your thoughts @rleir ! We agree that being more friendly to the colourblind or those with reduced vision is important.
This issue is a bit separate from that, as here we're trying to discuss whether colours for bases (nucleotide or amino acid) should always be the same, or dynamic (as they are now). 

However we have an issue for more colourblind friendly colourings [here](https://github.com/nextstrain/auspice/issues/932) with further thoughts [here](https://github.com/nextstrain/auspice/issues/974) - we'd love to hear if you have any comments on these issues! 

Unfortunately at the moment we haven't been able to implement this yet - we are pretty crazy busy at the moment - but it's not off our radar, and as soon as we can, we will pick it up!

rleir:
Hi Emma, thanks for the explanation. Yes, agreed that it makes sense to have the colors for bases constant. Thanks for the links!

Issue #745, opened 2019-07-19
`num_date` missing -> crash


rneher:
I  am working with a very bare-bones json (export v2) and it makes auspice crash bc num_date is missing

![image](https://user-images.githubusercontent.com/8379168/61519141-740ecd00-aa0b-11e9-978f-a468cb2b204d.png)

jameshadfield:
Huh. It _shouldn't_ because div-only trees are/were fine. Will fix.

emmahodcroft:
### Bug 1:

Hmmm... I think this is because in `export v2` we do not check whether the tree has date information - it is assumed. 
See line [#184](https://github.com/nextstrain/augur/blob/33721adb5cd60c9f3411e2189c11c55a1e31591a/augur/export_v2.py#L184)

This is reflected in your JSON @rneher, as under `colorings` you have `num_date`.

So, it's breaking in a different part than our previous fix to detect whether date information is there or not - this is actually earlier. Since there's no 'country' (default colorby), it's cycling through the other colorbys given (from export) - and breaks when it looks into `num_date` (as passed by export).

**Fix**: `export v2` needs to check for date info and only add as a colorby if it exists.

-------

Richard just said that even if you get rid of this in the JSON it still errors. 

### Bug 2:

I believe this was introduced in [this commit](https://github.com/nextstrain/auspice/commit/969a58fb92a31c3b6cbed6dc73d680a636d50235), which is replacing the old style `node.attr.num_date` with `node.num_date.value`. 

The big difference here is that `node.attr.num_date` had the value if present and was undefined if not. So, you could pass it around and check whenever you like whether it was undefined. 

With `node.num_date.value` - `node.num_date` holds `value` and `confidence` (list of 2) if dates are present, and is undefined if not. If it's undefined, you get errors anytime you try to access `value`!

This can be seen in things like:
_Old code_:
`const nodeDate = node.attr.num_date;`
Then later we check whether `!nodeDate` and break (or whether it's > or < something, if it has value)

_New code_:
`const nodeDate = node.num_date.value;`
We can try to check later if `!nodeDate` - but we have already errored by tracking to access `value` from the undefined `node.num_date`...

**Fix**: Need to chase these replacements through relevant files and update them so we don't try to access `value` before we check if undefined.

emmahodcroft:
I pushed a PR for the second bug, and I think I've figured out the first too - but going Rhine swimming now - will push later! 🏊‍♀ 😎

[...]

PR #747, opened 2019-07-19
`node.num_date.value` Undefined Fix

emmahodcroft:
Should address **bug 2** of #745. 

Where changes to how we store `num_date` meant that we need to more explicitly check if it's undefined before trying to access it's value. See the bug comments for more information. 

@jameshadfield  May be worth checking if there are more instances where this may occur? I had a look and didn't see more, but this would be worth investigating slightly more in-depth I think...

### DIFF
--- a/src/components/tree/infoPanels/click.js
+++ b/src/components/tree/infoPanels/click.js
@@ -148,7 +148,8 @@ const displayPublicationInfo = (info) => {
 
 const TipClickedPanel = ({tip, goAwayCallback}) => {
   if (!tip) {return null;}
-  const showUncertainty = tip.n.num_date && tip.n.num_date.confidence && tip.n.num_date.confidence[0] !== tip.n.num_date.confidence[1];
+  const showDates = "num_date" in tip.n;
+  const showUncertainty = showDates && tip.n.num_date.confidence && tip.n.num_date.confidence[0] !== tip.n.num_date.confidence[1];
 
   return (
     <div style={infoPanelStyles.modalContainer} onClick={() => goAwayCallback(tip)}>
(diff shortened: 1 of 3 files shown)


emmahodcroft:
! If you click a tip, auspice crashes. Will investigate.

emmahodcroft:
Ok - same problem, different place. In this dataset, there is _no node data at all_. Previously, as detailed in the bug, we were able to skim by because `node.num_date` was undefined and this allows some behaviour it probably shouldn't have. On `master`, using a v1 dataset where there are no `num_dates`, note that 'Collection Date' still displays but just has no value (the value is 'undefined'):
image

This comes from this line at ~163 in click.js:
```
{item(uncertainty ? "Inferred collection date" : "Collection date", prettyString(tip.n.attr.date))}
```
`tip.n.attr.date` was undefined, but the code worked generally (no errors) - though it probably shouldn't have been showing this field at all.

In the new setup, the line is this (roughly same place):
```
showUncertainty ? "Inferred collection date" : "Collection date", prettyString(numericToCalendar(tip.n.num_date.value))
```

First I had to fix `showUncertainty` as this was also coming up as `undefined` from further up. But even when this is correctly set to `false`, `tip.n.num_date` is undefined, so trying to access `value` throws an error.

I have fixed this by creating extra variable `showDates` which is used to set `showUncertainty` properly, and then used to decide whether to even try to include the 'Dates' line that uses the above code.

This is the result for Richard's dataset:
image

rneher:
Thanks. Works for me now. But having silly colorBys in the json still crashes auspice.

emmahodcroft:
On the `export v2` end, this should be fixed with my latest PR to `augur`. But, if these have gotten in some other way, yea, they'll crash Auspice. Don't know if we want to handle this better...

jameshadfield:
@emmahodcroft my intention with `v2` auspice was that all places in the code which want a "trait" -- be that author, num_date, country, whatever, should call `getTraitFromNode` which can handle all the complexity. I'll take a look if this is applicable here.

rneher:
This fixes the crashes for a bare-bones json specifying unsupported colorbys

emmahodcroft:
@jameshadfield I haven't tested, but this looks a better solution! Thanks! :)

PR #750, opened 2019-07-24
Upgrade docusarus, lodash and move eslint to devDepedencies

trvrb:
* Docusaurus and lodash versions were throwing GitHub security errors. This should mostly address this.
* eslint is only used in development. It should be safe to move to devDependencies and slim down the resulting bundle.


jameshadfield:
slipped through the cracks -- thanks @trvrb

PR #751, opened 2019-07-24
Pie charts

rneher:


rneher:
this still requires some logic to determine when to draw what

trvrb:
Thanks for starting a PR here Richard. Some thoughts on design direction:

There are obvious situations where the piecharts are a huge improve, like for flu clades:

flu-circles

flu-piecharts

But there are other situations where the piecharts add noise rather than clarifying. This seems to perhaps be the case for continuous variables like Zika sampling date:

zika-circles

zika-piecharts

I'd recommend that we add a "PIE CHARTS" vs "CIRCLES" toggle to the upper right of the panel like we have with "ENTROPY" vs "EVENTS" in the Diversity panel. We could even default to one vs the other when we have a continuous vs discrete colorby.

jameshadfield:
@trvrb -- @rneher and I had a chat last night and we thought to implement them only for non-continuous variables. 

If we want to do them for continuous variables -- which I think might be really good -- we'd want to use the same binning & ordering that we do in the tree legend. (That is, we already have the concept of a continuous -> discrete mapping in auspice.)

rneher:
properly sorted, pie charts might make sense for continuous variables, but otherwise, they are just noise. For date, they would essentially encode an epi-curve (lots of blue==early, red==peak is late). Wouldn't be hard to do, but we would need to replace d3.pie by our own function.

rneher:
looks great!

jameshadfield:
I think this is ready for a bit of testing & merge into `v6` 🎉 -- it's looking great and a really useful improvement. I didn't explore using the binning of continuous values as in the legend values, instead opting to keep continuous variables rendering a single circle with a blended colour. We can explore that later on if desired.

## Visual changes
Top: H3N2 / HA / 3y geo res: country, color-by: clade
Bottom: H3N2 / HA / 3y geo res: region, color-by: genotype (142)

image

Ebola publications ⭐️ 👍 
image

## Speed reduction
There's been a slight speed reduction between v1 & v2, and a further one here (not for continuous variables). We should try to improve this -- perhaps in a separate PR, ideally before v2 release. Here are the results of a built bundle with timers included (`animation tick` is effectively the frame rate).

									pie		v2		v1
`phylotree.change` 					41ms	47ms	45ms
`map.updateDemesAndTransmissions` 	19ms	14ms	10ms
`animation tick` 					        111ms	123ms	104ms

## Code changes
@rneher I ended up changing a bunch of stuff (comments in the commit messages), but bigger points:
* All arc data is now within `demeData` -- this made it easier to pass around. It's only calculated when pie charts are to be drawn, and color averaging is only done when blended circles are shown.
* pie chart d3 selections are now `state.d3elems.demes` instead of passing around a 3rd selection.
* map position (e.g. if you've panned / zoomed) doesn't reset when you change geo resolution etc (nice improvement)
* removed the d3 transitions in the map code. This means we don't get as gradual of a change when changing color-bys, but it should speed things up a bit & caused bugs when trying to transition arc-paths.
* generally more commenting of code, improving some functions in the map code

rneher:
thanks, James. Your changes/refactor make a lot of sense.  The only issue that I don't understand right now is how the order of slices is preserved. my understanding is that d3.pie will order slices according to abundance. But changing visibility seems to changes slices in place without reordering them when small ones get big. It is probably better the way it is, but I don't understand how it works ;)

Other than that. I am happy with this.

trvrb:
This truly fantastic work @jameshadfield and @rneher. I just have a single a aesthetic / design issue that came up my review:

* I think that the lack of `d3.transition` on circles / pie charts is a significant loss. I find the resulting animation of say Ebola or Zika (what I often look at) to become far too jagged. Even when framerate is good the transitions between pie chart / circle diameters is too great and so it results in a juddery animation.

This is the only blocker for me to merge. Again, a huge improvement in so many circumstances.

I would also like to return to the binning of continuous traits and include a toggle, but this can certainly wait.

jameshadfield:
> The only issue that I don't understand right now is how the order of slices is preserved. my understanding is that d3.pie will order slices according to abundance. But changing visibility seems to changes slices in place without reordering them when small ones get big. It is probably better the way it is, but I don't understand how it works ;)

So `pie()` is only called once -- https://github.com/nextstrain/auspice/blob/pie_charts/src/components/map/mapHelpersLatLong.js#L126 -- with _all_ the potential colours (this required refactoring `mapDemes()` into `getColorsForAllDemes()` as I got super confused there). Then on update -- https://github.com/nextstrain/auspice/blob/pie_charts/src/components/map/mapHelpersLatLong.js#L422 -- we modify the arc angles ourselves rather than using d3. I should do some more checking here, but I check that the updating colour matches the arc colour and it _seems_ to be good so fat. 

> My only blocker is the d3.transition point as mentioned above.

No problem. I'll look into this -- I can't work out what a 200ms transition was really doing on a frame rate of 10-20fps, but clearly something. It proved problematic when d3 wanted to update the arc path and a transition hadn't finished, but this didn't seem to have been a problem for radius / colour updates.

trvrb:
>No problem. I'll look into this -- I can't work out what a 200ms transition was really doing on a frame rate of 10-20fps, but clearly something. 

I thought that it was legitimate to just keep calling `d3.transition` even if you call the next transition before the first has finished. And that in this case, D3 just starts a fresh transition to the new value that was handed to it. But I may well have this wrong.

jameshadfield:
> I thought that it was legitimate to just keep calling d3.transition even if you call the next transition before the first has finished. And that in this case, D3 just starts a fresh transition to the new value that was handed to it. But I may well have this wrong.

For some things, but it's always caused problems with path transitions (in the tree & now with pie charts). Might look nicer if we try to match the transition time to ~the frame rate... (one day!)

jameshadfield:
Looking into it a bit more I think ordering the arcs according to the legend is quite important. Currently the ordering is defined by the initial (only) `pie()` call, which orders by abundance. So if you then change the visibility (e.g. time slice) it preserves that initial ordering... which is different to if you refreshed the page at that point as that will trigger a new `pie()` call and different ordering. 

Screenshots: same parameters, same arcs (I think!), different ordering
image
image

trvrb:
Many of our discrete traits are effectively ordered. For https://nextstrain.org/WNV/NA?c=host for example, we'd prefer the `bird-` hosts to be next to each other and the `Mosquito-` hosts to be next to each other. Is there a way to force the initial ordering to be prespecified rather than be based on abundance?

rneher:
you can sort the pie...
https://github.com/d3/d3-shape#pie_sort

but if we are updated the arcs anyway, we might just dump d3.pie and use one function for both.

jameshadfield:
Pie charts now sorted by legend order 😄 (I've dumped `d3.pie` as it's easy to calculate ourselves). Still the outstanding issue of jittery updates during animation (pie charts only), and I fear it's not an easy solution.

image

rneher:
I really like this. great work. Personally, I am not too bothered by the jitter in animation as I next to never use them. jitter is certainly more noticeable when running animations on pie charts. But when running the animation in pie-chart mode without diversity (say genotype with only a rare mutation) I am not even sure there is an actual difference. just a quick impression though.

trvrb:
@jameshadfield, @rneher ---

This is excellent work. This is good to merge from my perspective. A couple thoughts however:

1. It still would be really nice to get `d3.transition` working on the pie charts. I believe this could be accomplished by making a `g` element to hold each separate pie chart and then translating and scaling this `g` element, rather than the current iteration through the atomized arcs. Also, we'd need to make 0 width arcs rather than having arcs that are created and destroyed. 
&nbsp;
However, I don't think this is important enough to block the merge. The majority of the animation use cases are when colorBy is matched to geoResolution (like the default viz state for Zika, Ebola and WNV) and these correctly transition using just circles. I'd shelve this out into a separate issue.

2. Can we merge this into `master` instead of `v2`? There is more development work needed before we can launch augur v6 / auspice v2 and I don't see why we should could back on launching this feature. It will be immediately hugely useful for flu.

trvrb:
Also, just for fun, I had experimented with some of D3's other options here. The branch `bubble_chart` implements [D3's bubble chart](https://bl.ocks.org/alokkshukla/3d6be4be0ef9f6977ec6718b2916d168). 

Here's Zika authors for pie chart vs bubble chart:

zika-pie

zika-bubble

And here's flu clades for pie chart vs bubble chart:

flu-pie

flu-bubble

I'm not a fan of how bubble chart turned out, but thought I'd share the experiment.

jameshadfield:
> It still would be really nice to get d3.transition working on the pie charts. 

> I believe this could be accomplished by making a g element to hold each separate pie chart and then translating and scaling this g element, rather than only iterating through the atomized arcs. 

I don't think I understand... How do you translate & scale to mimic changes in arc angles within the pie chart?

> Also, we'd need to make 0 width arcs rather than having arcs that are created and destroyed.

Yup -- we do this currently 😄 

---

> Can we merge this into master instead of v2?

Technically, yes (although it wouldn't be the nicest merge). I was hoping that as `v2` didn't have many visually visual changes that this could be a big selling point.

trvrb:
Thanks James.

1. Maybe this doesn't make any sense. But it seemed to my eye that the thing making adding 
```
d3elems.demes
      .data(individualArcs)
      .transition()
      .duration(2000)
      .ease(easeLinear)
      .attr("d", (d) => arc()(d))
      .style("fill", (d) => { return d.color; })
      .style("stroke", (d) => { return d.color; });
```
to `updateVisibility` not look good was that the different arcs "travel" to their eventual locations in different paths (with some translation that appears to occur). I thought that enclosing in a `g` element could reduce this.

2. I see... that's _just_ fine. If you'd like to keep this as a v2 feature, please go ahead.

jameshadfield:
> the thing making ... updateVisibility not look good was that the different arcs "travel" to their eventual locations in different paths (with some translation that appears to occur). I thought that enclosing in a g element could reduce this.

I think it's a d3 issue to do with `.attr("d", (d) => arc()(d))`. When an animation tick (or just scrubbing through the date range) causes a d3 update call and there's already a previous one transitioning, then there seems to be a problem with updating the path attr. Doesn't seem to happen with calls to change the radius, stroke, fill etc as we do this with the blended circles. Might be fixable.

Currently, the number of arcs (and what each arc represents -- i.e. which pie chart, which color-by value) is fixed & doesn't change when you update things. (It changes when you change color-by or geo resolution as we recalculate things.)

image

trvrb:
Hmm... Yeah, I don't know what to do about this. You may have found this already, but this bl.ock was the most relevant thing I could find with Googling: http://bl.ocks.org/mbostock/5100636

trvrb:
Let's merge this into `v2`. Further development for transitions etc... are happening on branches off of `pie_charts` anyway.

PR #752, opened 2019-07-29
Recalculate radial and unrooted tree layout on clade zoom

trvrb:
This PR started with the direct goal of recalculating radial and unrooted tree layout on clade zoom. This was basically making `radialLayou`t and `unrootedLayout` depend on node visibility to set bounds. These changes required some additional modifications however. These were:
* Inclusion of a new node visibility flag to distinguish nodes that are filtered from nodes that outside of the "shell". Zooming into a radial or unrooted clade and applying layout to other branches resulted in strange layouts of out of view nodes. This had the added benefit of cleaning up the thin branches that interfered with the x-axes when zooming to clade in rectangular layout.
* A new change arg called `updateLayout` is passed through reactD3Interface when zooming to clades that causes layout recalculation in a modifySVG action.

I believe this is a slight improvement over existing behavior (but definitely not as much as I was hoping for when I started on it), but please @jameshadfield, @rneher, @emmahodcroft , etc... let me know if you think existing zoom behavior is preferable. Or if you have other directions you think to take this. There are a few things (like dynamically updating radial grid lines) that need cleaning up in this PR.

Zika radial zoom before and after:

zika-before

zika-after

Zika unrooted zoom before and after:

unrooted-before

unrooted-after

The zoom animation does come out looking rather fun for these.

_And I see now that this is mostly a reimplementation of @jameshadfield's `js-v-vals` branch. I think I prefer the current behavior of clade zoom for rectangular trees, but keeping more context around for unrooted and radial trees seems helpful._

### DIFF
--- a/src/components/map/mapHelpersLatLong.js
+++ b/src/components/map/mapHelpersLatLong.js
@@ -4,7 +4,7 @@ import _minBy from "lodash/minBy";
 import { interpolateNumber } from "d3-interpolate";
 import { averageColors } from "../../util/colorHelpers";
 import { bezier } from "./transmissionBezier";
-import { NODE_VISIBLE_TO_MAP_ONLY, NODE_VISIBLE } from "../../util/globals";
+import { NODE_NOT_VISIBLE } from "../../util/globals";
 import { errorNotification } from "../../actions/notifications";
 
 /* global L */
@@ -64,8 +64,7 @@ const setupDemeData = (nodes, visibility, geoResolution, nodeColors, triplicate,
   // second pass to fill vectors
   nodes.forEach((n, i) => {
     /* demes only count terminal nodes */
-    if (!n.children && visibility[i] === NODE_VISIBLE_TO_MAP_ONLY
-      || !n.children && visibility[i] === NODE_VISIBLE) {
+    if (!n.children && visibility[i] !== NODE_NOT_VISIBLE) {
       // if tip and visible, push
       if (n.attr[geoResolution]) { // check for undefined
         demeMap[n.attr[geoResolution]].push(nodeColors[i]);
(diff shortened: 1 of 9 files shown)


jameshadfield:
Cool! These directions are valuable & fun to play with. 

What you've done here I think brings radial & unrooted views much closer to the current rectangular tree behavior. By this I mean that zooming into "clades" takes all the other branches & tips off screen -- looking at clade A1b/131K means you can no longer see the other ones. Changing the filtering / date range has no effect on the layout (it turns branches into skeleton ones, but the layout is the same).

There are some d3 transition details that I think would need changing. In rectangular trees, zooming into a clade is like bringing the camera into that part of the tree, so you understand what's going on. But with radial / unrooted trees it's hard to get a feeling for what's happening when you zoom into a branch -- i.e. you loose the relationship of how this clade fitted into the bigger tree.

---

Re: `js-y-vals` -- one day I'd love to revisit this. It was a bit different in the sense that I wanted to keep the entire tree in the field of view at all times, but squash nodes that weren't "visible" (e.g. filtered as well as zoomed). The idea being that you would always be able to see how your selection fitted into the entire tree. This worked really well in the y-direction but not so well in the x-direction.

trvrb:
You're right that these don't feel like a camera zoom and so are rather disorientating. For both radial and unrooted, I'd like it to feel like the camera is zooming into the clade of interest and alongside this, the clade unspools a bit to fill the frame better. I'll keep tinkering with this.

trvrb:
@jameshadfield: Try unrooted zoom here now. I think this achieves most of what I wanted with 
>zooming into the clade of interest and alongside this, the clade unspools a bit to fill the frame better
I've reverted radial zoom while I try to get it working.

trvrb:
I'm going to rebase this without all the cruft as a separate PR targeted just at unrooted zoom.

PR #754, opened 2019-07-29
Unfurl clades during unrooted clade zoom

trvrb:
This PR is specific for the unrooted aspect of PR #752.

This alters unrooted tree layout to give more radial territory to tips that are "inView" compared to tips that are out of view. This causes clade zoom to unfurl clades when zooming. Besides altering unrootedLayout, this commit adds an "updateLayout" flag that causes layouts to be recalculated on clade zoom to account for this. I believe this new behavior is now a definite improvement over previous behavior.

Previous, example Zika clade zoom:

zika-old

New zoom:

zika-new

There are separate improvements to make to margins, but I thought better to shelve them off from this PR.

### DIFF
--- a/src/components/tree/phyloTree/change.js
+++ b/src/components/tree/phyloTree/change.js
@@ -250,6 +250,7 @@ export const change = function change({
   /* change these things to provided value */
   newDistance = undefined,
   newLayout = undefined,
+  updateLayout = undefined,
   newBranchLabellingKey = undefined,
   /* arrays of data (the same length as nodes) */
   branchStroke = undefined,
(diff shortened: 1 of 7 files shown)


jameshadfield:
Looks awesome 🎉 

(i) _completely unrelated to this PR and too much work_ I've just realised that the tree is still rooted, in the sense that clicking on a node (e.g. zika top right) that's close to the root zooms into the rest of the tree, not that clade. 

(ii) _also beyond this PR_ we should add in a scale to the unrooted view!

(iii) _also beyond this PR_ the clade label text should be made to not capture mouse events. You can't click on a branch behind text! (I probably didn't realise this as in rectangular trees the text isn't over branches)

(iv) code looks good. I removed `addLeafCount` from the initial phylotree setup as it's only ever used in unrooted layouts and each call to `unrootedLayout` will call it.

trvrb:
Thanks for the review James and for the code updates. This definitely made me want a scale in this view. And there might be a way to be smarter about text placement.

For (i), I don't immediately see how to avoid a "rooting" for these clade selections. If you point at a random branch we need to know whether the clade in question is to one side or the other of this branch. Without a rooting, I don't see how a single mouse click could specify right or left.

jameshadfield:
> I don't immediately see how to avoid a "rooting" for these clade selections. If you point at a random branch we need to know whether the clade in question is to one side or the other of this branch. Without a rooting, I don't see how a single mouse click could specify right or left.

Yeah, and there's no way it's worth the effort it would take to sort out. Typically, in unrooted views I think you'd want to zoom into the side with fewer tips. But again, not a good investment of time for us.

PR #761, opened 2019-07-31
prototype of "build source" in header

jameshadfield:
@trvrb here's a skeleton prototype of displaying a GitHub link for all builds (ours & community) as discussed in Slack. Simply provides the basic mechanism for us to iterate on 😄 

image


trvrb:
@jameshadfield: I'm so sorry that I missed this / forgot about this in diving in with #821.

Issue #760, opened 2019-07-31
Investigate how strings are being displayed


jameshadfield:
One of the bigger (but subtle) changes intended for v2 JSONs relates to how strings are displayed in auspice. Currently (v1 & v2) `prettyString()` performs a number of different string manipulations intended to improve trait display in auspice, e.g. "usvi" -> "USVI". 

As much as possible these changes should be moved upstream to `augur export v2` and the schema made to support that where needed. `prettyString` can be preserved on the auspice server to convert v1 JSONs to v2.

This means that trait values (e.g. "usvi", "usa") should be stored in the v2 JSON as they wish to be displayed in auspice. For authors this is `JSON.author.value`.

Currently used:
* in author download TSV creation. Should use the values as supplied by the v2 JSON.
* in strain download TSV creation. Should use the values as supplied by the v2 JSON.
* in footer "filters" (both title & items). Title should use the coloring title (as does the legend & dropdown). Items should display as found in the JSON.
* frequencies panel labels & hover infobox. Should display as found in JSON.
* info header. Same as filter display.
* tree click & hover panels: represent values as found in the JSON, with the potential exceptions of
    * stylising dates (which are already converted from numeric to calendar within auspice).
    * stylising divergence / numerical values
* legend values. Display as per JSON.

joverlee521:
I've pushed all of my changes to branch `use-json-strings`. 

There's one thing I can't figure out what do to: 
When a filter is not a colorings option, I'm currently just displaying filter name, which has not been "prettied" in the conversion function. See `country` below:

![Screen Shot 2019-08-16 at 5 36 57 PM](https://user-images.githubusercontent.com/40774762/63204442-8866d980-c04c-11e9-9f57-9d5d916ac177.png)

I'm not sure if we should "pretty" these filter names (when I tried to "pretty" the filters lots of things broke; it might just be my lack of understanding of how everything is connected within Auspice).

jameshadfield:
Thanks @joverlee521 -- this work has been merged into v6 with ec924d762946fbd9db7567f5201997b99dd681cf

Good find re: filter names which are not available as colorings and therefore don't have a "title" for display. In this case I think displaying the raw string is fine 👍

PR #764, opened 2019-08-06
Improve V2 author display

trvrb:
This is tested to improve simultaneous display of author fields in V1 JSONs and in V2 JSONs

This resolves:
* Don't CamelCase to "Et Al"
* Properly display author, title and journal in clicked info panel

This can be tested on http://staging.nextstrain.org/zika-v2.json.

### DIFF
--- a/src/components/framework/footer.js
+++ b/src/components/framework/footer.js
@@ -321,6 +321,13 @@ class Footer extends React.Component {
                   {" et al (" + totalStateCount[itemName] + ")"}
                 </span>
               );
+            } else if (filterName === "author") {
+              display = (
+                <span>
+                  {itemName}
+                  {" (" + totalStateCount[itemName] + ")"}
+                </span>
+              );
             } else {
               display = (
                 <span>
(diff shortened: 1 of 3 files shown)


jameshadfield:
@trvrb this might be related: https://github.com/nextstrain/auspice/issues/760 (or might not)

In auspice v2, the string as it appears in `author.value` will be displayed as is with no string manipulation.

trvrb:
@jameshadfield --- I like this direction. This PR makes this change to direct display explicitly for `author.value`.

jameshadfield:
@trvrb best to leave this until after @joverlee521 has closed https://github.com/nextstrain/auspice/issues/760 I think, mainly because the client will no longer see "authors" after this - the server will convert "authors" into "author" and run the `prettyString` function during the conversion.

P.S> I'd rather we use `getTraitFromNode(tip.n, "author")` to access node traits as it allows us to not have the structure of the node information hardcoded across auspice.

jameshadfield:
I think the current behavior is what's desired, but please re-open if I'm missing something.

Currently (https://nextstrain-dev.herokuapp.com/staging/WNV/NA) we have
`WestNile4K`, `Anderson et al` etc

Which are taken directly from the JSON via a `node -> author -> value` lookup. In this example we have nodes with
```
"author":{..., "value":"Anderson et al"},
``` 
and 
```
"author":{"title":"WestNile 4K, ...,  "value":"WestNile4K"}},
```

PR #771, opened 2019-08-15
Improvements to Codon and Nuc Bars & Mouseovers

emmahodcroft:
I got this started just to fix item 1 below, which James noticed was wrong, but then I got rolling and did some stuff that I've been wishing for for a while.

1. Negative strand genes now plot codons in the right position (10 codons from the end/right rather than 10 codons from the start/left).
Compare 10 AA (orange) and 30AA (grey) mutation in this gene which is on the negative strand before:
image

    And after:
  image
(Sorry, the 10AA is no longer orange, but it's the one to the far right.)

2. While doing this I got annoyed about how codon and nucleotide boxes are drawn. Here, the orange box is a codon for positions 5326-5328. It's just kind of floating randomly between 5325 and 5330.
The one to the left is worse - it's for positions 5323-5325, but it's plotted starting at 5325 and going to the right!
   image

   This now plots like this. The orange actually covers 5326-5328. The one to the left now looks like it sits nicely on 5323-5325:
   image

   I did the same thing for nucs - currently they sit to the right of what they are. Here the orange one is nuc 5325:
   image

    Now it sits actually on the number!:
   image

3. Finally, I wanted to make it easier to connect between the codons and the nucleotides. This is something I've dreamed of for a long while - knowing an interesting codon is number 103 in a gene in the middle of your genome can be not-so-helpful for more closely examining the associated nucleotides...

   So now, on a nucleotide in a gene, you can see what codon it is:
   image

   And for a codon, you can see what nucleotides they are!:
   image

    (Also shows the strand) This works on negative strand too, and is possibly even more useful, since this is super not-intuitive:
  
   image
  
   image

I wanted to write all this up before I forget (also because I'm very excited by it, this would be so useful for me), but welcome comments! Will have to be modified when the GFF changes are merged to look for `+` and `-` instead of `-1` and `+1` for strand. 

And to preempt maybe a comment on how the 'fat' codon view above looks - if you aren't really zoomed in the codons look very similar to what we have now. Only when you really zoom in do they look 'fat'. But this means their actual position is shown *so much more clearly*! 😄

### DIFF
--- a/src/components/entropy/entropyD3.js
+++ b/src/components/entropy/entropyD3.js
@@ -107,7 +107,7 @@ EntropyChart.prototype._aaToNtCoord = function _aaToNtCoord(gene, aaPos) {
   if (this.geneMap[gene].strand === "-") {
     return this.geneMap[gene].end - aaPos * 3 + 1;
   }
-  return this.geneMap[gene].start + aaPos * 3;
+  return this.geneMap[gene].start + aaPos * 3 - 2; // Plot from 1st codon position, not last.
 };
 
 EntropyChart.prototype._getZoomCoordinates = function _getZoomCoordinates(parsed, geneMap) {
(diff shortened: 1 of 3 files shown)


emmahodcroft:
Was slightly off on coordinates, noticed when making those screenshots. Codons sit better now:
image

rneher:
Thank you! Great improvement!

jameshadfield:
Nice @emmahodcroft, thanks!
I think i've got everything working with the GFF merges, but please double check (via the "v2" branch) with your Walker et al. dataset

emmahodcroft:
Looks good! Tested on full-genome EVD68 v2, TB_walker v1 and v2, and VP1-only EVD68 v1 and v2. 

I also checked how the new bars look when you have codons/nucs right at the beginning and end of your genome display, and this looks good :)

One very tiny small thing, which is probably not worth addressing but since we've been into it:
image

All our one-based numbering is beautiful, but we are still drawing genes from 0 (for full-genomes this is generally a non-issue due to UTRs at the beginning). Not sure it's worth any time being spent on this, but it's recorded for posterity now :)

PR #777, opened 2019-08-26
Update frequencies tooltips and include projection pivot

trvrb:
@jameshadfield, @rneher  ---

This PR is motivated by @huddlej's recent flu forecasting work. We thought it best to not mess with the underlying data structure of `tip-frequencies.json`. We've kept this JSON structure exactly as before but just included an additional attribute of `projection_pivot`:
```
{
 "A/Aberystwyth/5949/2018": {
  "frequencies": [
   0.0,
   0.0,
   0.0,
...
  ]
 },
...
 "pivots": [
  2017.67,
  2017.75,
  2017.83,
...
 ],
 "projection_pivot": 2019.67
}
```

This way the frequency streamplot can be drawn as before, but we just mark down with a dashed line what derives from empirical `augur frequencies` and what derives from a forecast.

Screen Shot 2019-08-26 at 9 41 25 AM

I've added `state.frequencies.projection_pivot` to redux to accomplish this. If you want a dataset to demo this with use: https://nextstrain.org/staging/flu/seasonal/h3n2/ha/2y

Importantly, this is backward compatible with previous versions of auspice. The forecast will still show correctly, but we just won't get the dashed projection pivot.

------------------------------------

Additionally, this improves frequencies tooltips to be specific to the frequency bin rather than show a full vertical line, as so:

Screen Shot 2019-08-26 at 9 44 51 AM

And makes a few aesthetic improvements (white text on black background and tooltip placement).

### DIFF
--- a/src/components/frequencies/functions.js
+++ b/src/components/frequencies/functions.js
@@ -74,6 +74,10 @@ const removeYAxis = (svg) => {
   svg.selectAll(".y.axis").remove();
 };
 
+const removeProjectionPivot = (svg) => {
+  svg.selectAll(".projection-pivot").remove();
+};
+
 export const drawXAxis = (svg, chartGeom, scales) => {
   removeXAxis(svg);
   svg.append("g")
@@ -83,6 +87,7 @@ export const drawXAxis = (svg, chartGeom, scales) => {
     .style("font-size", "12px")
     .call(axisBottom(scales.x).ticks(scales.numTicksX, ".1f"));
 };
+
 export const drawYAxis = (svg, chartGeom, scales) => {
   removeYAxis(svg);
   svg.append("g")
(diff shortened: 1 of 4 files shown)


jameshadfield:
Looks good -- excited to see what @huddlej's cooking up

Tested and confirmed no regressions with current datasets (i.e. those lacking `projection_pivot`). Tested new functionality by adding `"projection_pivot": 2019.25` to http://staging.nextstrain.org/flu_seasonal_h3n2_ha_2y_tip-frequencies.json (it wasn't there, BTW) and behaves as per your screenshots. 

Did you investigate changing the color / opacity of the streams to the right of the `projection_pivot`, or adding some text directly above the graph to the right of the dashed line? It wouldn't be obvious to me what the line represents, especially when the dates to the right of it are within the date range of the dataset.

jameshadfield:
Happy for this to be merged by @trvrb / @huddlej. I'll release new auspice / nextstrain.org when I see it's been merged.

trvrb:
Thanks for the feedback James. I agree that the fact this is a projection is opaque. I'll experiment with strategies to visually distinguish before and after.

trvrb:
Okay. I made a couple minor changes to try to make projection more obvious. These were to:
1. Add a "Projection" label centered over the projected portion of the frequencies streamgraph
2. Change tooltip labeling to "Projected frequency" when appropriate

Screen Shot 2019-08-28 at 9 40 01 PM

trvrb:
@jameshadfield: I'm going to merge this. Thanks again. Any additional aesthetic changes can happen on top of `master`.

Issue #753, opened 2019-07-29
plan to get rid of magics from data JSON


jameshadfield:
Over various iterations of the JSON sent to the client (not the JSON produced by augur; [see here](https://github.com/nextstrain/auspice/pull/720#issue-272469144) for background) we have had 3 keys injected in various guises. As we move to having the dataset JSONs handed directly to the client (no server modifications) we need a plan to get rid of these.

## Tree name.

See #776
~Variously injected as `_treeName` or `treeName`.~

~This is used by auspice's choose-second-tree dropdown selector, as it matches the dataset name against all others in the available dataset which differ only in tree name.~

~Proposed solution: provide a list of possible second-trees in the available dataset JSON. The dropdown simply renders these. Goes nicely with the proposal to represent tangle trees in the URL via `/flu/seasonal/h3n2/ha/2y:/flu/seasonal/h3n2/pb1/2y`~

## Display URL
Currently injected as `_url` (v1) or `auspice_url_should_be` (v2)

This is how a request for https://nextstrain.org/flu/ goes to https://nextstrain.org/flu/seasonal/h3n2/ha/3y. 

Proposed solution: Server can issue a redirect to the desired URL, and then auspice can refetch from this URL.

## source
Currently injected as `_source`.

Largely a historical legacy and only used by the dataset-display sidebar part of auspice. This is how auspice knows to display the github URL in the sidebar as opposed to a listing of available datasets. 

Proposed solution: Write logic in auspice to decided this from the URL. Allow build-time customisations to modify this.

jameshadfield:
Tree name -> #776
Display URL -> now competed & in branch v2
Source -> #783

PR #778, opened 2019-08-26
Implement server redirects in nextstrain.org to handle incomplete request urls

kairstenfay:
The part of issue #753 pertaining to Auspice v1 magic `_url`s is addressed via this PR.
See nextstrain/nextstrain.org/pull/25 for the nextstrain.org side of things.

Auspice `v2` doesn't quite work with Nextstrain.org yet, so I branched off of `master` instead for this work.

Note that it doesn't appear that changing the browser window path has to go through Redux at all. This is deeply suspicious to me, but so far all my testing suggests this works just fine. I would love some additional eyes on it.

### DIFF
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -84,15 +84,17 @@ const fetchDataAndDispatch = (dispatch, url, query, narrativeBlocks) => {
     fetchExtras += `&deprecatedSecondTree=${query.tt}`;
   }
 
-
   // fetchJSON(`${charonAPIAddress}request=mainJSON&url=${url}${fetchExtras}`)
+  let pathnameShouldBe;
   getDataset(`${url}${fetchExtras}`)
     .then((res) => {
+      pathnameShouldBe = queryString.parse(res.url.split("?")[1]).prefix;
       return res.json();
     })
     .then((json) => {
       dispatch({
         type: types.CLEAN_START,
+        pathnameShouldBe,
         ...createStateFromQueryOrJSONs({json, query, narrativeBlocks})
       });
       return {
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -12,6 +12,18 @@ import { hasExtension, getExtension } from "../util/extensions";
 /* TODO: make a default auspice server (not charon) and make charon the nextstrain server. Or vice versa. */
 const serverAddress = hasExtension("serverAddress") ? getExtension("serverAddress") : charonAPIAddress;
 
+/**
+ * Sends a GET response to the `/charon` web API endpoint requesting data.
+ * Throws an `Error` if the response is not successful or is not a redirect.
+ *
+ * Returns a `Promise` containing the `Response` object. JSON data must be
+ * accessed from the `Response` object using the `.json()` method.
+ *
+ * @param {String} prefix: the main dataset information pertaining to the query,
+ *  e.g. 'flu'
+ * @param {Object} additionalQueries: additional information to be parsed as a
+ *  query string such as `type` (`String`) or `narrative` (`Boolean`).
+ */
 const getDatasetFromCharon = (prefix, {type, narrative=false}={}) => {
   let path = `${serverAddress}/${narrative?"getNarrative":"getDataset"}`;
   path += `?prefix=${prefix}`;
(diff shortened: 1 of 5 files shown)


src/middleware/changeURL.js
```diff
@@ -174,10 +172,7 @@ export const changeURLMiddleware = (store) => (next) => (action) => {
   if (pathname !== window.location.pathname || search !== window.location.search) {
     let newURLString = pathname;
     if (search) {newURLString += search;}
-    // if (pathname !== window.location.pathname) {console.log(pathname, window.location.pathname)}
-    // if (window.location.search !== search) {console.log(window.location.search, search)}
-    // console.log(`Action ${action.type} Changing URL from ${window.location.href} -> ${newURLString} (pushState: ${action.pushState})`);
-    if (action.pushState === true) {
+    if (action.pushState) {
       window.history.pushState({}, "", newURLString);
```
jameshadfield: > it doesn't appear that changing the browser window path has to go through Redux at all

Are you referring to this line?

jameshadfield:
Thanks @kairstenfay 👍 

How can I test the redirect functionality? Is there an associated PR on the nextstrain.org repo?

src/middleware/changeURL.js
jameshadfield: > but I'm not sure all what that function is doing

This is the way by which the URL updates when (e.g.) you change the color-by, or if you change dataset via the dropdowns, or how it used to change via the `_url` property. From comment at top of that file: 

What is this middleware?
This middleware acts to keep the app state and the URL query state in sync by intercepting actions and updating the URL accordingly.

jameshadfield: Good to know! I anticipated that middleware may be confusing (not many people use it, although plenty of libraries do underneath the surface). Could you amend the comment at the top of `changeURL.js` so that it's less confusing? Thanks

src/actions/loadData.js
```diff
@@ -12,42 +12,73 @@ import { hasExtension, getExtension } from "../util/extensions";
...
+ */
 const getDatasetFromCharon = (prefix, {type, narrative=false}={}) => {
   let path = `${serverAddress}/${narrative?"getNarrative":"getDataset"}`;
   path += `?prefix=${prefix}`;
   if (type) path += `&type=${type}`;
 
   const p = fetch(path)
     .then((res) => {
-      if (res.status !== 200) {
+      if (!(res.status === 200 || res.status === 301)) {
```
jameshadfield: In all my testing, the client never actually sees the redirect (301), so should this still be here?

src/actions/loadData.js
```diff
@@ -61,9 +92,13 @@ const fetchDataAndDispatch = (dispatch, url, query, narrativeBlocks) => {
     fetchExtras += `&deprecatedSecondTree=${query.tt}`;
   }
 
-
-  // fetchJSON(`${charonAPIAddress}request=mainJSON&url=${url}${fetchExtras}`)
   getDataset(`${url}${fetchExtras}`)
+    .then((response) => {
+      url = response.url.split('prefix=')[1];
```
jameshadfield: Currently `prefix=X` is the last part of the query, so this technique works, but if there were to be another part to the query this wouldn't work as desired. Better to use `queryString` to parse the query I think.

src/actions/loadData.js
```diff
@@ -88,6 +97,9 @@ const fetchDataAndDispatch = (dispatch, url, query, narrativeBlocks) => {
   // fetchJSON(`${charonAPIAddress}request=mainJSON&url=${url}${fetchExtras}`)
   getDataset(`${url}${fetchExtras}`)
     .then((response) => {
+      url = response.url.split('prefix=')[1];
+      window.history.replaceState({}, "", url);
+
```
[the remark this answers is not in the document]
jameshadfield: > Note that it doesn't appear that changing the browser window path has to go through Redux at all. 

Correct -- one way to think of them is that they are independent sources of "state". It's on us to keep these two bits of state in sync, and there is always the potential for bugs! This is why all of the URL manipulation is currently happening in one place - the `changeURL.js` middleware. 

Because of this, this should be moved into `changeURL.js` -- this can be achieved by adding the `url` to the dispatch on line 105 and picking it up in the middleware. Note that this is very similar to how the `_url` magic key currently works, except we're no longer getting it from the JSON anymore 🎉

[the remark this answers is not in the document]
jameshadfield: I know there's some confusion about `changeURL.js` so I made a [small example branch](https://github.com/nextstrain/auspice/tree/tmp-redirect-example) with a single commit 872872337880c1de8ec6690a824272a52dcaf2df intended to show what I am talking about.

It works with your nextstrain.org server changes (https://github.com/nextstrain/nextstrain.org/pull/25), and uses your approach to separating out the response and conversion to JSON, but performs the URL modification within `changeURL.js`.

docs-src/docs/installation.md
```diff
@@ -34,7 +34,14 @@ npm install -g . # make "auspice" available globally
 ```
 
 Note that (at least on MacOS) this symlinks the source directory into the global `node_modules` folder so that changes to source code are automatically reflected in the globally available `auspice` command.
-We have had good success using a conda environment for development.
+
+To separate the npm-distributed and your development version of auspice, we recommend running Auspice inside of a virtual environment such as [Conda](https://conda.io/en/latest/).
```
jameshadfield: Good addition -- this is also detailed on the nextstrain.org docs: https://nextstrain.org/docs/getting-started/local-installation#install-auspice-from-source. Can you include this link here? (or just point users to those docs)

kairstenfay:
Done again incorporating your changes and using your example branch. Thanks!

jameshadfield:
Thanks @kairstenfay -- i've put these changes onto the v2 branch, and you can see the result by going to nextstrain-dev.herokuapp.com/flu/ and seeing it redirect  👍

Issue #774, opened 2019-08-16
ColorBys with Missing Tip Values & Not in `colors.tsv` Break Auspice 2.0


emmahodcroft:
Running Auspice `v2` branch, with JSONs generated from `augur` `v6` branch using `export v2`. 

The current Enterovirus D68 build as on the Nextstrain website, I've used for lots of testing - all colorbys work great here with all the v2/v6 stuff. 

However, today I tried to add 2 new traits, "source"  and "symptom". The important thing to know about these is that *not all tips have a value for this* (they have missing tip values) and *they have no value in `colors.tsv`*. 

This makes them different from other colorbys (which all work):

Country and region are categorical, have no missing tip values, and are in `colors.tsv`. 🔴✔️✔️
Clade is categorical, has no missing tip values, and is _not_ in `colors.tsv`. 🔴✔️❌ 
Host age is continuous, has missing tip values, and is _not_ in `colors.tsv` 🔵❌❌ 
Age range 1, 2, and 3, are categorical, have missing tip values, and are in `colors.tsv` 🔴❌✔️

Source and symptom are categorical, have missing tip values, and are _not_ in `colors.tsv`  🔴❌❌

_Clade gives the value "Unassigned" to those not in a clade, so does not technically have any missing values_
_(I'm excluding GT and date because these are a bit different and the same rules don't always apply.)_

I can break `age_range3` in the same way by removing it from `colors.tsv`. I can also make it break by including boolean data (true/false, 0/1, etc) that's not in `colors.tsv` if it has any missing values. (It works if it has no missing values.)

So this is the bad combo: categorical or boolean, missing tip values, not in `colors.tsv`.

----

The error occurs when trying to create the pie-chart circles:
![image](https://user-images.githubusercontent.com/14290674/63185370-244b1200-c05a-11e9-8073-aa85124f2ca9.png)
(Line numbers are off because of my debug. Real line is line [97](https://github.com/nextstrain/auspice/blob/998d46e271e4638b3afe35849253d21eb24b2aa8/src/components/map/mapHelpersLatLong.js#L97) in `mapHelptersLatLong.js`)

----

I _believe_ the problem is due to the fact that there are no `undefined` counts, which mean that when counting up the number of traits in a pie-chart, they do not equal to the total number of things in that pie-chart. 

So as an example with `age_range2`, `legendValues`, which is used to map counts of each possible option, has four values, including `undefined`:
[...]

PR #780, opened 2019-08-28
bugfix when calculating map arcs if nodes were missing coloring data

jameshadfield:


emmahodcroft:
This fixes my error - thanks!

jameshadfield:
Cheers @emmahodcroft -- will merge into v2 & test with the (live) enterovirus example from https://github.com/nextstrain/auspice/issues/774#issuecomment-525747727

PR #781, opened 2019-08-28
cli: Listen on only localhost by default instead of all interfaces

tsibley:
It is a safer to not expose services over the network by default.

The host to listen on can be overridden by setting the HOST environment
variable, which complements the existing PORT environment variable.


tsibley:
Done.

Issue #726, opened 2019-05-25
ordinal color scales


rneher:
we allow discrete and ordinal color scales in v2 schema. the order of discrete scales is determined by abundance. ordinal scales should be sorted by value. This is currently not implemented but would easy to add here:

https://github.com/nextstrain/auspice/blob/master/src/util/colorScale.js#L137

any thoughts?

jameshadfield:
Hi @rneher -- this is definately needed, however it should happen on the v2 branch. Currently (v2 branch, for v2 JSONs) ordinal & discrete both use the same scale,

PR #786, opened 2019-09-03
Ordinal vs categorical

rneher:
This implements ordinal scales (integer) for quantities like RBS or epitope mutations and falls back on continuous scales when the number of values exceeds the longest defined sequential colorscale. Addresses issue #726.

It leaves categorical scales untouched, but they should probably use a categorical rather than a sequential scale (as currently the case for clades).

### DIFF
--- a/src/components/framework/footer.js
+++ b/src/components/framework/footer.js
@@ -329,11 +329,11 @@ class Footer extends React.Component {
         {`Filter by ${filterTitle}`}
         {this.props.activeFilters[filterName].length ? removeFiltersButton(this.props.dispatch, [filterName], "inlineRight", `Clear ${filterName} filter`) : null}
         <Flex wrap="wrap" justifyContent="flex-start" alignItems="center" style={styles.citationList}>
-          {Object.keys(totalStateCount).sort().map((itemName) => {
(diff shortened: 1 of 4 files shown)


jameshadfield:
@rneher thanks for this! I've left the ideas here unchanged, but made some changes:

1. All the ordinal code is now in a single block -- this added a bit of code duplication but makes the logic easier to follow.
2. Some modifications to how we count traits across the tree to allow integer / numeric values (before they were coerced into strings).
3. I added a test build to augur v6 using @emmahodcroft's TB data -- see https://github.com/nextstrain/augur/commit/63e97a7dcfa8ceff0b86749127d916a243cb82f9. This works well. Please test on your flu datasets too! 🙏  (@emmahodcroft please modify this as needed -- I also changed sequence_traits to export numerical values https://github.com/nextstrain/augur/commit/5a7cab1ed83f7c0fa214a36fcde006912660ca13 addressing a TODO comment in the code there).

image

Note that this only implements ordinal scales for traits whose values are all integers. A categorical scale is used if this check fails. For the future, we may want some way of including ordering in the JSON, allowing things such as https://nextstrain.org/enterovirus/d68/genome?c=age_range1 to work without providing a color scale, but this is very low priority.

Issue #776, opened 2019-08-20
Add available tangle-tree options to available datasets JSON


jameshadfield:
## Current behaviour (v1 auspice)

The "tree name" (e.g. `na` or `ha`) is injected by the nextstrain.org server into the dataset JSON which is delivered to the client, under the top-level property `_treeName` (or `treeName`?). The auspice client uses this information to work out which other datasets are suitable candidates for a second (tangle) tree, and displays them in the dropdown. Logic here: https://github.com/nextstrain/auspice/blob/8790e3d27b2513b60adb7e482256e7b5d52b0c9a/src/components/controls/choose-second-tree.js#L33. 

Note 1: this information is _not_ required to actually access the tree - i.e. you can leave this information out and type in the correct URL (e.g. `na:ha`) and it'll work fine. It's only used to prepare a list of candidates for the sidebar drop-down. 

Note 2: Since the "tree name" is not actually stored in the JSONs anywhere, the server often has to guess at the name. This is what we want to avoid.

## Desired behaviour
The candidate trees should be detailed in the JSON returned by the `getAvailable` API request. Note that auspice makes this request every time it accesses any dataset -- that is, the whenever auspice makes a `charon/getDataset` request, there is also a `charon/gerAvailable` request. 

The `getAvailable` request is handled by the nextstrain.org server here: https://github.com/nextstrain/nextstrain.org/blob/master/auspice/server/getAvailable.js#L6. The generation of the data which is returned happens earlier. You can see what's currently returned by this request via: https://nextstrain.org/charon/getAvailable or localhost:5000/charon/getAvailable (the latter is running `npm run server` from the nextstrain.org repo). 

Ultimately, as you trace things upwards, you find that the data (for the datasets part of `getAvailable`) comes from a "manifest" json which the server fetches & parses. This can be found here: http://data.nextstrain.org/manifest_guest.json. That file is fetched & parsed here: https://github.com/nextstrain/nextstrain.org/blob/master/auspice/server/setAvailableDatasets.js#L67. 
The schema of this JSON document may need modifications, but I hope not. If so, please make them in a backwards-compatible fashion. 

This request should include, for each dataset, a list of possible second trees which the client can then display in the dropdown. The code relating to fetching of the second tree shouldn't need to be changed. 

[...]

PR #792, opened 2019-09-18
Tangle tree options

joverlee521:
This PR addresses issue #776, can be tested with nextstrain.org [branch](https://github.com/nextstrain/nextstrain.org/tree/tangle-tree-options).

In the `getAvailable` request, for each dataset, a list of possible second trees is provided. This list is rendered in the second tree dropdown. 

To allow for rendering of second trees that differ on parameters other than segment, this PR also changes the format of the url:

- Previously: `http://localhost:4000/flu/seasonal/h3n2/ha:na/3y`

- Now: `http://localhost:4000/flu/seasonal/h3n2/ha/3y:flu/seasonal/h3n2/na/3y`)

### DIFF
--- a/cli/server/getDatasetHelpers.js
+++ b/cli/server/getDatasetHelpers.js
@@ -114,10 +114,42 @@ const sendJson = async (res, info) => {
   }
 }
 
+const findAvailableTangleTreeOptions = (currentDatasetUrl, availableDatasetUrls) => {
+  const currentDatasetUrlArr = currentDatasetUrl.split('/');
+
+  const availableTangleTreeOptions = availableDatasetUrls.filter((datasetUrl) => {
+    const datasetUrlArr = datasetUrl.split('/');
+    // Do not return the same dataset
+    if (currentDatasetUrl === datasetUrl) return null;
+
+    // Do not return dataset if the URLs have different number of parameters
+    if (currentDatasetUrlArr.length !== datasetUrlArr.length) return null;
+
+    // Do not return dataset if different pathogens
+    if (currentDatasetUrlArr[0] !== datasetUrlArr[0]) return null;
+
+    // Find differences between the two dataset URLs
+    const urlDifferences = currentDatasetUrlArr.filter((param, index) => {
+      if (datasetUrlArr[index] !== param) {
+        return param;
+      }
+      return null;
+    });
+
+    // Do not return dataset if more than one parameter is different
+    if (urlDifferences.length > 1) return null;
+
+    return datasetUrl;
+  });
+
+  return availableTangleTreeOptions;
+};
+
 module.exports = {
   interpretRequest,
   extendDataPathsToMatchAvailiable,
   makeFetchAddresses,
   handleError,
-  sendJson
+  sendJson,
+  findAvailableTangleTreeOptions
 };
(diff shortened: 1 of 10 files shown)


jameshadfield:
Thanks @joverlee521 👍 I'll make comments below

tsibley:
> For nextstrain.org (seperate PR) we need to make sure that […] we only have "flu/seasonal/vic/ha/6y" as an available option. This mirrors what you see currently on nextstrain.org. […] Note that one should be able to access any of these datasets by manually entering them into the URL, but they shouldn't appear in the dropdown.

@jameshadfield Can you explain the thinking here a little more? I'm not sure I understand why we want to limit to just different segments and not allow different time frames as well.

I do see that comparing subtypes isn't going to make sense because there are no shared tips, but I guess it seems like this behaviour brings us back to hard-coding what's allowed and what's not (like how segment names are hard-coded currently before this PR). It seems like it'd be nice to avoid that if we can.

trvrb:
I could imagine a more general UI where you ask for a second tree and then dial in a dataset. For flu, it will be most common to want to compare segments, but comparing resolutions is possible. I think a general common situation would also be something like `zika_2019-06-01` vs `zika_2019-09-01` so you can see new data that's been added. I don't know exactly what this UI looks like however,

cli/server/getDatasetHelpers.js
```diff
@@ -109,12 +109,15 @@ const sendJson = async (res, info) => {
...
     /* v1 JSONs don't define a tree name, so we try to guess it here */
     const mainTreeName = info.parts.join('/');
-    const v2Json = convertFromV1({tree, meta, treeName: mainTreeName});
+    const v2Json = convertFromV1({tree, meta});
     return res.json(v2Json)
   }
 }
 
 const findAvailableTangleTreeOptions = (currentDatasetUrl, availableDatasetUrls) => {
+  const expectedDifferences = new Set(["HA", "NA", "PB1", "PB2", "PA", "NP", "NS", "MP", "L", "S",
```
jameshadfield: Hey @joverlee521 -- we don't want this for the auspice server (see https://github.com/nextstrain/auspice/pull/792#issuecomment-532606922). Having all the options displayed for the auspice server is fine!

For the nextstrain.org server you can use the manifest JSON (see comment in original issue) to work out which options to display.

jameshadfield:
> I'm not sure I understand why we want to limit to just different segments and not allow different time frames as well.

We could relax this to include the time frame also - but there's often very few tips shared here due to the way we independently subsample. 

> I do see that comparing subtypes isn't going to make sense because there are no shared tips, but I guess it seems like this behaviour brings us back to hard-coding what's allowed and what's not 

What solution are you proposing? Issue #776 was a request to remove the injected JSON key without changing behavior.

tsibley:
@jameshadfield

> some kind of hardcoding is always going to be present, else you have a tangle-tree-dropdown with every other dataset

I think there are ways to avoid hardcoding by applying light semantic conventions to dataset paths and using more general heuristics. This seems important, because hardcoding is not workable as we consider how to support tangletrees for non-core datasets (community, hub/groups). We can table this discussion for now, though, as you're right that this PR is for a smaller scoped change.

jameshadfield:
Merged. Thanks @joverlee521.

Issue #785, opened 2019-09-02
Unassigned Branches get a Colour


emmahodcroft:
I am not 100% sure if this is `augur`, `auspice` or both. I think it is `auspice` - seems to happen for older datasets too - but I haven't tested this super-extensively.
Seems to be new behaviour from sometime last week?

Branches with unassigned attributes are getting a colour other than grey:
![image](https://user-images.githubusercontent.com/14290674/64121053-0beb2f00-cd9e-11e9-94c6-7f274ec7a7ca.png)

Also true for other traits:
![image](https://user-images.githubusercontent.com/14290674/64121084-1efdff00-cd9e-11e9-882a-e0015236abb9.png)

Is also reflected in the pie charts:
![image](https://user-images.githubusercontent.com/14290674/64121561-7355ae80-cd9f-11e9-803b-78b0fb059237.png)

I am running the latest v2 `auspice` (or 2.0), and latest `augur` `master` and `v6`. 

This occurs both when I use `augur export` in `master` and `augur export v1` in `v6`.  

It does *not* occur if the colours have been assigned via `colors.tsv`:
![image](https://user-images.githubusercontent.com/14290674/64121324-c5e29b00-cd9e-11e9-8a31-69e2e059749d.png)

Happens on the Heroku Server thingy: (*test dataset to replicate*)
(slightly different tree to what's pictured above, but basically same dataset)
https://nextstrain-dev.herokuapp.com/community/emmahodcroft/echo30/vp1/all

jameshadfield:
Closed in 162824bbed4789a58e1f45930e0dfdc12806fd43. Thanks @emmahodcroft 👍

PR #793, opened 2019-09-19
Fix Undefined Value Colouring Bug

emmahodcroft:
Should resolve bug #785 , where a) branches without a value get a colour and b) undefined values show up in the legend.

You can see the before examples on [Heroku](https://nextstrain-dev.herokuapp.com/enterovirus/d68/vp1).
I've tested this with a bunch of EV-D68 stuff but it would benefit from someone having a good play with it on something else, esp ensuring filters etc work.

Compare before and after (before always above, after always below):

For categorical without specified colours:
image
image

image
image

For continuous:
image
image

For categorical with specified colors:
image
image
(Tree looks the same)

For clades ('Unassigned' wasn't being detected properly):
image
image

------

As a bonus, managed to get of annoying pie chart lines when filtering so demes go to one colour only:

Before
image

After
image

### DIFF
--- a/src/components/map/mapHelpersLatLong.js
+++ b/src/components/map/mapHelpersLatLong.js
@@ -8,6 +8,7 @@ import { NODE_NOT_VISIBLE } from "../../util/globals";
 import { getTraitFromNode } from "../../util/treeMiscHelpers";
 import { isColorByGenotype } from "../../util/getGenotype";
 import { errorNotification } from "../../actions/notifications";
+import { countTraitsAcrossTree } from "../../util/treeCountingHelpers";
 
 /* global L */
 // L is global in scope and placed by leaflet()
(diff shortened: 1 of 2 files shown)


src/components/map/mapHelpersLatLong.js
```diff
@@ -73,38 +74,47 @@ const getVisibleNodesPerLocation = (nodes, visibility, geoResolution) => {
...
   });
+
   /* turn counts into arc angles (radians) */
   let startAngle = 0;
   arcs.forEach((a) => {
     a.startAngle = startAngle;
     startAngle += 2*Math.PI*a._count/visibleNodes.length;
     a.endAngle = startAngle;
+    if (a.startAngle === a.endAngle) {
+      // this prevents drawing a 'line' for 'empty' slices
```
jameshadfield: 🥇

src/components/map/mapHelpersLatLong.js
```diff
@@ -445,6 +457,8 @@ UPDATE DEMES & TRANSMISSIONS
 
 const updateDemeDataColAndVis = (demeData, demeIndices, nodes, visibility, geoResolution, nodeColors, pieChart, colorBy, legendValues) => {
   const demeDataCopy = demeData.slice();
+  // get all possible values from tree rather than relying on legend (doesn't include 'undefined', or "", etc)
+  const allValues = Object.keys(countTraitsAcrossTree(nodes, [colorBy], false, false)[colorBy]);
```
jameshadfield: Haven't had a change to test this / read through everything, but I _think_ you'll have to sort this to keep the nice behavior where the arc colours match the order in the legend values.

emmahodcroft: Oh! Hadn't even realised it was doing this. Ok, can do!

emmahodcroft:
This had previously broken colorby Genotype. Now should be fixed...

src/components/map/mapHelpersLatLong.js
emmahodcroft: Should be working now - thanks for pointing this out!

emmahodcroft:
Pies should now draw in order of legends, with unknowns at the end.

Found a bug in continuous colouring in `v2` JSONs; should be fixed now.

jameshadfield:
Hey @emmahodcroft -- thanks for the super detailed examples, they were very helpful. I took a good look at this after merging Richard's ordinal color scales and, partly due to the merge conflicts, ended up implementing a slightly different fix (which was super easy since you did all the hard work to identify what was wrong!). Let me talk you through the 3 commits (comments welcome):

---

The main commit was https://github.com/nextstrain/auspice/commit/162824bbed4789a58e1f45930e0dfdc12806fd43. 

Instead of things like `(val === "undefined" || val === undefined || val === false || val === "")` I've  used the `isValueValid` function. This helps keep all the related logic in one place. 

By pruning all "invalid" values from the legend we almost solved all the problems -- the only remaining thing was the arcs, as there were now gaps where a grey arc should be! I fixed this by creating an arc at the end of the list of arcs & assigning tip values which weren't present in the legend to this arc. This lets us avoid extra calls to `countTraitsAcrossTree` :) 

---

https://github.com/nextstrain/auspice/commit/fd961d20891b09e20654ba00abf9ff06b3ec092c is your fix for the arcs with no tips assigned 💯 

---

https://github.com/nextstrain/auspice/commit/50faf2d8689aa4ee55bcb9e0856034f33955c8c6
adds "unassigned" (not case sensitive) to the set of "invalid" values.

emmahodcroft:
Ah ha! I had an idea there might be a better way to do this, which is one reason I was keen for someone else to take a look! This sounds good to me, and more robust :)

Issue #787, opened 2019-09-03
Option to Filter to 'No Value'


emmahodcroft:
In `auspice` 2.0, if not all tips have a trait, and this is set to be a filter option, there's a weird filter to only show those that *don't* have that trait:
![image](https://user-images.githubusercontent.com/14290674/64190485-ccdcdc80-ce76-11e9-9fd7-781cc90b040f.png)

This does not show if every tip has a trait:
![image](https://user-images.githubusercontent.com/14290674/64190506-d8c89e80-ce76-11e9-9f68-618e8b5f7d88.png)

If you click on it, it does indeed filter to those without the trait:
![image](https://user-images.githubusercontent.com/14290674/64190696-3a890880-ce77-11e9-87bd-9a065caf62f5.png)
(note the branch colour is because of issue #785 where unassigned branches get a colour)

The above examples can be see on the Heroku thing:
https://nextstrain-dev.herokuapp.com/community/emmahodcroft/echo30/vp1/all-clin

In `auspice` 1.0, this is not the case:
![image](https://user-images.githubusercontent.com/14290674/64190566-f990f400-ce76-11e9-9d55-d0c5faad6f47.png)

This is the same dataset, but can be seen using 'real' Nextstrain.org:
https://nextstrain.org/community/emmahodcroft/echo30/vp1/all-clin?c=group

---

I don't know if we want to start giving an option to filter to those without a trait? (I can't think of a use-case here.) If so, we should make this clearer. If not, we should get rid of this behaviour. Possibly tied to #785 ?

jameshadfield:
I think this is now fixed 😄 see https://nextstrain-dev.herokuapp.com/community/emmahodcroft/echo30/vp1/all-clin

Please reopen if not.

PR #794, opened 2019-09-19
Prettify and Sort Filters

emmahodcroft:
Should address issue #787 and also fixes sorting of filters to match legend (only applies if colour values have been supplied, thus legend is in that order).

I addressed this by *allowing* filtering by unknown values - I thought, maybe that's actually useful if you'd like to filter by those you don't have a value for and colour by another value. 

But, this required making the whole thing more official....

**This is tested with `v1` and `v2` export JSONS for values where in the metadata file they were recoreded as `""` (blank) or `NA`. Other values (`null`, `undefined`, etc) in metadata - especially in `v1` untested!**
More testing & thoughts would be GOOD

First, addresses original two issues: Unknown values are now 'official' and sit at back of list, and other values which are sorted in the legend (because of provided colours) are sorted in same order:
For legend:
image
(note the grey box is fixed in my other PR today)

Old way:
image
New way:
image

If colours/order are assigned but there are values that don't have a colour, they're added to the end:
Kenya just came in, I haven't yet added it to colors.txt:
image
So it also sits at the end in filter:
image

**The below applies to `v1` JSONS**
I needed a way to process this through the URL - this is tricky because the value for these nodes is "", and this doesn't work in the URL processing (URL filters with values of "" are seen as empty and thus deleted). I also needed a better way to store the active filter keys - if you store them as "" or "unk" then you get a same-key warning.

So this is handled a little complexly - I refer you to the code for details. In short, values on the tree/nodes remain "". Value in the filters and URL is "unk" (`trait` (ex country) is also stored/specified in these places, so this is unambiguous). When comparing to the tree, values of "unk" are converted to "". In the `key` of active filters, which are *not* unambiguous across `trait`s, they are stored as "unk_[trait]" (ex: "unk_country", "unk_samptype"). 

I added some prettifying so that the summary at the top looks nice. Some screenshots:

image

image

-------

image

image

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -416,9 +416,10 @@ const checkAndCorrectErrorsInState = (state, metadata, query, tree) => {
   /* are filters valid? */
   const activeFilters = Object.keys(state.filters).filter((f) => f.length);
   const stateCounts = countTraitsAcrossTree(tree.nodes, activeFilters, false, true);
+  // Also allow filter "unk" for unknown values!
   for (const filterType of activeFilters) {
     const validValues = state.filters[filterType]
-      .filter((filterValue) => filterValue in stateCounts[filterType]);
+      .filter((filterValue) => filterValue in stateCounts[filterType] || filterValue === "unk");
     state.filters[filterType] = validValues;
     if (!validValues.length) {
       delete query[`f_${filterType}`];
--- a/src/actions/tree.js
+++ b/src/actions/tree.js
@@ -241,17 +241,20 @@ export const updateTipRadii = (
  * @param {Array of strings} values the values (see above)
  */
 export const applyFilter = (mode, trait, values) => {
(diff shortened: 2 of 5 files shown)


emmahodcroft:
`v2` JSONS handle this better because we do this better in `export v2` - traits without a value are simply not on the node, so they are all `undefined`. I left this this way for the moment as it's easier, but did change how it displays in the summary at the top to be more clear *which* undefined is being filtered to. Here's how it looks:

image
image

jameshadfield:
@emmahodcroft -- it seems that by fixing issue #785 then bug #787 was also fixed. In light of that, how do you advise we proceed here?

emmahodcroft:
Ah ha. This, I did not consider. Let me get caught up with everything and have a think!

emmahodcroft:
I am going to close this PR and open a new one with just the 2 things not fixed by other changes!

PR #798, opened 2019-09-27
Sort Filters and Prettify Some Filter Buttons

emmahodcroft:
This takes just the parts of PR #794 which were not fixed via other changes.

First, prettifies the 'clear filter' button by using `filterTitle` instead of `filterName`.
Before:
image

After:
image

Second, sorts filters by the order specified in a `color.txt` file (order of the legend), if that trait has such an ordering.
Before:
image

After:
image

Finally, prettifies the filter buttons at the top when filtering by 'undefined' values, to show *which* undefined values are being filtered to.
Before:
image

After:
image

We could change this to 'Undefined' if we wanted to strictly match what shows up in the filter lists down at the bottom.

### DIFF
--- a/src/components/framework/footer.js
+++ b/src/components/framework/footer.js
@@ -324,12 +324,21 @@ class Footer extends React.Component {
   displayFilter(styles, filterName) {
     const totalStateCount = this.props.totalStateCounts[filterName];
     const filterTitle = this.props.metadata.colorings[filterName] ? this.props.metadata.colorings[filterName].title : filterName;
+    let stateKeysOrdered;
+    // If has ordered colorings, use this to sort
+    if (this.props.metadata.colorings[filterName] && this.props.metadata.colorings[filterName].scale) {
+      const colorValues = this.props.metadata.colorings[filterName].scale.map(function(x) { return x[0];});
+      // Add any extra values without a set color to the end
+      stateKeysOrdered = _.union(colorValues, Array.from(totalStateCount.keys()));
+    } else { // Otherwise, sort by value
+      stateKeysOrdered = Array.from(totalStateCount.keys()).sort();
+    }
     return (
       <div>
         {`Filter by ${filterTitle}`}
-        {this.props.activeFilters[filterName].length ? removeFiltersButton(this.props.dispatch, [filterName], "inlineRight", `Clear ${filterName} filter`) : null}
+        {this.props.activeFilters[filterName].length ? removeFiltersButton(this.props.dispatch, [filterName], "inlineRight", `Clear ${filterTitle} filter`) : null}
         <Flex wrap="wrap" justifyContent="flex-start" alignItems="center" style={styles.citationList}>
-          {Array.from(totalStateCount.keys()).sort().map((itemName) => {
+          {stateKeysOrdered.map((itemName) => {
             const display = (
               <span>
                 {itemName}
--- a/src/components/info/info.js
+++ b/src/components/info/info.js
@@ -173,10 +173,12 @@ class Info extends React.Component {
     );
   }
   addNonAuthorFilterButton(buttons, filterName) {
+    // Get filterTitle so we can show which unknown is being filtered by
+    const filterTitle = this.props.metadata.colorings[filterName] ? this.props.metadata.colorings[filterName].title : filterName;
     this.props.filters[filterName].sort().forEach((itemName) => {
       const display = (
         <span>
-          {itemName}
+          {itemName !== "undefined" ? itemName : "Unknown "+filterTitle}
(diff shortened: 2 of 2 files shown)


emmahodcroft:
One note, which isn't introduced by this PR, but might be the best place to discuss here:

If you filter by two undefined things, you get a warning in the console:
image

A zoom-in for readability:
image

Is this ok? Should we resolve this? From my testing, it does seem like it can tell these apart - it always clears the correct thing. But perhaps that's not stable/guaranteed?

Also, it seems you cannot load URLs what contain 'undefined' as a filter, with that filter (the filters disappear). I don't know if we want to support this or not.
https://nextstrain-dev.herokuapp.com/enterovirus/d68/vp1?f_age_range1=undefined&f_clade_membership=undefined

It also seems to takes with it any other (not 'undefined') filters you may have had in the URL:
https://nextstrain-dev.herokuapp.com/enterovirus/d68/vp1?f_clade_membership=undefined&f_country=Switzerland,Japan

jameshadfield:
> prettifies the 'clear filter' button by using `filterTitle` instead of `filterName`.

Seems like a good idea! 

> sorts filters by the order specified in a color.txt file (order of the legend)

Another good idea. For the code, can we just always sort according to the legend entries? Can't think where we wouldn't want them to match up. (For continuous traits this won't currently work, but one day the filters should probably match up with the legend items anyway.)

re: adding in undefined filters -- if this is beneficial for your datasets then I think it's not too controversial to add it. There'll be a few bugs to sort out with this as you've identified.

jameshadfield:
### re: undefined filters

I see now that this is part of the current v2 branch. I'm not sure if this is a feature or a bug (I lean towards bug). Do we want this?

emmahodcroft:
> Another good idea. For the code, can we just always sort according to the legend entries? Can't think where we wouldn't want them to match up. (For continuous traits this won't currently work, but one day the filters should probably match up with the legend items anyway.)

Yes, I considered this and I think it would work. I just wasn't sure if it would break something else, or if there was another subtle reason why we shouldn't try to do this if there wasn't a user-specified order. Would be pretty easy to change in the code, I think. Should I do this?

> I see now that this is part of the current v2 branch. I'm not sure if this is a feature or a bug (I lean towards bug). Do we want this?

Yes - I don't know! At first I thought it was a bug, but then I wasn't sure? Can anyone imagine a use-case where this might be useful?
Might it be interesting to filter by (for ex.) those who don't have age data, but colour by country, or gender - to see if there's any pattern in those where some data is missing?

I've never needed this so far, but it might not be so crazy?

jameshadfield:
Notes from after today's meeting:

* let's sort all non-continous traits which are color-bys by their ordering in the legend. 

* Let's include "undefined" filters. I can sort out the react warnings seen above. Perhaps easier to keep them out of the URL at this stage.

emmahodcroft:
> Let's include "undefined" filters. I can sort out the react warnings seen above. Perhaps easier to keep them out of the URL at this stage.

So it turns out the that the URL filter bug was (seemingly) unrelated to this branch (and didn't work with any value, not just `undefined`), I just happened to spot it while working on this. At some point between my branching this from `v2` and now, it's been fixed, and now works for all values, including `undefined`.

So given that it's working, are we happy to leave it like this, or should I still remove `undefined` from the URL?

jameshadfield:
> So I've had a look, but I believe legends are only created when you switch colorby, and only exist for the selected colorby. So you can't easily sort the filters by the legends - they don't exist for most of them (or possibly any of them) when you are creating the footer.

Yes, however the legend ordering would be really easy to pull out into a function & reuse in the footer. The legend order depends on knowing `stateCount` and is implemented here: https://github.com/nextstrain/auspice/blob/v2/src/util/colorScale.js#L44. The footer has access to `stateCount` but currently orders alphabetically: https://github.com/nextstrain/auspice/blob/v2/src/components/framework/footer.js#L332. 

> So given that [unknown filters in the URL are] working, are we happy to leave it like this, or should I still remove undefined from the URL?

Happy to leave it as is. I believe e7ff73c9f431e3f73d8de0a3c9ff850a8dde586f fixed this.

emmahodcroft:
> Yes, however the legend ordering would be really easy to pull out into a function & reuse in the footer. The legend order depends on knowing stateCount and is implemented here: 

Yes, I can do this, but probably not until end of this week or beginning of next. Depends how things are looking with other aspects, but don't think this should necessarily hold up pushing `v2` out if everything else is ready! We can leave this open for now and I'll try to come back to it soon as I can, but if it ends up holding things up, I can make an issue that I'll tackle afterwards.

PR #799, opened 2019-09-29
Gap, X, N Bases Are Always Grey & Last in Legend

emmahodcroft:
This is partial implemention of PR #739 . Deciding exactly what colours to use for which nucleotides/AA will require more discussion, and should be tackled after releasing `v2`. 

However, it seems that the idea of colouring gaps, X, and N (in nucleotides) grey is widely accepted.

This PR implements just this change. Gaps (-), X, and N characters will always appear at the end of the legend, and be coloured in shades of grey. This can make it much easier to see how much data is missing. 

I work with a few datasets where this is a big issue, so this colouring/ordering would be very helpful for me.

This changes from [this](https://nextstrain.org/community/emmahodcroft/echo30/vp1/all-clin?c=gt-VP1_285):
image

To this:
image

Or [this](https://nextstrain.org/enterovirus/d68/genome?c=gt-nuc_7324):
image

To this:
image

Grey colours may go too dark too quickly - I used existing code for missing discrete values, but the range could be easily adjusted.

### DIFF
--- a/src/util/colorScale.js
+++ b/src/util/colorScale.js
@@ -123,13 +123,17 @@ export const calcColorScale = (colorBy, controls, tree, treeToo, metadata) => {
     legendValues = orderOfGenotypeAppearance(tree.nodes, controls.mutType);
     const trueValues = controls.mutType === "nuc" ? legendValues.filter((x) => x !== "X" && x !== "-" && x !== "N") :
       legendValues.filter((x) => x !== "X" && x !== "-");
-    const unkValues = controls.mutType === "nuc" ? legendValues.filter((x) => x === "X" || x === "-" || x === "N") :
-      legendValues.filter((x) => x === "X" || x === "-");
-    let domain = [undefined, ...trueValues];
-    let range = [unknownColor, ...genotypeColors.slice(0, trueValues.length)];
-    if (unkValues.length) { // we must add these to the domain + provide a value in the range
-      domain = domain.concat(unkValues);
-      range = range.concat(createListOfColors(unkValues.length, [rgb(192, 192, 192), rgb(32, 32, 32)]));
+    const domain = [undefined, ...legendValues];
+    const range = [unknownColor, ...genotypeColors.slice(0, trueValues.length)];
+    // Bases are returned by orderOfGenotypeAppearance in order, unknowns at end
+    if (legendValues.indexOf("-") !== -1) {
+      range.push(rgb(217, 217, 217));
+    }
+    if (legendValues.indexOf("N") !== -1 && controls.mutType === "nuc") {
+      range.push(rgb(153, 153, 153));
+    }
+    if (legendValues.indexOf("X") !== -1) {
+      range.push(rgb(102, 102, 102));
     }
     colorScale = scaleOrdinal()
           .domain(domain)
--- a/src/util/colorScale.js
+++ b/src/util/colorScale.js
@@ -120,10 +120,20 @@ export const calcColorScale = (colorBy, controls, tree, treeToo, metadata) => {
     console.error("calcColorScale called before tree is ready.");
     error = true;
   } else if (genotype) { /* G E N O T Y P E */
-    legendValues = orderOfGenotypeAppearance(tree.nodes);
+    legendValues = orderOfGenotypeAppearance(tree.nodes, controls.mutType);
+    const trueValues = controls.mutType === "nuc" ? legendValues.filter((x) => x !== "X" && x !== "-" && x !== "N") :
+      legendValues.filter((x) => x !== "X" && x !== "-");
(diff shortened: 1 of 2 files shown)


src/util/colorScale.js
```diff
@@ -120,10 +120,20 @@ export const calcColorScale = (colorBy, controls, tree, treeToo, metadata) => {
...
   } else if (genotype) { /* G E N O T Y P E */
-    legendValues = orderOfGenotypeAppearance(tree.nodes);
+    legendValues = orderOfGenotypeAppearance(tree.nodes, controls.mutType);
+    const trueValues = controls.mutType === "nuc" ? legendValues.filter((x) => x !== "X" && x !== "-" && x !== "N") :
+      legendValues.filter((x) => x !== "X" && x !== "-");
+    const unkValues = controls.mutType === "nuc" ? legendValues.filter((x) => x === "X" || x === "-" || x === "N") :
+      legendValues.filter((x) => x === "X" || x === "-");
+    let domain = [undefined, ...trueValues];
+    let range = [unknownColor, ...genotypeColors.slice(0, trueValues.length)];
+    if (unkValues.length) { // we must add these to the domain + provide a value in the range
```
jameshadfield: An alternate way to do this would be 
```
if (legendValues.includes("X") || legendValues.includes("-") || (controls.mutType === "nuc" && legendValues.includes("N"))
```

src/util/colorScale.js
```diff
@@ -120,10 +120,20 @@ export const calcColorScale = (colorBy, controls, tree, treeToo, metadata) => {
...
+      domain = domain.concat(unkValues);
+      range = range.concat(createListOfColors(unkValues.length, [rgb(192, 192, 192), rgb(32, 32, 32)]));
```
jameshadfield: Instead of inferring colours between 2 values why don't we define 3 shades of grey for the three potential unknown "bases"? That way the colours for `N`, `X` and `-` will be the same over different datasets

src/util/colorScale.js
```diff
@@ -120,10 +120,20 @@ export const calcColorScale = (colorBy, controls, tree, treeToo, metadata) => {
     console.error("calcColorScale called before tree is ready.");
     error = true;
   } else if (genotype) { /* G E N O T Y P E */
-    legendValues = orderOfGenotypeAppearance(tree.nodes);
+    legendValues = orderOfGenotypeAppearance(tree.nodes, controls.mutType);
+    const trueValues = controls.mutType === "nuc" ? legendValues.filter((x) => x !== "X" && x !== "-" && x !== "N") :
+      legendValues.filter((x) => x !== "X" && x !== "-");
+    const unkValues = controls.mutType === "nuc" ? legendValues.filter((x) => x === "X" || x === "-" || x === "N") :
+      legendValues.filter((x) => x === "X" || x === "-");
+    let domain = [undefined, ...trueValues];
```
jameshadfield: I can't remember why we have this `undefined` leading entry, but it seems unnecessary (at least for my datasets) -- could you try your datasets with the following & see if it still works as expected?
```
let domain = [...trueValues];
let range = [...genotypeColors.slice(0, trueValues.length)];
```

emmahodcroft:
Thanks James! Will try to address these ASAP - was caught up with flu stuff today.

trvrb:
Nice improvement here.

src/util/colorScale.js
emmahodcroft: So for my couple of test sets this doesn't seem to make a difference, but I don't know if there are situations where this is possible? It seems unlikely....
I tried to test with some of the test builds that are div- and time-tree only etc, but can't seem to open them in Auspice right now. I will leave it in until I can test but if someone else is more confident that this won't hurt anything feel free to remove.

src/util/colorScale.js
emmahodcroft: To add specific colours per gap, X or N, I ended up taking a different route. But it seems like there must be a more efficient way to do this...

emmahodcroft:
I had trouble finding greys that are easily distinguishable but don't have a big jump in darkness. Those that look pretty different when in boxes directly next to each other in Inkscape end up looking identical when small circles far apart in Nextstrain....

So, I ended up still having one that's pretty dark. I unfortunately couldn't find a site where I could compare all three against each other (I generally don't have any sequence data that contains Xs). And unfortunately wasn't able to quickly find one with both N and X for a direct comparison. Here's an example of the colours now: 

image

image

Welcome to take improvements on that if someone has some better greys! I'm certainly no expert in colours...

emmahodcroft:
If nobody has better suggestions for the grey colours I will merge this in the next couple of days to try and get `v2` out the door.

jameshadfield:
Tested and can find no regressions. Merged into v2 before release. Thanks @emmahodcroft ⭐️

Issue #521, opened 2018-03-02
Add map zoom / position bounds to URL state


jameshadfield:
or encode map center and zoom in the URL

jameshadfield:
With the addition of zoom-to-filtered-demes this functionality is no longer high priority. We may reopen this in the future.

PR #802, opened 2019-10-02
Geo zoom

emmahodcroft:
I am excited to propose this new functionality!! 😄 

Currently the zoom that's chosen for the map when you load up a dataset is the zoom level that will encompass (as much as possible) ALL the `geo_resolution` information you've passed in. At every level. 

So, if you have country-level `geo` info, even if you're loading up with the default `geo` set to something much smaller (a set of cities in one country), or specifying `geo` as such in the URL, you'll still have a zoom level that matches ALL available country info.

So below, I have world-wide country info, but even when I specify that I want `geo` to be for Quarters (city-areas) in Basel, my zoom is like this:
image

(This is the default loading, with `geo` to country - so you can see why the map zooms to here:
image
)

Then, you have to click like a zillion times to get close into Basel to actually see what you want. And, if you want to send someone this view, or set it as default view, they will also have to click a zillion times.

I found that we can make this work quite beautifully with a simple change so that on loading, you only take into account the lat-long values for the *currently set `geo` level*.

Now it loads like this! (I've also set color-by to `Quarter` so it looks prettier...):
image

The default load still looks the same - since default `geo` is country here, without any URL specifications, it'll load up zoomed out to country, as above. 

-----

But of course this introduces perhaps a problem - if you load up zoomed into `geo`=Quarter and then switch to `geo`=country, you can't see anything, and now you have to click a zillion times to zoom out (or reload the page):
image

To counter this I introduced that little button on the bottom left... `Reset Zoom` (maybe should be instead `Auto Zoom`?).

Click this and it will re-zoom the map to whatever you currently have selected for `geo`. Here's a nice gif!
resetZoom

----

So here's the motivation behind this... zooming in Narratives! It's not perfect, because you've got to trigger a map redraw to zoom (if buttons are disabled, which they are in narrative view). There's probably a nice solution to this, but in the meantime, removing and adding the map is a demo-worthy work-around.

Here I start with `geo`=country. I get rid of the map to force re-zoom, then bring it back with `geo`=quarter` and then `geo`=block:
(I recorded just half of the narrative screen, not showing the narrative text on the left.)
resetZoom-narrative

----

The people for whom we are working on this would like to share a narrative like this with collaborators etc, so it would be great if we can see (if all agree) if this could be implemented :)

### DIFF
--- a/src/components/map/map.js
+++ b/src/components/map/map.js
@@ -75,6 +75,7 @@ class Map extends React.Component {
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#es6-classes
     this.playPauseButtonClicked = this.playPauseButtonClicked.bind(this);
     this.resetButtonClicked = this.resetButtonClicked.bind(this);
+    this.resetZoomButtonClicked = this.resetZoomButtonClicked.bind(this);
   }
 
   componentWillMount() {
@@ -505,12 +506,40 @@ class Map extends React.Component {
     return (<div/>);
   }
 
+  resetZoomButton() {
+    if (this.props.narrativeMode) return null;
+    const buttonBaseStyle = {
+      color: "#FFFFFF",
+      fontWeight: 400,
+      fontSize: 12,
+      borderRadius: 3,
+      padding: 12,
+      border: "none",
+      zIndex: 900,
+      position: "absolute",
+      textTransform: "uppercase"
+    };
+
+    return (
+      <div>
+        <button
+          style={{...buttonBaseStyle, bottom: 27, right: 50, width: 60, backgroundColor: goColor}}
+          onClick={this.resetZoomButtonClicked}
+        >
+          Reset Zoom
+        </button>
+      </div>
+    );
+
+  }
+
   maybeCreateMapDiv() {
     let container = null;
     if (this.state.responsive) {
       container = (
         <div style={{position: "relative"}}>
           {this.animationButtons()}
+          {this.resetZoomButton()}
           <div id="map"
             style={{
               height: this.state.responsive.height,
@@ -532,6 +561,12 @@ class Map extends React.Component {
   resetButtonClicked() {
     this.props.dispatch({type: MAP_ANIMATION_PLAY_PAUSE_BUTTON, data: "Play"});
     this.props.dispatch(changeDateFilter({newMin: this.props.absoluteDateMin, newMax: this.props.absoluteDateMax, quickdraw: false}));
+  }
+  resetZoomButtonClicked() {
+    const SWNE = this.getGeoRange();
+    // L. available because leaflet() was called in componentWillMount
+    this.state.map.fitBounds(L.latLngBounds(SWNE[0], SWNE[1]));
+    this.maybeDrawDemesAndTransmissions();
   }
   render() {
     const transmissionsExist = this.state.transmissionData && this.state.transmissionData.length;
--- a/src/components/map/map.js
+++ b/src/components/map/map.js
@@ -300,11 +300,13 @@ class Map extends React.Component {
     const longitudes = [];
 
     this.props.metadata.geoResolutions.forEach((geoData) => {
-      const demeToLatLongs = geoData.demes;
-      Object.keys(demeToLatLongs).forEach((deme) => {
-        latitudes.push(demeToLatLongs[deme].latitude);
-        longitudes.push(demeToLatLongs[deme].longitude);
-      });
+      if (geoData.key === this.props.geoResolution) {
+        const demeToLatLongs = geoData.demes;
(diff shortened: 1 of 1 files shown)


emmahodcroft:
For fancy show-off demo purposes this afternoon I made the narrative zoom super-slick by just re-drawing the map every time you change geo resolution, but this isn't a good actual fix, since I don't think we want this to happen automatically. Will consider how we could trigger this in narratives with a little URL tag that just makes the map re-draw.

trvrb:
I'm 100% behind this proposal. I've been annoyed with the current default zoom in other situations (eg including country as a DRC Ebola geo was breaking default zoom).

I suspect there's a more natural UI element to indicate "reset", but this is an easy fix. Should already exist in Leaflet.

tsibley:
Instead of (or in addition to) a button to reset the zoom, could we automatically trigger just a zoom recalc when the map/geo option changes? Automatically re-zooming to fit the data extent is pretty standard behaviour on web maps.

emmahodcroft:
So I considered this but I wasn't sure if we want it. You could want to
switch geo res and not necessarily want to change the zoom. I have this
enabled at the moment for demo but I think it's probably not the solution
we want, as you can't (for example) be zoomed in to a region and switch to
country without being thrown fully out...

trvrb:
That was my initial thought as well.

One other thing: for narratives probably the more expressive solution would be to encode map center and zoom in the URL.

jameshadfield:
> One other thing: for narratives probably the more expressive solution would be to encode map center and zoom in the URL.

Yes, I think this is the best solution for narratives (and URL sharing in general). This is #521.

But I do think there's a place for automatic zooming when interacting with auspice. auspice v1 would reset the map zoom whenever the geoResolution changed, but this was really annoying & so removed from auspice v2. @tsibley's suggestion to auto-zoom "if no data remains in the map extent" seems sensible.

emmahodcroft:
I think it's a question of how much control we want to take from the user. @tsibley's idea is a good one, but will also control behaviour, maybe also in ways that are annoying. If I switch from 'quarter' or 'block' to 'country', maybe I only want to zoom out to look at a few surrounding countries, but an auto-zoom would take me back to the whole world. 

So, that's why I went with a button - it allows a beautiful auto-zoom, but not ever unexpectedly. Also, it allows for nice fine-scale adjustments like between block and quarter, where the difference isn't much (wouldn't qualify for auto-zoom, still plenty in frame) but the map adjusts super nicely. 

I did a bunch of smart-zoom work for the entropy view, which I was very proud of at the time, and now I find it driving me mad about 50% of the time - that's part of the reason I'm leaning more towards letting users decide when they want these things to happen rather than trying to guess... 

> Currently, I believe we only ever have 1 "default" map view, which is calculated only on dataset load for whatever geoResolution is set at that time.

From what I was working on today and yesterday, I think the default view is whatever encapsulates *all* available geo locations, not just whatever geo is set. That was the first change I made in this PR, and definitely makes a big difference!

jameshadfield:
Ok. Understood. Can we make the button aesthetically the same as the "RESET LAYOUT" button at the top right of the tree. 

I think #521 should be high priority after v2 is released.

emmahodcroft:
I made the button to match the 'reset layout' above the tree - this was a great idea James, I think it looks much better!
image

I've retained the skeleton of the functionality to have this button be 'greyed out' and 'greyed in' like the 'reset layout' button, but we'd need to decide on some way of measuring zoom level and position to decide when to grey-in and grey-out.
This is kind of hard to decide, as it's an easy way to recenter from zooming out, being slightly sideways, or whatever - it's not super clear how 'different' the map should be before we decide it's worth being able to use the 'reset zoom' button....

Worth noting that the 'reset layout' button above the phylogeny does change colours and the mouse pointer changes, but it's never actually "disabled" - you can click on it and trigger a tree re-draw even when it's greyed-out.  (This maybe means it's not a big deal if we don't have this (even without the greying) on the map side, either.)

trvrb:
This looks great Emma.

emmahodcroft:
If nobody has objections to this functionality so far (with plans to improve this in narratives after `v2` release), I'd like to merge this, as a collaborator would like to be able to use this to share (it'll have to be via the Heroku for now, but that's ok).

emmahodcroft:
I'd really love to get a basic version of this working for narratives, where something in the URL could just trigger a map re-draw. (Currently the only way to do this is to remove the map and then put it back - making for awkward narratives.) 

I've given this two goes but for the life of me can't figure it out. @jameshadfield, would you have any tips or ideas for how one might trigger a map re-draw using existing stuff? I don't know if it's possible it could be as easy as knowing some trick for being able to call 'resetZoomButtonClicked()` (`src/components/map/map.js`) from `recomputeReduxState.js` (I can't seem to make that a thing), or if some other existing action could be triggered...?

jameshadfield:
Will spend an hour on this today & report back findings...

PR #804, opened 2019-10-03
improve axis labels to include months or epi-weeks

jameshadfield:
This changes our default numerical view to show months for timespans of under 4 years and epi weeks for timespans of under 1 year. (These thresholds may need tweaking.) 

Currently online at nextstrain-dev.herokuapp.com

Future improvements:
* ensure that the major grid lines line up with the start of an epi week (and the minor gridlines could be days)
* recompute grid lines for views where the timespan is under ~10 days to show dates (and have no minor grid lines)

image

image

cc @rneher @trvrb @alliblk

### DIFF
--- a/src/components/tree/phyloTree/grid.js
+++ b/src/components/tree/phyloTree/grid.js
@@ -40,6 +40,61 @@ const calculateMajorGridSeperation = (range) => {
   return step;
 };
 
+const computeXGridPoints = (xmin, xmax, step, layout, distanceMeasure, minorTicksTimeTree, minorTicks) => {
+  const majorGridPoints = [];
+  const minorGridPoints = [];
+  const gridMin = Math.floor(xmin/step)*step;
+  const minVis = layout==="radial" ? xmin : gridMin;
+  const maxVis = xmax;
+  const precisionX = Math.max(0, -Math.floor(Math.log10(step)));
+  for (let ii = 0; ii <= (xmax - gridMin)/step+3; ii++) {
+    const pos = gridMin + step*ii;
+    majorGridPoints.push({
+      position: pos,
+      name: String(pos.toFixed(precisionX)),
+      visibility: ((pos<minVis) || (pos>maxVis)) ? "hidden" : "visible",
+      axis: "x"
+    });
+  }
+  let numMinorTicks = distanceMeasure === "num_date" ? minorTicksTimeTree : minorTicks;
+  if (step===5 || step===10) {
+    numMinorTicks = 5;
+  }
+  const minorStep = step / numMinorTicks;
+  for (let ii = 0; ii <= (xmax - gridMin)/minorStep+30; ii++) {
+    const pos = gridMin + minorStep*ii;
+    minorGridPoints.push({
+      position: pos,
+      name: String(pos.toFixed(precisionX)),
+      visibility: ((pos<minVis) || (pos>maxVis+minorStep)) ? "hidden" : "visible",
+      axis: "x"
+    });
+  }
+  console.log("MAJOR:", majorGridPoints);
+  console.log("MINOR:", minorGridPoints);
+  return {majorGridPoints, minorGridPoints};
+};
+
+const computeYGridPoints = (ymin, ymax) => {
+  const majorGridPoints = [];
+  let yStep = 0;
+  yStep = calculateMajorGridSeperation(ymax-ymin);
+  const precisionY = Math.max(0, -Math.floor(Math.log10(yStep)));
+  const gridYMin = Math.floor(ymin/yStep)*yStep;
+  const maxYVis = ymax;
+  const minYVis = gridYMin;
+  for (let ii = 1; ii <= (ymax - gridYMin)/yStep+10; ii++) {
+    const pos = gridYMin + yStep*ii;
+    majorGridPoints.push({
+      position: pos,
+      name: pos.toFixed(precisionY),
+      visibility: ((pos<minYVis)||(pos>maxYVis)) ? "hidden" : "visible",
+      axis: "y"
+    });
+  }
+  return {majorGridPoints};
+};
+
 /**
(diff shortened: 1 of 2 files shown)


rneher:
I like this. But I'd be hesitant to include the dependence `epi-week` which is not maintained and hasn't been changed in 7y. Simple code-snippets like this one would do the trick:
https://www.epoch-calendar.com/support/getting_iso_week.html

tsibley:
My understanding is that CDC and WHO **do not** use the same week definitions. The CDC defines their own "MMWR week" (which the `epi-week` package is using) while the WHO uses ISO week definitions. A good summary of this is in the documentation of the Python `epiweeks` package: https://epiweeks.readthedocs.io/#overview.

Auspice should be very clear about which week system is displayed.

@rneher The dense code snippet on https://www.epoch-calendar.com/support/getting_iso_week.html to get the ISO week seems worse than using a proper datetime library that's maintained. For example, the [Luxon](https://moment.github.io/luxon/) library can be used to get ISO weeks:

```js
const { DateTime } = require("luxon");
const now = DateTime.local();
console.log(now.toFormat("kkkk-'W'WW")); // 2019-W40
```

rneher:
yes, that code snippet is not ideal. but it does show that its not all that hard a problem. 

My vote is for ISO week (which is used for epi purposes in much of the world) via a well-supported library.

tsibley:
Yeah, I'd also vote for ISO week or a toggle between CDC and ISO week.

> but it does show that its not all that hard a problem.

True! :-)

trvrb:
@jameshadfield: Thanks for the PR! This is a fantastic direction.

Regarding epi weeks: I wonder if this won't just add confusion overall. Everyone knows calendar dates, only a subset of people know (and like) epi weeks. Roughly this would be my thinking:

https://bl.ocks.org/mbostock/34f08d5e11952a80609169b7917d4172

Notice that under 1 month, you get Sep 03, Sep 10, Sep 17, etc...

Also, the double row labeling seems helpful to not have to repeat "2019" over and over, like so:

Screen Shot 2019-10-03 at 11 30 53 AM

Though this certainly fits more in "nice to have".

jameshadfield:
No problem -- I won't have time to work on this for the next week or so, so if someone would like to take it on then that would be very appreciated.

tsibley:
@trvrb It's my understanding that public health folks really like to use epi weeks. Would you say that's accurate?

trvrb:
Seems like yes and no. A small sampling:

* [WHO DRC Ebola Sit Rep](https://apps.who.int/iris/bitstream/handle/10665/326015/SITREP_EVD_DRC_20190721-eng.pdf?ua=1) uses calendar dates
* [WHO West Africa Ebola Sit Rep](http://apps.who.int/ebola/sites/default/files/atoms/files//who_ebola_situation_report_30-12-2015.pdf?ua=1&ua=1) also uses calendar dates
* [WHO MERS Sit Rep](http://applications.emro.who.int/docs/EMROPub_MERS_August_2019_EN.pdf) uses epi weeks
* [WHO Zika Sit Rep](https://apps.who.int/iris/bitstream/handle/10665/246222/zikasitrep14Jul16-eng.pdf?sequence=1) only reports monthly summaries

I'd conclude from this that there is not consensus and calendar dates are common.

tsibley:
Thanks for that info!

alliblk:
Public health folks really do like Epi weeks. That is a lot of the surveillance data I've seen at the regional level is reported. E.G USVI Zika surveillance was reported in epi weeks, WA pertussis uses both dates and epi weeks etc.

While I absolutely agree that having epi weeks adds a potential layer of confusion, having them I think is an important step for moulding nextstrain into something that is more familiar to field epidemiologists, if that is something that is important to us (obviously my two bits is that it is).

rneher:
could we just label as `2019-cw43` to make it clear that things are calendar weeks? and once we have a toggle, we can switch between `2019-cw43` and `2019-ew44`?

trvrb:
>could we just label as 2019-cw43 to make it clear that things are calendar weeks? and once we have a toggle, we can switch between 2019-cw43 and 2019-ew44?
I wouldn't use `2019-cw43`. I'd instead use "29 Oct" for calendar dates.

tsibley:
@rneher The problem with epi weeks is that there isn't a single agreed upon definition of calendar weeks. (I'd also prefer not to make up date notations like `2019-cw43`. `2019-W40` is an ISO standard format, for example.)

:+1: for showing day + month name + (year when January) for now, as Trevor described.

rneher:
It is pretty common to refer to particular calendar weeks in this part of the world -- most calendars have ISO week numbers.

trvrb:
Hmm... my thinking would be:

1. We now know that there is considerable ambiguity over what a "week" is. We have two separate week definitions, "ISO week" used by much of the world and by the WHO for their epi weeks, and we also have "MMWR week" used by CDC and state agencies in the US for their epi weeks (leave it to the Americans...)
2. I understand that epi folks often like using epi weeks, but this practice seems far from standard. It will also only be epi folks who really think in weeks rather than calendar dates (I believe, but could be wrong).
3. A "good enough" solution would then be calendar dates that then (A) don't suffer from the above ambiguity and (B) are readable by everyone.

Does this seem like reasonable logic?

Later on, we can implement a toggle to swap between "calendar date", "ISO week" and "MMRW week", but this to me seems secondary.

tsibley:
I'm on board with that plan.

alliblk:
Agreed with plan to just do calendar dates for now, with later development of epi week stuff as needed.

jameshadfield:
Thanks for the comments 💯 

When the grid-line spacing is between a month & a year we display "YYYY-MON", when the grid-line spacing is under a month we show "YYYY-MON-DD". 

The major grid spacing has been modified to better line up with months where appropriate.

PR #807, opened 2019-10-14
Remove twitter

jameshadfield:
This PR removes 
* The loading of the twitter widget -- This was able to render tweets from narratives "nicely" however it is prudent to remove it for security reasons. An issue will be created regarding the narrative functionality. (To my knowledge no narratives relied on this.)
* Twitter-related CSS


tsibley:
The "Intro to narratives" narrative calls out Twittering embeding: https://nextstrain.org/narratives/intro-to-narratives?n=5

(That also makes me notice that `?n=5` is preserved but doesn't load with the 5th section shown.)

PR #819, opened 2019-11-05
Include strain metadata accession

joverlee521:
Includes the GISAID EPI Isolate ID in the strain metadata download.


jameshadfield:
Hi @joverlee521, thanks for addressing this bug introduced in the v2 release. I think it's great that there is good feedback from the flu community and that you can address bugs like these 👍 

I've fixed a couple of additional bugs related to your one. A strain's `url` wasn't being converted to the v2 format either (identical issue to `accession`). And neither `accession` nor `url` were being displayed in the tip-clicked info box, which is a regression from v1.

PR #821, opened 2019-11-12
Display repo and maintainers as byline

trvrb:
This PR aims to do two things:

1. Surface repository / build URL. It's nice to point people to GitHub readmes that have details behind how the analysis was generated.
2. Better highlight maintainers. In the continued effort to give everyone credit, probably better to move maintainers to the top of the page rather than the very bottom.

This looks like this for the Zika Tutorial build:

zika

This is what this looks like for the Ebola DRC build:

ebola

(There are some more mocks in Slack here: https://bedfordlab.slack.com/archives/C116R9PR7/p1564668779044000)

This would require groups to use custom GitHub profile photos if they want their logo shared.

This uses `metadata.repository`, but could easily be changed to `metadata.buildUrl` or something else (https://github.com/nextstrain/augur/pull/408#issuecomment-552731571).

I'd imagine this working with existing community functionality in that if `metadata.repository` is not specified and charon `getAvailable` has shared `buildUrl` with auspice, then `metadata.repository` would use the URL passed in from community. `metadata.repository` would override passed in community `buildUrl` if both exist.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -56,6 +56,7 @@ rules:
   prefer-destructuring: off
   implicit-arrow-linebreak: off
   jsx-quotes: off
+  no-use-before-define: ["error", { "functions": false }]
 parserOptions:
   ecmaVersion: 6
   sourceType: module
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -524,8 +524,8 @@ const createMetadataStateFromJSON = (json) => {
   if (json.meta.maintainers) {
     metadata.maintainers = json.meta.maintainers;
   }
-  if (json.meta.repository) {
-    metadata.repository = json.meta.repository;
+  if (json.meta.build_url) {
+    metadata.buildUrl = json.meta.build_url;
   }
   if (json.meta.genome_annotations) {
     metadata.genomeAnnotations = json.meta.genome_annotations;
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -524,6 +524,9 @@ const createMetadataStateFromJSON = (json) => {
   if (json.meta.maintainers) {
     metadata.maintainers = json.meta.maintainers;
   }
+  if (json.meta.repository) {
+    metadata.repository = json.meta.repository;
+  }
   if (json.meta.genome_annotations) {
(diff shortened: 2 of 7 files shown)


tsibley:
The UI here looks great :sparkles:, and I :heart: that we're better highlighting build maintainers with a byline.

jameshadfield:
I'm 100% behind this. For completeness, see also this PR from July: https://github.com/nextstrain/auspice/pull/761

image

trvrb:
Thanks Tom and thanks James! I've just requested a review from you @jameshadfield. I suspect some of my JS use here could be improved.

tsibley:
>     2\. I'll take a look at this as we'll have to cover cases where a "build url" is provided both from the dataset JSON & the `getAvailable` API call. In this case, the former should take precedence.

I think it'd be nice to remove `buildUrl` from `/charon/getAvailable` and converge the Auspice client on reading it only from the dataset JSON. The server-side can then just opportunistically fill in a missing `buildUrl` for data it fetches from community repos.

jameshadfield:
@trvrb i've made a few changes & think this is pretty good to go once we decide on the key name for the dataset JSON (currently `build_url`). Please see the commit titles/messages for what was done. Happy to explain any of them in more detail. 

I've changed the font size & weight as it looked much heavier on my setup than yours, but this needs testing (feel free to get rid of this commit if you want).

Screen Shot 2019-11-20 at 7 36 44 PM
Screen Shot 2019-11-20 at 7 37 09 PM

> I think it'd be nice to remove buildUrl from /charon/getAvailable and converge the Auspice client on reading it only from the dataset JSON. 

This is specified in the v2 auspice API here: https://nextstrain.github.io/auspice/server/api#charon-getavailable. It won't be used if a `build_url` is supplied in the dataset JSON. Can be removed in auspice v3 if people wish.

trvrb:
Thanks James! Changes look great. I agree byline works better as a separate React component. The only suggestion I had was to split the difference with aesthetic changes. I just pushed a small change there. I'm happy merging, but I do think we should wait on decision on `build_url` name.

jameshadfield:
> I do think we should wait on decision on build_url name

Agreed. No huge hurry here as there are no JSONs which have it yet! 😉

> I just pushed a small change there.

Great! We already request weight 500 fonts so no extra overhead here.

PR #822, opened 2019-11-13
Add logos to flu footer

jameshadfield:
@trvrb feel free to change styling if desired (I tried above the text & in-between paragraphs, but this felt the best). We could also add the logos for the 4 titer-contributing centers if this is desired?

image


trvrb:
Awesome! Thanks James. Looks great. I just slightly adjusted horizontal spacing.

PR #825, opened 2019-11-25
read (& send) JSONs as streams where possible

jameshadfield:
Background: The auspice server (and the nextstrain.org server, but that's not part of this PR) read JSON data into memory during `getDataset` request handling and then send the data as a response to the client. For all non-v1 JSONs these data are passed through unmodified by the server.

This PR shifts to reading in data as a stream and piping this data to the response (which is also a stream). This results in some nice memory savings (green vs orange lines):

image
_H3N2 datasets, all v2 JSONs, no tip-frequencies_


trvrb:
Awesome!

Issue #714, opened 2019-04-17
Improve offline functionality


jameshadfield:
Currently auspice can run “locally” such that datasets are sourced from the user’s computer. This is fantastic, but requires an internet connection to load a number of cross-origin resources (see below). This includes mapbox tiles, meaning the map isn’t usable when offline. This functionality would be particularly desirable for conference talks etc where network connections are often unavailable, as well as developing / running auspice while offline.

Here are the cross-origin requests made by auspice. These should be either made available offline — where allowed — or auspice should gracefully handle their non availability. (It may be that, when an internet connection is available, some are preferable to load from other domains for performance reasons, or as specific requests that can be cached.)

### Fonts / styles
* `leaflet.css` (unpkg.com) — sourced from `index.html`. Available in a HTTP cache for 24 hours.
* `font-awseome.min.css` (maxcdn.bootstrapcdn.com)  — sourced from `index.html` & HTTP cached for up to 1 year. 
* Lato fonts description — not the fonts themselves, fonts.googleapis.com, sourced from `index.html` HTTP cached for up to 1 year.

The number of font files — typically `font/woff2` files — requested varies according to which ones (and weights) are used in the current page. All seem to have cache limits of 1 year. They’re requested from `font-awseome.min.css`, the lato fonts description, and a bundled leaflet script. Number of requests varies from 2 (auspice splash page) to 15 (zika). 

### Scripts
* Google analytics (google-analytics.com) — cached for 2 hours, but somewhat pointless when offline! (Ditto for the requests the script will make.) This doesn’t affect performance / functionality in any way.
* A twitter script (platform.twitter.com) is fetched (which fetches another few resources), but this functionality is only used for the narrative. It shouldn’t be needed when offline, and ideally should only fetched when a narrative with tweets is loaded. It’s reasonably common to block these scripts, so we should ensure the narrative view looks ok without this. 

### Mapbox tiles

Mapbox requests a number of `png` map tiles (16 for zika before zooming/panning), all cached for up to 12 hours. Storing the files in the git repo isn’t allowed, but [this thread](https://github.com/mapbox/mapbox-gl-js/issues/4326#issue-210039601) provides a solution using service workers and one using `CacheStorage` (which is a service worker API but is apparently also available in window scope).

### Service workers: the solution?
At a glance, service workers are my preferred solution as we could use them for other things as well (but `CacheStorage` might be simpler).

[...]

PR #826, opened 2019-11-25
Replace remote styles and fonts with the ones served locally

ivan-aksamentov:
This PR attempts to address some of the points in issue #714 (Improve offline functionality):

 - `leaflet`'s stylesheets are now bundled with webpack.
 - `font-awesome`'s stylesheets and fonts are are now bundled with webpack. Official `font-awesome` package is added as a dependency.
 - "Lato" font is bundled and served statically, so there no need to hit Google Font API anymore. Dependency to [`typeface-lato`](https://github.com/KyleAMathews/typefaces) is added. This package includes `woff` and `woff2` formats and this should cover the majority of target browsers, including IE 9 and Safari 5 (see: [caniuse.com/woff](https://caniuse.com/woff))

### Verification

 - Check out this pull request:
    ```bash
    git clone https://github.com/nextstrain/auspice
    cd auspice
    git pull origin pull/826/head
  
    ```
 - Add data and narratives, install packages, launch dev server as usual.
 - Open app in the chrome. In dev tools Network tab, type into filter input box  `-domain:localhost` (meaning "not",  "domain", "localhost").
 - Make sure that no requests are shown other than to `api.mapbox.com` and, eventually, `chrome-extension://*/backend.js` from installed Chrome extensions.

### DIFF
--- a/index.html
+++ b/index.html
@@ -23,7 +23,6 @@
     -->
     <link rel="stylesheet" type="text/css" href="https://fonts.googleapis.com/css?family=Lato:100,200,300,400,500,700">
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css">
-    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.0.1/dist/leaflet.css" />
 
   </body>
 </html>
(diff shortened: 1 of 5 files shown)


trvrb:
Thanks so much for the PR @ivan-aksamentov! 🙌This looks like a nice improvement. Brief perusal looks great to me, but I'll defer to @jameshadfield for full review.

PR #827, opened 2019-11-26
[WIP] Cache map tiles locally server-side

ivan-aksamentov:
This PR attempts to implement a file-based local server-side cache for map tiles, addressing some of the concerns of #714 

### Motivation

Together with #826 this PR allows to run local server without internet connection. When reliable connection is available, user may selectively visit regions of interest on the map in order to build up the cache, and then go offline, with these same regions still being available.

### Implementation details

In order to not violate terms of currently used MapBox service, this PR switches to OpenStreetMap provider, which provides free tiles and allows (even encourages) caching.

Instead of fetching tiles directly, leaflet layer now requests tiles from `/tiles/{s}/{x}/{y}/{z}` (when running server locally, this resolves to localhost). On server side, this request is routed to `TileCache` service that initially proxies requests to the remote tile provider of choice, currently OpenStreetMap. 

The image tiles resulting from this request are sent back to client, but are also asynchronously written to filesystem. If previously seen tile is requested again, `TileCache` will return corresponding cached file instead of making requests to remote tile provider.
When remote tiles cannot be successfully retrieved, `TileCache` currently returns status 404. Leaflet in this case renders a grey tile.

### Discussion and further improvements

 - OpenStreetMap [Tile Usage Policy](https://operations.osmfoundation.org/policies/tiles/#technical-usage-requirements) allows caching and proxying, but they forbid setting HTTP user agent.  However, requesting tiles with `node-fetch` always returns status 429, therefore, in this PR we forward `User-Agent` from the original client (browser) request. We have to check if this behavior is allowed.

 - Although tile images are small (10-20kb), they come in bulk numbers and overall cache size may reach limits of available disk space. We may implement various cache eviction policies (e.g. LRU) to free up disk space on user's machine.

 - Similarly, caching can be implemented client-side. The tiles can be stored in LocalStorage or in IndexedDB. This is more tricky, as it will involve modifying leaflet's layer, but is doable.

### DIFF
--- a/cli/server/TileCache.js
+++ b/cli/server/TileCache.js
@@ -1,0 +1,133 @@
+const _ = require('lodash');
+const { default: fetch, Headers } = require('node-fetch');
+const fs = require('fs-extra');
+const path = require('path');
+
+const utils = require('../utils');
+
+const TILES_CACHE_DIR = 'cache/tiles';
+
+
+function areValidCoordinates(coord) {
+  return _.every(coord, (c) => _.isInteger(_.toNumber(c)));
+}
+
+/**
+ * Caches map tiles requested by the client, so that previously seen tiles are
+ * available from localhost.
+ */
+class TileCache {
+  constructor() {
+    // TODO: setup transpiling server code to avoid .bind(this) and for other syntactic perks
+    this.requestHandler = this.requestHandler.bind(this);
+    this.getTile = this.getTile.bind(this);
+  }
+
+  /**
+   * Handles network requests
+   */
+  async requestHandler(req, res) {
+    if (!this.validateParams(req, res)) {
+      return res.status(400)
+        .type('json')
+        .send({ error: 'Bad request' });
+    }
+
+    // Send tile
+    const { s, x, y, z } = req.params;
+    try {
+      const image = await this.getTile({ s, x, y, z, headers: req.headers });
+
+      if (!image) {
+        return res.sendStatus(404);
+      }
+
+      // TODO: better performance we may implement streaming response
+      return res.type('png')
+        .send(image);
+
+
+    } catch (error) {
+      const details = error.message || 'unknown';
+      utils.error(`TileCache.requestHandler: ${details}`);
+      return res.status(500)
+        .type('json')
+        .send({ error: `server error: ${details}` });
+    }
+  }
+
+  validateParams(req) {
+    const { s, x, y, z } = req.params;
+
+    // Check `s` param
+    if (!_.isString(s) || s.length === 0) {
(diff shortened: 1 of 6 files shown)


cli/server/TileCache.js
```diff
@@ -0,0 +1,133 @@
...
+  return _.every(coord, (c) => _.isInteger(_.toNumber(c)));
+}
+
+/**
+ * Caches map tiles requested by the client, so that previously seen tiles are
+ * available from localhost.
+ */
+class TileCache {
+  constructor() {
+    // TODO: setup transpiling server code to avoid .bind(this) and for other syntactic perks
```
jameshadfield: I'm not a fan of transpiling the server code as I value the simplicity of being able to run `node server.js`, although it does mean some code is a little bit more verbose. For background, auspice used to transpile the server code and it's been a much nicer dev experience since we stopped doing this.

jameshadfield:
Thanks @ivan-aksamentov -- this is really cool & happy to see an implementation of this 😄 

I haven't yet had a chance to really test the implementation but wanted to discuss two bigger points:

## Tile provider
This PR switches to OpenStreetMap which allows caching by the (auspice) server. This is a big aesthetic change for auspice which we'll need to discuss internally:

image
image

As mentioned in #714, MapBox tiles can be cached by a service worker in the browser, so that approach should work if we choose to keep MapBox. A nice side-effect of this PR would be the ability for implementations of auspice which use a custom server to define different map tile sources -- meaning that nextstrain.org could continue to use MapBox tiles by implementing a custom `/tiles` handler.

## New API call

*Background:*
The auspice npm package / this github repo provides both a server (e.g. `auspice view`) and a client (built using `auspice build`). The auspice client is designed to be used by other projects (such as nextstrain.org), and they may choose to run the auspice server or design their own. This design allows projects to implement their own server handlers for the auspice-client-API calls which allows a great deal of flexibility. See [these docs](https://nextstrain.github.io/auspice/server/introduction) for more info. 

This PR would require a major version bump and for all custom servers to implement a new handler for `/tiles`. (A lot of the code, such as `TileCache.js` can be exposed by auspice and therefore imported by such servers.) Note that we just released v2 a few weeks ago!

P.S. The actual API address should use `getServerAddress()` (https://github.com/nextstrain/auspice/blob/master/src/util/globals.js#L143), which by default is`/charon`. This allows it to be modified by a build-time customisation to auspice ([docs here](https://nextstrain.github.io/auspice/customise-client/api#specifying-the-api-server-address)).

---

I'll make sure we discuss these bigger points internally over the coming two weeks & thanks again for the help here!

Issue #689, opened 2019-01-16
library imports fail when auspice is locally (not globall) installed


jameshadfield:
Webpack should allow library imports which are part of

jameshadfield:
How do you guarantee you're getting the same version (e.g. of `react`) used inside auspice?

tsibley:
If the customized build doesn't declare a dep on `react` itself but relies on the transitive dependency through `auspice`, then it will use the same version. This is currently what's going on in nextstrain/nextstrain.org#27 where I tried out this change. The benefits are that the version of Auspice the customizations rely on is declared as a normal dep and building the customizations doesn't have to overwrite the global Auspice install every single time.

jameshadfield:
Cool 😄 

Some things for us to double check:

* what happens if there are multiple declared dependencies which themselves declare `react`. E.g. in nextstrain.org, both gatsby and auspice will install (different) copies of react.

* What happens if there's a declared dependency on react? Does auspice still import the correct version?

* Does the customisation code work if you are using a global auspice install or running auspice from outside that directory (e.g. developing auspice and running the customisations of nextstrain.org)?

tsibley:
Currently the static site's deps are self-contained since it has its own package.json. The version of React that Gatsby wants should not affect the version Auspice wants. It we assimilate the static site's deps into the top-level package.json, then [install order comes into play](https://npm.github.io/how-npm-works-docs/npm3/how-npm3-works.html). From a clean-slate install, Auspice will determine the top-level version of React¹ and any conflicting version of React required by Gatsby will be nested under Gatsby itself. The customizations will continue to see Auspice's version. Clean-slate installs are handy for reproducible builds, and can be performed using [`npm ci`](https://docs.npmjs.com/cli/ci.html) to install deps instead of `npm install`.

If we declare a top-level dep on React, then it should be compatible with Auspice's required version. If it is not, then Auspice's version will get nested underneath it and I believe (but am not certain) that the customizations will end up with a different version. We could test this case if we're worried about, but I don't think we need to be.

The customizations should still work when manually built using a global Auspice install to do `auspice build` or `auspice develop`. They did when I used `develop` a few times last week.

¹ Because by happy happenstance, Auspice is early in an alphabetical ordering.

eharkins:
[...]

PR #830, opened 2019-11-27
Fix webpack transforms for different install methods

jameshadfield:
The directory structure of npm dependencies is different
depending on whether auspice is installed from source, a
global npm install or a project's dependency. E.g. for webpack
to process the `font-awesome` dependency, then:
Global install & from source: `${__dirname}/node_modules/font-awesome`
Project dependency:              `${__dirname}/../font-awesome`

This bug comes from #826 which specified the first of the above
two scenarios.

Tested with projects using auspice (incl client-side customisations)
via global npm auspice install, auspice as a npm dependency and  auspice via `npm link <path>`

Note that this PR implements a short-term solution. In general, there are a a lot of
complex issues related to `npm` and using auspice to build custom bundles, including
but not limited to #689. I plan to write up a detailed issue for these shortly.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -14,6 +14,22 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
   /* which directories should be parsed by babel and other loaders? */
   const directoriesToTransform = [path.join(__dirname, 'src')];
 
+  /* which font modules need to be able to be loaded by auspice? */
+  const fontModules = ["font-awesome", "leaflet", "typeface-lato"];
+  /* note that the directory structure is different depending on whether auspice
+  is installed from source, a global npm install or a project's dependency. E.g.
+  Global install & from source: `${__dirname}/node_modules/font-awesome`
+  Project dependency:           `${__dirname}/../font-awesome`
+  This implementation is a short-term solution and should be improved */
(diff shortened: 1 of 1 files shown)


tsibley:
@jameshadfield I think a simpler solution to this is to remove the `include` condition for `file-loader` all together:

```diff
diff --git a/webpack.config.js b/webpack.config.js
index 8c34bdf3..50b8b2de 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -118,13 +118,7 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
         },
         {
           test: /\.(gif|png|jpe?g|svg|woff2?|eot|otf|ttf)$/i,
-          use: "file-loader",
-          include: [
-            ...directoriesToTransform,
-            path.join(__dirname, 'node_modules/font-awesome'),
-            path.join(__dirname, 'node_modules/leaflet'),
-            path.join(__dirname, 'node_modules/typeface-lato')
-          ]
+          use: "file-loader"
         }
       ]
     }
```

Not entirely sure why the `include` was there in the first place, but by removing it, I believe this lets the `file-loader` process any import which matches the `test`, not just those within the listed dirs. This seems like fine behaviour.

jameshadfield:
Thanks @tsibley!

PR #831, opened 2019-11-27
Implementing Custom Clade Labels for Zoom!

emmahodcroft:
This branch is 1 commit behind master ( it's parallel to [this commit](https://github.com/nextstrain/auspice/commit/32b2688c1ca56b17ffd3ae96c57a8fead2756653)) , as the latest merge commit (#826) breaks Auspice for me...

This PR is to create a way to have custom 'clade' labels in Auspice that can be used to zoom. It's already possible in the V2 JSON format to specify multiple `labels` for a branch under `branch_attrs`. Currently these are commonly used to specify `clade`, which gives the clade designations from `augur clade`, and `aa` which gives the AA mut labels. 

At the moment, Auspice only recognises `clade` for the purposes of changing the URL and thus only things labelled by `clade` can be specified for zooming via URL (and thus in narratives!).

This PR changes the functionality so that any "'label' 'selected'" pair can be used to identify nodes, change the URL, and thus be zoomed. 
This generally works by, instead of passing and expecting things like `clade=A2`, passing and expecting `clade={label: "clade", selected: "A2"}`
This allows the user to use whatever combination of branch labels they like. 

----

At the moment in this PR Auspice will only detect a branch as having a possible label (and thus able to go into the URL) if it **only** has _one_ branch label. And, it does not count or consider the `aa` labels (they look really ugly in the URL and I also think this isn't their purpose, so I disregard them). 

However, you can always manually change the URL to zoom to any branch, with any number of labels - they'll all end up in the same place. (There's no restriction on this end.)

This _could_ be changed to be allowed for branches with 2 labels -- we'd just have to pick one to put in the URL when the user clicks! Unsure what behaviour we want - feedback appreciated! It might be better to have 'something' than inexplicably be unable to specify a label you can see on a branch as a zoom point...?

---

One of the coolest things about this feature (I think) is that you can now have 'invisible zoom'. As in, you can zoom in to clusters that have been labelled via something other than `clade`, without having to clutter up the tree view by actually adding more text (via `clade`). You can of course also change the branch labels to show the other views, but you can have any branch label (or none) visible, and the zoom will still work for any of them.
_(This is actually what inspired me to do this - I really would like to send some more zoomed in views to collaborators, but I don't want to confuse them/clutter the tree by adding more clade annotations that aren't really clades.)_

---

I haven't tested this extensively with V1 to ensure it works fine - it should, but would be worth someone checking.
Would also benefit from some extra general testing. You can easily add extra labels to existing V2 JSONS by adding fields under "branch_attrs" "labels" in the format ex: "cluster": "African" etc. 
Would also appreciate extra eyes on the code to ensure I'm not doing things super-inefficiently :)

There's a little bit of existing debug to clean up once we are happy with the code.

**The next step** for this direction would be to update `augur clades` and `augur export` so that we can specify the clade 'name' (defaulting to 'clade' to preserve exact functionality with existing builds) and export these straight into the JSON. I don't expect this to be too difficult - but might wait until after `v6` release for everyone's sanity!  (Unless there's a strong preference otherwise.) I'd be happy to label a few things manually to zoom via URL until then :D

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -445,7 +445,7 @@ const modifyTreeStateVisAndBranchThickness = (oldState, tipSelected, cladeSelect
     oldState.selectedStrain = tipSelected;
   }
   if (cladeSelected) {
-    const cladeSelectedIdx = cladeSelected === 'root' ? 0 : getIdxMatchingLabel(oldState.nodes, "clade", cladeSelected);
+    const cladeSelectedIdx = cladeSelected === 'root' ? 0 : getIdxMatchingLabel(oldState.nodes, cladeSelected["label"], cladeSelected["selected"]);
     oldState.selectedClade = cladeSelected;
     newIdxRoot = applyInViewNodesToTree(cladeSelectedIdx, oldState); // tipSelectedIdx, oldState);
   }
@@ -630,6 +630,18 @@ export const createStateFromQueryOrJSONs = ({
     controls.colorByConfidence = doesColorByHaveConfidence(controls, controls.colorBy);
     tree.nodeColorsVersion = colorScale.version;
     tree.nodeColors = nodeColors;
+  }
+
+  // 'clade' zoom could now be under any label - check for first available query key that
+  // matches available branch labels, and convert it to be 'query.clade'
+  // This currently leaves any other 'clade zoom' requests in the URL dangling - they do nothing but remain
+  // Don't know if there's a better way to handle this?
+  const queryKeys = Object.keys(query);
+  for (const possibleClade of queryKeys) {
+    if (tree.availableBranchLabels.includes(possibleClade)) {
+      query.clade = {label: possibleClade, selected: query[possibleClade]};
+      break;
+    }
   }
 
   if (query.clade) {
--- a/src/actions/tree.js
+++ b/src/actions/tree.js
@@ -77,6 +77,13 @@ export const updateVisibleTipsAndBranchThicknesses = (
 ) => {
   return (dispatch, getState) => {
     const { tree, treeToo, controls, frequencies } = getState();
+    console.log("tree.selectedClade", tree.selectedClade);
+    // If going back to root or to unlabelled branch from 'custom' label branch, need to 
+    // expliticly set that 'label' to undefined, so it gets deleted from URL!
+    if (tree.selectedClade && !cladeSelected) {
+      cladeSelected = tree.selectedClade;
+      cladeSelected["selected"] = undefined;
+    }
     if (root[0] === undefined && !cladeSelected && tree.selectedClade) {
       /* if not resetting tree to root, maintain previous selectedClade if one exists */
       cladeSelected = tree.selectedClade;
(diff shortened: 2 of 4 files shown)


jameshadfield:
Really cool @emmahodcroft -- Alli & I struggled with exactly this in the DRC & ended up annotating trees with a whole lot of arbitrary clades in order to zoom to them in narratives. This will be a _massive_ improvement.

In terms of the overall concept, there are some things I think we should change / work out.

1. The current special treatment of the label `clade` is understandable but regrettable. It currently is set as the default to be displayed (this isn't too bad, we can add a `display_default` to modify this) and gets a URL query key of it's own (`clade`) which isn't great. I think it'd be better for the "zoom branch" to get a URL key of its own where a label is available. E.g. if a node has `myLabel: "myValue"` then you'd have `?label=myLabel:myValue` (or similar - key name & separator up for grabs), rather than `?myLabel=myValue`. This avoids issues where a label name clashes with another query key.  For labels called `clade` this would now become be `?label=clade:A1b/137F`; we could use `?clade=clade:A1b/137F` but this seems overly confusing.

2. We currently don't put anything in the URL for "hidden" states -- i.e. everything in the URL currently reflects something you can see. This would change this, by setting non-selected (and unseen) labels as URL query parameters. I think it's probably OK to make an exception here.

3. AA labels. While these currently look poor in the URL, we could use these with some modifications. E.g. an AA label of `NS3: T215A; NS4B: H15Q` currently goes to `?aa=NS3%3A T215A%3B NS4B%3A H15Q` (!) but by modifying [this code](https://github.com/nextstrain/auspice/blob/master/src/middleware/changeURL.js#L177) I think we can make it look close to the label string, although spaces aren't allowed. (This is how we get `?clade=A1b/137F` looking like it does rather than `?clade:A1b%2F137F`; technically, this seems not to be a valid URL but it's been working for a long time.)

4. What to do about multiple labels? Labels already have an ordering, in that they appear in the dropdown in a certain order (although how it's actually set & parsed might be a bit ad-hoc). If a node has multiple labels it seems sensible to use the first.

5. What to do about multiple labels with the same value in a tree? There are lots of situations where this will happen. Can't think of a good solution here at the moment.

emmahodcroft:
Thank you for these thoughts James - they are good! I will have to put this aside for a few days but hope to get back to it soon.

Just one quick though - my other hesitation with the AA labels being allowed in the URL is that in many trees, almost every node has one. This means that when clicking around the tree you will always have a rather hefty URL at the top, possibly distracting from the other URL things, which you might care more about (at the moment, it's all quite beautifully clear in the URL, which is kinda nice).

It also seems like these labels might be even more tenuous than clade labels when it comes to new data/re-runs - you can guard against this kind of thing _usually_ fairly easily in `clades` by giving enough mutations that it resists a few re-runs - but if you're zooming to a clade with 1 AA mutation, this seems like it could really jump around your tree. And, as you say, there could be multiple of them (without being able to specify in any way that the user should try to guard against this, which we could do in `clades`).

jameshadfield:
Closing as superseded by https://github.com/nextstrain/auspice/pull/861

PR #834, opened 2019-12-11
Footer text from dataset JSON

### LINKED ISSUE #707: Allow footer text to derive from meta.json

joverlee521:
This PR closes #707 

Uses `meta.description` from the dataset JSON as the footer text, if available. If not provided, then it will fall back to the hard-coded footer text. 

[marked](https://github.com/markedjs/marked) is used to convert Markdown to HTML. 
[DOMPurify](https://github.com/cure53/DOMPurify) is used to sanitized the generated HTML. It is very customizable, and I'm currently using a very explicit approach by listing all the tags and attributes that we allow.


jameshadfield:
Nice, thanks @joverlee521 -- if you're going to leave the hardcoded footers there could you move them to a separate file & match them to only nextstrain.org pages (this is https://github.com/nextstrain/auspice/commit/b894bac3a0d732874c5b798c4e89c5f1d9de2f69#commitcomment-36388052).

Could you provide some JSONs to test this out on? (Or branches of certain augur builds which generate these).

I won't comment on the sanitizer settings as I believe you and Tom have discussed this.

cli/server/convertJsonSchemas.js
```diff
@@ -237,6 +237,10 @@ const setMiscMetaProperties = (v2, meta) => {
       v2.geo_resolutions.push({key, demes: prettyDemes});
     }
   }
+  // FOOTER DESCRIPTION
```
jameshadfield: Since we are deliberately not making this part of v1 JSONs (i.e. meta + tree JSONs) I don't think this is necessary.

joverlee521:
> Could you provide some JSONs to test this out on? (Or branches of certain augur builds which generate these).

I've made a [PR](https://github.com/nextstrain/augur/pull/) in augur that completes the flow. You should be able to test with any build using the new `--description` option.

jameshadfield:
> I've made a PR in augur that completes the flow. You should be able to test with any build using the new --description option.

Just about -- could you convert the hardcoded footers into suitable markdown files in their various repos (at least, the ones which we are in charge of). This will complete the flow and allow some good testing 😄

joverlee521:
> > I've made a PR in augur that completes the flow. You should be able to test with any build using the new --description option.
> 
> Just about -- could you convert the hardcoded footers into suitable markdown files in their various repos (at least, the ones which we are in charge of). This will complete the flow and allow some good testing 😄

Yup, in the process of writing the Markdown files. So far I've only done it for [seasonal-flu](https://github.com/nextstrain/seasonal-flu/tree/test-footer).

jameshadfield:
Using a test-build within auspice (https://github.com/nextstrain/augur/commit/fa98bb2b04b04107ddb029182c331a77f1ff6be7) this PR works as expected and correctly strips out `<script>` tags. 

---

Regarding the two questions in https://github.com/nextstrain/auspice/issues/707#issuecomment-564711082:

> I believe the link would have to be in raw HTML within the Markdown to allow links to open in new tabs.

This doesn't work (see example above) & I'm not in favor of making all links open in new tabs. Happy to leave it as is. 

> unsure about how styling can be done without raw HTML within Markdown.

The current solution allows this which I think is a good solution.

---

As we talked about yesterday, the styling for the footer isn't great for headers as there were none in the hard-coded footers. Please wrap the footer in a styled component to make the headings thinner & smaller. If you implement this then minor stylistic changes can easily be added by others later.

image

tsibley:
@jameshadfield 

> if you're going to leave the hardcoded footers there could you move them to a separate file & match them to only nextstrain.org pages (this is [b894bac#commitcomment-36388052](https://github.com/nextstrain/auspice/commit/b894bac3a0d732874c5b798c4e89c5f1d9de2f69#commitcomment-36388052)).

Why leave these in Auspice and match them to only nextstrain.org pages instead of moving the hardcoded descriptions into nextstrain.org itself? That leaves Auspice clean.

jameshadfield:
I've thought about this and decided that I don't care. @joverlee521 please make sure that if you decide to implement a solution in the nextstrain.org repo that it works with streaming (https://github.com/nextstrain/nextstrain.org/pull/47 which I plan to implement this week). I don't know how to do this, but I'm assuming that it's possible.

joverlee521:
@jameshadfield Update: @tsibley agrees to keep the hard-coded text in Auspice. It _is_ possible to inject into the stream, but way more work than it's worth. 

I'll move the hard-coded text to a separate file within Auspice so we can delete them easily in the future when we want.

src/components/framework/footer.js
```diff
@@ -1,251 +1,135 @@
...
+      // Find nodes that contain images and add imageContainer class to update styling
+      if (([...node.childNodes].filter((child) => child.localName === 'img')).length > 0) {
+        node.className += ' imageContainer';
+      }
+    });
+
+    const sanitizer = dompurify.sanitize;
+    const sanitizerConfig = {
+      ALLOWED_TAGS: ['div', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p', 'em', 'strong', 'del', 'ol', 'ul', 'li', 'a', 'img', '#text'],
+      ALLOWED_ATTR: ['style', 'href', 'src', 'width', 'height', 'alt'],
```
tsibley: Note that allowing `style` will allow clickjacking. Is it absolutely necessary, especially considering images are now put into an `.imageContainer`?

joverlee521: Ah, right. I will remove it. I guess if people really want to customize their footer image styling, then they can change the styling within Auspice rather than the Markdown file.

joverlee521:
Thanks for catching that bug @trvrb! 
One question about code, do we want to include syntax highlighting? There is a [library](https://www.npmjs.com/package/highlight.js) to use for it if we want to include it.

tsibley:
> One question about code, do we want to include syntax highlighting?

This seems unnecessary to me.

trvrb:
Yes. Seems unnecessary to have syntax highlighting.

trvrb:
I just wanted to add that I like current behavior of injecting "This work is made possible by the open sharing of genetic data..." (https://github.com/nextstrain/auspice/pull/834/files#diff-2247b10275f828893a0c9035f6805927R172) to all nextstrain.org pages that lack `metadata.description`. This keeps us from needing to add `description.md` for pages like https://github.com/nextstrain/dengue.

PR #840, opened 2020-01-03
info-panel: round floating point numbers in info panels to 4 signific…

rneher:
…ant digits. This will change the info panels from this 

image

to this
tmp1

The >10 significant digit accuracy doesn't reflect reality and makes the panel hard to read. I am not sure there might be circumstances where we want to see floating point numbers at higher accuracy though.


jameshadfield:
👍 much better!

PR #841, opened 2020-01-04
Fix back button breaks

### LINKED ISSUE #815: Page history broken during navigation
There's some weird behaviour with Auspice modifying page history such that the back button breaks. For example, go to https://nextstrain.org/staging, observe you get a list of staging datasets and no 404 message, click a dataset, click the back button. You end up with a 404 message and a list of "live" datasets, instead of where you were (but the URL is intact). This manifests with other pages as well. I suspect this is because Auspice is being too greedy about what it considers a dataset for display.

Discussion about this started here: https://bedfordlab.slack.com/archives/C116R9PR7/p1568322725029100

joverlee521:
This PR closes #815.

Use existing function `chooseDisplayComponentFromURL` to determine what to set as the `displayComponent` instead of hard coding to "datasetLoader". 

This function includes known urls that are hardcoded to display the Auspice splash page and I've added "group" to this list.

### DIFF
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -12,11 +12,9 @@ export const chooseDisplayComponentFromURL = (url) => {
   if (
     !parts.length ||
     (parts.length === 1 && parts[0] === "") ||
-    (parts.length === 1 && parts[0] === "local") ||
     (parts.length === 1 && parts[0] === "staging") ||
     (parts.length === 1 && parts[0] === "community") ||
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @joverlee521, really great to have this bug fixed 🎉 😄 👍 

I've added a commit to remove the logic surrounding pathnames starting with "local" which haven't been used in a long time. 

In the long run this logic (which is largely specific to nextstrain.org) shouldn't be in the auspice codebase, but this fix is a big improvement.

PR #846, opened 2020-01-22
Date axis

jameshadfield:
Here we change the concept of minor grid lines for temporal trees to represent "sensible" periods of time rather than fitting a fixed number between major grid points. This will greatly aid in epidemiological understanding of trees. Major grid lines are also chosen using a new heuristic which attempts to pick sensible periods, and is responsive to screen sizes to avoid a squashed grid on small screens.

No modifications are made to the calculation of grids for divergence trees.

Examples:

image
☝️ minor grid lines represent days here

image
☝️ small screens look better, and the minor gridlines are still meaningful (here each minor grid is 1 week)

image
☝️ large screens can show lots of information. Here minor gridlines represent months.

image
☝️ a potentially unwanted side effect -- here weeks don't fit nicely into multi-month major gridlines, so the spacing of minor gridlines is less aesthetically pleasing. I'd argue it's more epidemiologically informative. This can be changed - for instance here we could have minor grid lines show months instead.

image
☝️ a improvement common to a lot of datasets - minor gridlines represent months

image
☝️ often you see fewer grid lines, but they are easier to understand

### DIFF
--- a/scripts/get-data.sh
+++ b/scripts/get-data.sh
@@ -55,6 +55,7 @@ data_files=(
   "flu_seasonal_yam_na_12y.json" "flu_seasonal_yam_na_12y_tip-frequencies.json" \
   "enterovirus_d68_genome_meta.json" "enterovirus_d68_genome_tree.json" \
   "enterovirus_d68_vp1_meta.json" "enterovirus_d68_vp1_tree.json" \
+  "ncov.json" \
 )
 
 rm -rf data/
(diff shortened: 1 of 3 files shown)


trvrb:
This is fantastic @jameshadfield! I think a huge improvement. Comparing old behavior and new behavior side by side, I'm completely sold.

I just made one small addition to slightly lighten grid lines. I was actually meaning to propose this for a while, but this ended being a nice time. I think this looks better especially when there are lots of minor lines corresponding to days.

emmahodcroft:
This is amazing James!! Thanks so much. I have spent far too long adjusting figures to make the date-markers appear in reasonable places. So happy you tackled it! :D

PR #849, opened 2020-01-25
Styling changes

trvrb:
This PR does:
1. Makes min map zoom slightly wider to better fit the globe in a grid view
2. Hides sidebar if window path contains `gisaid`
3. Always puts `author` in the tip hover panel

(2) Should be revisited to allow sidebar state in the URL, but I think this is a good solution for the time being.

### DIFF
--- a/src/components/main/index.js
+++ b/src/components/main/index.js
@@ -52,6 +52,9 @@ class Main extends React.Component {
       showSpinner: !(this.props.metadataLoaded && this.props.treeLoaded)
     };
     analyticsNewPage();
+    if (window.location.pathname.includes("gisaid")) { // TODO fix this hack by moving sidebar state to URL
+      this.state.sidebarOpen = false;
+    }
   }
   static propTypes = {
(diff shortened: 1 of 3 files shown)


dpark01:
Hi @trvrb -- out of curiosity, what was the motivation for hiding the sidebar if the window path contains `gisaid`?

trvrb:
@dpark01 ---

This was at the request of GISAID to make https://www.gisaid.org/epiflu-applications/next-hcov-19-app/ work nicely via an iFrame. They link to https://nextstrain.org/ncov/gisaid, which is exactly https://nextstrain.org/ncov except dropping the title. We didn't mean for this to catch things like https://auspice.broadinstitute.org/sars-cov-2/boston/gisaid-4k. We should narrow down the regex URL match.

cmloreth:
Hi @trvrb just wondering if there were any updates on being able to narrow down the regex URL match for paths containing gisaid? thanks!

PR #852, opened 2020-01-27
Fix bugs in redux state of pathname

### LINKED ISSUE #851: back button broken for community dataset pages in certain situations
This bug can be reproduced via the following steps:
1. Go to https://nextstrain.org/community/jameshadfield/basel-flu (or any community URL where a "default" dataset is not present)
2. Load one of the listed datasets
3. click the back button
4. The listed datasets are for the core source, not the community source.

jameshadfield:
We were getting bugs due to the redux state of `general.pathname` set to `undefined` when it shouldn't be as a result of `PAGE_CHANGE` actions with no `action.path` data. This resulted in incorrect API requests and the display of results which didn't match the URL.

This should close #851 but needs more testing via nextstrain.org

### DIFF
--- a/src/reducers/general.js
+++ b/src/reducers/general.js
@@ -20,11 +20,14 @@ const general = (state = {
 }, action) => {
   switch (action.type) {
     case types.PAGE_CHANGE:
-      return Object.assign({}, state, {
-        pathname: action.path,
+      const stateUpdate = {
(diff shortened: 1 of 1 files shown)


jameshadfield:
Tested locally (auspice) & on nextstrain (using review app)

PR #853, opened 2020-01-27
Pass width to nav-bar as prop

jameshadfield:
This wasn't defined in the API but was expected by the nextstrain nav-bar.

How nextstrain currently looks:
image

How nextstrain.org was designed to look (and will look with this PR):
image


trvrb:
This will be a nice improvement.

PR #855, opened 2020-01-27
hover: fix bug when authorInfo is null

### LINKED ISSUE #854: Screen blanks to white after loading for flu builds on live site
There is a bug with auspice in the live Nextstrain site where after loading the 2y flu build (https://nextstrain.org/flu/seasonal/h3n2/ha/2y) and hovering over the tree, the page goes blank.

The console view shows the following errors when the screen blanks:

image

I recreated this problem on my laptop with auspice 2.3.4 and found that the screen blanks at the same time the debugging message below about getting available datasets and narratives is printed.

```
GET DATASET query received: prefix=/flu/seasonal/h3n2/ha/2y
GET DATASET query received: prefix=/flu/seasonal/h3n2/ha/2y&type=tip-frequencies
GET AVAILABLE returning locally available datasets & narratives
```

I downgraded stepwise to previous versions of auspice and found the issue starts with v2.3.0. This bug affects the other flu segments and resolutions besides HA and 2y. The issue does not appear to affect other pathogen builds (at least not Zika, measles, or Ebola).

joverlee521:
This PR closes #854 

Check for `authorInfo` instead of `authorInfo.value` since`authorInfo` can be returned as `undefined`.

### DIFF
--- a/src/components/tree/infoPanels/hover.js
+++ b/src/components/tree/infoPanels/hover.js
@@ -269,7 +269,7 @@ const VaccineInfo = ({node}) => {
 const AttributionInfo = ({node}) => {
   const renderElements = [];
   const authorInfo = getFullAuthorInfoFromNode(node);
-  if (authorInfo.value) {
+  if (authorInfo) {
     renderElements.push(<InfoLine name="Author:" value={authorInfo.value} key="author"/>);
   }
 


jameshadfield:
Will set up a nextstrain.org review app now to test this

huddlej:
This fixes the issue for me locally. Thank you, @joverlee521!

jameshadfield:
Thanks @joverlee521 @huddlej. Tested & can confirm the fix.

trvrb:
This was my fault. Thanks so much for fixing the issue.

PR #857, opened 2020-01-28
prototype implementation of narratives for mobile

jameshadfield:
Current status: very early prototype.

Needs extensive testing and addition of features. Not tested with sidebar (theme) customisations.

### DIFF
--- a/src/components/narrative/MobileNarrativeDisplay.js
+++ b/src/components/narrative/MobileNarrativeDisplay.js
@@ -2,7 +2,13 @@ import React from "react";
 import { connect } from "react-redux";
 import queryString from "query-string";
 import { changePage, EXPERIMENTAL_showMainDisplayMarkdown } from "../../actions/navigation";
-import { MobileBannerTop, MobileBannerBottom, MobileContentContainer } from "./styles";
+import { TOGGLE_NARRATIVE } from "../../actions/types";
+import {
+  linkStyles,
+  MobileBannerTop,
+  MobileBannerBottom,
+  MobileContentContainer
+} from "./styles";
 import Tree from "../tree";
 import Map from "../map/map";
 import MainDisplayMarkdown from "./MainDisplayMarkdown";
@@ -30,6 +36,9 @@ class MobileNarrativeDisplay extends React.Component {
     super(props);
     this.state = {
       showingEndOfNarrativePage: false
+    };
+    this.exitNarrativeMode = () => {
+      this.props.dispatch({type: TOGGLE_NARRATIVE, display: false});
     };
     this.goToNextPage = () => {
       if (this.state.showingEndOfNarrativePage) return; // no-op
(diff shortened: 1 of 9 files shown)


src/components/narrative/MobileNarrativeDisplay.js
```diff
@@ -127,6 +127,19 @@ class MobileNarrativeDisplay extends React.Component {
           </MobileContentContainer>
         </>
       );
+    } else if (this.props.currentInFocusBlockIdx === 0) {
+      return (
+        <>
+          <MobileContentContainer height={contentHeight}>
+            {this.pageNarrativeContent()}
```
jameshadfield: block 0 is the title page & shouldn't display viz / main-markdown (see sketch file for more notes if needed). It'll probably need special styling too to correctly display the header / authors etc.

src/components/narrative/MobileNarrativeDisplay.js
```diff
@@ -153,13 +153,25 @@ class MobileNarrativeDisplay extends React.Component {
...
     );
   }
 
+  NextButton = (props) => (
+    <MobileBannerBottom height={this.state.bannerHeight} onClick={this.goToNextPage}>
+      {props.children}
+    </MobileBannerBottom>
+  )
+
+  PreviousButton = (props) => (
```
jameshadfield: This is cool - I didn't know fat-arrow functions would bind `this` when used inside classes.

src/components/narrative/MobileNarrativeDisplay.js
```diff
@@ -89,9 +89,7 @@ class MobileNarrativeDisplay extends React.Component {
     }
     const block = this.props.blocks[this.props.currentInFocusBlockIdx];
     return (
-      <div
-        id={`MobileNarrativeBlock_${this.props.currentInFocusBlockIdx}`}
```
jameshadfield: This is used for scrolling to a particular block via the URL query parameter. While this doesn't always work (bugs separate to this work) I'd like to preserve these id's here please.

jameshadfield:
Just talked through this with Kairsten:

Priorities (we can release after these are done + testing):
* Rebase onto master to pick up yesterday's narratives bugfixes
* Make reloading of a page with `?n=<x>` go to page `<x>` (this now works on non-mobile view, but mobile uses a different implementation)
* Change the explanation text on p0 to reflect mobile behavior
* Fix broken final page (no progress dot, broken links etc)
* Ensure things look OK using nextstrain.org theme extensions

Secondary tasks:
* Tweak the nextstrain.org theme (e.g. add banner colours etc)
* Test breakpoints to switch into mobile view

src/components/narrative/MobileNarrativeDisplay.js
kairstenfay: Is this still relevant? I don't know that I fully understand what you're saying here. Sorry I missed it earlier.

src/components/narrative/MobileNarrativeDisplay.js
kairstenfay: Me neither!

jameshadfield:
Thanks for this work @kairstenfay. I'm merging & releasing this now as it's a huge improvement over previous versions. Will make issues for the remaining bugs / work to do.

PR #859, opened 2020-01-28
Hide tip labels during animation

### LINKED ISSUE #848: Tree flickers oddly during animation
Looks like maybe an edge case for small trees, where the default label rendering logic is creating an odd flicker.

bug

jameshadfield:
Closes #848


trvrb:
Excellent! Looks great to me. Thanks for providing the review app.

There's a small bit of additional jank for nCoV due to windowing of the animation relative to available timesteps. I believe this is because the animation wants to cut up time into 0.075 intervals (relative to width of date slider), but these don't fall nicely on days. I don't think it's very important however.

jameshadfield:
Yeah -- separate to this issue (tip labels during animation) I think datasets like nCoV should disable animation (they can't do this at the moment).

PR #860, opened 2020-01-28
Display "subs per genome per year" as appropriate

### LINKED ISSUE #858: Regression rate calculation is wrong for mutation-scaled divergence
Please see https://github.com/nextstrain/ncov/issues/17.

Given that we don't specify in the Auspice JSON what units divergence has, my best idea would be to:

1. For the time being remove regression rate calculation from Auspice.
2. Have `augur refine` include subs rate estimate in `branch-lengths.json`.
3. Package up this rate estimate when running `augur export`
4. Use this rate estimate in Auspice rather than running a regression

This should be more accurate as well as the rate estimate produced by TreeTime is not just a regression.

jameshadfield:
Closes #858

### DIFF
--- a/src/components/tree/infoPanels/hover.js
+++ b/src/components/tree/infoPanels/hover.js
@@ -5,6 +5,7 @@ import { getTipColorAttribute } from "../../../util/colorHelpers";
 import { isColorByGenotype, decodeColorByGenotype } from "../../../util/getGenotype";
 import { getTraitFromNode, getDivFromNode, getVaccineFromNode, getFullAuthorInfoFromNode } from "../../../util/treeMiscHelpers";
 import { isValueValid } from "../../../util/globals";
+import { formatDivergence } from "../phyloTree/helpers";
 
 const InfoLine = ({name, value, padBelow=false}) => {
   const renderValues = () => {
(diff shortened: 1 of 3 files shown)


jameshadfield:
@emmahodcroft / @trvrb available for testing via:
https://auspice-fix-rate-info-h1f5bmaz.herokuapp.com/ncov/2020-01-23?l=clock&m=num_date
https://auspice-fix-rate-info-h1f5bmaz.herokuapp.com/ncov/2020-01-25?l=clock&m=num_date
https://auspice-fix-rate-info-h1f5bmaz.herokuapp.com/dengue/all?l=clock&m=num_date

trvrb:
Awesome! I have one small suggestion: "subs per genome per year" could be wrong for a something like HA in flu with mutation divergence or some other fragment in a different bug. I'd suggest just saying "subs per year" or "subs per sequence per year" (though I slightly prefer the former).

emmahodcroft:
This looks great @jameshadfield! Thanks for getting it up so quickly. I don't have strong feelings between Trevor's suggestions of either 'subs per year' or 'subs per sequence per year'.

trvrb:
I've elected to just shorted to "subs per year". Will go ahead and merge. Thanks again James.

PR #861, opened 2020-01-29
Custom Clade Labels for Zoom - Revisited

emmahodcroft:
I wanted to revisit this idea, which I originally brought up in Nov 2019, in PR #831 . I think, in particular, it could come in handy very soon for the nCoV narratives, as we'll have more structure in the tree soon and it may be helpful to be able to do fancy zooming.

Most of this is about the same as previously but I've addressed almost all of the requests that @jameshadfield [made on the previous PR](https://github.com/nextstrain/auspice/pull/831#issuecomment-559288177):

1. `clade` is no longer specially treated. The URL now used is, as was suggested by James, `?label=myLabel:myValue`. For example, for clades: `?label=clade:A` or for something else: `?label=grouping:hospital` . Labels and values can be anything at all. `clade` is still 'specially treated' in other senses, in that it shows up by default (if present) in the Auspice view, etc.

2. As before, you can now zoom to a 'hidden' (not displayed) clade label. This will show in the URL if there's a clade label in the tree but not displayed. Unlike before, you can now zoom to a node with multiple labels - they are ordered, as suggested, by the oder in the drop-down on the controls. So, if `clade` is present, it'll be first-choice for the zoom URL, for example. 

3. I still haven't included AA labels. Not only is the URL parsing much trickier, but this ends up meaning almost every node in the tree has a 'label', so every single branch you click on changes the URL (and often in a really long way). It's also not uncommon to have clashes (same value in multiple nodes) here. I'm just not sure it's useful, but we can definitely revisit this later, perhaps with a bit more thinking through.

4. See 2 - if a node has multiple labels, they're ordered by the order in the drop-down menu (excluding `aa`). This is despite what's visible - so for a branch with `clade:A` and `groupings:hospital`, even if you have `grouping` selected for the branch labels, you'll still get `clade:A` as the URL. (You can still put `groupings:hospital` in the URL yourself and get the same zoom result.) I think we could address in future for perfect behaviour but I don't think this should stop us implementing this now.

5. When clicking to zoom, there is no awareness of whether there are multiple labels with the same value in the tree. (This is also true with current behaviour of clade zooming.) It will just load the values of the node that's been clicked, even if they are also elsewhere. Again we could perfect this in future but it's the same as current and I think fine for now. 

6. If you specify a `label` value that exists two place in the tree, via URL, it currently throws a `console.error` and doesn't zoom at all. This is *better* than current behaviour with clades, which just zooms to whatever is encountered first when traversing the tree, with no warning. **However** I will now push a commit which makes a pop-up warning for the user to make this clearer. I'm doing this separately as I'm not 100% this is a good way to do it, so it could be undone easily if it's too much and better to stick with the basics for now, and come back to this later.

-----

As before, the coolest thing about this is that you can now put custom 'clade' labels in the JSON and use them to zoom via URL without having to display a bunch of additional `clade` labels all over your tree.  I think this will be *great* for narratives. 

As before, this currently has to be done manually as there's no corresponding `augur` function to do anything but the 'official' `clades` values, but I can start working on this next. For the moment, even though it's manual, I still think it'll be super useful for us!

----

Hopefully this should create some kind of Heroku version that can be tested out!

### DIFF
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -181,7 +181,8 @@ const fetchDataAndDispatch = async (dispatch, url, query, narrativeBlocks) => {
         query,
         narrativeBlocks,
         mainTreeName: secondTreeUrl ? mainDatasetUrl : null,
-        secondTreeName: secondTreeUrl ? secondTreeUrl : null
+        secondTreeName: secondTreeUrl ? secondTreeUrl : null,
+        dispatch
       })
     });
 
@@ -233,7 +234,7 @@ export const loadSecondTree = (secondTreeUrl, firstTreeUrl) => async (dispatch, 
     return;
   }
   const oldState = getState();
-  const newState = createTreeTooState({treeTooJSON: secondJson.tree, oldState, originalTreeUrl: firstTreeUrl, secondTreeUrl: secondTreeUrl});
+  const newState = createTreeTooState({treeTooJSON: secondJson.tree, oldState, originalTreeUrl: firstTreeUrl, secondTreeUrl: secondTreeUrl, dispatch});
   dispatch({type: types.TREE_TOO_DATA, ...newState});
 };
 
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -445,7 +445,7 @@ const modifyTreeStateVisAndBranchThickness = (oldState, tipSelected, cladeSelect
     oldState.selectedStrain = tipSelected;
   }
   if (cladeSelected) {
-    const cladeSelectedIdx = cladeSelected === 'root' ? 0 : getIdxMatchingLabel(oldState.nodes, "clade", cladeSelected);
+    const cladeSelectedIdx = cladeSelected === 'root' ? 0 : getIdxMatchingLabel(oldState.nodes, cladeSelected["label"], cladeSelected["selected"]);
     oldState.selectedClade = cladeSelected;
     newIdxRoot = applyInViewNodesToTree(cladeSelectedIdx, oldState); // tipSelectedIdx, oldState);
   }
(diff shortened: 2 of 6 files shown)


emmahodcroft:
Latest commit implements pop-up warnings to users in two cases:
* The label value cannot be found (`?label=clade:T`, when there is no `clade` label `T` in the tree)
image
(Currently just the `console.error` is shown - not visible to most users.)

* There are multiple nodes in the tree with the requested label value (`?label=clade:A`, when two branches are labelled `clade A`. 
image


If you put a completely non-existing label into the tree (`?label=pie:pumpkin` and there is no `pie` label), no warning is shown, but the query is removed from the URL and no zoom is performed.

jameshadfield:
Looks great! Will test over the next 24 hours

emmahodcroft:
Thanks @jameshadfield !

emmahodcroft:
Thanks James! I agree about combining with Map Zoom to make some super cool stuff. 
I will try to send you some testing stuff ASAP and will look into all your other points today and tomorrow.

Think I can fix the backwards compatibility as I had both working for a while during testing :) But will try to add something to 'convert' it to the new format to stop it persisting...

Couple questions:
- When you say the random labels you added to Zika don't render, what do you mean? They don't show up visually? Were these using labels other than 'clade'? I think I can imagine how that might not be working properly...
- For label values not being anything but strings, what behaviour do we want here? When we do eventually including something to add other labels in `augur` we can include in verification on that end, but for anything added in between, shall we just ignore? Or error with a warning?

If you can send me the data/narratives for your testing as well, I will check them all out too!

emmahodcroft:
The error I am getting for the ebola narrative is related to two of the slides having the same title. However, this doesn't come up with `master`, so perhaps was something fixed/modified in between when I branched this and now? I'll just update the narrative for now so I can focus on the next error, but writing this mostly to myself, to remind me to test this again later.

emmahodcroft:
The most recent commit should fix 
- backwards compatibility issue - can now say `clade=B` like the old style and it works (and convert it in the URL). This also fixes the Ebola narrative
- clicking on branches with no clades giving a bad URL (it seems this was related to the above)

One thing I noticed that's true both in the old and new functionality is that in a narrative if you are zooming to clades (like the Ebola one) and go 'back' to a slide with the original full tree, it doesn't zoom out again.
You can force this by putting `&clade=root` in the URL, but we should probably try to make this default behaviour if no label/clade specified. 
Given this is current behaviour this is maybe low priority for this PR.

jameshadfield:
> For label values not being anything but strings, what behaviour do we want here? 

Let's deal with this in a separate PR (it's due to queries always being strings but labels can be numeric). Right now it'll have a warning banner saying "... not found in tree". This is fine.

> The error I am getting for the ebola narrative is related to two of the slides having the same title. However, this doesn't come up with master

I can no longer recreate my original error (!). P.S. The "key" error should now be fixed in master, and react warnings such as these only show up in dev mode, which is why you don't see it on nextstrain.org etc.

> One thing I noticed that's true both in the old and new functionality is that in a narrative if you are zooming to clades (like the Ebola one) and go 'back' to a slide with the original full tree, it doesn't zoom out again.

This is a bug. Let's move to a separate issue.

> When you say the random labels you added to Zika don't render, what do you mean? They don't show up visually? 

Correct -- you can't see them (but they are there). Not sure if there are DOM elements.

emmahodcroft:
Ok!

So the Ebola narrative is working now, plus the other things either working or moved to separate issues?

For the last (?) remaining thing of the missing labels - still a little unsure how to recreate however I found this:

If something is `clade` then it's treated specially in that it's always displayed, no matter how the tree is zoomed, or how 'crowded' the tree is. However, this isn't true for any other 'label' (aa included). If it's anything else, and it's deemed by some measure to be too 'crowded' to show it, it won't show it until you zoom in. You can see this with most trees and aa labels, but I also tested it out with some fake labels on a larger tree. 

Could this be what you were seeing, or something else? If not it, let me know a little more so I can recreate it!

Apart from that, I think this might be good?

jameshadfield:
All commits & bugfixes now in https://github.com/nextstrain/auspice/pull/886 🎉

Issue #842, opened 2020-01-08
info boxes too big


rneher:
when lots of metadata fields are available, tip info boxes get quite large -- sometimes large enough that there is no space left to 'click outside the box' to get rid of it.

 * should we have a way to restrict data shown there? if so, at the level of the `export` and in the json, or dynamically via a UI
 * limit the size of the box? make it scrollable? separate close button...

trvrb:
I generally like the idea of a close button on the modals (while still keeping the click outside behavior). The download modal does scroll, so that could be a target. Also, we might think about centering the tip modal on the screen (like the download data modal), in part to give more real estate. The map and diversity panels are pretty uninteresting with one tip selected anyway.

jameshadfield:
We can definitely add a `x` button to the modal right now so that this isn't problematic in the short term. Let me know & I can do this ASAP.

More generally, the tip modal box isn't ideal, especially with narratives. We want some way of seeing this information without obscuring the tree. Reloading a tree with a tip selected also doesn't bring up this modal (e.g. https://nextstrain.org/zika?s=Angola/2017/PoHuZV634959) -- this is a bug which is somewhat helpful for narratives.

jameshadfield:
Closed by #862 -- thanks @emmahodcroft 🎉

PR #862, opened 2020-01-29
Make Infobox Y-Scroll and Limit Its Size

emmahodcroft:
This attempts to address #842 . Probably the info-box could still be improved, but this should make things work a bit better, anyway!

Simple adjustments that restrict the tip-click info box height and width, and allow Y-scrolling (if needed). As a side-effect, gets rid of the previously unusable but present scrolls bars.

In my tests there was always enough room around the box to click outside of it and exit, but this should be confirmed. Otherwise (or either way) the X-box that James mentioned could be added.

I currently do the height restriction in `src/components/tree/infoPanel/click.js` by copying the `infoPanelStyles.panel` object from `src/globalStyles.js`, as I wasn't sure if anything else might use this style, and didn't want to accidentally cause unintentional changes elsewhere. This code could be simplified to just changing the `globalStyles.js` file, if it this won't mess up anything else!

------

Before:
image
(note that though a scroll bar is visible, it doesn't work)

After:
image

-----

Before, when zoom very high - infobox takes up whole screen - can't click away to close!
image

After - same zoom. Infobox now scan scroll, and doesn't take up whole screen, so can click away!
image

### DIFF
--- a/src/components/tree/infoPanels/click.js
+++ b/src/components/tree/infoPanels/click.js
@@ -209,10 +209,12 @@ const Trait = ({node, trait, colorings}) => {
  */
 const TipClickedPanel = ({tip, goAwayCallback, colorings}) => {
   if (!tip) {return null;}
+  const panelStyle = { ...infoPanelStyles.panel};
+  panelStyle.height = "70%";
   const node = tip.n;
   return (
     <div style={infoPanelStyles.modalContainer} onClick={() => goAwayCallback(tip)}>
-      <div className={"panel"} style={infoPanelStyles.panel} onClick={(e) => stopProp(e)}>
+      <div className={"panel"} style={panelStyle} onClick={(e) => stopProp(e)}>
         <StrainName>{node.name}</StrainName>
         <table>
           <tbody>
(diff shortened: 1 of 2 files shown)


trvrb:
I think this is a big improvement. Thanks Emma!

I do have one thought: I think it would make more sense to have the gray click box centered in the main panel area rather than restricted to just the tree panel. In these cases, the map is not so helpful:

Screen Shot 2020-01-29 at 5 22 26 PM

I'm imagining it would look more like it does with "full" view (obscuring both tree and map):

Screen Shot 2020-01-29 at 5 23 26 PM

This could also be done in a separate PR. Just issue I found in the current PR:

Screen Shot 2020-01-29 at 5 24 11 PM

The click box extends all the way to the bottom of the page when it doesn't need to.

jameshadfield:
Thanks @emmahodcroft -- nice improvements here. @trvrb i'm going to move your comment to a separate issue (and one that's not urgent), which is how to deal with info-boxes more generally (and one that will work with narratives).

emmahodcroft:
@trvrb Thanks for spotting the little bug where the box extends to the bottom of the page when it doesn't need to! It seem this is an easy fix - just set `maxHeight="70%"` rather than `height`. I will push this little adjustment to master and it should look better in full view :)

PR #869, opened 2020-01-30
'Reset Zoom' Works when Filters are Set

### LINKED ISSUE #863: Reset map zoom doesn't take into account filters
Take a look at https://nextstrain.org/ncov?f_country=China&r=location. This should "reset" to just the locations in China and be zoomed appropriately. 

~However, "reset zoom" zooms way out, I believe because it's trying to include all three sets of lat/longs (both the central, but also the left-shifted and right-shifted locations). I suspect the fix here is to only use the central set of lat/longs when calculating default bounds.~

I first thought this was `map_triplicate`. It's not, reset zoom works with map triplicate as it should. However, it doesn't seem to take into account filtered state.

emmahodcroft:
Should resolve #863 !

When deciding what lat-longs should be added to set the map resolution, it now checks the `counts` for each deme - if 0, it's not added. 

So, this means you can now select a filter (or multiple) and then click 'reset zoom' and it'll auto-adjust! Most visible with selection regions and stuff, but also works with other filters, too:
map-zoom-with-filter

At the moment this *doesn't* auto zoom to the right resolution if you load up with a filter in the URL:
map-zoom-with-filter-loading

This requires changing the initial rendering order, and I was unsure if this is ok, or might break something else. I will put this functionality in, in a separate commit, so that it can be reversed or modified if needed!

### DIFF
--- a/src/components/map/map.js
+++ b/src/components/map/map.js
@@ -199,11 +199,6 @@ class Map extends React.Component {
     if (mapIsDrawn && allDataPresent && demesTransmissionsNotComputed) {
       timerStart("drawDemesAndTransmissions");
       /* data structures to feed to d3 latLongs = { tips: [{}, {}], transmissions: [{}, {}] } */
-      if (!this.state.boundsSet) { // we are doing the initial render -> set map to the range of the data
-        const SWNE = this.getGeoRange();
-        // L. available because leaflet() was called in componentWillMount
-        this.state.map.fitBounds(L.latLngBounds(SWNE[0], SWNE[1]));
-      }
 
       const {demeData, transmissionData, demeIndices, transmissionIndices} = createDemeAndTransmissionData(
         this.props.nodes,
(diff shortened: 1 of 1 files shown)


emmahodcroft:
Ok! So with the latest commit, the map resolution should load up from the beginning at the right level of zoom:
map-zoom-with-filter-loading2

This requires changing the order of the code in `maybeDrawDemesAndTransmissions` in `src/components/map/map.js`. It doesn't seem to affect anything for me - but wanted to do this separately so @jameshadfield can check this isn't going to break anything!

This should unlock some cool stuff! :)

emmahodcroft:
This seems it has a bug - can't get it to work with ncov/2020-01-30 that I downloaded from Nextstrain.org today.  (Works with other local datasets I am using.) I will try to check into this more later today. Sorry.

jameshadfield:
Closing this as I believe this is included in #870 -- please reopen if I've got this wrong

PR #870, opened 2020-01-30
Totally Dynamic Map Zooming in Narratives

emmahodcroft:
This is a super-master PR for all the great Zoom work!

It incorporates my PR for the 'Reset Zoom' working with filter changes. And @jameshadfield's PR for detecting geo_resolution changes in Narratives and changing the zoom. (And my tiny push to master for the tip-boxes - accidentally.)

PLUS something that takes advantage of all of this - and means we can now auto-zoom in narratives if the filter changes (and this would change the geo boundaries).

Here's an example that first uses Jame's auto-switch for changing the geo-resolution, then mine for detecting a filter change!:
narrative-all-auto-zoom

We can start doing some very cool stuff with this! It would be great to use this, and the cluster zoom, to do some neat stuff in the next nCoV narrative update :)

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -642,7 +642,7 @@ export const createStateFromQueryOrJSONs = ({
 
 
   /* calculate colours if loading from JSONs or if the query demands change */
-  if (json || controls.colorBy !== oldState.colorBy) {
+  if (json || controls.colorBy !== oldState.controls.colorBy) {
     const colorScale = calcColorScale(controls.colorBy, controls, tree, treeToo, metadata);
     const nodeColors = calcNodeColor(tree, colorScale);
     controls.colorScale = colorScale;
--- a/src/components/map/map.js
+++ b/src/components/map/map.js
@@ -75,7 +75,6 @@ class Map extends React.Component {
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#es6-classes
     this.playPauseButtonClicked = this.playPauseButtonClicked.bind(this);
     this.resetButtonClicked = this.resetButtonClicked.bind(this);
-    this.resetZoomButtonClicked = this.resetZoomButtonClicked.bind(this);
     this.fitMapBoundsToData = this.fitMapBoundsToData.bind(this);
   }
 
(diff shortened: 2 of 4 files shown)


emmahodcroft:
This seems it has a bug - can't get it to work with ncov/2020-01-30 that I downloaded from Nextstrain.org today.  (Works with other local datasets I am using.) I will try to check into this more later today. Sorry.

This is because it contains same changes as the `map_zoom_w_filters` branch.

jameshadfield:
@emmahodcroft i've added 4 commits here and wanted to explain what's going on. All in all, i'm super happy with where we've got to and think the behavior is fantastic. I've tried to write detailed commit messages to explain what each one's doing and why they're here.

Current behavior (needs a bit more testing on different datasets):
* On map load, the map bounds are set to the visible data
* If a user _has not_ interacted (pan/zoom) with the map, then:
    * enabling filters / changing geo res will cause the map to resize accordingly.
(If they have panned/zoomed, then these things have no effect on the map bounds, to avoid the annoying situation of zooming into some country and then having the map zoom somewhere else)
* in narratives mode, subsequent pages can set filters / change geoRes and the map updates nicely.


There are lots of console logs left in in case you want to tweak behavior - they should be removed before merge.

jameshadfield:
Unfortunately this is broken with animation :( -- see https://auspice-both-zooms-pro2wfzdt97.herokuapp.com/narratives/twenty-years-of-WNV?n=2

emmahodcroft:
@jameshadfield You are like an angel in the night, leaving me amazing things to wake up to! Super excited to play with this today :)

emmahodcroft:
James, this is awesome!! Sorry I've been slow to test, it's a little crazy, as I'm sure you know. 

I will continue testing but just a couple of quick notes - overall it looks great!
- If at any point you press 'Reset Zoom' (perhaps mistakenly trying to remove a filter - I did this without thinking) then this apparently counts as 'moving the map' and so from that point, it won't auto-zoom anymore. Not sure if we can tell these two apart, but if possible it would be nice if (in absence of any other panning/zooming) this 'didn't count' as moving the map.
- What would be your thoughts on making the 'Reset Zoom' button also 'reset' if someone has panned and zoomed? So maybe someone pans & zooms for a while, but then they want to go back to whatever the default zoom is, so they hit 'Reset Zoom' - should the site then resume the behaviour of auto-zooming with filters, etc? I'm slightly leaning towards this but haven't thoroughly turned it over.

I will try to test more as well - sorry, I keep getting pulled away. But I tried out with multiple data sets and a couple narratives, and all is working for me so far!

jameshadfield:
> What would be your thoughts on making the 'Reset Zoom' button also 'reset'

It should reset everything -- that's what this line was supposed to be doing https://github.com/nextstrain/auspice/pull/870/files#diff-1e9bedd1f09454a766be711b9756a86dR639, but the `onClick` handler must be interfering.

jameshadfield:
Superseded by https://github.com/nextstrain/auspice/pull/886 🎉

Issue #655, opened 2018-09-05
Blue dots bunch to left on IE/Edge/iPhone


emmahodcroft:
In Internet Explorer, Edge, and on iPhone (maybe others?) the blue dots at the top of narratives bunch to the left.

![image](https://user-images.githubusercontent.com/14290674/45081545-b41cd800-b0f7-11e8-8228-bd2f443492fa.png)

kairstenfay:
Closed by #875 thanks to @ivan-aksamentov

PR #875, opened 2020-01-30
fix(styles): navigation progress buttons to not align properly on IE

ivan-aksamentov:
@kairstenfay
Hi Kairsten,

Pardon me for this blatant intrusion into your private branch.

I was [playing with IE fixes today](https://github.com/nextstrain/nextstrain.org/pull/109) and noticed that one of the styling bugs is much easier to fix on your branch (compared to master). It probably will be merged soon(ish) anyway, right? So I decided to base my fix here, rather that dealing with master, which has the component in question in a bit of a hairy state.

Anyways, this commit fixes navigation progress buttons to not align properly in flexbox container when browsing using Internet Explorer (any version).

This is, what I believe to be a browser bug, so called "flexbug-15" as described in:
https://github.com/philipwalton/flexbugs#flexbug-15
I applied the workaround as described there.

Originally reported by @emmahodcroft at: https://github.com/nextstrain/auspice/issues/655

There are still styling issues in IE, but at least working navigation makes narrative usable for these folks who still use IE.

Before: 
before

After: 
after

Do you think you can merge these changes along with yours?

### DIFF
--- a/src/components/narrative/styles.js
+++ b/src/components/narrative/styles.js
@@ -163,6 +163,11 @@ export const ProgressButton = styled.div`
   background: #74a9cf;
   border-radius: 50%;
   cursor: pointer;
+
+  // BEGIN: mitigate ie flexbug-15: https://github.com/philipwalton/flexbugs#flexbug-15
+  align-self: center;
+  margin: auto;
+  // END mitigate ie flexbug-15
 `;
 
 const baseBannerStyles = css`


ivan-aksamentov:
I also submitted same fix for the master branch anyways: https://github.com/nextstrain/auspice/pull/876
The style is inline there, so it's not really the same fix, I guess.

kairstenfay:
Thanks for submitting this, @ivan-aksamentov! This resolves #655.

PR #881, opened 2020-02-01
Narrative: explore data & leave narrative 

### LINKED ISSUE #871: Clicking "Leave the narrative & explore the data yourself" leads to an incomplete page
Repro:
 - Go to the latest situation report on 2019-ncov, currently at https://nextstrain.org/narratives/ncov/sit-rep/2020-01-30
 - Scroll to the very last page
 - Click "Leave the narrative & explore the data yourself"
 - You will be redirected to the latest https://nextstrain.org/ncov results, however, the page does not show either phylogeny tree or map and the sidebar seems rather empty
 - Refresh the page. The page is now complete, as expected.

I think this is a bug and the complete page with phylogeny tree and map should be shown from the beginning.

joverlee521:
This PR resolves #871 

1. Hide "Explore the data yourself" button when on main display markdown page
2. Clicking on "Leave the narrative..." button completely exits the narrative and changes to the dataset page.

### DIFF
--- a/src/components/main/index.js
+++ b/src/components/main/index.js
@@ -128,7 +128,9 @@ class Main extends React.Component {
           navBarHandler={() => {this.setState({sidebarOpen: !this.state.sidebarOpen});}}
         />
         <PanelsContainer width={availableWidth} height={availableHeight} left={this.state.sidebarOpen ? sidebarWidth : 0}>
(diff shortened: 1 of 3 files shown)


jameshadfield:
Works well, thanks @joverlee521.

If we ever want to revisit this, since the data is already loaded we should be able to show the dataset view & modify state via the JSON-set defaults without needing to reload the entire page. But this would only be a minor improvement & would require more complex code.

PR #882, opened 2020-02-01
fix: ensure Lato font correctly falls back to sans-serif

ivan-aksamentov:
Currently, some of the text elements, for example `<a/>`, are forced to be rendered using Lato font (which is a sans-serif font). When text contains Unicode characters (for example, from Chinese and Cyrillic alphabets) which are missing in our current version of Lato font, these characters are automatically substituted from other fonts.

However, this substitution is unpredictable (depends on browser implementation, platform, font and characters in question) and in some cases leads to inconsistent results.

This adds Helvetica and sans-serif alternatives to the places where only Lato was specified, thus at least ensuring that Unicode will properly falls back to sans.

Example with Russian:

Before - Cyrillic characters in these `<a/>` tags on my machine are rendered with Tinos font, which is serif and looks smaller:
ru_serif

After - Cyrillic characters in these `<a/>` tags are rendered with Arimo font (sans):
ru_sans

### DIFF
--- a/src/components/narrative/styles.js
+++ b/src/components/narrative/styles.js
@@ -144,7 +144,7 @@ export const linkStyles = { // would be better to get CSS specificity working
   color: "#5097BA",
   textDecoration: "none",
   cursor: "pointer",
-  fontFamily: "Lato",
+  fontFamily: `"Lato", "Helvetica Neue", "Helvetica", "sans-serif"`,
   fontWeight: "400",
   fontSize: "1.8em"
 };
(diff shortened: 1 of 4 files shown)


jameshadfield:
Thanks @ivan-aksamentov ⭐️ 

In the future I'd like to move more and more styling to styled components where we can use themes to set these in a single place. (Note that nextstrain.org modifies auspice at build time to use its own fonts -- https://github.com/nextstrain/nextstrain.org/blob/master/auspice/client/config.json#L6, although these happen to be the same as auspice's -- and these modifications are designed to be disseminated within auspice via a styled components theme. As you can see, we haven't been consistent about using this technique!)

PR #891, opened 2020-02-07
max-zoom when fitting to data depends on geographic resolution

jameshadfield:
To avoid setting the bounds too small (e.g. if restricted to one country
      then we don't want to be at maximum zoom) we use hardcoded zoom levels which
      depend on the geo resolution. In the future these may be settable via the dataset
      JSON. Note that this does not change the max zoom level available when interacting
      with the map. See https://leafletjs.com/examples/zoom-levels/ for what a zoom
      level represents.

### DIFF
--- a/src/components/map/map.js
+++ b/src/components/map/map.js
@@ -592,12 +592,34 @@ class Map extends React.Component {
       }
     }
   }
-
+  getMaxZoomForFittingMapToData() {
+    /* To avoid setting the bounds too small (e.g. if restricted to one country
+      then we don't want to be at maximum zoom) we use hardcoded zoom levels which
+      depend on the geo resolution. In the future these may be settable via the dataset
+      JSON. Note that this does not change the max zoom level available when interacting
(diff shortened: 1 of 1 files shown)


jameshadfield:
This implementation is _better_ than the current version of auspice (which was better than the previous version), but is still not perfect. I'm holding off letting the JSONs set this until we are sure it's the right approach.

## Country zoom is targeted towards bigger countries:
https://auspice-max-zoom-when-f-jkvpar.herokuapp.com/ebola?f_country=Liberia&r=country isn't great vs https://auspice-max-zoom-when-f-jkvpar.herokuapp.com/zika?f_country=USA

## Fallback for geo resolutions not included here
E.g. see division, which wasn't included in this PR as it has different meanings in different datasets http://localhost:4000/mumps/na?f_division=Ontario

trvrb:
Thanks for this @jameshadfield! I played around with the Heroku review app and I think this is indeed a good compromise. Please merge whenever you'd like.

PR #892, opened 2020-02-09
Make Auspice / NCoV map more translatable

mapmeld:
This project has been really interesting to follow! I saw in https://github.com/nextstrain/ncov/issues/32 that it's difficult to change the language of instructions in the NCoV map because they are hardcoded here in the Auspice repo.

Here's what this pull request does:
- Creates one locales.json file with all languages
- Adds a getPreferredLanguage function - that you can overwrite in config - which looks at the browser's preferred languages and narrative URLs
- Has English as a fallback
- Replaces the narrative instructions and a few other labels with the preferrd language

Why is ```my_locale``` set at render and not at the beginning of the file? When I was testing, arriving on the homepage set the language to EN. Selecting a Spanish narrative did not re-check the language.

Possible changes: I could see my_locale being changed to a function, so you could call ```my_locale('intro_paragraph')``` and it could fall back to English if a language is missing translations of individual phrases

Update, more possible ideas:
- having getPreferredLanguage return my_locale object, with English fallback if any key are missing
- for zh-CN, zh-HK, and zh-TW to have separate locales, you'd need to change how the browser locale is read in


src/util/preferredLanguage.js
```diff
@@ -0,0 +1,15 @@
...
+
+export const getPreferredLanguage = hasExtension("getPreferredLanguage") ?
+  getExtension("getPreferredLanguage")
+  : () => {
+  let browserLang = window.navigator.language.split("-")[0],
+      narrativeLang = "",
+      urlParts = window.location.pathname.toLowerCase().split("/");
+
+  if (urlParts.length > 0 && urlParts[1] === "narratives") {
+      narrativeLang = urlParts[urlParts.length - 2];
```
tsibley: This is also a bit of logic specific to our 2019-nCoV narratives which doesn't have any fundamental basis in Auspice's view of things. I think we'd want to keep that separation in the default implementation of this function.

src/util/preferredLanguage.js
```diff
@@ -0,0 +1,15 @@
+import { hasExtension, getExtension } from "./extensions";
+
+export const getPreferredLanguage = hasExtension("getPreferredLanguage") ?
+  getExtension("getPreferredLanguage")
+  : () => {
+  let browserLang = window.navigator.language.split("-")[0],
```
tsibley: I think important to note that this only considers the most-preferred language, but it's critical to consider all the user's preferred languages ([`window.navigator.languages`](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages)) as their second or third preference might be the only one we support. Using a library like [accept-language](https://www.npmjs.com/package/accept-language) can help with this matching.

jameshadfield:
Adding support for different languages into auspice would be a fantastic addition.

Currently, the nextstrain.org server attempts to guess the desired language and redirect auspice to sourcing a translated narrative file, if available (https://github.com/nextstrain/nextstrain.org/blob/master/redirects.js#L37). This is separate to auspice. 

@tsibley could the nextstrain.org server response (either the redirect, or the subsequent `getNarrative` call) return some metadata indicating the language? Adding language support via this approach would keep the auspice-derived elements and the narrative content in sync, and allow you to switch back to english if desired, which is important. 

More long term, the selected language needs to selectable via the UI and be stored in redux state as to allow users to switch between them. And this needs to be in sync with the language dropdown currently injected by the custom nextstrain.org nav-bar.

mapmeld:
I agree on using ```i18next``` and ```accept-language``` for a more robust system. I'll take a stab at it in the next few days / maybe weekend, and figure out how to make the translations extensible for NextStrain. If anyone else is working on this already let me know if there is a Slack or something.

src/util/preferredLanguage.js
```diff
@@ -0,0 +1,15 @@
+import { hasExtension, getExtension } from "./extensions";
+
+export const getPreferredLanguage = hasExtension("getPreferredLanguage") ?
+  getExtension("getPreferredLanguage")
```
jameshadfield: Extensions are currently stored as a JSON file and so can't define functions. (One day we will be able to provide functions, but this is not currently the case.)

jameshadfield:
After talking with @tsibley here's what I propose:

Interpret the language via inspection of the URL, as in this PR. The code which does this should include comments stating that it's a temporary solution (see below). Store this information in redux state (`state.controls.language` or similar). Each component can then use a `getText(key, language)` function to fetch the text, which falls back to english if the key has not been set in the JSON for that language. 

In the future, we will expose this language support via a UI in auspice, but there are larger questions about how this will tie in with server requests / responses which we want to ensure we get right. 

Happy to help with any aspects as desired - thanks for your work here!

tsibley:
:+1: I do still think it'd be a good idea to use `i18next` and `accept-language` now if we can.

mapmeld:
I finally returned to this pull request! Hope that this version is more useful
- uses i18next's react-i18next's ```<Trans>``` tag, if the inner content is a key it will look it up in the JSON file, otherwise use ```<Trans i18nKey="shorter_key">```
- uses [i18next's language detector](https://github.com/i18next/i18next-browser-languageDetector) which looks at language options (including navigator.languages) and has overrides (including URL params and cookies)

PR #894, opened 2020-02-12
Include proper MapBox attribution for tiles

trvrb:
See here for recommendation: https://docs.mapbox.com/help/how-mapbox-works/attribution/


mapmeld:
Note on credits: you can hide the Leaflet label with ```map.attributionControl.setPrefix("")``` and credit it elsewhere

jameshadfield:
Thanks @trvrb 🙌

Issue #896, opened 2020-02-14
Blend branch coloring


trvrb:
Right now, a branch inherits tip color as shown here:

![Screen Shot 2020-02-13 at 4 13 15 PM](https://user-images.githubusercontent.com/1176109/74489925-c158fd00-4e7b-11ea-9d77-c1e02ceff83f.png)

However, this color is only really true at the right-most edge of the branch. Ie each of these tips is definitely in their place of sampling at the time of sampling but as we go back things should get more uncertain. In fact the left-most edge of the branch should be the same gray as the right-most edge of the parent branch. This should appear as a continuous process.

I actually think it shouldn't be too difficult to have a SVG gradient for each branch line (https://stackoverflow.com/questions/42874203/linear-gradient-across-svg-line). The gradient end points would be chosen as:
- right: current branch color
- left: parent branch color

I suspect this will result in a more natural, interpretable visualization.

jameshadfield:
Yes -- the way we draw branches can often be misleading like this (see discussion on https://github.com/nextstrain/auspice/issues/200 for our thoughts 2 years ago!). SVG computation / rendering is a sizeable bottleneck in animations, so perhaps we'd have to fall-back to the current method if this slows things down, but I don't see a problem there. [Seems to be pretty simple](http://bl.ocks.org/pnavarrc/20950640812489f13246) to implement in d3.

trvrb:
I didn't think of this animation issue. Thanks James!

jameshadfield:
Update -- this issue has a partial solution, but is in need of further work to complete, as detailed in https://github.com/nextstrain/auspice/pull/897#issuecomment-588014096. Contributions on top of this PR #897 are very welcome -- please leave a comment there if you start work on it!

cpsandbox:
Contributed pr #947 for this

cpsandbox:
updated pr #947

jameshadfield:
closed by the merge of #947

PR #897, opened 2020-02-17
Prototype of SVG gradient for tree branches

jameshadfield:
See #896 for intention.

image

This requires using SVG `rect`angles for the branch stem as gradients are overly complex to add to `path`s. As such, lots of things are broken (e.g. all layouts except rect), and some functionality (branch hover colour changes, branch click to zoom) is disabled here.

I think it looks really good, but wanted to discuss this before we commit to cleaning up the code so that previous functionality is restored. cc @trvrb @rneher

### DIFF
--- a/src/components/tree/phyloTree/layouts.js
+++ b/src/components/tree/phyloTree/layouts.js
@@ -370,8 +370,11 @@ export const mapToScreen = function mapToScreen() {
     this.nodes.forEach((d) => {
       const stem_offset = 0.5*(d.parent["stroke-width"] - d["stroke-width"]) || 0.0;
       const childrenY = [this.yScale(d.yRange[0]), this.yScale(d.yRange[1])];
+      // Note that a branch cannot be perfectly horizontal and also have a (linear) gradient applied to it
+      // So we add a tiny amount of jitter (e.g 1/1000px) to the horizontal line (d.branch[0])
+      // see https://stackoverflow.com/questions/13223636/svg-gradient-for-perfectly-horizontal-path
       d.branch =[
-        [` M ${d.xBase - stem_offset},${d.yBase} L ${d.xTip},${d.yTip}`],
+        [` M ${d.xBase - stem_offset},${d.yBase} L ${d.xTip},${d.yTip+0.001}`],
         [` M ${d.xTip},${childrenY[0]} L ${d.xTip},${childrenY[1]}`]
       ];
       if (this.params.confidence) {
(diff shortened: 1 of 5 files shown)


trvrb:
Fantastic work @jameshadfield! I'm super happy with this direction. I find the more continuous behavior to be more aesthetically pleasing, but I'd also note that people read a lot into these colors and (perhaps rightfully so) interpret an extended line of a particular color as arriving in that region at that moment, [like so](https://twitter.com/MaddyLove2018/status/1228858803803717632). Another good example of this would be the arrival of mumps into BC. The [solid red line here](https://nextstrain.org/mumps/na) makes it look like the arrival was 2013, when that's definitely not the case.

old new

I experimented with this with basically all the live builds and I think it's a consistent improvement. My only concerns at this point are:
1. Possible performance issues as you mentioned previously (might deserve a small bit of testing to confirm it's not horribly slow)
2. Loss of pleasing curvature at elbow joints in going from path to rect. (In original nextflu auspice [we did `polyline` for this](https://github.com/blab/nextflu/blob/master/auspice/js/auspice/tree.js#L330).) But I bet this can be worked out with thinking about placement and branch thickness.

jameshadfield:
Great -- I'll find time to restore the functionality so we can use this 👍 
Completely agree about the elbow joints & am confident we can find a solution here, I just didn't spend any time looking (e.g. rectangles can have rounded corners).

tsibley:
It seems like it'd be helpful to stick with paths instead of rects if possible? My understanding is that gradients are very non-trivial for curved paths, but that they are relatively straightforward for straight paths at any given angle using a gradient transform. It seems to me that the branches we want to apply gradients to on any of our layouts are straight paths?

trvrb:
I don't know best solution between `rect`, `polyline`, `path`, etc... I would note that the only curved lines we use are in the radial layout. These are equivalent to the vertical lines in the rectangular layout and don't get a gradient assigned.

jameshadfield:
Update:
* Rectangular trees now working with paths. 
* Branch hover updated to use gradients & the way I emphasize colours on hover has changed

To do:
* Gradient axis needs to be updated depending on branch direction -- currently it is only Left - right, which is why it doesn't look like it should on non-rectangular phylogenies
* Ensure no branches are rendered as vertical or horizontal as you cannot apply linear gradients to such paths (this was why I initially didn't use paths).

tsibley:
> Ensure no branches are rendered as vertical or horizontal as you cannot apply linear gradients to such paths (this was why I initially didn't use paths).

If the jittering of such paths is problematic, an alternate solution is to switch the coordinate system of the gradients into user space instead of using the path bounding box. I don't think this would be troublesome to specify since I think we already know the coordinates of the branch in user space. Jittering may be simpler though if it doesn't cause other issues.

enjalot:
I don't know if this is helpful, it seems like a complex solution but could be helpful:
https://bl.ocks.org/mbostock/4163057

another idea for the rectangular version would be to simply render a circle under the intersection to smooth things out a little bit

jameshadfield:
@enjalot all the paths to have gradient applied are straight, so "normal" SVG gradients work (and are working in this PR) as long as the line isn't perfectly vertical / horizontal.

jameshadfield:
PR supersceeded by https://github.com/nextstrain/auspice/pull/947

jameshadfield:
Closed by the merge of #947

Issue #899, opened 2020-02-20
poor legend display for long names


jameshadfield:
e.g. 
![image](https://user-images.githubusercontent.com/8350992/74978378-561db500-5491-11ea-8e75-a8b3447e4af8.png)

Issue #885, opened 2020-02-04
Improve Compatibility with Darkreader


emmahodcroft:
See email to hello@nextstrain from Henri on 3rd Feb 2020.

Some with low vision benefit from inverting website displays, so they have a black background and white text. Extensions like Darkreader aid with this. 

Using Darkreader with nextstrain.org/ncov renders like this:
![image](https://user-images.githubusercontent.com/14290674/73741873-75a3a680-474b-11ea-8086-c56d7be7983b.png)

Mostly ok - but the tip labels on the tree don't show up at all. The mouseover still works. I suspect this may be because the tree is rendered slightly differently from the rest of the page, so isn't responding to however this works?

Would be good to test a little more thoroughly to ensure everything else is working, too.

jameshadfield:
Tip labels are SVG text, which would be my guess at why they're not inverted.

joachimschmidt557:
This issue has been raised before with DarkReader. See https://github.com/darkreader/darkreader/issues/1963

The solution there was to create some exceptions for the site in question, ngrok.com. 

This could also be achieved on nextstrain.org with:

```css
.tipLabel {
    fill: #e8e6e3;
}
```

PR #912, opened 2020-03-01
Handle long names in tree legend

mapmeld:
Potential fix for #899 - long names in the legend of the tree view. I considered long labels with and without any whitespace

The current menu (SVG, absolute positions) is replaced with HTML + CSS.

Chrome-based browsers work best - in the middle left column you can see samples which carry over to a second line:

Screen Shot 2020-03-01 at 2 42 14 AM

Firefox and Safari, with identical settings, would move a long label without whitespace over to the next line:

Screen Shot 2020-03-01 at 3 02 13 AM

With some CSS for compatibility, Firefox and Safari are now level with the other labels. The downside is they will break a word if it overflows

Screen Shot 2020-03-01 at 4 09 17 AM

### DIFF
--- a/src/components/tree/legend/item.js
+++ b/src/components/tree/legend/item.js
@@ -4,7 +4,6 @@ import { dataFont, darkGrey } from "../../../globalStyles";
 
 const LegendItem = ({
   dispatch,
-  transform,
   legendRectSize,
   legendSpacing,
   rectStroke,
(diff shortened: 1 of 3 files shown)


jameshadfield:
Thanks @mapmeld -- legends with lots of entries are now being truncated (granted the previous implementation is problematic here too, but for different reasons):

image

Could we truncate the string being displayed such that it is broadly similar to current master but doesn't overflow horizontally, and perhaps display an info-box on hover?

jameshadfield:
@mapmeld I've merged #914 instead, but am going to tag #885 as moving to HTML text is needed eventually. Really appreciate your contribution, it's amazing to see the community helping out with auspice & nextstrain 🙌

Issue #887, opened 2020-02-04
Narrative tables (md or html) need styling


jameshadfield:
(Moved out of #883 for clarity)

@emmahodcroft:

I currently can't seem to get tables (either Markdown or HTML) working in the auspiceMainDisplayMarkdown part working. This might be nice for formatting lists of links with images, for example. Or lists of previous situation reports (or other languages)

@jameshadfield:

I presume HTML tables don't work because of https://github.com/nextstrain/auspice/blob/master/src/components/narrative/MainDisplayMarkdown.js#L136 (this code was copied from the footer markdown parsing). If markdown tables don't work then there may be an option we can add to https://github.com/nextstrain/auspice/blob/master/cli/server/parseNarrative.js#L13 to improve this.

_UPDATE: Thanks to @CameronDevine (#924) tables now display in `auspiceMainDisplayMarkdown` sections. These things are still needed:_

- [ ] The tables in `auspiceMainDisplayMarkdown` sections need to be styled appropriately
- [x] "Normal" (sidebar) narrative text should be able to display tables (low priority)

jameshadfield:
To reproduce / work on a fix:

* Install auspice from source, as per https://nextstrain.github.io/auspice/introduction/install or similar
* Obtain example datasets + narratives via `npm run get-data && npm run get-narratives`
* Add markdown tables to a narrative which is now present in `./narratives`
* Test via `auspice develop --verbose`

CameronDevine:
I was able to get tables in the `auspiceMainDisplayMarkdown` section working. However, I think some CSS work may also be necessary. As this is not my area of expertise, I will leave this work to someone else.

jameshadfield:
Can't seem to get this working - could you paste an example narrative markdown file (or section of) which shows this working?

CameronDevine:
@jameshadfield Tables in the sidebar also work (This was working before #924). For example:

```
### Further Reading:

| test | table |
| --- | --- |
| 1 | 2 |
| 3 | 4 |
```

![image](https://user-images.githubusercontent.com/10394896/76251331-14686a80-6204-11ea-9f39-9c2c8f9df37e.png)

joachimschmidt557:
I would be happy to work on styling tables.

jameshadfield:
Closed by #966 -- thanks @joachimschmidt557 🎉

PR #919, opened 2020-03-05
[documentation] Fix typos + update fixture data download instructions

hydrosquall:
## Summary

- Ran both welcome documents through basic spellcheck.
- Updated the welcome instructions to download data to the `data/` folder, which matches the behavior of the `get-data` script. This will reduce the number of PRs where someone accidentally commits the `datasets/` folde,  if they were following the README.md.
- Didn't know about `npm run get-data` until seeing the directions in https://github.com/nextstrain/auspice/issues/887#issuecomment-594429320, seems link a command that would be useful to document in either `README.md` or `CONTRIBUTING.md` as it would lower the startup cost for new developers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@ title: Contributing to Auspice
 ---
 
 
-> We have recieved a number of generous offers to contribute developer effort to nextstrain (and auspice).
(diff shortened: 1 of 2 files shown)


jameshadfield:
wonderful - thanks @hydrosquall ⭐️

PR #920, opened 2020-03-05
[feature] Improve smoothness of temporalWindow animation over short time frames

### LINKED ISSUE #918: nCoV animation should be smooth
This is due to us hardcoding some minimum amount of time to be shifted each tick. The tree is small enough that it should be animating super smoothly.

hydrosquall:
## Motivation

- Fixes #918, which described a bug with temporal animations in the tree view. Users would get choppy results when the time interval spanned months or shorter instead of years.

## Changes

- Video with updated performance using `d3.transition` instead of destroying/creating the rectangles on every iteration of `addTemporalSlice`: https://a.cl.ly/jkuKeJn7
- Rewrote rectangle movement to use CSS transforms instead of setting the x and y attributes
- Adjusted the "transitioned" set of properties to the minimal set (x position), since it's simpler if height/width don't need to be modified on every frame.

## Testing

```bash
mkdir data
curl http://data.nextstrain.org/ncov.json --compressed -o data/ncov.json
curl http://data.nextstrain.org/zika.json --compressed -o data/zika.json
node auspice view --datasetDir data
```

Feel free to test with other layouts or datasets, I tested exclusively with `zika` and `nkov` using the links provided by @trvrb [here](https://github.com/nextstrain/auspice/issues/918#issuecomment-594968336).

- http://localhost:4000/zika?animate=2005-12-15,2019-07-06,1,0,30000&p=grid
- http://localhost:4000/ncov?animate=2019-12-12,2020-03-06,0,0,30000&d=tree&p=full

### DIFF
--- a/src/components/tree/phyloTree/grid.js
+++ b/src/components/tree/phyloTree/grid.js
@@ -357,7 +357,7 @@ export const temporalWindowTransition = transition('temporalWindowTransition')
  */
 export const showTemporalSlice = function showTemporalSlice() {
   if (this.layout !== "rect" || this.distance !== "num_date") {
-    this.groups.temporalWindow.style('visibility', 'hidden');
+    this.hideTemporalSlice();
     return;
   }
 
@@ -381,7 +381,7 @@ export const showTemporalSlice = function showTemporalSlice() {
       translateX_startRegion = xWindow[0];
     }
 
-    const wasStartRegionVisible = this.groups.temporalWindowStart.attr('opacity') === 1;
+    const wasStartRegionVisible = this.groups.temporalWindowStart.attr('opacity') === '1';
 
     this.groups.temporalWindowStart
       .attr('opacity', 1)
@@ -405,31 +405,36 @@ export const showTemporalSlice = function showTemporalSlice() {
   let xStart_endRegion = xWindow[1]; // starting X coordinate of the "end" rectangle
   let width_endRegion = totalWidth - this.params.margins.right - xWindow[1];
 
+  let transform_endRegion = `translate(${totalWidth - this.params.margins.right},0) scale(-1,1)`;
   // With a right hand tree, the coordinate system flips (right to left)
   if (rightHandTree) {
     xStart_endRegion = this.params.margins.right;
     width_endRegion = xWindow[1] - this.params.margins.right;
+    transform_endRegion = `translate(${xStart_endRegion},0)`;
   }
 
   if (width_endRegion > minPxThreshold) {
-    const wasEndRegionVisible = this.groups.temporalWindowEnd.attr('opacity') === 1;
+    const wasEndRegionVisible = this.groups.temporalWindowEnd.attr('opacity') === '1';
 
     this.groups.temporalWindowEnd
       .attr('opacity', 1)
       .attr("height", height)
-      .attr("fill", fill);
+      .attr("fill", fill)
+      .attr("transform", transform_endRegion);
+
+
     // Only apply animation if rectangle was already visible in the previous frame.
     // Unlike the startingRegion, this panel cannot depend
     // on letting the SVG boundaries clip part of the rectangle.
-    // As a result, we'll have to animate width AND height.
+    // As a result, we'll have to animate width instead of position
+    // If performance becomes an issue, try add a custom clip-path with
+    // a fixed-width region instead.
     if (wasEndRegionVisible) {
       this.groups.temporalWindowEnd
         .transition('temporalWindowTransition')
-        .attr("transform", `translate(${xStart_endRegion},0)`)
         .attr("width", width_endRegion);
     } else {
       this.groups.temporalWindowEnd
-        .attr("transform", `translate(${xStart_endRegion},0)`)
         .attr("width", width_endRegion);
     }
   } else {
--- a/src/components/tree/phyloTree/grid.js
+++ b/src/components/tree/phyloTree/grid.js
@@ -1,7 +1,9 @@
 /* eslint-disable space-infix-ops */
 import { min, max } from "d3-array";
+import { transition } from "d3-transition";
+import { easeLinear } from "d3-ease";
 import { timerStart, timerEnd } from "../../../util/perf";
-import { months } from "../../../util/globals";
+import { months, animationInterpolationDuration } from "../../../util/globals";
 import { numericToCalendar } from "../../../util/dateHelpers";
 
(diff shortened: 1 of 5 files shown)


src/components/tree/phyloTree/grid.js
```diff
@@ -334,17 +342,20 @@ export const addGrid = function addGrid() {
   timerEnd("addGrid");
 };
 
-
-export const removeTemporalSlice = function removeTemporalSlice() {
```
jameshadfield: This removed function is still set as a prototype in `phyloTree.js` and called by the function `modifySVGInStages`. This will result in a complete app crash when (e.g.) the tree layout it changed

src/components/tree/phyloTree/grid.js
```diff
@@ -188,7 +196,7 @@ const computeYGridPoints = (ymin, ymax) => {
  */
 export const addGrid = function addGrid() {
   const layout = this.layout;
-  addSVGGroupsIfNeeded(this.groups, this.svg);
+  addSVGGroupsIfNeeded(this.groups, this.svg, this.temporalWindow);
```
jameshadfield: `addSVGGroupsIfNeeded` still takes only 2 arguments, but 3 are provided here

src/components/tree/phyloTree/grid.js
```diff
@@ -357,25 +368,29 @@ export const addTemporalSlice = function addTemporalSlice() {
...
-      .attr("y", 0)
+    const xEnd_start = rightHandTree ? totalWidth - xWindow[0] : xWindow[0]; // ending X coordinate of the "start" rectangle
+    const translateX_start = -totalWidth + xEnd_start; // translation distance for starting temporalWindow
+    this.groups.temporalWindowStart
       .attr("height", height)
-      .attr("fill", fill);
+      .attr("width", totalWidth)
+      .attr("fill", fill)
+      .transition('temporalWindowTransition')
+      .attr("transform", `translate(${translateX_start},0)`);
```
jameshadfield: Cool! Am i right in understanding that on each update, we modify the rect's position via the transform attr, and in-between transforms the transition smooths things?

src/components/tree/phyloTree/grid.js
```diff
@@ -357,25 +368,29 @@ export const addTemporalSlice = function addTemporalSlice() {
 
   /* the gray region between the root (ish) and the minimum date */
   if (Math.abs(xWindow[0]-rootXPos) > minPxThreshold) { /* don't render anything less than this num of px */
-    this.groups.temporalWindow.append("rect")
-      .attr("x", rightHandTree ? xWindow[0] : 0)
-      .attr("width", rightHandTree ? totalWidth-xWindow[0]: xWindow[0])
-      .attr("y", 0)
+    const xEnd_start = rightHandTree ? totalWidth - xWindow[0] : xWindow[0]; // ending X coordinate of the "start" rectangle
```
jameshadfield: It's important to address the case of two-trees, where the x-axis is flipped. Currently (e.g.) localhost:4000/ncov/2020-02-11:ncov/2020-02-09?dmax=2020-01-26&m=num_date is broken as somewhere in this PR (probably not this line) we are assuming x-values are left-to-right

src/components/tree/phyloTree/grid.js
hydrosquall: Oops, I didn't realize this was used on the outside, I should have checked given that it was `export`ed. I'll bring this back.

src/components/tree/phyloTree/grid.js
hydrosquall: I had tried an alternative solution which had involved modifying the signature, and accidentally left this in - good catch.

src/components/tree/phyloTree/grid.js
hydrosquall: Correct, this is in contrast to the current behavior of creating a new rectangle on every frame.

Alternately, we have the option of fixing the position, and instead modifying the width on each frame. However, modifying position is cheaper to animate than modifying the width.

hydrosquall:
Thanks for the thorough feedback @jameshadfield ! I've moved your comments into a checklist for you or another reviewer to cross-validate my initial results. I believe the latest changes solve all the regressions identified, but please let me know if you end up finding a different result in your local testing.

- [ ] Two trees don't work appropriately (the x-axis coords can get confusing here) -- e.g. load two ncov trees (see localhost link in line-comment) and modulate the date sliders (regression).

I didn't know how to test this earlier, but I opened this up side-by-side with `nextstrain.org` with the help of your link, and think I have something matching now. Please let me know if you notice more differences.

- [ ] Sliding the max-date slider to the end doesn't clear the grey shading (regression).

This should be fixed as of the latest version, I added code to hide the regions if they don't pass the "minimum width" test.

- [ ] Clearing the max date filter (via the "x" in the info box in the header of the page) doesn't clear the grey shading (regression).

I think this should be fixed now.

- [ ] Sometimes, when the animation is "reset" (i.e. stopped), the grey regions remain, when they should disappear (regression).

I wasn't able to reproduce this, can you let me know if you still observe this one?

- [ ] Behavior at the start of /zika animation is weird when the left grey area comes into view. Didn't happen with the ncov dataset.

I suspect this was one of the animations was animating into its initial position instead of appearing instantly, I've added some custom logic to avoid doing the interpolation when the left gray area first appears. Let me know if this reoccurs.

- [ ] Sometimes (happens for me with /zika, but not /ncov) the grey areas remain at the end of the animation (regression).

I think the issue was related to a change in the visibility toggling logic, I think it should be better now (I've been testing in both datasets).

- [ ] Performance will need to be tested (but wouldn't expect this to add much overhead)

To handle the edge cases at the start/end of the animation, I had to animate width for the `endRegion` panel instead of just transforming the x position as I had originally hoped. After I did this, performance is slightly worse than when I first opened this PR, and you can see some animation slowdown in the busiest parts of the `/zika` tree.

If it's too much of a slowdown, there's another technique we can try (applying a clip path and using a fixed width panel). The problem is that we would need to figure out when to regenerate the clip path, so didn't want to pursue that approach until it becomes absolutely necessary. I didn't notice a performance degradation on the `ncov` animation.

jameshadfield:
Quick note to say thanks for this & that I'll try to test / merge this as soon as possible!

hydrosquall:
No worries/rush @jameshadfield ! Let me know if there are issues and I'm happy to go through whatever rounds of iteration are necessary.

Issue #900, opened 2020-02-20
Add legend open/close state to URL query


jameshadfield:
Will be very beneficial to iframe embedding and narratives

emmahodcroft:
@jameshadfield you are a mindreader, I had a rant about this just the other day and it's on my 'auspice wish list' door, but I never got around to making an issue 😆

jameshadfield:
The best example to follow for this would be to follow how we currently store the state for being able to control the sidebar open/closed state via the URL query (see https://github.com/nextstrain/auspice/search?q=sidebarOpen&unscoped_q=sidebarOpen). Please note that it's not applicable for this to be set via the dataset JSON. 

The default state for the legend open/closed should be set as it currently is, and any user-interaction will result in a URL query (via the redux middleware). For instance `?legend=open` or `?legend=closed`.

ncallaway:
I'm starting work on this now, I expect I'll be able to have a PR for you shortly.

jameshadfield:
Closed by #923

PR #923, opened 2020-03-05
Issue #900: Load legend state from URL

ncallaway:
This should resolve issue #900  by loading the legend state from the URL.

**Notes**
* This adds a new state parameter to the redux-control `legendOpen`, which behaves similarly to `sidebarOpen`
* This preserves existing behavior where the legend changes its default behavior based on the size of the browser window, or the number of items in the legend.
** If the default due to the browser window size would be closed, but the URL parameter is `open`, then the URL parameter wins.
* This adds one additional error to the lint output (we go from 161 problems, 143 errors up to 162 problems, 144 errors). Because of the number of eslint errors, I haven't been able to find the one issue that I introduced. However, the newly introduced lint failure is autofixable with the `--fix` option. Because there are `96` errors that will be fixed with the `--fix` option, I want to leave that out of this PR. However, I can make a follow-up PR that runs lint with the `--fix` option.

**PR Questions**
* Should we also persist changes to the legend state back to the URL (the same way the `c` parameter gets set and updated in the URL)?
* Is there anywhere that URL parameters that would be useful for embedding or external consumers are documented? Where can I add documentation about this feature?


jameshadfield:
Thanks @ncallaway - this is excellent! I've merged this with two additional commits (both of which you mentioned in your comment!)

1. d9628c64a819699f160e6ab895753d8ad14753cc removes the URL query upon user interaction, so that the query + app state don't diverge. In the future we may explore changing the URL query upon all user interactions, but for now this is great. 

2. 014b8396272523954b1b4462c7645c20dce4ec82 adds documentation to https://nextstrain.github.io/auspice/advanced-functionality/view-settings (or, it will, once I rebuild the docs!) 

Re: Linting errors. These are a long-term endeavor to improve. I get them flagged up in VS code and have been slowly fixing them as time allows. PRs would be great!

Thanks again 🙏

ncallaway:
@jameshadfield That looks great. If I make any other changes I'll make sure to update the documentation.

I've made the PR that runs the autolinting here: https://github.com/nextstrain/auspice/pull/931.

After running it, I did find the linting error that this introduced and saw that its fixed in that PR.

tsibley:
@ncallaway A little late in this instance, perhaps, but for whatever it's worth, I wrote a small tool to help me find eslint errors I introduce into projects which aren't eslint-clean yet: https://github.com/tsibley/git-eslint. Not great, but gets the job done. :-)

PR #924, opened 2020-03-06
Added table tags to narrative main display allowed tags list.

CameronDevine:
Some CSS work is still required to make the tables look nice.

This is a partial solution to #887


jameshadfield:
Can't seem to get this working - could you paste an example narrative markdown file (or section of) which shows this working?

PR #927, opened 2020-03-06
Only truncate legend labels if they are strings

### LINKED ISSUE #926: Color by legend is broken for numeric traits
Selecting a color by option that has numerical values produces an error that "something has gone wrong". For example, selecting "Age" in the novel coronavirus build (https://nextstrain.org/ncov?c=age) produces this output:

image

Using git bisect, I tracked this error down to the [truncation of legend labels using the substring method](https://github.com/nextstrain/auspice/commit/07d94a4434fec69014bf6f5a828d37d266d4d469#diff-caeef2e2c6f35f6443be8162ba918ba8R38).

I'm not familiar enough with React to implement a solution here, but it seems like we need to check the type of the label prior to applying substring.

huddlej:
Fixes a bug where the tree legend could not be displayed for numeric color by
values because those values do not have a substring function.

Closes #926.

### DIFF
--- a/src/components/tree/legend/item.js
+++ b/src/components/tree/legend/item.js
@@ -35,7 +35,7 @@ const LegendItem = ({
       style={{fontSize: 12, fill: darkGrey, fontFamily: dataFont}}
     >
       <title>{label}</title>
-      {label.substring(0, legendMaxLength)}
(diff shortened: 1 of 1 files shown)


emmahodcroft:
Amazing, @huddlej ! Thank you for catching & fixing this!

PR #928, opened 2020-03-07
[webpack] Transpile non-es5 third party dependencies

### LINKED ISSUE #902: Auspice extensions are not being (fully) transpiled, causing ES6 code to be leaked into the bundle.
This issue affect both nextstrain.org and auspice.

## Description

Recently a dependency to 'iso-639-1' package was introduced in nextstrain.org, required for language picker functionaly, implemented as a part of nextstrain's extension to auspice. This uncovered a defect in auspice extension system. 'iso-639-1' package contains untranspiled ES6 classes, and they go (minified) straight into the bundle. Due to dynamic nature of extensions, it seems that either they are not being transpiled at all, or at least, their dependency packages aren't.

Basically, current users of extension system have to ensure that their modules are compatible with their desired target browsers. one way is to transpile them separately.

## Attempted fix

Adding `.browserslistrc`

```
[development]
Last 2 versions

[production]
cover 99%
ie >= 9

```

and whitelisting `iso-639-1` for transpilation in `webpack.config.js`:

```js

  /* which directories should be parsed by babel and other loaders? */
  const directoriesToTransform = [
    path.join(__dirname, 'src'),
    path.join(__dirname, 'node_modules/iso-639-1'),
  ];

```

somehow did not work.

I still only understand the extension system vaguely, and is super-puzzled right now.

## Context

This was uncovered as another "neststrain.org is broken Internet Explorer again" issue, but this time it is a bit deeper than usual and also affects at least all users with browsers that do not support ES6 classes, which are all the browsers prior to ~2016-2017, which corresponds to ~7% of global share (for instance, see caniuse.com/#feat=es6-class).

Note how Firefox 45 is listed as green on caniuse for ES6 classes, but still fails to render, probably due to other untranspiled code fragments in the bundle.

Here are the screenshots of IE10, Chrome 54 and Firefox 45 (versions are chosen semi-randomly):

This issue affects both nextstrain.org and auspice.

## Description

Recently a dependency to ['iso-639-1' package](https://github.com/meikidd/iso-639-1) was introduced in nextstrain.org, required for language picker functionality, implemented as a part of nextstrain's extension to auspice:
[nextstrain/nextstrain.org/auspice/client/languageSelector.js#L2](https://github.com/nextstrain/nextstrain.org/blob/34a468ae9615b3e5ae4dde7662b8163fcd09d2e4/auspice/client/languageSelector.js#L2)

This uncovered a defect in auspice extension system. The `iso-639-1` package distribution contains transpiled ES5 bundle and original ES6 modules:
https://github.com/meikidd/iso-639-1/blob/master/package.json#L5-L6

Auspice's webpack somehow picks ES6 code and fails to transpile it, so that ES6 classes (in particular, `class` keyword), go (minified) straight into the bundle.

This could be due to:
 - extensions are not being transpiled at all
 - only their dependency packages are not being transpiled 
 - both of the above

Now, normally, dependencies are excluded in `babel-loader`, but one may need to selectively whitelist some of the packages, if package module target does not meet the requirements. I tried to whitelist it, in auspice, but that did not work (see below).

Note that in the case of nextstrain.or we know what to whitelist because the auspice and extension come from the same team. I am sure we will find a way to patch it. However, for a third party exstensions, auspice can not know (at runtime) which packages are being used by the extension, therefore it can not know what to whitelist.

Basically, current users of the extension system have to ensure that their modules are compatible with their desired target browsers. One way is to transpile the modules as a separate step.

## Attempted fix

Adding `.browserslistrc`

```
[development]
Last 2 versions

[production]
cover 99%
ie >= 9

```

and whitelisting `iso-639-1` for transpilation in `webpack.config.js`:

```js

  /* which directories should be parsed by babel and other loaders? */
  const directoriesToTransform = [
    path.join(__dirname, 'src'),
    path.join(__dirname, 'node_modules/iso-639-1'),
  ];

```

That somehow did not work. `class` keyword is still in the bundle.

I still only understand the extension system vaguely, and is puzzled right now.

## Context

This was uncovered by @emmahodcroft  as another "nextstrain.org is broken in Internet Explorer again" issue, but this time it is a bit deeper than usual and also affects at least all the users with browsers that do not support ES6 classes, which are all the browsers prior to ~2016-2017, ~(7% of global share, see [caniuse.com/#feat=es6-class](https://caniuse.com/#feat=es6-class)).

Note how Firefox 45 is listed as green on caniuse for ES6 classes, but still fails to render, probably due to other untranspiled code fragments appearing in the bundle later on.

hydrosquall:
## Summary

This a pass at using the `are-you-es5` library (https://github.com/obahareth/are-you-es5 ) to determine which of the current dependencies need to be transpiled to ES5 to support older browsers. It appears that we have some dependencies that don't have non-ES5 entrypoints even before getting the extension system involved. 

If this works, we could apply the `are-you-es5` script programmatically to the extensions directory, and can in the longer term help with #902 .

The fuller solution would be to run this script as a postinstall as described in [this article](https://developer.epages.com/blog/coding/how-to-transpile-node-modules-with-babel-and-webpack-in-a-monorepo/), but I wanted to first confirm whether transpiling these specific modules actually solved the issue in older browsers (which I currently don't have access to).

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -112,6 +112,16 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
             cwd: path.resolve(__dirname)
           }
         },
+        // To regenerate the regular expression for selecting non-ES5 dependencies, run:
+        // npx are-you-es5 check . -r | tail -n 1
+        {
+          test: /\.js$/,
+          loader: 'babel-loader',
+          exclude: /[\\/]node_modules[\\/](?!(chalk|express-naked-redirect|node-fetch|typeface-lato)[\\/])/,
+          options: {
+            cwd: path.resolve(__dirname)
+          }
+        },
         {
           test: /\.css$/,
           use: ["style-loader", "css-loader"]


jameshadfield:
Thanks @hydrosquall! We currently use modules such as `chalk` only in the auspice server code, which should never be transpiled, as they're never run in-browser (webpack is only used for the auspice client). Is there a way to make webpack only follow the imports which trace back to the [entry](https://github.com/nextstrain/auspice/blob/52798801186ccd31519d6afd90831051091f962b/webpack.config.js#L68)?

hydrosquall:
Hi @jameshadfield , I see! In that case, then I see how the commit I added wouldn't affect the browser bundle, since all of the excluded dependencies (`node-fetch`, `express-naked-redirect`, etc) are just for the server. I'll close this PR, thanks for taking a look.

From my reading of @obhareth's comment [here](https://github.com/obahareth/are-you-es5/issues/25#issuecomment-596087426), it looks like this library either checks `package.json` or `node_modules`, and doesn't necessarily check the runtime-level imports. 

I wonder if moving the server-only dependencies to `devDependencies` instead of `dependencies` would make `are-you-es5` skip these server-side only libraries- probably something to investigate in a separate issue.

jameshadfield:
I haven't dug deep into webpack since I wrote auspice's extension machinery, but I think a good audit of exactly which dependencies are getting pulled into auspice is needed.

What's desired is that webpack (via babel) transpiles
1. all code imported by the entrypoint (i.e. the client), and doesn't touch the packages in `node_modules` which are only used by the auspice server
2. all code imported by client extensions (but not other modules within the `node_modules` in the extension directory). This is complicated by the fact that if an extension depends on `react` then we want to ensure we are using auspice's version of react!

jameshadfield:
Thanks for your help @hydrosquall, we really appreciate it!

hydrosquall:
> all code imported by `entrypoint` || `client_extensions`

It would be more involved, but I suspect that we could use one of these libraries like this to check imports at runtime, then pass the imported listed of 3p dependencies to the "check if es5" functionality of `are-you-es5`: https://github.com/dyatko/arkit#relevant-projects

I won't be able to pursue this extra search in the near term, so if anybody else reading this feels inspired, I have some resources about working with JS dependency graphs that may be useful: https://github.com/hydrosquall/code-maps-frontend

PR #930, opened 2020-03-09
Search highlights tips as you type

### LINKED ISSUE #929: Highlight tips on strain search
Quite often it's helpful to just highlight some strains in the tree without doing a full "filter". This has traditionally been accomplished by mousing over a legend item. For example, go here: https://nextstrain.org/ncov?c=country and mouseover "USA" to get:

mouseover

with USA tips increasing in size to be more obvious.

However, with additional items in the legend, it's difficult to have this interaction. For example, with default "division" coloring, we have:

division

where many of the categories are unreachable because of the count of legend items.

We already have a "strain search" that matches live to multiple strains:

strain-search

The feature request is to make tips in the tree highlight the way that legend mouseover highlights for all strains that match when typing into "strain search".

The current behavior of filtering to a single strain when clicking on a strain search should be maintained.

This would really help with the sorts of screen shot interpretations I'm trying to provide. Thank you! 🙏

mapmeld:
Fixes #929 
As you are typing in the "Search Strains" autocomplete, it will highlight the tips of matching strains. Just like the autocomplete, the highlight goes away if you have 0-1 characters, or pick one of the nodes

Screenshot:
Screen Shot 2020-03-08 at 11 15 23 PM

### DIFF
--- a/src/actions/tree.js
+++ b/src/actions/tree.js
@@ -204,7 +204,7 @@ export const changeDateFilter = ({newMin = false, newMax = false, quickdraw = fa
  * @return {null} side-effects: a single action
  */
 export const updateTipRadii = (
-  {tipSelectedIdx = false, selectedLegendItem = false} = {}
+  {tipSelectedIdx = false, selectedLegendItem = false, searchNodes = false} = {}
 ) => {
   return (dispatch, getState) => {
     const { controls, tree, treeToo } = getState();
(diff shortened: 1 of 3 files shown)


jameshadfield:
@mapmeld this is outstanding. Tested and working perfectly ⭐️

trvrb:
Hugely appreciated. Thank you!

Issue #818, opened 2019-10-30
add legend to map if tree not displayed


jameshadfield:
Similar to #817, when a map is displayed without a tree, the legend should be displayed to indicate the meaning of colours.

ncallaway:
I'm currently coordinating with another developer on the drafting of some mockups for this issue.

colinmegill:
@ncallaway really nice! I like the idea of a timeline across the bottom a lot. That will work in the case of:
* mobile, when the sidebar is closed
* narrative, when there is no sidebar or tree to offer a time axis
* divergence, when tree x axis is mutations

![divergence](https://user-images.githubusercontent.com/1770265/76059122-0d173780-5f4c-11ea-80a3-74abf447dc2f.gif)

ncallaway:
@jameshadfield @colinmegill I can start implementing today or tomorrow if the design works. Only thing I need to know is: with or without padding? If with padding, should that padding also be applied to the tree view?

jameshadfield:
Brilliant, thanks @ncallaway 

There is a reason why padding shouldn't be introduced in the tree: the legend sometimes occludes parts of the tree, and we want to work on making this better (e.g. see screenshot of current ncov dataset in divergence view & note that already we can't hover/click on the top-most tips [1]). 

Your mocks don't show the current color-by name and chevron, which are part of the legend on the tree & should be used for the map also. This title has the effect of adding padding, although there is technically none. This is what I think we should do for the map - that is, no padding but with a title+chevron above the legend entries. My default is no horizontal padding, but I'll leave that up to you as to whatever looks best.

> This just indicates one spot where the animation timeline controls could live when that future issue is implemented.

Like this a lot.

>  This option just indicates that the map legend will not be visible when the tree view is visible.

This seems sensible. The exception is that if the view is in "full" mode (i.e. tree on top of map) then the map should show a legend, as to view it you've scrolled away from the map's legend. Happy for this to come later / separate PR as needed.


Also note that #923 shifted the legend open/close state into redux, and this could be leveraged (with minor modifications) to keep things in sync. At the very least the beginning open-close state of the map legend should check this value (which can be set via URL query), as it does in the tree.

Thanks for all your work here 😄 

---

[1] Legend obscuring tips
![image](https://user-images.githubusercontent.com/8350992/76255306-d8401300-62b2-11ea-8b88-31cfc7521246.png)

ncallaway:
[...]

PR #935, opened 2020-03-10
Issue 818: Add legend to map view

ncallaway:
This adds the legend to the map view as described in #818.

**Notes**
* The legend is displayed:
  * If the tree view is not displayed, OR
  * If the tree view and map view are displayed "Full" such that they are stacked.
* The map legend state is stored separately from the tree legend state, so the legend of one view does not impact the other.
* Toggling _any_ legend state removes the parameter from the URL (whether you toggle the `map` legend or the `tree` legend)

**PR Questions**
* The thing I'm least happy about is how the `<Legend>` gets tied to the different redux state
  * The `Legend` is taking a `for` parameter, which is expected to be one of `tree` or `map`
  * The `Legend` uses that parameter to decide which legend control state variable to use from redux
  * That variable selection must match the logic in the reducer that sets the two variables.
  * An alternative I considered was having the legend state variable be an object with keys that would match the `for` parameter, which would reduce the logic coupling, but make the state slightly more complicated.
  * If you have a preference for the alternate implementation, let me know and I can make the change. Or, if you have a better suggestion, I'm happy to make any changes there.
* I checked the documentation that was added in the PR #923 and it seemed generic enough to not require a change with this addition. Let me know if you'd like me to make any changes there.

### DIFF
--- a/src/components/main/index.js
+++ b/src/components/main/index.js
@@ -77,6 +77,14 @@ class Main extends React.Component {
   toggleSidebar() {
     this.props.dispatch({type: TOGGLE_SIDEBAR, value: !this.props.sidebarOpen});
   }
+
+  shouldShowMapLegend() {
+    const showingTree = this.props.panelsToDisplay.includes("tree");
+    const inGrid = this.props.panelLayout !== "grid";
+
+    return !showingTree || inGrid;
+  }
+
   render() {
     if (this.state.showSpinner) {
       return (<Spinner/>);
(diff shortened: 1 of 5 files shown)


jameshadfield:
I've merged this in for release with one extra commit:

971ee320a409506916cc360eee94b124779b594d: This removes the "for" parameter, resulting in the open/close state of the legends being in-sync. I consider this a feature not a bug. Additionally, the code is simpler & the URL query for "legend" works again.

Thanks for the work here Noah, this is really nice 👍

ncallaway:
👍 Sounds great. Sorry I've gone a little dark on nextstrain work for the last couple of weeks. Work-work has been really busy, but I hope to be back picking up some tasks in the next week or so.

I'm also interested in helping with NextTrace in any way, so if there's anything y'all need help with there, let me know.

Issue #933, opened 2020-03-09
Extend narrative parsing to include translators


jameshadfield:
Currently the YAML frontmatter at the start of each narrative markdown file is parsed [by this code](https://github.com/nextstrain/auspice/blob/master/cli/server/parseNarrative.js#L23) into HTML to be rendered by the client as the fist page of the narrative. E.g. [this YAML](https://github.com/nextstrain/narratives/blob/master/ncov_sit-rep_2020-03-05.md) tuns into [this narrative page](https://nextstrain.org/narratives/ncov/sit-rep/2020-03-05).

Currently we display YAML fields `title`, `authors`, `authorLinks`, `affiliations`, `date` and `updated`.

Given the fantastic work by the community to translate the nCoV narratives, we should extend this to also parse: `translators` -> array of strings, each containing an optional URL or email. E.g. the current Spanish translation would include:

```yaml
translators:
  - Julián Villabona-Arenas <https://twitter.com/Chjulian>
```

And auspice should display this as a line below the author affiliations

ncallaway:
If no one else has started working on this, I'll take a look at this tonight.

jameshadfield:
Closed by #937

Issue #934, opened 2020-03-09
Extend narrative authors field to use arrays


jameshadfield:
Currently the YAML frontmatter at the start of each narrative markdown file is parsed [by this code](https://github.com/nextstrain/auspice/blob/master/cli/server/parseNarrative.js#L23) into HTML to be rendered by the client as the fist page of the narrative. E.g. [this YAML](https://github.com/nextstrain/narratives/blob/master/ncov_sit-rep_2020-03-05.md) tuns into [this narrative page](https://nextstrain.org/narratives/ncov/sit-rep/2020-03-05).

Currently we display YAML fields `title`, `authors`, `authorLinks`, `affiliations`, `date` and `updated`.

Currently `authors` and `authorLinks` must be strings. This issue asks that `authors` be extended to an array such as

```yaml
authors:
  - Author 1 Name <author URL>
  - Author 2 Name <author URL>
```

And in this case, `authorLinks` will not be used (for backward compatablity we should maintain the current behavior of allowing `authors` to be a string and, if so, using `authorLinks` if it is also a string. 

The Auspice client should then display them similar to now but with a series of in-line links. 

I'm not sure how `affiliations` should behave in this setting, as we also want it linked to authors. Ideas welcome!

jameshadfield:
Will be closed by #937

PR #937, opened 2020-03-11
Issue 933: Add translators to narrative parsing

ncallaway:
This adds translators to the narrative frontmatter parsing as described in #933.

**Notes**
*  I noticed there was a warning for `authors` as arrays not being implemented yet. I decided that the best path forward would be to implement `authors` as arrays or strings, then share that code with the `translators` parsing. The end result is that `authors` and `translators` have the same semantics, and can be given as either a single string, or an array.
* While the issue doesn't specify it, sharing the same code means `translators` also has a `translatorLinks` key, which can be used to link to translator (a personal blog, mailto, twitter account, etc). `translatorLinks` are not necessary, and can be omitted for any/all translators.
* This adds validation around the links.
  * If the `authors` attribute is a string, then `authorLinks` must be a string. If `authors` is an array, then `authorLinks` must be an array of the same size (`null` can be used for any element that doesn't have a link).
  * The same is true for `translators` and `translatorLinks`.
  * If the validation fails, the link is omitted and a warning message is displayed that explains specifically what failed the validation.
* The only documentation I found on `authors` and `authorLinks` appeared to be in a tutorial. I didn't see reference documentation. Because I didn't want to bog down a tutorial with a more obscure attribute, I didn't add documentation for this. If there’s reference documentation anywhere point me at it and I can the documentation for these new attributes.

### DIFF
--- a/cli/server/parseNarrative.js
+++ b/cli/server/parseNarrative.js
@@ -28,18 +28,19 @@ const makeFrontMatterBlock = (frontMatter) => {
   const markdown = [];
   markdown.push(`# ${frontMatter.title}`);
   if (frontMatter.authors) {
-    if (typeof frontMatter.authors === 'object' && Array.isArray(frontMatter.authors)) {
-      utils.warn(`Narrative parsing -- can't do author arrays yet`);
-    } else if (typeof frontMatter.authors === 'string') {
-      if (frontMatter.authorLinks && typeof frontMatter.authorLinks === "string") {
-        markdown.push(`### Author: [${frontMatter.authors}](${frontMatter.authorLinks})`);
-      } else {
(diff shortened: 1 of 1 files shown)


jameshadfield:
Hi @ncallaway -- thanks again for more great work here. This is really fantastic and allows us to credit all the contributions from translators. Breaking into functions is very wise. I've merged with one extra commit which addresses linting errors (c24aef6aa4fa365617d644904d17627289922cd9). 

> implement authors as arrays or strings, then share that code with the translators parsing. 

This is great. I had separated this out into a separate issue (#934) which this PR will now close 🎉 

> The only documentation I found ...

Yeah - Unfortunately documentation is somewhat lacking for this (and not only this!). If you have ideas of how you think this should be done please let us know (via an issue).

Thanks again!

Issue #264, opened 2017-03-15
Tooltips on map


colinmegill:
- [ ] nodes
- [ ] branches

jameshadfield:
Previous discussion (#266) on what node (deme) hover / click behavior should be:

@trvrb 
>I think that mousing over map demes should act just like mousing over legend items. Legend items don't click to filter. Clicking on map demes should also not filter.

@jameshadfield 
>we should be able to click on the legend icons to filter. For discrete traits this is identical to the filters as they exist. For continuous variables, we already create a mapping of the (continuous) variable to a discrete number of ranges, each of which is represented as a single colour in the legend. I propose clicking this should filter to that range. With the new info / title paragraph this can be effectively conveyed to the user.

>The map demes should always represent geographical filters - i.e. even if they are coloured by some other variable (e.g. date), clicking on a country should filter to that country.

kumikokashii:
@jameshadfield Hi, I can take this if the issue can be clarified. May I ask where the legend for the map can be seen?

PR #940, opened 2020-03-12
Hover on map deme will highlight tree tips

mapmeld:
My attempt at fixing #264 

> mousing over map demes should act just like mousing over legend items

The first commit does this now!

> The map demes should always represent geographical filters

The second commit modifies the clade-filtering so if you click a node on the map, you will see only nodes matching the location, and their parent branches. I didn't find this mode to be as helpful as expected, because the context is hidden? But the good news is, it works on both zika and ncov datasets (country or location attribute)

Screen Shot 2020-03-11 at 11 04 29 PM

### DIFF
--- a/src/actions/tree.js
+++ b/src/actions/tree.js
@@ -204,7 +204,7 @@ export const changeDateFilter = ({newMin = false, newMax = false, quickdraw = fa
  * @return {null} side-effects: a single action
  */
 export const updateTipRadii = (
-  {tipSelectedIdx = false, selectedLegendItem = false, searchNodes = false} = {}
+  {tipSelectedIdx = false, selectedLegendItem = false, geoFilter = false, searchNodes = false} = {}
 ) => {
   return (dispatch, getState) => {
     const { controls, tree, treeToo } = getState();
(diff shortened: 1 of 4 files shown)


mapmeld:
I added a commit to support the datasets that I saw in the deploy.

- Adds support for "pie chart" maps (such as seasonal flu), separating map hover from legend hover
- Uses location, division, or country attribute in that order (for the ebola map).
- If you click a geo deme to filter, but no locations match, no filtering on the tree

jameshadfield:
Thanks @mapmeld -- I'll try to test this out in the next couple of days!

mapmeld:
Hi - I want to re-up this. Should I change clicking on a geo to use the same functionality as clicking on the geo filter links (Filter by Location, etc) further down the page?  I didn't notice them until later.

mapmeld:
OK I've taken out the node click code, so this should contain hover code which works on /ncov-type nodes or /dengue-type pie charts

jameshadfield:
Thanks @mapmeld, this is a really nice feature we've wanted for a long time!

I added 2 commits on top of this PR:
1. dc91b9f43d416791ac7c0b04892118c10291aad2 - Updates the algorithm to use `geoResolution` so we can check the corresponding value on all tips and match against the value of the hovered deme
2. ef81d51ab60c82ff8b94fcdc49c63eb62756a176 - remove map deme cursor=pointer which was left over from a version of the PR which had click behavior

PR #947, opened 2020-03-13
Update to PR #897 fixing rotation issues

cpsandbox:
This should be PR #897 with an addition of a rotation added to the gradient definition to account for layouts other than rectangular.
Showing reversed gradient:
Pre

Showing rotated gradient:
Post

### DIFF
--- a/src/components/tree/phyloTree/change.js
+++ b/src/components/tree/phyloTree/change.js
@@ -322,7 +322,8 @@ export const change = function change({
     elemsToUpdate.add(".vaccineCross").add(".vaccineDottedLine").add(".conf");
     elemsToUpdate.add('.branchLabel').add('.tipLabel');
     elemsToUpdate.add(".grid").add(".regression");
-    svgPropsToUpdate.add("cx").add("cy").add("d").add("opacity").add("visibility");
+    svgPropsToUpdate.add("cx").add("cy").add("d").add("opacity")
+.add("visibility");
   }
 
   /* change the requested properties on the nodes */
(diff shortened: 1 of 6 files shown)


cpsandbox:
Got it. I didn't know that double view was possible! Let me look into this before you do any more.

cpsandbox:
The double tree view inserts two linear gradients with the same id value. Why the gradient would be reversed in the double rectangular view is that the one branch would be referencing a gradient with the wrong rotation. Using the debugId of the tree in making the linearGradient id solves this. This would make the debugId a dependency.

The color-by change is implemented by directly setting the stroke instead of going through the calculation that happens when the layout is changed. I'm looking at moving all of the logic to an updateStroke() function that will determine if a gradient is required and if so, create it or update it if it exists, or remove it if not required, and return a value to be used as the style stroke value(either a rgb spec or a gradient URL). Everything will be modified to use this instead of calculating the stroke in multiple places.

By adding ids to the stops of a gradient the stop-color can be updated with the emphasis colors instead of creating a new gradient on hover and destroying it on leave.

jameshadfield:
Quick note to say thanks for this work and that I'll try to test / merge this as soon as possible!

jameshadfield:
Hey Clint -- this is great work, thanks for this. It's going to have a big impact on how people interpret the data.

Really like the implementation which changes the gradient values in place 😄 

I added a few minor commits on top of this PR which renamed `debugId` (as it's now a dependency) and added some more comments. 

I'm going to release this functionality in a new version of auspice ASAP (along with quite a few other PRs), and I'll deploy on the dev.nextstrain.org for ~24 hours before updating the live site in case there are bugs I haven't found.

PR #949, opened 2020-03-13
Remove github.com from byline

trvrb:


jameshadfield:
Tested via review app & works well 👍

PR #950, opened 2020-03-14
Issue 945: fix jumpy narrative map

### LINKED ISSUE #945: Map pan is broken in narratives
This issue occurs in narratives, seemingly when changing between map-only display -> map+tree display. It causes the map to reposition itself incorrectly. Note that this happens after a 1000 or 1500ms delay implying that a `setTimeout` is involved. 

Mar-13-2020 19-34-46

Steps to reproduce:
1. [Install auspice from source](https://nextstrain.github.io/auspice/introduction/install#installing-from-source)
2. Obtain hCoV-19 datasets & narratives via cloning github.com/nextstrain/ncov
3. Run auspice via `auspice develop --datasetDir <ncov/auspice> --narrativeDir <ncov/narratives>`

Run the latest narrative (currently unpublicised) and switch between http://localhost:4000/narratives/ncov/sit-rep/2020-03-13?n=3 and the following slide.

mapmeld:
Fixes #945 

On the current site, when you resize the map by resizing the browser, switching between full <=> grid layout, or switch slides in a narrative, ```maybeInvalidateMapSize``` sets a 1500ms timer to call ```map.invalidateSize()``` and fill in blank map space.  This can clash with the narrative's goal to focus the map "viewport" to contain the data.

Proposed solution in narrative mode, ```map.invalidateSize()``` happens immediately when we arrive on a new slide. The Leaflet ```resize``` event listener will be called after the resize is completed, and this is when we fit viewport to the data.

In free (not-narrative?) mode, the only change is I halved the resize-browser to resize-map delay to 750ms. If you prefer it can be un-changed.

### DIFF
--- a/src/components/map/map.js
+++ b/src/components/map/map.js
@@ -142,7 +142,15 @@ class Map extends React.Component {
   maybeInvalidateMapSize(nextProps) {
     /* when we procedurally change the size of the card, for instance, when we swap from grid to full */
     if (this.state.map && (this.props.width !== nextProps.width || this.props.height !== nextProps.height)) {
-      window.setTimeout(this.invalidateMapSize.bind(this), 1500);
+      // first, clear any existing timeout
+      if (this.map_timeout) {
+        window.clearTimeout(this.map_timeout);
+      }
+      // delay to resize map (when complete, narrative will re-focus map on data)
+      this.map_timeout = window.setTimeout(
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @mapmeld -- tested and it fixes the issue with no apparent regressions. The map logic is complicated, so thanks for tracking down a fix here ⭐️

PR #953, opened 2020-03-16
Add Language Support to Auspice

### LINKED ISSUE #944: Language support for auspice
There is currently no language support in Nextstrain. This issue is cross cutting between auspice & nextstrain.org, but is detailed here for both.

Please see https://github.com/nextstrain/nextstrain.org/issues/130 for a description of this issue, as it applies to both codebases. Please also see PR https://github.com/nextstrain/auspice/pull/892 for a potential start to this work.

crwilcox:
Fixes #944 by adding i18next and localizing sidebar fields.

Provides an implementation of Spanish. To add additional languages, mirror the files in en or es. Also add an option to the languages selector.


crwilcox:
image

jameshadfield:
Thanks @crwilcox -- I'll create a review app so everyone can look 👍  Could you comment on how this implementation should interface with (i) language selection in nextstrain.org and (ii) loading of different datasets/narratives (as per https://github.com/nextstrain/nextstrain.org/issues/130 which #944 points to)? Thanks

src/components/framework/head.js
```diff
@@ -3,7 +3,25 @@ import { connect } from "react-redux";
...
+import i18n from "i18next";
+import { initReactI18next } from "react-i18next";
+
+const Head = ({metadata, general}) => {
+  let lang = general.language;
+
+  i18n
+  .use(initReactI18next)
+  .init({
+    resources: require("i18next-resource-store-loader!../../locales/index.js"),
```
jameshadfield: Review app error here:

```
ERROR in ./src/components/framework/head.js

Module not found: Error: Can't resolve '../../locales/index.js' in '/home/travis/build/nextstrain/auspice/src/components/framework'
...
```

src/components/controls/controls.js
```diff
@@ -12,21 +12,28 @@ import MapAnimationControls from "./map-animation";
 import PanelToggles from "./panel-toggles";
 import SearchStrains from "./search";
 import ToggleTangle from "./toggle-tangle";
+import Language from "./language";
```
jameshadfield: Review app error here:

```
ERROR in ./src/components/controls/controls.js

Module not found: Error: Can't resolve './language' in '/home/travis/build/nextstrain/auspice/src/components/controls'
```

src/components/framework/head.js
crwilcox: Oh, you know, I should include the language files in the PR 😅

crwilcox:
Language can be controlled with the query parameter 'lang=es|en|future-lang'. For instance, https://auspice-loc-hhtbnimsdcn4kjsgk2.herokuapp.com/ncov/2020-03-13?lang=es

If the language chosen doesn't exist, the fallback is 'en' as configured, so that would be the degraded experience if the language wasn't yet provided in Auspice.

joverlee521:
@jameshadfield  I took a quick look, this shouldn't interfere with the language selection on nextstrain.org because this PR only adds language support to the auspice control panel. My work in nextstrain.org is very specific to the hCoV narratives.

jameshadfield:
@crwilcox -- broadly speaking this looks great. Thankyou! 

* There are strings outside of the sidebar which should be updated to use the machinery introduced here. This includes: The header ("Build with", Maintained by", "Showing xx genomes sampled between ..."). The titles of each panel ("Phylogeny", "Transmissions", "Diversity") as well as text within each ("Reset Zoom", "Events" etc). The footer ("Filter by", "Data Updated", "Download data") and also the text within the "Download Data" modal.
* It's fine to not have all the translations for each of these fields (we can crowd source this after this PR is done 😄 ), but we should add them to the English locales JSON and extend the code to call that instead of the hardcoded string. 
* Question related to previous point: If a call is made for, say, `t(keyName)` but the current language doesn't include that `keyName` in the locales JSON, does it nicely fallback to English?
* URL query looks good 👍 
* I've added a figure to https://github.com/nextstrain/nextstrain.org/issues/130 to head towards a plan to make languages work well between the different parts of nextstrain.org (of which auspice is one). I believe this PR can work nicely with that & if you have thoughts here please comment! @joverlee521 do you have comments on the bigger picture there?

colinmegill:
Exciting to see this @crwilcox! 

I assume if you grab the browser language and pass it through as a default, that's a quick fix? If so, could be in this PR, if more complicated, could be in a subsequent?

crwilcox:
@jameshadfield 
- Those other strings make sense. Let me find those and add them to the base english version and encode them
- 100% on the crowdsourcing the other languages. After speaking with @colinmegill last week I thought doing this now would enable us to scale out to more folks.
- If t(not-in-spanish) is made, it will use the fallback language (english)

crwilcox:
@jameshadfield I think that addresses the strings you brought up and a few extra. 

@colinmegill this could be made to use the locale that is present in the browser. I think it is likely straightforward to add as well. My concern is without many languages this is going to currently fall back to english most of the time. I am happy to proceed either way.

colinmegill:
+1 for fallback to default English when no strings for browser language present.

colinmegill:
We also have to consider what happens when the translations are out of date, and how the system will know that they are out of date.

crwilcox:
@colinmegill there exist tools to help show what is/isn't done that could be leveraged. That will help for the initial fleshing out. After a string exists, the thing to do is probably to file, at the time of changing the string in english, bugs for the other languages to be completed.

jameshadfield:
Language guessing (i.e. inference via browser lang) is important, but please leave out of this PR. I'll write more on  nextstrain/nextstrain.org#130 about our plan to do this! 

> We also have to consider what happens when the translations are out of date, and how the system will know that they are out of date.

@colinmegill could you say more here? I don't see how the strings in auspice can become out of date

tsibley:
@jameshadfield As the UI copy of Auspice changes, the translations need to be updated. Since this isn't instantaneous and we have a continuous release cycle, translations will usually always be lagging behind by varying amounts.

crwilcox:
@tsibley this is my experience with other applications also. Though, if the work is tracked it can be minimized.

src/components/info/info.js
```diff
@@ -61,13 +63,19 @@ const arrayToSentence = (arr, {prefix=undefined, suffix=undefined, capatalise=tr
   return ret + " ";
 };
 
-export const createSummary = (mainTreeNumTips, nodes, filters, visibility, visibleStateCounts, branchLengthsToDisplay) => {
+export const createSummary = (mainTreeNumTips, nodes, filters, visibility, visibleStateCounts, branchLengthsToDisplay, t) => {
   const nSelectedSamples = getNumSelectedTips(nodes, visibility);
   const sampledDateRange = getVisibleDateRange(nodes, visibility);
   /* Number of genomes & their date range */
-  let summary = `Showing ${nSelectedSamples} of ${mainTreeNumTips} genomes`;
+  let summary = t(
```
jameshadfield: `TypeError: "t is not a function"` Which prevents the Download modal opening (it imports this function, but hasn't been updated to pass in `t`)

src/middleware/changeURL.js
```diff
@@ -64,6 +64,10 @@ export const changeURLMiddleware = (store) => (next) => (action) => {
       query.r = action.data === state.controls.defaults.geoResolution ? undefined : action.data;
       break;
     }
+    case types.CHANGE_LANGUAGE: {
```
jameshadfield: In order to allow more time for discussion on the interaction between language support in nextstrain.org & auspice, please comment out this code & leave a comment explaining that it is to be reintroduced after further discussions. I'm 95% sure what you've done here is the right direction but don't want to introduce a new query that we then remove in a week or two.

src/reducers/general.js
```diff
@@ -32,6 +43,10 @@ const general = (state = {
       return Object.assign({}, state, {
         pathname: action.pathname
       });
+    case types.CHANGE_LANGUAGE:
```
jameshadfield: same as comment above -- please comment out this code for the time being.

cli/develop.js
```diff
@@ -85,5 +86,6 @@ const run = (args) => {
 
 module.exports = {
   addParser,
-  run
+  run,
+  i18next
```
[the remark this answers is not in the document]
crwilcox: Good catch. This isn't used anywhere. I added this early on and never cleaned it up.

src/reducers/general.js
crwilcox: I can't comment this out without breaking the language selector in the sidebar.

src/components/info/info.js
crwilcox: I had t added, but didn't notice that the one path didn't pass branchLengths so I had t, but to the wrong arg index.

src/components/download/downloadModal.js
```diff
@@ -192,10 +193,14 @@ class DownloadModal extends React.Component {
       this.props.nodes,
       this.props.filters,
       this.props.visibility,
-      this.props.visibleStateCounts
+      this.props.visibleStateCounts,
+      undefined, // this.props.branchLengthsToDisplay,
```
crwilcox: @jameshadfield this was the issue with t and downloadModal.

jameshadfield:
Really excited to have this in auspice 🎉 I added a few commits on top of this to fix some linting errors, remove the `lang` query, and remove console logs for development. 

> managing things getting out of date

Let's adhere to keeping the (English) keys the same as the text. We should be able to write some simple scripts to compare the EN keys with those in other languages, which will identify both missing strings and now-unused strings. 

>  crowd sourcing

We've got an amazing community helping us translate sit-reps (coordinated via @cassiawag) so if @crwilcox you could write a guide on how to add language support for new languages we could get this done in quick order 🎉 

Thanks again @crwilcox

crwilcox:
Thanks James! I am going to write a guide this week :). I also have a group of coworkers that are interested in helping as well as we scale out :)

PR #958, opened 2020-03-17
Fix #946 by waiting to change the map bounds in narrative mode

### LINKED ISSUE #946: Map bounds show hysterisis in narratives
This issue occurs in narratives, where approaching a page from one direction causes the map bounds to be different from approaching the same page from the other direction. This is best explained in a GIF!

map_hysterisis

Steps to reproduce:
1. [Install auspice from source](https://nextstrain.github.io/auspice/introduction/install#installing-from-source)
2. Obtain hCoV-19 datasets & narratives via cloning github.com/nextstrain/ncov
3. Run auspice via `auspice develop --datasetDir <ncov/auspice> --narrativeDir <ncov/narratives>`

Run the latest narrative (currently unpublicised) via http://localhost:4000/narratives/ncov/sit-rep/2020-03-13?n=6. Go to the following page and then return to page 6. Note how the map bounds have changed.

CameronDevine:
This pull request fixes #946 by waiting a set amount of time before changing the map bounds. This is similar to the `maybeInvalidateMapSize` function.

It appears that this is a problem only when changing between different pages in the narrative which both have maps on them.

### DIFF
--- a/src/components/map/map.js
+++ b/src/components/map/map.js
@@ -631,7 +631,18 @@ class Map extends React.Component {
     // window.L available because leaflet() was called in componentWillMount
     this.state.currentBounds = window.L.latLngBounds(SWNE[0], SWNE[1]);
     const maxZoom = this.getMaxZoomForFittingMapToData();
-    this.state.map.fitBounds(window.L.latLngBounds(SWNE[0], SWNE[1]), {maxZoom});
+    // first, clear any existing timeout
+    if (this.bounds_timeout) {
+      window.clearTimeout(this.bounds_timeout);
+    }
+    // delay to change map bounds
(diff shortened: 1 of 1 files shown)


jameshadfield:
Quick note to say thanks for this work and that I'll try to test / merge this as soon as possible!

eharkins:
Just tested this and it:
- fixes the issue as seen in the [narrative in question in the original issue](https://github.com/nextstrain/auspice/issues/946#issuecomment-618735334)
- looks consistent with existing timeouts that deal with the map in narrative mode
- doesn't touch other parts of the code

so I'm going ahead and merging. Thanks a lot @CameronDevine!

Issue #956, opened 2020-03-17
Allow scrolling within a narrative


jameshadfield:
When viewing narratives in auspice in non-mobile view, the text content of the "narrative sidebar" cannot overflow the available space (any overflow is hidden). This is best explained by viewing a narrative locally at a small screen resolution -- the content in the "narrative sidebar" is too big for the available space but isn't allowed to scroll and is therefore un-viewable.

![narrative_overflow](https://user-images.githubusercontent.com/8350992/76817765-a4dc2600-6868-11ea-856b-dafa9ae3008b.png)

The sidebar content is rendered via [this code](https://github.com/nextstrain/auspice/blob/master/src/components/narrative/index.js#L178). 

My intention for the "narrative sidebar" is that both scroll actions and clicking on the chevrons can move to the surrounding slides. In the case that the content overflows, scrolling should bring it into view rather than changing the slide shown. Moving to the surrounding slides via scroll actions should only happen after the content of the current slide has been fully exposed. 

Please see https://github.com/nextstrain/auspice for how to get auspice installed locally  (from source) with local datasets & narratives. The above example is then viewable via http://localhost:4000/narratives/ncov/sit-rep/2020-03-13?n=3.

jameshadfield:
This issue now has a PR (#959)

tsibley:
@jameshadfield 

> My intention for the "narrative sidebar" is that both scroll actions and clicking on the chevrons can move to the surrounding slides. In the case that the content overflows, scrolling should bring it into view rather than changing the slide shown. Moving to the surrounding slides via scroll actions should only happen after the content of the current slide has been fully exposed.

Have you thought about having a continuous scroll instead of slide-based paging?

jameshadfield:
> Have you thought about having a continuous scroll instead of slide-based paging?

Yes - this was considered during the original design implementation and then more recently with the mobile implementation. Happy for it to be reconsidered in a future implementation.

Closed by #959

PR #959, opened 2020-03-17
Allow narrative sidebar scrolling

CameronDevine:
This pull request implements scrolling of the narrative sidebar as requested in #956. This is accomplished by first setting `overflow-y=auto` on the sidebar page element to allow scrolling. Next, the narrative page is only changed when the content is fully scrolled to the top or bottom (see https://stackoverflow.com/questions/2481350/how-to-get-scrollbar-position-with-javascript). Finally, to avoid accidentally changing narrative pages when trying to scroll through a given pages content, the magnitude of the `WheelEvent`'s `event.deltaY` property must be over a given threshold to allow the page to be changed.

Note:

1. Arrow keys do not scroll, they only change pages (I don't know why this is the case).
2. **This has not been tested on mobile/touch screen devices.**
3. Some styling tweaks may be needed.

### DIFF
--- a/src/components/narrative/ReactPageScroller.js
+++ b/src/components/narrative/ReactPageScroller.js
@@ -105,19 +105,22 @@ export default class ReactPageScroller extends React.Component {
         this[scrollWindowUp] = () => {
             if (!_.isNil(this["container_" + (this.state.componentIndex - 1)]) && !this[scrolling]) {
 
-                this[scrolling] = true;
-                this._pageContainer.style.transform = `translate3d(0, ${(this.state.componentIndex - 1) * -100}%, 0)`;
+                var element = this["container_" + this.state.componentIndex].getElementsByTagName('div')[0];
+                if (element.scrollTop === 0) {
+                  this[scrolling] = true;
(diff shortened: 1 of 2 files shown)


jameshadfield:
Thanks @CameronDevine 🎉 -- will review this when I can. Quick answers to 2 of your notes now:

> Arrow keys do not scroll, they only change pages (I don't know why this is the case).

These are set [here](https://github.com/nextstrain/auspice/blob/master/src/components/narrative/index.js#L93) and important for [automatic narrative - PDF conversion](https://nextstrain.github.io/auspice/narratives/create-pdf). But we could explore changing those keys to a different one since ideally up/down would scroll. 

> This has not been tested on mobile/touch screen devices.

Mobile display is rendered via a [separate react component](https://github.com/nextstrain/auspice/blob/master/src/components/narrative/MobileNarrativeDisplay.js) so I don't expect this to be a problem :)

CameronDevine:
So is the function, https://github.com/nextstrain/auspice/blob/1274e88c9a92e648fd013c14d24eb3a5d70c92b8/src/components/narrative/ReactPageScroller.js#L69-L76 never called?

PR #960, opened 2020-03-18
Enabled right to left text to be properly justified in narratives

### LINKED ISSUE #954: Make narratives work with R-->L languages
To broaden access to reports on hCov-19, we are translating narratives into Farsi and Arabic, both languages read from R-->L. Currently, narratives do not handle this well, see for example: https://nextstrain.org/narratives/ncov/sit-rep/fa/2020-03-13.

I'm imagining a right-to-left flag that right aligns all text, headers, bullet points, and punctuation. For example, a period would be on the left-side of the final word in a sentence. Bullet points would be on the right side of the screen with the text just to the left of the bullet point.

In an ideal world, this flag would also flip the entire report, so what's on the L-side of the screen would be on the R and vice a versa. Flipping the report would likely be hard, might take a bit of work, and is much lower priority than R-aligning the text.

CameronDevine:
This pull request should close #954.

The `dir="auto"` property was added to each narrative block, and to the `MainDisplayMarkdown` container.

### DIFF
--- a/src/components/narrative/MainDisplayMarkdown.js
+++ b/src/components/narrative/MainDisplayMarkdown.js
@@ -101,7 +101,7 @@ const EXPERIMENTAL_MainDisplayMarkdown = ({narrativeBlock, width, mobile}) => {
   const cleanHTML = mdToHtml(narrativeBlock.mainDisplayMarkdown);
   return (
     <Container width={width} mobile={mobile}>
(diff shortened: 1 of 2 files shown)


tsibley:
Thank you for picking this up, @CameronDevine!

tsibley:
This looks good on https://auspice-rtl-narrative-mxudrved.herokuapp.com/narratives/ncov/sit-rep/fa/2020-03-13?n=1.

PR #962, opened 2020-03-20
make shell scripts conform to shellcheck standards

joachimschmidt557:
This basically just includes missing shebangs, some quoting and removing reference to a variable never assigned to.

### DIFF
--- a/scripts/create-data-dir.sh
+++ b/scripts/create-data-dir.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 if ! [ -d "data" ]; then
   echo "Creating empty data directory"
   mkdir data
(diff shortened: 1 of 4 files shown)


jameshadfield:
@tsibley could you check these shebangs are ok with you? (I normally use `usr/bin/env bash` but that may have it's own problems). thanks

tsibley:
@jameshadfield `/bin/bash` is basically universal and what I use for Bash shebangs. I'd stick with it.

I would expect `/usr/bin/env bash` to also work nearly all the time, but the use cases for allowing `PATH` to determine which `bash` to run seem vanishingly small and in terms of hardcoding paths it just swaps one for another.

PR #963, opened 2020-03-21
Fix installation instructions for source

JensWehner:
changed `checkout` to `clone`


jameshadfield:
Ha! Thanks for catching this Jens

PR #964, opened 2020-03-21
fix file modes for non-executable files

joachimschmidt557:
It is a good idea to remove the executable permission for files which aren't meant to be executable.


jameshadfield:
Thanks!

Issue #936, opened 2020-03-11
Enable zoom out from tree after zooming in


huddlej:
## Motivating user story

John is interested in the hCoV-19 strains circulating in Washington State. He clicks the [base node of the primary Washington State clade to zoom into that phylogenetic context](https://nextstrain.org/ncov?label=clade:B1&m=div) (shown below). Next, John wants to view [the Washington clade in the context of its sibling clades](https://nextstrain.org/ncov?label=clade:B&m=div). He cannot zoom out from the current context by a single step up the tree. He has to click “Reset Layout” to return to the whole tree view, find his place in the tree again, and click the appropriate node for the ancestor of the Washington clade’s node.

![image](https://user-images.githubusercontent.com/85372/76369695-984c5080-62f1-11ea-8e33-f07bb11e9270.png)

## Proposed solution

Provide an interface to allow users to “zoom out” of the tree in steps (or “step up the tree toward the root”). 

The definition of a “step” up the tree can differ depending on the selected “Branch Length” mode. The figure above shows the “divergence” view of the tree where multiple strains are identical to their ancestral node (also known as a “polytomy”). For example, USA/WA1/2020 and Fujian/8/2020 and the ancestral node all have 3 mutations from the root of the tree. One step up the tree will take the user to the parent node of the ancestral node in the view above.

If the user has selected the “time” branch length view, they will see a tree like the one below instead.

![image](https://user-images.githubusercontent.com/85372/76369708-9edac800-62f1-11ea-8e2c-63210bac3e28.png)

If the user zoomed into the ancestral node of the USA/WA1/2020 sample, they would see:

![image](https://user-images.githubusercontent.com/85372/76369714-a4d0a900-62f1-11ea-8809-a1e58572f9a0.png)

One step up the tree from that node would be:

![image](https://user-images.githubusercontent.com/85372/76369719-ac904d80-62f1-11ea-8d5e-4fb7032e4e68.png)

## Proposed implementation ideas

Add an affordance to the root (or left-most ancestral) node in the current view to alert users that they can step up the tree by clicking something (e.g., the root again, or a nearby button, etc.).

Add a left arrow button (like https://fontawesome.com/icons/arrow-circle-left) that would allow the user to step up the tree once per click.

[...]

PR #965, opened 2020-03-21
936 tree zoom out

lukas-eu:
Solves 936

Zooming out one step when the root branch is clicked

As a visual cue for the user, the text in the bottom of the hover tooltip was changed for the root node
zoom_out


jameshadfield:
Thanks @lukas-eu -- this is a great start, but polytomies (and branches with very small amounts of divergence or time) need to be taken into account as per #936 and via this gif 👇 cc @huddlej 

Mar-23-2020 12-12-38

huddlej:
This is super cool, @lukas-eu! The relevant parts of [the original issue](https://github.com/nextstrain/auspice/issues/936) for dealing with polytomies are:

> Consider implementing one “step” based on the divergence view of the tree regardless of the currently selected branch length view. This implementation could be more biologically relevant and easier to actually implement in the existing tree component.

and

> Alternately, consider implementing one step as a delta of the existing branch length view (e.g., divergence of 1 or time of 1 month).

I think both of these suggestions would require potentially walking farther up the tree than the parent of the root node. For example, if following the divergence tree, I think you'd want to walk up to the first parent with a different divergence value than the root node.

lukas-eu:
Hey, sorry I didn't fully get the desired behavior previously. I think this should fix it.
Divergence value float comparison is done with a tolerance value of 0.0001 https://github.com/nextstrain/auspice/pull/965/files#diff-7bc7fe459ddcc463b80cd9a73a6c05b0R114

huddlej:
I think this is working much closer to how I had imagined. There is a weird edge case I noticed with the original example from the issue where zooming into the Washington hCoV-19 clade and back out again hides the first Washington sample that used to be visible.

zoom-issue

Do you think this could be related to the divergence tolerance value setting?

huddlej:
@lukas-eu that last change fixed things! @jameshadfield does this work for your polytomy example, too? 

I noticed one regression at the bottom of the internal node tooltips (below). Is this maybe related to separate translation work?

image

lukas-eu:
@huddlej Also my bad, messed that up when I was resolving merge conflicts with the master. 
Should be fixed with the next commit.

jameshadfield:
Thanks @lukas-eu -- i've added a bit of work on top of this, which is now PR #1001. Would you mind taking a look at that?

PR #967, opened 2020-03-21
Fix some broken links in documentation

joachimschmidt557:
These links were pointing to GitHub resources which have since moved slightly.


jameshadfield:
Thanks @joachimschmidt557 - much appreciated!

Issue #921, opened 2020-03-05
Various usability issues/suggestions


Xkeeper0:
Putting a handful of accessibility and usability notes in this issue.

----

The text on the bottom of this page is extremely difficult to read. The links are "#5097BA" (contrast ratio 3.24), but the font they are rendered in is too thin to even fill a full pixel. The text itself is "#888" (CR 3.54) and has the same font thickness problem. This seems to only directly affect the data page, and not the narrative sections. (https://nextstrain.org/ncov/2020-03-04)

![image](https://user-images.githubusercontent.com/1077535/76022635-5d56b100-5edc-11ea-906e-dc0eea48977b.png)

----

The pagination dots are very tiny and difficult to hit with a mouse. They appear to be about 6 pixels in diameter. There is also no hint as to where they go; a tooltip or other indicator upon hovering, indicating what section it represents, would help make it more clear. (It may also be worth changing from simple dots to something indicating that the pagination is referencing the section below, browser tabs.)

![image](https://user-images.githubusercontent.com/1077535/76023941-10280e80-5edf-11ea-9a2b-9d977ed6db92.png)

----

Long sections cannot be scrolled with the arrow keys. Even if you first click on the right side of the page, pressing the up or down arrow keys will still swap between sections. (Clicking on the right side and then using page up/down will scroll it, but not until the click has happened, and it will stop working if you change sections.)

![image](https://user-images.githubusercontent.com/1077535/76024154-76149600-5edf-11ea-88fe-f9133cc4442b.png)

----

If the page is zoomed in, it becomes impossible to read parts of the narrative. In this screenshot, the text below the "next section" arrow is not readable - scrolling will swap to the next section, revealing that the text is actually cut off entirely.

![image](https://user-images.githubusercontent.com/1077535/76024558-2c787b00-5ee0-11ea-9c1a-94c7088217ba.png)

![image](https://user-images.githubusercontent.com/1077535/76024694-66498180-5ee0-11ea-8856-42bf8c8a8e08.png)

jameshadfield:
Great suggestions -- thanks @Xkeeper0. Would you mind breaking these out into separate issues? Hopefully we can get people to help submit PRs for each of them.

One quick point regarding the right-hand-side of narratives dissapearing as you change sections. This is by design -- each left-hand-section is associated with a right-hand-section (either text or more commonly a view into the tree/map). When you move between sections, the right hand part should change. Your point about scroll-ability and being too greedy with binding the up/down arrows is spot on.

PR #972, opened 2020-03-22
Enlarge progress buttons when hovering

joachimschmidt557:
This may fix one issue raised by https://github.com/nextstrain/auspice/issues/921. Progress buttons are small and hard to pinpoint and click, so this pull request adds a temporary enlargement on hovering.


jameshadfield:
Thanks for this! 

Merging as it provides better feedback that the dots are clickable, but since you still need to hover over the small dot to get this enlargement I don't think it solves the underlying issue of them being hard to hover over in the first place.

PR #977, opened 2020-03-23
Add section to contributing about i18n and l18n

### LINKED ISSUE #976: [Docs] Author Guide for Contributing Additional Language Translations
A followup task to https://github.com/nextstrain/auspice/pull/953 is to create a guide to scale out efforts to localize Auspice to more languages.

crwilcox:
This PR fixes #976

This adds a short section to contributing to explain where to start to add translations for new locales.


crwilcox:
@jameshadfield lmk if there is more you are looking for here. Also, do we have an idea of the most desired languages? I figured probably follow the sitreps? (Korean, Russian, Arabic, Dutch, Japanese, Spanish, German, French, Portuguese, Mandarin, Italian)

jameshadfield:
Thanks Chris -- I've added this as commit 9f002c96a676e4603b7b9c06ef7df8a26be6d04c (with you as co-author) as PR #978 changed the name / format of the contributing markdown file.

> do we have an idea of the most desired languages

Yes -- if we can get auspice translated into the languages we've got sit-reps in that would be pretty great. Of those, Chinese (following https://nextstrain.org/narratives/ncov/sit-rep/zh/2020-03-20) and Japanese are perhaps the highest priority. We can leverage the fantastic community which has translated the sit-reps here, so when I get a moment I (or anyone else) will add issues to solicit these and then people (Google / wider community) can grab them 😄

PR #979, opened 2020-03-24
Fix #973 (Legend ordering)

### LINKED ISSUE #973: Legend ordering is not correct for traits without a supplied colour scale
For traits ("colorings") in a dataset which do not have a color-scale provided by the dataset JSON, auspice will generate a colour scale for them. One such trait in the current ncov data is "Exposure History". 

The ordering of the values are reversed in Firefox compared with Chrome (not tested on other browsers):

Firefox (MacOS):
image

Chrome (MacOS):
image

The ordering is via [this code](https://github.com/nextstrain/auspice/blob/master/src/util/colorScale.js#L43) and should start (dark blue colours) with the value with the most occurrences in the data and end with the value with the fewest occurrences (red colours). This means that neither of the browsers are correct!

This bug was first noticed in https://github.com/nextstrain/ncov/pull/239.

dlowder-salesforce:

- The legend ordering code was treating `stateCount` as an object when it is actually a Map.
- Added code to sort items alphabetically if they have equal counts.
- Refactored this code into its own module for easier testing.
- Added unit tests.

### Related issue(s)  

Related to #973 

### Testing

Verified the changes locally, and added Jest tests.

### DIFF
--- a/src/util/colorScale.js
+++ b/src/util/colorScale.js
@@ -8,6 +8,7 @@ import { getExtraVals } from "./colorHelpers";
 import { isColorByGenotype, decodeColorByGenotype } from "./getGenotype";
 import { setGenotype, orderOfGenotypeAppearance } from "./setGenotype";
 import { getTraitFromNode } from "./treeMiscHelpers";
+import { sortedDomain } from "./sortedDomain";
 
 const unknownColor = "#AAAAAA";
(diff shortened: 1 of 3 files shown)


jameshadfield:
Wonderful! and even includes tests 🎉

PR #994, opened 2020-03-24
Fixes #993 by creating separate translations for compound sentences

### LINKED ISSUE #993: [BUG] Some text can't be translated due to fields out of translations
Some entries are hard to translate as text is built up with separate contexts.

From #981 :

> Yes, they are numbers and dates and I can keep them. But there are some texts that are added outside the templates, such as (from https://github.com/nextstrain/auspice/blob/master/src/components/info/info.js#L91)
> 
>   const prefix = branchLengthsToDisplay !== "divOnly" ? t("and comprising") : t("comprising");
>   const filterText = arrayToSentence(filterTextArr, {prefix: prefix, capatalise: false});
>   if (filterText.length) {
>     summary += ` ${filterText}`;
>   } else {
>     summary += ". ";
>   }
> In this case, the code assumes that you can add filterText after the template and there's currently no way to control it.
>

crwilcox:
Fixes #993 by creating separate translations for compound sentences

### DIFF
--- a/src/components/info/info.js
+++ b/src/components/info/info.js
@@ -69,16 +69,23 @@ export const createSummary = (mainTreeNumTips, nodes, filters, visibility, visib
   const sampledDateRange = getVisibleDateRange(nodes, visibility);
 
   /* Number of genomes & their date range */
-  let summary = t(
-    "Showing {{x}} of {{y}} genomes",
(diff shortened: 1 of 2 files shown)


jameshadfield:
Thanks Chris! Merged & I've updated the two translations we currently have (Spanish & German).

PR #1002, opened 2020-03-26
Leaflet gesture handling

### LINKED ISSUE #260: Allow mouse scroll to zoom into map
The implementation must not interfere with normal page scroll in the situation where the user wants to access information below the map (other panels, filters etc)

tolot27:
With this PR it is possible to use the mouse scroll wheel in combination with Ctrl/Cmd to zoom in/out the map. It fixes #260.

I've tested it on desktop browsers as well as on Android phone.

The [Leaflet.Sleep](https://github.com/CliffCloud/Leaflet.Sleep) code was inactive, because `src/util/leaflet-plugins.js` was removed with commit f972a8a.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9033,9 +9033,14 @@
       }
     },
     "leaflet": {
-      "version": "1.5.1",
-      "resolved": "https://registry.npmjs.org/leaflet/-/leaflet-1.5.1.tgz",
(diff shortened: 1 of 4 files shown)


jameshadfield:
Just Works. Thanks @tolot27 this is a really nice improvement. Should be released in a new version today.

Issue #762, opened 2019-08-01
Conda install for auspice


jameshadfield:
This would be fantastic to have to remove the need to run `npm install --global auspice` once the conda environment is set up.

ccarson-publicsalesforce:
I have a proposal to sidestep the issue entirely - support the use of VSCode development containers. This wraps the global install for auspice to just the docker container, rather than the bare metal machine.

tsibley:
Resolved by https://github.com/bioconda/bioconda-recipes/pull/27188.

PR #1012, opened 2020-03-27
Strawman: VSCode dev container mode

ccarson-publicsalesforce:
#762 
I have a proposal to sidestep the conda vs global auspice install entirely - support the use of VSCode development containers.

This wraps the global install for auspice to just the docker container, rather than the bare metal machine.

To test, please give the updated README.md notes a try.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,22 @@
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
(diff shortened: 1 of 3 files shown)


jameshadfield:
Thanks Cory -- I haven't heard of VSContainers for development before, but will try to look into them a bit.

Conda is used very widely in the bioinformatics community, so the ability for a non developer to run `conda install -c bioconda auspice` (or similar) will help the project a lot. I don't think this PR changes that at all, but that doesn't mean this PR isn't a good idea!

genehack:
This PR has been untouched for over four years, and as James notes, we almost certainly need to support conda usage. I'm oging to close this out. 

We're trying to get a good handle on our open issues and PRs, so they can serve as a reliable backlog of things we actually will work on.

PR #1013, opened 2020-03-27
Prettier formatting and eslint fixes

dlowder-salesforce:

- Consistent source file formatting using `prettier`
- Fix all `eslint` warnings and errors

### Testing

Execute `yarn run lint`


dlowder-salesforce:
@jameshadfield I just merged from upstream to fix conflicts with master

dlowder-salesforce:
@jameshadfield I just pushed fixes for the merge conflicts, plus a package.json change

jameshadfield:
Hi Douglas -- thanks for this. I've not used prettier, but I know it's extremely popular and another PR here suggested using it on every commit. I'm a big fan of linting (although the rules are up for grabs) and current master has zero linting errors.

Briefly looking at the changes in this PR, most of which I assume are the product of automatically running `prettier`, they all look harmless. There are a few that are overly verbose, which I can learn to live with, and a few that I don't like (e.g. ternary `?` and `:` characters moving from the end of the line to the start, fat arrow function arguments not using brackets). 

As a group we've talked about spaces and brackets between arithmetic operators and agreed that this should be definable by the author, i.e. this change isn't desirable
```diff
-    top: navHeight/2-8,
+    top: navHeight / 2 - 8,
```

I know prettier is an "opinionated code formatter" so I'm not sure if all of these are configurable? I would like to explore this PR, especially in the context of running `prettier` automatically on each commit / PR.

dlowder-salesforce:
@jameshadfield thanks for the feedback! I'll set up prettier configuration and address the issues you flagged.

dlowder-salesforce:
@jameshadfield I'm going to close this and open a new PR once I've fixed the above issues, as there are more PRs going in and the conflicts will get too complex to deal with.

jameshadfield:
Sounds good -- could you perhaps add a PR which just defined the rules, dependencies & instructions for how to run, but not the actual JS changes? or at lease separate out the JS changes into a separate commit

Issue #872, opened 2020-01-30
"END OF NARRATIVE" page should have its own pagination dot


ivan-aksamentov:
Right now in 2019-ncov narrative (currently https://nextstrain.org/narratives/ncov/sit-rep/2020-01-30) one can jump to the "Detailed scientific credit" page (which is the one before last), by clicking on the *last* pagination dot icon.

However, there is no way to jump to "END OF NARRATIVE" page.

I believe it would be a better UX if we'd allow that. Thoughts?

tsibley:
I agree. I noticed that the "end of narrative" page also doesn't get an updated `?n=…` URL parameter.

joachimschmidt557:
I also agree that the "end of narrative" page should be counted as an actual page. This would fix the issue that scrolling up from the "end of narrative" page goes to the second-last page, not the last page. I will work on a PR for this.

PR #1016, opened 2020-03-28
Enable navigation to End-of-narrative page

joachimschmidt557:

This PR addresses part of issue #872. On the desktop (non-mobile) view of a narrative, the "End of Narrative" page now has a dot in the progress bar (the last one) and navigation to that page should be possible with a `?n=<last narrative block + 1>`

A PR for this change in the mobile view will follow.

### Related issue(s)  
Fixes part of #872

### DIFF
--- a/src/components/narrative/MobileNarrativeDisplay.js
+++ b/src/components/narrative/MobileNarrativeDisplay.js
@@ -65,13 +65,7 @@ class MobileNarrativeDisplay extends React.Component {
     };
 
     this.goToNextPage = () => {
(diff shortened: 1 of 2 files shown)


joachimschmidt557:
Ok, @jameshadfield, I'll include the changes to mobile view in this PR as well.

joachimschmidt557:
I also had the idea of making the "End of Narrative" page a "real" page, but I thought it would be just as much effort to adapt to that. Well, I was wrong. I looked over #1063 and it looks waaaay better than my version of patchworking the internals. Therefore, I'll close this PR and endorse @jameshadfield's version.

Issue #955, opened 2020-03-17
Request for performance audit


jameshadfield:
This issue describes a general request for a performance audit of auspice, due to the (wonderful!) contributions being offered by the open source community 🎉 (I'm happy for this issue to be broken into separate issues if needed.)

### Background

I've checked on performance somewhat regularly over the past few years. The main bottlenecks are exposed during animation which requires multiple tree traversals, expensive d3 updates & SVG rendering. Please also see https://github.com/nextstrain/auspice/issues/702 for some possible improvements which are related to this issue. 

This issue calls for an audit of performance and creation of issues detailing bottlenecks and areas which can be improved.

### Notes 

* There are JS timing functions throughout the auspice codebase, which are by default removed at compile time. Run `auspice build --includeTiming` to keep them in and then get logs of the JS bottlenecks 😄 
* I think there are some unnecessary react-component updates which I believe can be identified and removed. 
* Comments related to available improvements in SVG/d3 code are welcome, as are thoughts on whether we should move away from SVG. I note that painting is a big bottleneck in my testing, but I am unsure how to improve this.
* Ideally, trees will be performant up to c. 5k tips, but this is a bit of a guess (they become slow currently at around 2k tips on my laptop). Display of more tips than 5k is a problem to be solved using innovative visualisation rather than improved JS speed!

avin-kavish:
I'm working on this.

rleir:
Hi Avin, You probably already looked at 
https://nextstrain.org/charon/getDataset?prefix=/ncov/global
It loads 1.5M JSON information, 
the request headers include Accept-Encoding: gzip, deflate, br
the response headers include Content-Encoding: gzip
but it took 36 seconds to download this item.
1/ does it really need to be this big? Some citation info could be lazy-loaded. Only data  which is immediately needed for the graphics needs to be in this JSON, so the first paint could be quick.
2/ is the charon server overloaded? That is understandable.
I hope this helps
cheers -- Rick

avin-kavish:
No longer working on this :P

rleir:
Hi James @jameshadfield Just a question: the server (charon?) would not be creating this JSON at response time from a dadtabase would it? A cache would normally respond faster.
Thanks
Rick

tsibley:
@rleir 

> it took 36 seconds to download this item.

[...]

Issue #702, opened 2019-02-04
List of performance improvements to be implemented


jameshadfield:
This issue summarises the available performance improvements that could be made to auspice. These can involve algorithmic improvements and reducing the bundle size(s). Each one would probably be a separate PR.

## Code/bundle reduction
* Don't load the Map component for datasets that don't need it (this is easy with code splitting but results in a blank pane for ~0.5s for datasets that show the map). 
* Remove font-awesome. We only it in the mobile close/open sidebar icons and can use SVG paths (as done in the static site within nextstrain.org)
* Remove bootstrap CSS --  I believe we are only using this on the status page and the splash page - doing this with Flexbox would save a large CSS fetch.
* Remove lodash -- this is 25kb of the initial bundle (160kb) and we only use a dozen easy-to-reimplement functions. 
* Investigate tree-shaking to remove unused code from the bundle. (https://github.com/nextstrain/nextstrain.org/commit/699e25864d3efc0e245144331a3b8a415e64db3a makes me think we're not doing this).
* Investigate browser specific bundles. Currently the initial 160kb bundle has (at least) 28kb of polyfills which shouldn't be needed for modern browsers. Additionally, native code should run faster than polyfilled code. 

## Algorithm improvements
* Color averaging performance -- #289

PR #1018, opened 2020-03-28
Reduce bundle size by cherry-picking lodash methods

avin-kavish:
Reduces bundle by cherry-picking lodash methods used in the final build. In my experience, this is the quickest and safest method. The two other methods being using `lodash-es` which an ESModule build of lodash that is tree shaking compatible, or to use per method lodash packages. But both of them require code-changes and re-testing. This method avoids the risks associated with them and shaves ~25kb of the bundle size.

Related to #955 #702 

### Testing
This is a battle tested library, highly unlikely to break something. I also did a build and launched it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,7 @@
     "yaml-front-matter": "^4.0.0"
   },
   "devDependencies": {
+    "babel-plugin-lodash": "^3.3.4",
     "chai": "^4.1.2",
     "chai-http": "^4.0.0",
     "eslint": "^5.14.1",
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "awesomplete": "^1.1.2",
     "babel-eslint": "^10.0.1",
     "babel-loader": "^8.0.4",
+    "babel-plugin-lodash": "^3.3.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -121,6 +121,7 @@
     "yaml-front-matter": "^4.0.0"
   },
   "devDependencies": {
+    "babel-plugin-lodash": "^3.3.4",
```
jameshadfield, line 124: I think this needs to be in `dependencies` as projects which use auspice as a dependency still need access to `[npx] auspice build` which uses babel.

avin-kavish:
It's based on the overall bundle size.

Without plugin,
Screenshot from 2020-04-02 08-23-01

With plugin,
Screenshot from 2020-04-02 08-21-52

The savings are from `auspice.bundle.js`.

jameshadfield:
Yes -- you were right, it does save ~25kb (gzipped!) from the main bundle. Great! Will test this a bit & release. If you also want to look at what's going on in bundle2 that would be great, but it's only 6.5kb gzipped and there are a lot of other things I think we could make bigger savings on.

PR #1023, opened 2020-03-29
another Node version in Travis CI

peterbecich:
Travis CI coverage for Node versions 10 and 12.  Updates Ubuntu distro.

### Testing
Builds succeed with those versions: https://travis-ci.com/github/peterbecich/auspice/builds/156292881

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 sudo: false
-distro: bionic
+dist: bionic
 node_js:
 - '10'
 - '12'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: node_js
 sudo: false
+distro: bionic
 node_js:
 - '10'
+- '12'
 install:
 - npm install
 - pip install --user awscli


.travis.yml
```diff
@@ -1,7 +1,9 @@
 language: node_js
 sudo: false
+distro: bionic
 node_js:
 - '10'
+- '12'
```
jameshadfield: Did not know we could test on different node versions in parallel 👍 
Right now CI essentially just tests for critical build errors, but other PRs are forthcoming which will expand our testing so this will take on more importance.

.travis.yml
```diff
@@ -1,7 +1,9 @@
 language: node_js
 sudo: false
+distro: bionic
```
jameshadfield: `distro` -> `dist`? https://docs.travis-ci.com/user/reference/overview/

peterbecich: Looks like both `dist` and `distro` work https://travis-ci.com/github/nextstrain/auspice/jobs/307552818#L9 , but good to know

PR #1025, opened 2020-03-29
Change date range immediately and only debounce chart updates

### LINKED ISSUE #1021: Date range Slider lags by 100ms due to incorrect debounce
When the Date range slider gets debounced the UI is not updated with the new position of the slider making at appear as if the UI is laggy. What should really happen is that the slider position gets updated in real time. Only the dispatch events to update the charts should be debounced. Also, it would be much more performant to handle debounce timing outside of react state. By using lodash debounce for instance.

https://github.com/nextstrain/auspice/blob/536f33ea42e3ccf1edd31545c9a68a21c6ab9b4a/src/components/controls/date-range-inputs.js#L59

I will write a fix for this.

avin-kavish:
This improves the responsiveness of the data range slider to not be laggy.

Fixes #1021   
Related to #955 

### Testing
Probably deploy and test


jameshadfield:
Apologies for not testing this yet. I've just reviewed a bunch of PRs, and hope to get to this on the next round!

src/actions/tree.js
```diff
@@ -164,35 +166,56 @@ export const changeDateFilter = ({newMin = false, newMax = false, quickdraw = fa
...
     }
 
-    /* D I S P A T C H */
-    dispatch(dispatchObj);
-    updateEntropyVisibility(dispatch, getState);
-    if (frequencies.loaded) {
-      updateFrequencyDataDebounced(dispatch, getState);
+    clearTimeout(timeout)
+    // Quickdraw also means debounce in this context, but better change it.
+    // TODO: Refactor to two seperate variables for quickdraw and debounce
```
eharkins, line 214: @avin-kavish Definitely agree with the TODO here, I think it would make this more readable. Could you add that refactor here please? Thanks!

avin-kavish:
Yeah I can do that. btw @eharkins I don't see a badge next to your name...

eharkins:
@avin-kavish 
> are you also maintaining this library?

Yep :)

PR #1027, opened 2020-03-30
Fixed trivial doc typo

terrycojones:


terrycojones:
In case it's not clear, the second hunk of the diff is just the addition of a newline.

PR #1029, opened 2020-03-30
Update index.js

### LINKED ISSUE #1019: Unnecessary require statement
What is the need for a require statement here? Can't we use good old ESModule imports?

https://github.com/nextstrain/auspice/blob/536f33ea42e3ccf1edd31545c9a68a21c6ab9b4a/src/index.js#L38

raftar2097:
Fixes #1019
Related to # Replacing "require" with "import"

### Related issue(s)  

Fixes #  
Related to #  

### Testing

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,7 @@ import { Provider } from "react-redux";
 /* A U S P I C E   I M P O R T S */
 import configureStore from "./store";
 import { initialiseGoogleAnalyticsIfRequired } from "./util/googleAnalytics";
+import Root from "./root";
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @raftar2097 -- I think this should be absolutely fine but I want to test a bit over the coming days as I have a vague memory that this is/was needed for react hot-reloading functionality (used via `auspice develop`). Might be completely wrong about this.

jameshadfield:
Tested this using react-hot-reloader functionality and it all seems to still work. Thanks!

PR #1041, opened 2020-04-03
use font awesome chevrons instead of inline svg

joachimschmidt557:

Have the narrative navigation use chevron icons from font awesome instead of inline svg. auspice already uses font awesome chevrons for the sidebar, so it would make sense to include them here as well.

Icons used:
- https://fontawesome.com/icons/chevron-up?style=solid
- https://fontawesome.com/icons/chevron-down?style=solid

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "express-static-gzip": "^0.2.2",
     "file-loader": "^1.1.11",
     "font-awesome": "^4.7.0",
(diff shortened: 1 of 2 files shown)


joachimschmidt557:
Personally, I think it is best to use `react-icons`. That way we can still reap the performance benefits while also reducing clutter. But that's just my personal opinion.

Issue #237, opened 2017-02-28
Dynamically add tip labels


trvrb:
For SVG performance (especially when transitioning), we need to dynamically add and remove branchLabels and tipLabels from the DOM. With a zoomed tree with labels showing, I bet it's going to be safe to transition the ones that are visible.

jameshadfield:
Somewhat temporarily fixed in c347f582bd4a4af9a25be6824e46618f8ded1fb3 where I draw them iff < 50 tips inView, and they appear after the branches/tips have transitioned

jameshadfield:
- [ ] only active / "on" tips have labels
- [ ] new tips come into view via a L->R transition  
- [X] pan doesn't change active tips
- [ ] better tip in/out transitions

kairstenfay:
@jameshadfield should tip labels be re-calculated each time `drawTips()` is called?

kairstenfay:
I believe commit 579c524 addresses the final bullet point.

kairstenfay:
https://github.com/nextstrain/auspice/pull/728 addresses the parts of this issue where tip label visibility would not update after Date Range change, and where tip labels would not re-appear after layout change.

jameshadfield:
Closing in favor of #915 #916

PR #1043, opened 2020-04-05
fix #915: correctly determine tip counts based on visibility

### LINKED ISSUE #915: Tip labels should only count visible tips
## Background: 

Currently we only display tip labels when the total number of tips in-view is below a certain threshold. This prevents hundreds of tips being displayed on big trees, and allows small trees, or zoomed in sections of trees, to display labels. 

In auspice we refer to nodes / branches which are displayed on the screen as "in view". These tips may be "unselected" (e.g. filtering is applied), so that they are rendered but appear skeletal (very thin lines). This is referred to in the code to whether a tip is visible or not. 

(Potentially see #237 for further information)

### This issue

The hardcoded # of branches selected should only count "visibile" tips which are also in-view. For instance, filtering a large tree (many "in-view" nodes) to only have a few tips visible (e.g. select USA on the nCoV tree) should result in tip labels being displayed.

frogsquire:
### Description
Correctly determines whether to show tips based on the number of in-view, fully visible nodes (as opposed to partially visible, unselected nodes), not based on the number of in-view nodes in total.

### Related issue(s)  
Fixes #915 
Related to #237 

### Testing
1. Load the ncov dataset (or any of sufficient density).
2. Filter to any specific strain (or anything that selects only one node).
3. The tip for that strain should appear next to it, even on the fully zoomed-out tree.

### DIFF
--- a/src/components/tree/phyloTree/labels.js
+++ b/src/components/tree/phyloTree/labels.js
@@ -13,8 +13,11 @@ export const updateTipLabels = function updateTipLabels(dt) {
   const yPad = this.params.tipLabelPadY;
   const inViewTerminalNodes = this.nodes
     .filter((d) => d.terminal)
-    .filter((d) => d.inView);
+    .filter((d) => d.inView)
+    .filter((d) => d.visibility == NODE_VISIBLE);
+
   // console.log(`there are ${inViewTerminalNodes.length} nodes in view`)
+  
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks Maxwell -- this is great :) Merged into master with an additional commit to improve the font sizes (but plenty more work to be done re: picking font sizes).

PR #1045, opened 2020-04-07
Fixes #981 Add Japanese translation

### LINKED ISSUE #981: [i18n] Translate Auspice to Japanese
Following the guide at:
https://github.com/nextstrain/auspice/blob/master/DEV_DOCS.md#contributing-to-internationalization-and-localization-i18nl18n

yuryu:
Adding the Japanese locale files.

### Related issue(s)  

Fixes #981 

### Testing
Change the locale setting to Japanese and check translations.


yuryu:
Thanks for reviewing! I added translations for these literals. Also, I needed to adjust the original text for "more" because numbers need to come after "more" in Japanese.

Issue #917, opened 2020-03-04
Add E2E testing


jameshadfield:
Thoughts, implementations and advice welcome!

dnprock:
Testing frameworks in node.js are fragmented. I think we separate unit tests and integration (feature) tests. We use different frameworks for each test suite.

Currently, I think:

- enzyme unit tests for react client components
- jasmine unit tests for server components
- zombie.js for integration (feature) tests

jameshadfield:
I don't have the expertise here to comment, but am happy to have them separated. 

In general, I think feature (e2e?) tests are needed. Over the years we've added amazing capabilities to auspice, and we have hundreds of datasets which make different use of these varied features. As such, manually testing a decent proportion of the datasets at each PR has become untenable and this leads to bugs which often take a while to discover 😦  Creating a dev environment which can give us confidence in this area would be wonderful for nextstrain.

dnprock:
I'm happy to help out here. Can you write down the main test cases that you want to have coverage for? Especially the use cases where you think are most important.

For example:

Rendering
1. Minimal dataset rendering.
2. Complex dataset rendering.

Tree Layout
3. Change Color By option.
4. Change Tree Layout option.

Map Layout
5. Change Geographic resolution option.

hydrosquall:
For integration testing of general frontend logic, I've had good experiences with cypress ( https://www.cypress.io/ ), especially in contrast to previous experiences with setting up Selenium/PhantomJS based testing. It integrates well with CI, so it seems like a logical first step might be to add a Cypress test step to the existing Travis pipeline.

For visual regression testing, I've heard that others have had good experiences with percy.io, but I haven't personally used it.

dnprock:
cypress.io looks good. It seems like the framework to get started with now.

dnprock:
@tihuan thanks for the info, very useful. I always find node.js ecosystem a mess :). In these situations, I'd delegate the decision to whoever takes the lead. I'm happy to help out with writing some tests.

tihuan:
Thanks for the quick reply, @dnprock 😄

Yeah, not sure who's gonna lead the testing framework setup, but I can help prototype integration tests with jest + puppeteer if needed!

hydrosquall:
To clarify, the test recording limit is about sending the test results to
the cypress.io dashboard hosted service, which is optional - the test
runner itself can be used without the hosted dashboard and is still quite
helpful. Additionally they have an unlimited plan for qualifying open
source projects (
[...]

PR #1046, opened 2020-04-07
522 Add conf intervals to URL query string

### LINKED ISSUE #522: Add confidence intervals to URL state
For an introduction to how we store (and read) redux state to a URL, I suggest exploring how we do this when a color-by or geographic resolution changes. Happy to help more if anyone takes this issue on :)

eharkins:
Note: this is a work in progress and was submitted early as a PR for @jameshadfield and I to discuss and test in staging on Heroku. 

This will close #522 by adding the confidence intervals for the time trees to the URL query string so that they may be more easily shared, e.g. in narratives.

I was also hoping to include a fix to #1051 here as it seems relevant, so I will update the name of the PR accordingly when I push those commits.

### Related issue(s)  
#522
#1051 
#1053

### Testing
Given that the relevant part of the redux store ([`state.controls.temporalConfidence`](https://github.com/nextstrain/auspice/blob/master/src/reducers/controls.js#L48)) has these three pieces:
- `exists`: does the confidence interval exist for the value in question in the dataset?
- `display`: given the tree controls settings in the current redux state, does it make sense to have confidence intervals displayed on the tree?
- `on`: is the confidence intervals toggle switch on?
, I will need to test transitioning to/from as many permutations of these three options as I can manually by changing the sidebar controls - as @jameshadfield pointed out, this would be much accelerated by automated front-end testing, which it seems like is in development (#917)!


eharkins:
Another potential bug:

1. go to https://nextstrain.org/tb/global?m=div
2. select dengue
3. change branch length to time
4. conf intervals toggle breaks (everything in the redux state - and url query - seems to be what I'd expect, so perhaps something in the d3 code. 

Trying to sort out where exactly, so I can include a fix in this PR.

eharkins:
@jameshadfield I think this is ready to merge if you want to take one last look. 

- I tested the narrative you sent and it seems to work changing to and from the confidence slide in both directions.
- I can't figure out why `npm run develop` in `docs-src` (i.e. `docusaurus-start` in `docs-src/website`) is not letting me test the addition to the docs, but hopefully since it's a simple addition it's not a big deal for now.

PR #1054, opened 2020-04-10
Eslint fixes and adding a prettier config for JS formatting

douglowder:

- Fix some existing eslint errors and warnings
- Allow higher versions of node to be used
- Add a config for the `prettier` JS formatter that is minimally invasive and compatible with the current `.eslintrc` being used

### Testing

`yarn run lint` should return no errors.


jameshadfield:
Thanks for this work -- I've heard about prettify but never used it, so this PR have me a chance to play with this. I ended up rebasing this PR onto master as [branch "prettify"](https://github.com/nextstrain/auspice/tree/prettify). All in all, it changes pretty much every file in the repo! I need to read a bit more about this and have a think about whether this is the direction we want to go, vs enforcing a subset of these style rules via eslint.

PR #1057, opened 2020-04-11
test: text-based dataset smoke test

### LINKED ISSUE #1048: Add testing which simply loads datasets
There are other PRs which are implementing more advanced E2E tests etc, however a complementary (and simpler) approach which would be _extremely_ valuable for Nextstrain is to load a number of URLs and check that they work*. Having this easily runnable on every PR would save a lot of time and help prevent bugs. 

[*] By "work" I mean that the app loads without errors. I'm not actually sure how to programmatically check this, perhaps checking for any `console.error` calls, or if any ErrorBoundaries are run (they could be made to always trigger a `console.error`)? 

**Examples**  
I imagine maintaining a (perhaps quite large) list of URLs such as
```
localhost:4000/ncov
localhost:4000/zika
localhost:4000/zika?legend=open&c=region&r=region&l=clock&m=div
```

bcoe:

This pull request is an implementation of the work outlined in #1048. Given a list of URLs in `smoke-test/urls.txt`, it:

1. loads the page, and makes sure that a 200 response was received.
2. asserts that a list of target strings exist in the rendered text (it does this using chrome's `Page.captureSnapshot`, which is just a raw dump of all the text on page).

Step `2.`  is meant to be the most basic of smoke test, e.g., that `/ncov` has `novel coronavirus` in the body (compared to #917, which actually cares about page layout).

TODO:

- [x] write GitHub action that actually pulls down the datasets and runs this smoke test.
- [x] `smoke-test` directory should be skipped by `npm test`.

### Related issue(s)  

Fixes #1048
Related to #917 

### Testing

I will submit a GitHub action shortly that will run `smoke-test`, If you're happy with this approach, would love help making sure we wire it up appropriately.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - run: node --version
-      - run: npm install
+      - run: npm ci
       - run: npm test
   integration-test:
     runs-on: ubuntu-latest
@@ -28,10 +28,24 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node }}
-      - run: npm install
+      - run: npm ci
       - run: npm run get-data
       - run: npm run get-narratives
       - run: npm run integration-test:ci
+  smoke-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [10]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+      - run: npm ci
+      - run: npm run get-data
+      - run: npm run get-narratives
+      - run: npm run smoke-test:ci
   lint:
     runs-on: ubuntu-latest
     strategy:
--- a/DEV_DOCS.md
+++ b/DEV_DOCS.md
@@ -16,11 +16,33 @@ Code contributions are welcomed! [Please see the main auspice docs](https://next
 Please comment on an open issue if you are working on it.
 For changes unrelated to an open issue, please make an issue outlining what you would like to change/add.
 
-Please ensure there are no **linting** errors by running `npm run lint` (which uses [eslint](https://eslint.org/)).
-In the future we will make this a requirement for PRs or commits. 
-
 Where possible, **please rebase** your work onto master rather than merging changes from master into your PR.
 
+### Make sure tests are passing
+
+When you submit a pull request to the auspice repository, a variety of integration and unit tests will need to pass before it can be merged.
+
+You will likely want to run these tests locally before submitting:
+
+First, install the dependencies with `npm i`, then:
+
+#### For unit tests:
+
+Run `npm test`.
+
+#### For linting:
+
+Run `npm run lint`. If there are issues run `npm run lint:fix`.
+
+#### For integration tests:
+
+1. Fetch the datasets with `npm run get-data` and `npm run get-narratives`.
+2. Ensure you are **not** currently running the site locally, then run `npm run integration-test:ci`.
+
+#### For smoke tests:
+
+1. Fetch the datasets with `npm run get-data` and `npm run get-narratives`.
+2. Ensure you are **not** currently running the site locally, then run `npm run smoke-test:ci`.
 
 ## Contributing to Documentation
 
--- a/package.json
+++ b/package.json
@@ -32,8 +32,10 @@
     "gzip-and-upload": "env bash ./scripts/gzip-and-upload.sh",
     "build-docs": "echo 'see ./docs-src/README.md'",
     "test": "jest test/*.js",
-    "integration-test": "NODE_ENV=test ENV=dev jest --config puppeteer.config.js",
+    "integration-test": "NODE_ENV=test ENV=dev jest ./test/integration/*.js --config puppeteer.config.js",
     "integration-test:ci": "start-server-and-test server http://localhost:4000 integration-test",
+    "smoke-test": "NODE_ENV=test ENV=dev jest ./test/smoke-test/urls.test.js --config puppeteer.config.js",
+    "smoke-test:ci": "start-server-and-test server http://localhost:4000 smoke-test",
     "diff-lang": "./scripts/diff-lang.js"
   },
   "dependencies": {
@@ -41,9 +43,9 @@
     "@babel/plugin-proposal-class-properties": "^7.3.4",
     "@babel/plugin-proposal-decorators": "^7.3.0",
     "@babel/plugin-transform-runtime": "^7.8.3",
-    "@babel/runtime": "^7.8.7",
     "@babel/preset-env": "^7.3.4",
     "@babel/preset-react": "^7.0.0",
+    "@babel/runtime": "^7.8.7",
     "argparse": "^1.0.10",
     "awesomplete": "^1.1.2",
     "babel-eslint": "^10.0.1",
@@ -137,6 +139,7 @@
     "jest": "^25.1.0",
     "jest-puppeteer": "^4.4.0",
     "puppeteer": "^2.1.1",
+    "quoted-printable": "^1.0.1",
     "start-server-and-test": "^1.11.0",
     "webpack-bundle-analyzer": "^3.3.2"
   }
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,6 +7,7 @@ globals:
   browser: true
   context: true
   jestPuppeteer: true
+  BASE_URL: true
 rules:
   camelcase: off # require camel case names
   prefer-template: off
--- a/package-lock.json
+++ b/package-lock.json
@@ -11626,6 +11626,15 @@
       "resolved": "https://registry.npmjs.org/querystring-es3/-/querystring-es3-0.2.1.tgz",
       "integrity": "sha1-nsYfeQSYdXB9aUFFlv2Qek1xHnM="
     },
+    "quoted-printable": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/quoted-printable/-/quoted-printable-1.0.1.tgz",
+      "integrity": "sha1-nuv16z0R7vAismT9LStrK7O4TMM=",
+      "dev": true,
+      "requires": {
+        "utf8": "^2.1.0"
+      }
+    },
     "randombytes": {
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/randombytes/-/randombytes-2.1.0.tgz",
@@ -13815,6 +13824,12 @@
       "version": "3.1.1",
       "resolved": "https://registry.npmjs.org/use/-/use-3.1.1.tgz",
       "integrity": "sha512-cwESVXlO3url9YWlFW/TA9cshCEhtu7IKJ/p5soJ/gGpj7vbvFrAY/eIioQ6Dw23KjZhYgiIo8HOs1nQ2vr/oQ=="
+    },
+    "utf8": {
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/utf8/-/utf8-2.1.2.tgz",
+      "integrity": "sha1-H6DZJw6b6FDZsFAn9jUZv0ZFfZY=",
+      "dev": true
     },
     "util": {
       "version": "0.11.1",
--- a/test/smoke-test/urls.test.js
+++ b/test/smoke-test/urls.test.js
@@ -1,0 +1,49 @@
+const {readFileSync} = require('fs');
+const {decode} = require('quoted-printable');
+
+const testCases = [];
+
+// Load a suite of smoke tests from a simple text file. File is of the format:
+// [path], [comma separated text assertions]
+const text = readFileSync(require.resolve('./urls.txt'), 'utf8').trim();
+for (const line of text.split(/\r?\n/)) {
+  if (line.length === 0) continue; // allow for blank lines.
+  if (line.match(/^#/)) continue; // allow for comments.
+  const testCase = {assertions: []};
+  line.split(/,/).forEach((token, i) => {
+    if (i === 0) {
+      // The first token is the path that we wish to traverse the headless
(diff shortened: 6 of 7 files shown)


jameshadfield:
Thanks @bcoe -- excited to look at this. I think using GitHub actions to run these on each PR is the way forward (similarly with other tools like bundle size checks, linting etc). Will probably need your help with wiring everything up, but i'll explore the smoke-tests here first 😄

package.json
```diff
@@ -30,18 +30,20 @@
     "rebuild-docker-image": "env bash ./scripts/rebuild-docker-image.sh",
     "gzip-and-upload": "env bash ./scripts/gzip-and-upload.sh",
     "build-docs": "echo 'see ./docs-src/README.md'",
-    "test": "jest",
+    "test": "jest test/*.js",
```
bcoe, line 33: `npm test` was running the `test/integration` folder, and `smoke-test/`, which was causing errors (because the puppeteer variables aren't set). I think we just want to run the unit tests with this command.

package.json
```diff
@@ -30,18 +30,20 @@
...
     "integration-test": "NODE_ENV=test ENV=dev jest --config puppeteer.config.js",
+    "smoke-test": "NODE_ENV=test ENV=dev jest ./smoke-test/urls.test.js --config puppeteer.config.js",
+    "smoke-test:ci": "start-server-and-test server http://localhost:4000 smoke-test",
```
bcoe, line 36: `start-server-and-test` allows us to bring up and tear down the server in a GitHub action, without making it a background process.

smoke-test/urls.txt
```diff
@@ -0,0 +1,10 @@
+# First token represents the path to access, remaining tokens are a
+# comma separated list of text assertions (text we expect to find in the
+# rendered page):
+/ncov, novel coronavirus, Phylogeny, Diversity, Filter by Location
```
bcoe, line 4: Because of the single-page-app nature of auspice, I was thinking it's worth asserting against a couple rarely changing strings in the content, otherwise I was finding that we get `200` responses, even though the page might not have actually rendered.

My thinking is, if we pick the right strings, like `coronavirus`, it's unlikely that they change frequently, if they do change, we'll catch it in CI/CD, and just need to update this text file.

bcoe:
@jameshadfield I'm excited I found a place to help 😄 I know _literally nothing_ about genome research, but I know a healthy amount about developer tools.

Since I'm not a collaborator, GitHub actions I add won't run until you merge them, so I've created a copy of the repo [here](https://github.com/bcoe/test-auspice-action/pull/1) and have confirmed that the current actions configuration should work:

Screen Shot 2020-04-11 at 4 01 22 PM

I'd be happy to pitch in and help adding some other checks, such as linting 👍

smoke-test/urls.test.js
```diff
@@ -0,0 +1,52 @@
...
+      for (const assertion of testCase.assertions) {
+        expect(snapshotText).toEqual(expect.stringContaining(assertion));
+      }
+    });
+  }
+});
+
+// MIME HTML might have lines terminated with =\r\n, combine these lines
+// back together:
+function cleanSnapshotText(snapshotText) {
```
jameshadfield, line 46: This is nice. `snapshotText` also seems to be in [QP encoding](https://en.wikipedia.org/wiki/Quoted-printable) (not sure why) so it'd be nice to convert this to UTF-8 (I think). I couldn't find an out-of-the box node call to do this. At the lease we should replace `=3D` with `=` so we can do string contains checks such as `id="PhylogenyCard"` (although those quotations are escaped in `snapshotText` 🤷‍♂️ ). 

As a probably overly-complicated solution this could be parsed into a DOM tree, so then you could query things like (pseudocode) `hasDivWithId("PhylogenyCard")`. But I'm guessing this is getting outside the realm of smoke testing?!?

smoke-test/urls.txt
jameshadfield: Ensuring more than just `200` is good (and necessary). These strings on their own are fantastic, but I found myself wondering about the possibilities here, for instance:
If we define the keyword "rectangularTree" here can we check that the DOM has a SVG element with id "MainTree" and groups with ids A, B, C etc. I'm guessing this is going beyond the realm of smoke-testing?!?

Again, just checking for strings as you've done here is a _massive_ step forward for us, and I'm happy to implement this solution but wanted your thoughts on the above and how it fits into testing more generally.

package.json
jameshadfield: 👍

smoke-test/urls.test.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+describe("smoke test rendering of paths in urls.txt", () => {
+  for (const testCase of testCases) {
+    // eslint-disable-next-line no-loop-func
+    it(`it appropriately renders ${testCase.path}`, async () => {
+      const resp = await page.goto(`${BASE_URL}${testCase.path}`, { waitUntil: 'networkidle2' });
+      expect(resp.status()).toEqual(200);
+      // Create a Devtools session for fetching MIME HTML snapshot:
+      const client = await page.target().createCDPSession();
```
jameshadfield, line 32: (A note, not a request for this PR) Would be worth investigating [the runtime API](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#event-consoleAPICalled) to check for any `console.error()` calls during each test and failing on these.

smoke-test/urls.txt
bcoe: We could definitely build out the DSL a bit, make it so there are a few other rules.

A lot of this machinery was already in place in the integration tests, but my thinking (in agreement with yours) is it can be nice to just toss a new line in a text file for this type of assertion, vs., writing an integration tests for each. Also, I think the snapshot based approach, which looks at the whole text content, is better than plucking specific elements from the page (like in the current smoke tests).... **all this is to say**...

We could build out the integration tests further, so that you can programmatically do more complex assertions against the rendered page (and we should have this in mind, if our smoke tests get too complicated). But adding a few more simple rules that someone could place in the text file might be neat.

smoke-test/urls.test.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+  });
+  testCases.push(testCase);
+}
+
+describe("smoke test rendering of paths in urls.txt", () => {
+  for (const testCase of testCases) {
+    // eslint-disable-next-line no-loop-func
+    it(`it appropriately renders ${testCase.path}`, async () => {
+      const resp = await page.goto(`${BASE_URL}${testCase.path}`, { waitUntil: 'networkidle2' });
```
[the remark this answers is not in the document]
bcoe: `BASE_URL` is defined in the jest and puppeteer configuration I believe. We can just add the line `/* global BASE_URL */` at the top fo this file

smoke-test/urls.test.js
bcoe: As I understand it, the MHTML snapshot returns all the text content on the page (regardless of whether it's in a hidden div, frame, iframe, etc.) in MIME format. This is why it's QP Encoded (because it's a bunch of MIME content).

The nice thing, is you get a cheap snapshot of all of the visual text on the page.

> As a probably overly-complicated solution this could be parsed into a DOM tree

My concern with this, is the smoke tests start getting pretty complicated, and have to have semantic knowledge of the page structure, and as you change page structure these simple smoke tests need to change.

The integration tests in the `test/integration` folder also render the actual site using puppeteer, and are testing against specific elements in the page. It might be worth fleshing some of these tests out further, for specific examples like `hasDivWithId("PhylogenyCard")`, and if you do decide to have smoke tests as well, keep these assertions as simple as possible -- perhaps just checking for a single string in the whole page content that fingerprints the page.

jameshadfield: Thanks for humoring me 😄  I agree that it's best to keep smoke test nice and simple (especially since we can have them _right now_), and develop more complex logic in the integration tests.

smoke-test/urls.txt
jameshadfield: Agreed 👍

smoke-test/urls.test.js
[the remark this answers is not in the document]
tihuan: HAI! Sorry for the confusion, `BASE_URL` is defined in `puppeteer.config.js:4`. 

And since ESLint doesn't know about that, we'll need to whitelist that global variable in `.eslintrc:10`, like so:

```
globals:
  page: true
  browser: true
  context: true
  jestPuppeteer: true
  BASE_URL: true // <-- this
rules:
...
```

Then in theory we don't have to add `/* global BASE_URL */` for every file that uses `BASE_URL`

Thank you!

smoke-test/urls.test.js
bcoe: this might be interesting to try for sure, I can experiment.

.github/workflows/smoke-test.yaml
```diff
@@ -0,0 +1,21 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+name: smoke-test
```
jameshadfield, line 6: After this is rebased onto master can we add this job to the CI workflow (I think my last statement here was misleading).

bcoe: yes, I agree let's combine with `ci`.

smoke-test/urls.test.js
jameshadfield: Let's not worry about this here, was more a note for the future. I'm sure it will highlight a scary number of console errors!

bcoe: I wonder if we could consider doing this in a follow up? don't want this to get too complicated.

bcoe:
> re: `net::ERR_CONNECTION_REFUSED at http://localhost:4000/ncov`

I'm hesitant to suppress an error that occurs as a result of `page.goto`, as it might hide other problems. And I think the `ERR_CONNECTION_REFUSED` does a pretty good job of pointing people towards what they need to do.

What if we instead add a section to the README about running tests, which mentions integration and smoke tests need a server running?

> re: really long error messages

I pushed a change that throws this pretty error instead:

```
could not find text "Zika virus evolutiong" on page /zika
```

You will still see it in the context of the way Jest displays errors, i.e., with some additional context around the error. But, this is part for the course with test runners, and I think Jest is a really good choice for your test runner.

> re: quoted printable format strings

I pulled in the. `quoted-printable` library from `npm`, which is actually maintained by one of the same maintainers as puppeteer ... it seemed like I was going down the slippery slope of writing a crumby parser myself.

jameshadfield:
That all sounds great. Thanks @bcoe

bcoe:
@jameshadfield I think this is mostly ready, but I'll take a pass at updating the README tomorrow.

jameshadfield:
Adding a brief heads up to the `DEV_DOCS.md` would be great. One thing I noticed while looking at #1068 is that these files should probably move to `./test/smoke-test/`. That also confirmed that the approach you've taken here of just checking for certain strings in the snapshot is the right one. After this goes in it will allow us to add dozens (if not hundreds) of tests here which will really help make sure there are no unexpected breaking changes in PRs 🎉

bcoe:
> add dozens (if not hundreds) of tests here which will really help make sure there are no unexpected breaking changes in PRs

When we start to add assertions, we should have in mind that our goal is to test the different varieties of pages that might be loaded, e.g., pages with phylogenomic data that have traits that make them render slightly differently than other pages, pages loaded with GET parameters that change the page behavior.

If four diseases render identically, we might not need a test for all four 👍

.github/workflows/ci.yaml
```diff
@@ -32,6 +32,20 @@ jobs:
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [10]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+      - run: npm install
```
eharkins, line 45: @bcoe Any reason why we shouldn't use `npm ci` here?

bcoe: @eharkins looks like `npm ci` works like a charm 👍

DEV_DOCS.md
```diff
@@ -21,6 +21,33 @@ In the future we will make this a requirement for PRs or commits.
...
+2. Run the linter, `npm run lint`.
+  * _If there's an issue, run `npm run lint:fix`._
+
+_Integration tests:_
+
+1. Install dependencies, `npm install`.
+2. Fetch the datasets:
+  * `npm run get-data`.
+  * `npm run get-narratives`
+3. Start the server, `npm run start`.
```
eharkins, line 47: @bcoe looks like you created `integration-test:ci` for the GH action to automatically start a server as part of the tests, which seems to work perfectly. 

Does this mean steps 3 and 4 could be consolidated (more verbose but hopefully clearer?) into something like:
> If you have a local auspice server running on port 4000 (default), run: `npm run integration-test`.
Otherwise, you can start one and run integration tests in one command with: `npm run integration-test:ci`, or start a server separately  with `npm start`.

DEV_DOCS.md
```diff
@@ -21,6 +21,33 @@ In the future we will make this a requirement for PRs or commits.
...
+4. In another terminal, run the tests:
+  * `npm run integration-test`, or...
```
eharkins, line 49: @bcoe everything here works for me locally! `integration-test` also runs the smoke tests, so it seems like you would only run `npm run smoke-test` if you just want to run the smoke tests and not the rest of the integration tests; is that a correct interpretation? If so, we could leave out 
```
, or...
* `npm run smoke-test`.
```
if we just want folks to run everything on their contributions, right?

bcoe:
> integration-test also runs the smoke tests

I think you've pointed out a bug 😆I'd rather that we have it a bit more granular where `smoke-tests` run the smoke tests, `integration` runs the integration. I fixed this.

> Does this mean steps 3 and 4 could be consolidated (more verbose but hopefully clearer?) into something like:

I like the suggestion of simplifying the directions, and potentially just having folks run `npm run smoke-test:ci`, or `npm run `integration-test:ci`, I've updated accordingly.

PR #1059, opened 2020-04-12
build: adds GitHub Actions for test/integration/lint

bcoe:

Follow on from #1057, adds GitHub actions for `lint`, `test`, `integration`. If #1057 is landed, it might be worth combining it into the single `ci.yaml`.

### Related issue(s)  

Related to #1057 

### Testing

I've tested the configuration [here](https://github.com/bcoe/test-auspice-action/pull/1).

### Notes

**Relationship between integration tests and #1057:**

I noticed when wiring up `integration-test`, that it has many similarities with #1057 (in that, it requires the datasets, the server is running, asserts against text) -- the advantage is that folks can simply enumerate a variety of URLs in `urls.txt` (it's a nice little abstraction)...

I'd argue it might be worth dropping `integration-test`, in favor of the approach in #1057, and the E2E work happening in #917 around E2E visual testing.

**Requiring tests on PRs:**

I recommend enforcing _branch protection_ if this lands:

Screen Shot 2020-04-11 at 11 57 42 PM

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,40 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+name: ci
+jobs:
+  unit-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [10]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+      - run: node --version
+      - run: npm install
+      - run: npm test
+  integration-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
+      - run: npm install
+      - run: npm run get-data
+      - run: npm run get-narratives
+      - run: npm run integration-test:ci
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
+      - run: npm install
+      - run: npm run lint
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,21 +20,27 @@ jobs:
       - run: npm test
   integration-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [10]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 10
+          node-version: ${{ matrix.node }}
       - run: npm install
       - run: npm run get-data
       - run: npm run get-narratives
       - run: npm run integration-test:ci
   lint:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [10]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 10
+          node-version: ${{ matrix.node }}
       - run: npm install
       - run: npm run lint
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "build": "node auspice.js build --verbose",
     "prepare": "npm run build",
     "lint": "eslint src",
+    "fix": "eslint --fix src",
     "get-data": "env bash ./scripts/get-data.sh",
     "get-narratives": "env bash ./scripts/get-narratives.sh",
     "heroku-postbuild": "npm run build && npm run get-data && npm run get-narratives",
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,40 @@
...
+    branches:
+      - master
+  pull_request:
+name: ci
+jobs:
+  unit-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [10]
```
bcoe, line 12: example of how to run on a matrix of Node.js versions, if there's any wish to do so.

package.json
```diff
@@ -24,6 +24,7 @@
     "build": "node auspice.js build --verbose",
     "prepare": "npm run build",
     "lint": "eslint src",
+    "fix": "eslint --fix src",
```
bcoe, line 27: I like the pattern of using `npm run fix` to cleanup linting errors, can understand if you'd like to keep this out of the PR.

package.json
```diff
@@ -135,6 +137,7 @@
     "jest": "^25.1.0",
     "jest-puppeteer": "^4.4.0",
     "puppeteer": "^2.1.1",
+    "start-server-and-test": "^1.10.11",
```
bcoe, line 140: used the same trick as #1058, which has been working nicely (I've had the smoke tests pass in actions 10 or so times now, on my testing repo).

src/components/controls/language.js
```diff
@@ -30,7 +30,7 @@ class Language extends React.Component {
       {value: "lt", label: "Lietuvių"},
       {value: "pt", label: "Português"},
       {value: "fr", label: "Français"},
-      {value: "ja", label: "日本語"},
+      {value: "ja", label: "日本語"}
```
bcoe, line 33: gets lint actually passing.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - run: node --version
+      - run: npm install
+      - run: npm test
+  integration-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
```
jameshadfield, line 27: Think this (and the `lint` job) should use `${{ matrix.node }}`. Eventually I can see us testing on multiple node versions

jameshadfield:
This is really cool. I pushed up a copy of this branch as [this temporary PR](https://github.com/nextstrain/auspice/pull/1064) to test the CI. It is looking really good, but I have a couple of questions:

1. The workflow doesn't actually run on the PR itself, rather it runs on a merge of the PR into master (see this part of the actions/checkout@v2 docs](Checkout pull request HEAD commit instead of merge commit). This creates a confusing failure in the above PR, where linting fails on a recent introduction in master which isn't in this PR (https://github.com/nextstrain/auspice/pull/1064/checks?check_run_id=588632308#step:5:10). 

Is this normal practice? I imagine that after implementing this CI we avoid linting errors so this shouldn't be a problem, and in more complicated situations it favors a rebase-onto-master approach.

2. The unit tests fail CI with `ReferenceError: page is not define` which I think is addressed by [a comment of yours](https://github.com/nextstrain/auspice/pull/1057#discussion_r407120127) on #1057. Would you mind adding that to this PR?

> I recommend enforcing branch protection if this lands

👍  will do

bcoe:
> The workflow doesn't actually run on the PR itself, rather it runs on a merge of the PR into master 

I'm using a similar configuration to what I use on yargs ([seen here](https://github.com/yargs/yargs/blob/master/.github/workflows/ci.yaml)). I believe it will run on pull requests once merged, and part of the repos main workflow:

```yaml
on:
  push:
    branches:
      - master
  pull_request:
```

☝️ indicates, run on all `pull_request`, only run on `push` to `master`. This is so, for `pull_request`, you don't get a suite of tests for both the `push` and the PR changes.

jameshadfield:
Looking back at it my comment didn't quite explain what I was trying to get at:

My understanding is that the CI will run on each PR, and so you'll get a "checks" tab like you see in #1064. But the actual code the CI tests are running on is a merge of the PR into master, i.e. you're testing the world  that would exist were the PR to me merged. This is why you get failures such as 👇 where the line which fails has been introduced in master upstream of where this PR branched off.

image


P.S. Do you know why the CI is running on this PR? I thought  they didn't run for PRs from forks. Still trying to wrap my head around actions.

package.json
jameshadfield: Happy for this to be here but please rename to `lint:fix`

bcoe:
> i.e. you're testing the world that would exist were the PR to me merged

I think this is good default behavior, because you need to know that your PR doesn't fail tests if it's landed on top of master.

In practice, on my team's repos, we require that you're up to date with master before you can merge:

Screen Shot 2020-04-15 at 10 37 12 PM

So at any given time, it's equivalent to the PR having being rebased with master prior to testing.

----

I think I've addressed the other review 👌

bcoe:
> P.S. Do you know why the CI is running on this PR

I'm not 100% sure, I could have sworn it needed to be blessed by the maintainer by being merged first (maybe the fact that you forked the branch had an effect?).

.github/workflows/ci.yaml
bcoe: I added a matrix for both `lint` and `integration` 👍

PR #1061, opened 2020-04-14
Add license/copyright field to narratives

sidneymbell:
Adds a field for a license, copyright, etc. This uses the same function as the previous `parseContributors`, renamed to `parseAttributions` to be a bit more generic. This will probably be used for a single string +/- link most of the time, but allowing arrays enables adding both a license and a copyright.

### Testing
Usage:
```
license: "CC-BY"  
licenseLink: "https://creativecommons.org/licenses/by/4.0/"
```

Yields:
image


jameshadfield:
@sidneymbell this looks fine to me. Could you rebase onto master now that most of the changes in this PR have gone in via another PR?

PR #1063, opened 2020-04-15
Add "end of narrative" slide to blocks.

### LINKED ISSUE #872: "END OF NARRATIVE" page should have its own pagination dot

jameshadfield:
Previously, there was a disconnect between the redux state's storage of narrative blocks, and the display of a "end of narrative" (EON) slide, which wasn't part of the blocks. PR #1016 attempted to add a URL query to represent the EON, which proved problematic as we had a `blockIdx` outside of the range of the underlying blocks structure.

Here I take a different approach, by having auspice add the EON as a "normal" block on JSON parsing. Thus, we have a URL query etc without introducing lots of edge cases.

Closes #872

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -703,6 +703,7 @@ export const createStateFromQueryOrJSONs = ({
   only displaying the page number (e.g. ?n=3), but we can look up what (hidden)
   URL query this page defines via this information */
   if (narrativeBlocks) {
+    addEndOfNarrativeBlock(narrativeBlocks);
     narrative = narrativeBlocks;
     let n = parseInt(query.n, 10) || 0;
     /* If the query has defined a block which doesn't exist then default to n=0 */
@@ -830,3 +831,12 @@ export const createTreeTooState = ({
   // }
   return {tree, treeToo, controls};
 };
+
+function addEndOfNarrativeBlock(narrativeBlocks) {
+  const lastContentSlide = narrativeBlocks[narrativeBlocks.length-1];
+  const endOfNarrativeSlide = Object.assign({}, lastContentSlide, {
+    __html: undefined,
+    isEndOfNarrativeSlide: true
+  });
+  narrativeBlocks.push(endOfNarrativeSlide);
+}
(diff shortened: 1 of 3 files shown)


src/actions/recomputeReduxState.js
```diff
@@ -830,3 +831,16 @@ export const createTreeTooState = ({
...
 };
+
+function addEndOfNarrativeBlock(narrativeBlocks) {
+  const lastContentSlide = narrativeBlocks[narrativeBlocks.length-1];
+  const endOfNarrativeSlide = {
+    dataset: lastContentSlide.dataset,
+    query: lastContentSlide.query,
+    isEndOfNarrativeSlide: true
+  };
+  if (lastContentSlide.mainDisplayMarkdown) {
```
eharkins, line 842: @jameshadfield I don't know what all the possible narrative fields are that you might want to set to match (or not match) the lastContentSlide, but would it make sense to do it like this?:
```
  const endOfNarrativeSlide = Object.assign({}, lastContentSlide, {
    __html: undefined,
    isEndOfNarrativeSlide: true
  });
```

src/components/narrative/index.js
```diff
@@ -136,7 +126,26 @@ class Narrative extends React.Component {
     );
   }
   renderBlocks() {
-    const ret = this.props.blocks.map((b, i) => {
+    return this.props.blocks.map((b, i) => {
+
+      if (b.isEndOfNarrativeSlide) {
+        return (
+          <EndOfNarrative key="EON" id="EndOfNarrative">
```
eharkins, line 133: Does this need to get wrapped in a translate function for non-english narratives?

jameshadfield: It should I think, but this isn't a regression. There are a few strings in auspice which do not have translate calls, and which therefore haven't been added to the translation JSONs. Would be a good issue to make to track them all down.

src/actions/recomputeReduxState.js
jameshadfield: Yes -- this is better if we ever add another field! Changed to this approach.

Issue #1037, opened 2020-04-02
[RFC] Change bundle names to include content hashes


avin-kavish:
**Context**  
Currently, the bundle & chunk names used across builds are the same. This can cause caching issues on the client side where browsers may serve outdated versions of the bundle. It may also cause version mismatches across chunks. A common cache busting technique/practice is to include a unique hash calculated for each chunk during the build. A change in code will cause the hash to be different, thereby the file name to be different and in turn bypass the browser cache  

**Relevant Reading**
https://webpack.js.org/guides/caching/#output-filenames

**Possible Solution**
Add `[contenthash]` to output.filename and output.chunkFilename

HashTalmiz:
Hey, can I take this issue up?

jameshadfield:
Yes -- please do!

HashTalmiz:
I'm facing issue while trying to install Auspice locally. Just installed miniconda, which worked fine for the conda environment, but when i enter `npm install --global auspice`, this shows up:
![err](https://user-images.githubusercontent.com/55018280/78753585-a5616b80-7993-11ea-8e7a-44a03c3ad68f.jpg)

jameshadfield:
Hmm. Have not seen this particular error. To develop auspice I suggest installing from source via https://nextstrain.github.io/auspice/introduction/install#installing-from-source (apologies if you're doing this, `npm install --global auspice` looks to me like installing from npm)

HashTalmiz:
Thanks that worked.
I've made all the webpack changes required making sure the dependencies don't

HashTalmiz:
Thanks, that worked.
Many HTML files are dependent on a single `auspice.bundle.js` file, only problem is that they're all scattered in many directories. I've made a way around that, but it looks like the content-hashing isn't working ;-;
This is how the `output` object in `webpack.config.json` looks like
`
output       : {
			path          : outputPath,
			filename      : 'auspice.[contenthash].bundle.js',
			chunkFilename : 'auspice.chunk.[name].bundle.js',
			publicPath    : '/dist/'
		},
`
Can you check out the changes I've made in my forked repo?

jameshadfield:
Hey @HashTalmiz -- I believe that the main bundle (`./dist/auspice.bundle.js`) is only referenced by a single file, `index.html`. This bundle then lazily loads the other chunks as needed. If you submit a PR I'd be happy to take a look 😄

HashTalmiz:
[...]

PR #1069, opened 2020-04-16
Add bundlesize checks to CI

### LINKED ISSUE #1039: Add automated bundle size checks to PRs
Following from #1018 it would be great to report bundle size changes in each PR. Perhaps something like [bundlesize](https://github.com/siddharthkp/bundlesize) but I haven't looked into it.

jameshadfield:
This adds a bundlesize check to the CI with very liberal tolerances for most of the chunks. It should identify egregious bundle size expansion and provide an easier path for manual comparison of master vs PRs. 

Currently bundlesize doesn't work with GitHub actions out-of-the-box, but this is planned for their roadmap (and they've just got $10k from Google so I'm confident it will happen). Redux-saga seems to have this working via https://github.com/redux-saga/redux-saga/pull/1952/files. 

We can revisit the settings here after the chunk hashing https://github.com/nextstrain/auspice/issues/1037.

Closes #1039


bcoe:
@jameshadfield it looks like the `bundlesize` check worked as a GitHub action. I'm curious, what currently doesn't work on actions ?

jameshadfield:
Only a minor addition, in that some repos have the ability for the action to report size changes as a little graphic in the checks on each PR (e.g. 👇 , but that's a different tool). `bundlesize` has this but not for GitHub actions & it's a hassle to set up. This is minor & not worth investing time in at the moment. Thanks for reviewing!

image

Issue #1050, opened 2020-04-08
Enable a narrative to source multiple builds


cassiawag:
**Context**  
We publish weekly situation reports using narratives on the spread of COVID-19. These situation reports are broken into sections by world region. Latest report for example: https://nextstrain.org/narratives/ncov/sit-rep/2020-04-03

Currently, narratives can only source one dataset. When there was only one global `ncov` build, this wasn't a problem. Now that we have multiple subsampled builds (`ncov/global`, `ncov/europe`, `ncov/africa`...) using only one of these builds for the situation reports greatly limits the data we can cover.

**Description**   
We'd like to upgrade the narratives functionality, so it can source multiple datasets within a single narrative. For example, one slide could show the `ncov/global` build and the next slide could show the `ncov/north-american` build.

**Possible solution**   
There's been some limited discussion about the problem at issue: https://github.com/nextstrain/auspice/issues/890

jameshadfield:
This would be wonderful.

Currently the architecture of auspice is designed around loading a single dataset ([code which fetches the JSONs here](https://github.com/nextstrain/auspice/blob/master/src/actions/loadData.js#L124-L188)). This dataset JSON essentially generates _most_ of the redux state needed to display that dataset, via [createStateFromQueryOrJSONs](https://github.com/nextstrain/auspice/blob/4402d587a4102950ba70bb2dead6a84a5a200db7/src/actions/recomputeReduxState.js#L652). 

(The way narratives work is that they then use URL queries to modify this redux state when you move between slides, thus affecting how the data should be presented. This is via  `createStateFromQueryOrJSONs`.) 

To allow multiple datasets the redux state (via `createStateFromQueryOrJSONs`) will have to be re-created from the "new" dataset JSON when moving to that page in the narrative. In theory this isn't too bad, but there will be lots and lots of little bugs (I explored this a little bit when first building narratives, since I always wanted it, but hasn't been high priority until now). Since fetching JSONs is slow, any solution should pre-fetch the required JSONs for the narrative & store them in a redux cache or similar.

salvatore-fxpig:
Opened a WIP here https://github.com/nextstrain/auspice/pull/1071 but needs help for comments, test cases, debugging

jameshadfield:
Leaving a summary of a comment in #1071 here for posterity also:

[...]

PR #1071, opened 2020-04-16
Enable a narrative to source multiple builds

salvatore-fxpig:
WIP trying to address https://github.com/nextstrain/auspice/issues/1050
Requires test cases, comments, more work

Added a new script to pull narratives and data as well, and enabled local serve to serve .json data directly (as opposed to .md data).

To run the example case, run
```
npm install -g 
npm run get-data-new
```

Then launch dev or prod server as normal and browse to narrative `narratives/ncov/sit-rep/2020-04-10`

### Related issue(s)  
Fixes #  https://github.com/nextstrain/auspice/issues/1050
Related to #  

### Testing
It requires a plethora of test cases as I'm virtually positive that I've missed a fair bit of functionality in there, in fact I'm not even 100% sure how this is supposed to look like, my knowledge of genomics stops at "DNA is helix"

### DIFF
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -240,6 +240,44 @@ export const loadSecondTree = (secondTreeUrl, firstTreeUrl) => async (dispatch, 
 };
 
 
+const loadMultipleBlocks = (blocks) => async (dispatch) => {
+  const jsons = {};
+  const namesMainTree = [];
+  const namesTreeToo = [];
+
+  for (let i = 0; i < blocks.length; i++) {
+    [namesMainTree[i], namesTreeToo[i]] = collectDatasetFetchUrls(blocks[i].dataset);
+    if (namesMainTree[i]) {
+      // eslint-disable-next-line no-await-in-loop
+      if (!jsons[namesMainTree[i]]) {jsons[namesMainTree[i]] = await (await getDataset(namesMainTree[i])).json();}
+    }
+    if (namesTreeToo[i]) {
+      // eslint-disable-next-line no-await-in-loop
+      if (!jsons[namesTreeToo[i]]) {jsons[namesTreeToo[i]] = await (await getDataset(namesTreeToo[i])).json();}
+    }
+
+    const state = createStateFromQueryOrJSONs({
+      json: jsons[namesMainTree[i]],
+      secondTreeDataset: jsons[namesTreeToo[i]] || false,
+      mainTreeName: namesMainTree[i],
+      secondTreeName: namesTreeToo[i] || false,
+      narrativeBlocks: blocks,
+      query: { n: i },
+      dispatch
+    });
+
+    if (i === 0) {
+      dispatch({
+        type: types.CLEAN_START,
+        pathnameShouldBe: window.location.pathname,
+        ...state
+      });
+    }
+  }
+
+};
+
+
 export const loadJSONs = ({url = window.location.pathname, search = window.location.search} = {}) => {
   return (dispatch, getState) => {
     const { tree } = getState();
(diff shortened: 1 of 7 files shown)


cli/server/getAvailable.js
```diff
@@ -66,8 +66,8 @@ const getAvailableNarratives = async (localNarrativesPath) => {
   try {
     narratives = await readdir(localNarrativesPath);
     narratives = narratives
-      .filter((file) => file.endsWith(".md") && file!=="README.md")
-      .map((file) => file.replace(".md", ""))
+      .filter((file) => /\.(md|json)$/.test(file) && file !== "README.md")
```
jameshadfield, line 69: Narratives are markdown files only. See https://nextstrain.github.io/auspice/narratives/how-to-write fore more

scripts/get-data.js
```diff
@@ -0,0 +1,63 @@
+const rimraf = require("rimraf");
```
jameshadfield, line 1: I'm not sure of the point of this script, given that we already have (bash) scripts to get data and narratives. If you need datasets which `npm run get-data` doesn't currently download then they can be added to [get-data.sh](https://github.com/nextstrain/auspice/blob/master/scripts/get-data.sh).

Note that #1050 may appear SARS-CoV-2 specific but it's actually a more general problem of being able to change datasets within a single narrative.

src/actions/recomputeReduxState.js
```diff
@@ -327,7 +327,7 @@ const modifyControlsStateViaTree = (state, tree, treeToo, colorings) => {
       /* check mutations */
       if (node.branch_attrs && node.branch_attrs.mutations) {
         const keys = Object.keys(node.branch_attrs.mutations);
-        if (keys.length > 1 || (keys.length === 1 && keys[0]!=="nuc")) aaMuts = true;
+        if (keys.length > 1 || (keys.length === 1 && keys[0] !== "nuc")) aaMuts = true;
```
jameshadfield, line 330: There are quite a few stylistic changes such as these in this PR. I'm not necessarily against them, but would be easier to understand if they were different PRs as they are unrelated to narratives.

src/actions/loadData.js
```diff
@@ -240,6 +240,47 @@ export const loadSecondTree = (secondTreeUrl, firstTreeUrl) => async (dispatch,
 };
 
 
+const loadMultipleBlocks = (blocks) => async (dispatch) => {
+  const jsons = {};
+  const namesMainTree = [];
+  const namesTreeToo = [];
+
+  for (let i = 0; i < blocks.length; i++) {
+    namesMainTree[i] = blocks[i].dataset.split(":")[0];
```
jameshadfield, line 249: I think it'd be better to use `collectDatasetFetchUrls` rather than splitting on `:` here

src/actions/loadData.js
```diff
@@ -240,6 +240,47 @@ export const loadSecondTree = (secondTreeUrl, firstTreeUrl) => async (dispatch,
...
+
+    if (i === 0) {
+      dispatch({
+        type: types.CLEAN_START,
+        pathnameShouldBe: window.location.pathname,
+        ...state
+      });
+    }
+
+    window[`allBaseStates.${i}`] = state;
```
jameshadfield, line 278: Is there a reason this is better stored as a global rather than a (new piece of) redux state?

I also see potential issues associating it with the block number as in a narrative with blocks 1&2 being dataset A and blocks 3&4 being dataset B, then the user can jump straight from block 1 to block 4 (via the blue dots in the sidebar).

An alternative approach would be to fetch the JSONs behind each of the different datasets in a narrative and store them (as a global, or as redux state). Then, when going between blocks where the dataset differs, we run `createStateFromQueryOrJSONs` and replace the current redux state accordingly. This should be more memory efficient also.

salvatore-fxpig:
Hi @jameshadfield , 

Sorry but I was very busy these latest days so the whole thing fell of a cliff on my side. looking to pick it back up in the next couple of days.

src/actions/loadData.js
salvatore-fxpig: Good point!

src/actions/recomputeReduxState.js
salvatore-fxpig: Yeah thats my editor autoformatting everything xD
Will clean it up xD

salvatore-fxpig:
It actually took less than expected.

Tried it with your test .md case, also tried adding a couple of pages to that switching back and forth between mumps and zika with and without maps, navigating with dots and arrows. All good as far as I can tell but would obviously benefit from your review.

Re the windows vs redux thing, the way I've written it at least, it only really dumps data in one place and reloads them in the other, and the two functions are called directly from each other... Indeed I have rewritten it so that all the saving to window happens only in one function, this way it should be a very easy read. 

Given that this state is not "distributed" to components directly but dropped into redux and from there to components, I don't think there's any gain in terms of readability, memory management or efficiency to be made by wrapping it inside redux.

src/actions/loadData.js
salvatore-fxpig: See comments below, the state saving system is completely contained within one function, putting it into redux would make it less rather than more maintainable imho

salvatore-fxpig:
Rebased it and cleaned it up a bit because things had been getting fairly messy ':D

salvatore-fxpig:
Superseded by #1164

PR #1074, opened 2020-04-17
fix 1073; set up arrow height

### LINKED ISSUE #1073: handUp arrow (div) is clickable from any height on page
**Current Behavior**  
The narratives up arrow is clickable from any height on the page, and covers other links up in certain areas.

**Expected behavior**  
The up arrow should only be clickable when hovering over its icon.

**How to reproduce**  
1. Go to http://nextstrain.org/narratives/ncov/sit-rep/2020-04-10?n=1
2. Hover over the up arrow and trace your mouse down, keeping it in the same place in the lateral dimension, so that the mouse keeps showing the link style cursor hand.
3. You can click anywhere on the page within the lateral bounds of the up arrow (excluding clicking on the down arrow since it is inserted into the DOM on the same z-axis but after the up arrow - I think this is how it works) and it will go to the previous page.

**Possible solution**  
Add an explicit CSS heigh to the up arrow div (and thereby the down arrow div too since they are created programmatically with the same function). 

- Operating system: macOS High Sierra 10.13.4
- Browser: Chrome 80.0.3987.149 (Official Build) (64-bit)

**Additional context**
image

eharkins:
Fix #1073

### DIFF
--- a/src/components/narrative/index.js
+++ b/src/components/narrative/index.js
@@ -102,7 +102,8 @@ class Narrative extends React.Component {
       position: "absolute",
       cursor: "pointer",
       left: `${this.props.width/2 - width/2}px`,
-      fontSize: `${width}px`
+      fontSize: `${width}px`,
+      height: `${width}px`
     };
     if (pointUp) style.top = narrativeNavBarHeight + progressHeight;
     if (!pointUp) style.bottom = "5px";


eharkins:
Closing this for now because #1073 actually doesn't reproduce on nextstrain.org. My mistake!

jameshadfield:
This'll be (I think) introduced in  https://github.com/nextstrain/auspice/pull/1041 which is in auspice master but not auspice 2.12.0 which nextstrain.org is using

eharkins:
Yep @jameshadfield, checking out https://github.com/nextstrain/auspice/pull/1041/commits/5562785d76c4a2f063eeac5cce14f79dd33acb7f seems to be the earliest this issue is introduced. My fix here does seem to prevent the issue from happening.

PR #1084, opened 2020-04-24
Add screenshot testing (rebased)

jameshadfield:
This is a rebase of #1068 onto current master, which now has `git-lfs` set up to automatically track `*png` snapshot images. All work here by @tihuan.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,6 +25,8 @@ jobs:
         node: [10]
     steps:
       - uses: actions/checkout@v2
+        with:
+          lfs: true
       - uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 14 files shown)


tihuan:
Yay! Teamwork 🤩🙌🎉 Thanks to you, @bcoe, and @anescobar1991 for your help and great advices too!

jameshadfield:
I wondered if this would happen! The CI test fails with `Error: Invalid file signature` which is (I think) because without `git lfs` installed the `png` is just a pointer (text file?). Will look into how to add `git-lfs` support to the workflow

tihuan:
Ahh awesome! Thanks for fixing lfs for CI, @jameshadfield !

PR #1087, opened 2020-04-24
Allow SVG elements in markdown

jameshadfield:
This modifies our settings for DOMpurify to allow SVG tags (and tags used within SVG).

You can see a test narrative [here](https://github.com/nextstrain/narratives/blob/master/test_svg-in-md.md). This will be available vie the review app also.

For this to work we removed the `ALLOWED_ATTR` sanitizer config settings, but I note that the docs state "The included default configuration values are pretty good already". An alternate approach would be to also remove the `ALLOWED_TAGS` setting as the default allows SVG.

### DIFF
--- a/src/components/framework/footer.js
+++ b/src/components/framework/footer.js
@@ -1,7 +1,5 @@
 import React from "react";
 import { connect } from "react-redux";
-import marked from "marked";
-import dompurify from "dompurify";
 import styled from 'styled-components';
 import { withTranslation } from "react-i18next";
 import { FaDownload } from "react-icons/fa";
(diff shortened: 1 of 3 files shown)


joverlee521:
I'm definitely not an expert on script injection and not sure what "default configuration values are pretty good already" means. 

When I made these initial changes to the footer, @tsibley had [commented](https://github.com/nextstrain/auspice/pull/834#discussion_r357755220) that allowing `style` will allow clickjacking. Perhaps we want to add `{FORBID_ATTR: ['style']}`?

tsibley:
I don't think it's wise to remove our `ALLOWED_ATTR` or switch to only `FORBID_ATTR`. [DOMPurify's goals](https://github.com/cure53/DOMPurify/wiki/Security-Goals-&-Threat-Model) are very focused on XSS and not other issues arising from embedding untrusted content and its defaults for allowed content reflect this.

jameshadfield:
> I don't think it's wise to remove our ALLOWED_ATTR or switch to only FORBID_ATTR

I have no problem keeping `ALLOWED_ATTR` if it's possible to find a list of values which allow SVG rendering.

PR #1096, opened 2020-05-01
Do frequency dates like tree dates

eharkins:
@rneher pointed out 
> We currently have inconsistent date labeling between the tree (which is pretty) and the frequency panel which is just decimals. most readily seen on https://nextstrain.org/flu/seasonal/h3n2/ha/2y
How hard would it be to harmonize those?

This PR makes the tooltip for frequencies display ISO format dates (instead of decimals like 2018.5) and makes the x axis ticks display a nice version of ISO format dates that depends on the time scale (also instead of decimals like 2018.5). For short time scales it looks like this:
Screen Shot 2020-05-01 at 1 01 40 PM
(which does get slightly cut off on the right in mobile at times - this can be tweaked if it's an issue)
and like this for a long time scale:
Screen Shot 2020-05-01 at 1 12 20 PM

### Testing
I just manually tested with the flu datasets (which seem to be the only ones with frequency data) in Chrome with a variety of screen widths, so more testing may be necessary.

Also, I did not add any integration tests for this.

### DIFF
--- a/src/components/tree/phyloTree/grid.js
+++ b/src/components/tree/phyloTree/grid.js
@@ -70,7 +70,7 @@ const calculateMajorGridSeperationForDivergence = (range, minorTicks) => {
   }
   const minorStep = step / numMinorTicks;
 
-  return [step, minorStep];
+  return {step, minorStep};
 };
 
 /**
@@ -80,8 +80,7 @@ const calculateMajorGridSeperationForDivergence = (range, minorTicks) => {
  * @returns {array} [0] {numeric} space between major x-axis gridlines (measure of time)
  *                  [1] {numeric} space between minor x-axis gridlines (measure of time)
  */
-const calculateMajorGridSeperationForTime = (timeRange, pxAvailable) => {
-
+export const calculateMajorGridSeperationForTime = (timeRange, pxAvailable) => {
   const rountToNearest = (n, p) => Math.ceil(n/p)*p;
 
   const getMinorSpacing = (majorTimeStep) => {
@@ -121,7 +120,7 @@ const calculateMajorGridSeperationForTime = (timeRange, pxAvailable) => {
     majorTimeStep = 1/365.25;
   }
   const minorTimeStep = getMinorSpacing(majorTimeStep);
-  return [majorTimeStep, minorTimeStep];
+  return {majorTimeStep, minorTimeStep, nSteps};
 };
 
 /**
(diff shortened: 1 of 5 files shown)


rneher:
this looks very good! but it seems the date grid lines are gone...
image

both for flu and ncov on firefox and chrome. 

https://auspice-frequency-date-swmxcwg.herokuapp.com/flu/seasonal/h3n2/ha/2y

src/components/tree/phyloTree/grid.js
```diff
@@ -179,7 +178,7 @@ const computeXGridPoints = (xmin, xmax, layout, distanceMeasure, minorTicks, pxA
 const computeYGridPoints = (ymin, ymax) => {
   const majorGridPoints = [];
   let yStep = 0;
-  yStep = calculateMajorGridSeperationForDivergence(ymax-ymin)[0];
+  yStep = calculateMajorGridSeperationForDivergence(ymax-ymin).majorTimeStep;
```
jameshadfield, line 181: I believe you want to access the `step` property here

src/components/tree/phyloTree/grid.js
```diff
@@ -147,7 +146,7 @@ const computeXGridPoints = (xmin, xmax, layout, distanceMeasure, minorTicks, pxA
   const minorGridPoints = [];
 
   /* step is the amount (same units of xmax, xmin) of seperation between major grid lines */
-  const [step, minorStep] = distanceMeasure === "num_date" ?
+  const {step, minorStep} = distanceMeasure === "num_date" ?
```
jameshadfield, line 149: The 2 functions which could be called here return objects with different property (names). I think this is the reason the grid doesn't appear for temporal axes but does for divergence.

eharkins:
Thanks Richard and James! Should be fixed now. I didn't add integration tests as I'm still familiarizing myself with them, let me know if I should do that before we merge.

rneher:
Thanks! The grid is back for me :) 

One additional question: looking at the flu data sets, there are very few x-axis labels now:
image

is this something that can be easily tuned?
https://auspice-frequency-date-swmxcwg.herokuapp.com/flu/seasonal/h3n2/na/2y

eharkins:
@rneher Thanks for pointing that out it was because of a mistake in my last commit!

rneher:
excellent! I thought it might have bin a minor/major issue.

rneher:
@jameshadfield do you think this is ready?

PR #1099, opened 2020-05-02
Event listener

rleir:
Fix a memory leak. I am not a React expert, so please verify that this change is actually doing that.
For every addEventListener there should be a corresponding removeEventListener.

### Related issue(s)  

Fixes #  
Related to #955
Related to performance in that a leak of small items causes a reduction in cache hits.
However, this is probably not one of the performance 'low hanging fruit' and that is what I should be looking for.

### Testing
Yes please James, I would like help testing this. My guess is that there will be 'fireworks' in the console if this change is incorrect, so that is a test in a way. 

Also, DeepScan is no longer complaining about the code.
Thanks -- Rick


rleir:
Bug in this: the _throttle needs to be removed before unmount. You can see a warning in the console.

rleir:
This PR is a mistake.

PR #1103, opened 2020-05-03
Adds toggle for transmission lines

### LINKED ISSUE #1095: Provide toggle to display transmission lines in the map
**Context**  
With SARS-CoV-2 jumping around the globe so quickly, showing transmission lines on the map can be overwhelming in terms of information content and detract from overall understanding. See for example:

lines

There are frequently situations where we want "traits" reconstruction on phylogeny, but don't want to show these lines on the map.

**Description**   

Include a toggle in the sidebar under "Map Options" directly below "Geographic Resolution" that says "Show transmission lines". When toggled on, we'd have the current behavior. When toggled off, the deme circles / pie charts would remain but no transmission lines would be drawn on the map.

In addition, this toggle should update the URL and (eventually) this "show transmission lines" default behavior should be included in the Auspice JSON as something like:

```
    "display_defaults": {
      "transmission_lines": false
    }, 
```

But would be good to get working as UI element and URL before bothering to include in JSON.

marcaaron:
Adds toggle for transmission lines.

@jameshadfield not exactly sure if I've got the translation stuff set up right, but let me know how this looks so far

### Related issue(s)  
Fixes #1095 
Related to #1095 

### Testing
1. Load up the app and locate the "Map Options" section in the side bar
1. Verify that there is a toggle switch that says "Show transmission lines"
1. Verify that toggling the switch will remove the transmission lines from the map UI 

**Tests**
- Didn't write any new tests, but updated the snapshot tests as the UI change broke the existing tests


jameshadfield:
This looks great - and even works while an animation is in-progress 🚀 

There are a few changes which we need to address here:
* The strings to be translated (e.g. "Transmission lines") should be added to https://github.com/nextstrain/auspice/blob/master/src/locales/en/sidebar.json (if the keys are missing it falls back to english which is why it's not really noticeable at the moment). 
* I think we can remove the byline "Transmission lines" as the toggle has its own label.
* The toggle shouldn't appear for geographic resolutions which don't have transmissions available. We don't have an easy check for this at the moment -- we can check if the `transmissionData` array has any items in it but this is locked up in the state of the `<Map>` component so perhaps we need to implement a check for this in the map component's logic and a new (boolean) piece of redux state created (`currentGeoResHasTransmissions` or similar). There may be simpler ways. (You can see an example of this for nCoV by toggling the geoResolution to "location".)

The following points mentioned in #1095 aren't essential here and should probably be tackled in separate issues / PRs:
* Technique which skips calculation + drawing of transmissions 
* `showTransmissionLines` should be keyed off of geographic-resolution so it can differ according to the resolution currently displayed.
* JSON display_defaults can set the initial state
* URL query can set the initial state for the current geo-res.

--- 

I'm guessing the integration tests are failing because the snapshots are highly platform dependent, and yours probably differs from the one we use for GitHub CI. Will think about this as there are many failing PRs currently due to this.

trvrb:
Amazing work here @marcaaron. I agree with all of James's comments, except perhaps:
>`showTransmissionLines` should be keyed off of geographic-resolution so it can differ according to the resolution currently displayed.
I was thinking it would be simpler to just have a global toggle for whether auspice is displaying lines or not. You could then, for example, have transmission lines off and row through geo resolutions and you'd never see lines. Or you could have transmission lines on and then when rowing through geo resolutions with the current "global" build, you'd only see lines for "country" where the trait reconstruction actually exists.

PR #1105, opened 2020-05-04
1008 split by state - progress update

frogsquire:
**This PR is for discussion only, and is not intended as a final product.** 

Outstanding issues I am still working on:
- Branches not within subtrees need to be hidden, and branches that within subtrees need to be rendered correctly. I think this is the biggest issue remaining.
- Zooming the tree doesn't look quite right yet, partially because of the branches, I think.
- I haven't disabled this functionality when not using the rectangular layout (or forced the tree to a rectangular layout - which do we prefer?).
- I haven't tested situations like multiple trees enabled, or extensive zooming/filtering.
- I'm not yet really handling nodes that have no value for the given trait. They get shoved to the bottom for now; what do we want to do with them?

Some other points following up on our discussion:
- The subtrees are ordered from top to bottom by the `num_date` trait. This seems to work quite well visually, I think.
- It doesn't look like I need offsets to make the data look reasonable, at least with the ncov dataset. With more sparse datasets, perhaps it'd be helpful, but the tree resizing should take care of making it look nice.
- I'd appreciate any further suggestions on how to improve the efficiency of what I'm doing here; I've taken @jameshadfield's suggestions and tried to run with them, but I want to be sure I am producing something that will meet our needs. Specifically:
-- I couldn't find a way to run this code without updating the layout, and without using the `modifySVGInStages` approach. This makes the transition when flipping the checkbox on and off pretty slow. Would love to improve this.
-- Any optimizations of the tree traversal/subtree data gathering process would be appreciated.
-- I need to do some code cleanup/quality improvements (for example, using a dictionary or trait-indexed array instead of a custom object to store the subtrees), run the linter, etc. Would appreciate any suggestions on this front as well.

I will continue working on this over the coming days, but wanted to update on my progress again. Thanks!


frogsquire:
@jameshadfield  @evogytis  Just wanted to check in on this -- any thoughts?

evogytis:
Hey @frogsquire, I checked the branch as of 2870aa3a63f0a1aef58412b6341bae4c71eaadf5 and this is what I get (this is DENV1 dataset):
Screenshot 2020-05-11 at 08 43 57

What I'm seeing is missing internal nodes. This is what it should look like roughly (colours and sorting might differ, but the look of each subtree shouldn't):
download (26)

trvrb:
Very exciting to see this progress @frogsquire. 😀 I see the same issue as Gytis with deeper nodes missing connections. I think for ordering subtrees I would do:

1. First order based on legend ordering. In the attached it would be China, Hong Kong, Taiwan, etc...
2. Within these color blocks, order by `num_date` as you propose above.

Screen Shot 2020-05-11 at 10 23 18 PM

src/components/tree/phyloTree/helpers.js
```diff
@@ -101,6 +101,73 @@ export const setYValuesRecursively = (node, yCounter) => {
...
+    }    
+  }
+}
+
+/**
+ * setSplitTreeYValues - works similarly to setYValues above,
+ * but splits the tree by the given trait, grouping nodes with the
+ * same trait value together
+ * 
+ * todo: is 0 guaranteed to be root node?
```
jameshadfield, line 137: yes -- `nodes[0]` is always the root.

src/components/tree/phyloTree/helpers.js
```diff
@@ -101,6 +101,73 @@ export const setYValuesRecursively = (node, yCounter) => {
...
+    if (bDate > aDate) return 1;
+    return 0;
+  });
+
+  // set the y-values in each subtree
+  // such that oldest subtrees are at the top
+  let currentMaxY = 0;
+  while (subtreeStack.length) {
+    const nextSubtree = subtreeStack.shift();
+    nextSubtree.subtreeNodes.forEach(node => node.n.yvalue = currentMaxY++)
```
jameshadfield, line 167: This isn't using the same technique as `setYValues` which recursively walks the tree (from the root to the tips). It _may be_ that the order of the `nodes` is the same as if you were to recurse the tree but I don't think so. Either way, a preorder traversal should be done here to set the yvalue.

src/components/tree/phyloTree/helpers.js
```diff
@@ -101,6 +101,73 @@ export const setYValuesRecursively = (node, yCounter) => {
...
 export const setYValues = (nodes) => setYValuesRecursively(nodes[0], 0);
 
+// sorts all child nodes from a subtree into their subtrees
+const collectSubtrees = (startNode, subtree, trait, currentTraitValue, subtreeStack) => {
+  if (!startNode.children) return;
+
+  for (let i = startNode.children.length-1; i >= 0; i--) {
+    let currentNode = startNode.children[i];
+    let thisNodeTraitValue = getTraitFromNode(currentNode.n, trait);
+    // todo: nodes with no trait value? should they be grouped together?
```
jameshadfield, line 111: I believe they should be ignored (cc @evogytis). This isn't quite possible to do currently for tips in auspice, but for branches we can set a `hidden` property so they won't be rendered.

src/components/tree/phyloTree/helpers.js
```diff
@@ -101,6 +101,73 @@ export const setYValuesRecursively = (node, yCounter) => {
...
+    if (a.traitValue > b.traitValue) return 1;
+    return 0;
+  });
+
+  // todo: sort the subtrees by num_date
+  subtreeStack.sort((a, b) => {
+    if (a.traitValue == b.traitValue)
+      return 0;
+
+    let aDate = new Date(getTraitFromNode(a.subtreeNodes[0].n, "num_date"));
```
jameshadfield, line 155: `"num_date"` means numerical date so it'd be faster to sort these numerically rather than converting into `Date` objects.

src/components/tree/phyloTree/helpers.js
evogytis: Yes, I reckon for the sake of simplicity just ignore nodes without trait values.

src/components/tree/phyloTree/helpers.js
jameshadfield: image
This screenshot shows that the internal nodes of a single subtree are being assigned y-values before the tips, so the subtree doesn't appear connected

src/components/tree/reactD3Interface/change.js
```diff
@@ -92,6 +92,18 @@ export const changePhyloTreeViaPropsComparison = (mainTree, phylotree, oldProps,
     args.svgHasChangedDimensions = true;
   }
 
+  /* split tree by colored-by trait */
+  /* just copies into tree; will be null/empty string/not defined if no trait */
+  if (newProps.tree.splitTreeByTrait !== oldProps.tree.splitTreeByTrait)
+  {
+    args.splitTreeByTrait = newProps.tree.splitTreeByTrait;
+    if (args.splitTreeByTrait == null) {// if the split is being unset, reset the layout
+    // todo: could use updateLayout instead?
```
jameshadfield, line 101: Let's revisit this, and exactly how we're using d3 to update the DOM (e.g. transitions) once the algorithm is working correctly

src/components/tree/phyloTree/renderers.js
```diff
@@ -172,7 +172,7 @@ export const drawBranches = function drawBranches() {
   if (!("branchTee" in this.groups)) {
     this.groups.branchTee = this.svg.append("g").attr("id", "branchTee");
   }
-  if (this.layout === "clock" || this.layout === "unrooted") {
+  if (this.layout === "clock" || this.layout === "unrooted" || params.isTreeSplitByTrait) {
```
jameshadfield, line 175: This is causing the "T" branches (the vertical part of a branch) not to be drawn, which is one reason why the subtrees aren't joining up. Drawing the T branches in this PR doesn't look good as it stands, I think because the y-values of parents and children are far apart when they should be similar, and the coordinates of the T-branch are such that it spans the y-values of the nodes children (see other comments).

src/components/tree/phyloTree/helpers.js
frogsquire: @jameshadfield thanks very much for your thoughts. I think this is the core of the issues I'm having. I'll work on this in the next couple of days, and update.

frogsquire:
@jameshadfield @evogytis Substantial progress on this over the course of this weekend. Branches are now working correctly, except that I'm not yet detecting all cases where a should be hidden (for instance, when sorting by country on the denv1 dataset, some trees have "tails" at their roots that are gray intermediate branches). Mostly, it looks good, though, and in particular the example @evogytis pointed out (denv1 by denv genotype) looks great!

I'll continue working on this, but wanted to check in again on further objectives regarding efficiency, the d3 transition, etc., and to make sure this is getting closer to meeting your needs.

Did we want to adopt the ordering @trvrb suggested - rather than strict ordering by `num_date`? If so, that may require some additional work to get the ordering in the key.

src/components/tree/phyloTree/helpers.js
frogsquire: I think this should now be resolved. The trees now match what I expect based on @evogytis's screenshots/mockup, and nodes should be getting assigned y-values in the correct order.

src/components/tree/phyloTree/helpers.js
frogsquire: Because ignoring those nodes resulted in them all being clustered at the top of the graph, I am actually explicitly hiding them (as well as the intermediate nodes that aren't relevant to the subtrees). I think this makes things look much cleaner.

src/components/tree/reactD3Interface/change.js
frogsquire: Update on this - the d3 transition is working reasonably well for turning the toggle off. For turning it on, it still looks a bit jerky.

frogsquire:
Here's my list of what needs to be resolved in order for me to consider this ready for full review (commenting here for my own benefit as much as anyone else's): 

- Rebase off of current auspice master and make sure nothing breaks
- Fix the broken integration test (probably breaking because there's a new UI control)
- Get rid of the last "trailing" gray branches that show up in some views
- When in split-by-state mode, changing the color-by state should keep split mode enabled
- When in split-by-state mode, changing the type of tree (or doing anything else to the tree) should turn the checkbox off
- When enabling the checkbox, transform the tree back to a rectangular layout if it isn't in one
- Re-run linter, tests, etc.

I'm presuming a review will include a discussion about efficiency of the algorithm, d3 stuff, etc. If desired, we could also discuss adding tests.

@jameshadfield @trvrb In the interests of time, can we split "Reorder the subtrees by order in the legend, then by `num_date`" to a separate issue?.

frogsquire:
@jameshadfield Please see the attached image diffs from the integration tests. The only differences I can see is that there's some issues with font aliasing between my machine and the reference images, and that the UI has been altered because of the new toggle.

If these look acceptable to you, let me know, and I can work on the font issue (according to [this](https://support.google.com/chrome/thread/27556542?hl=en) and the terminal output I've gotten, I may just need to reinstall some fonts) and provide new reference images.

frogsquire:
Base changed and changed back because I ran into [this issue](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch).

src/components/tree/reactD3Interface/change.js
frogsquire: @jameshadfield, if you feel we are at this point, do you have any thoughts on why the transition is still not smooth when enabling the toggle (though it is smooth at all other times)?

jameshadfield: I'll look into this shortly, as I do think we're at this point 😄

Issue #957, opened 2020-03-17
RFC: Live editing of narratives


jameshadfield:
This issue is a request for comments relating to functionality that we would like in auspice (or perhaps it's better as part of nextstrain.org?). For a background to narratives please see https://nextstrain.github.io/auspice/narratives/how-to-write. 

### The current situation
Currently to create a narrative you must take one of two paths:
1.  install auspice locally, download the dataset you wish to write a narrative for, create a markdown file, edit the markdown file whilst viewing the result in auspice locally, then upload the markdown file when you wish to publish it. 
2. Same as 1, but don't test locally and hope that it Just Works in nextstrain when you upload it!

It would be great to have a live-editing functionality whereby you could essentially write the markdown in the browser and have it displayed on-the-fly. There would then be some functionality to either (a) download the markdown file when done and/or (b) publish it (currently publishing means uploading to a GitHub repo). 

In order to start discussion I've created these two (preliminary!) mocks for how something could look:

![narratives_p1](https://user-images.githubusercontent.com/8350992/76820706-c3deb600-6870-11ea-9f2f-98a556825b8a.png)

![narratives_p2](https://user-images.githubusercontent.com/8350992/76820708-c4774c80-6870-11ea-9b92-fe83e5120554.png)

### Discussion points:

* It's not clear how simple it will be to programmatically import `auspice` to view the current markdown slide and update it as it changes. I didn't build auspice with this in mind! There may be different solutions here.
* @cassiawag hopefully this will really help with translations. What are some of the points needed to be taken into consideration for this? I'm imagining (a) starting with the content "filled in" from a previous markdown file and (b) publishing to a GitHub branch / fork...
* @enjalot you mentioned you have experience here, so I'm tagging you in the hope you can guide some of this!

enjalot:
I've had good experience using [CodeMirror](https://codemirror.net/) in other projects as a markdown IDE.
It sounds like the pieces are all there, but as you note there may be some considerations around live-updating the auspice view. I'll see if I can dig into the code and understand the flow.

Some other things to consider:
* maintaining temporary state while editing, so that browser crashes/accidental navigating away doesn't destroy work (could use localstorage or indexdb for this)
* different editing modes: split screen, vs edit mode vs full preview
* another thought: the markdown and the query could be treated as separate editing modes, so that you can iterate on how your query will render full screen, but when you edit markdown you are seeing the output update live (since it will always be displayed narrowly). I may need to sketch this to communicate the idea better.

[...]

PR #1117, opened 2020-05-09
Live editing of narrative files

jameshadfield:
This PR introduces a prototype of live editing of narrative files. It's not ready for merge into master [1] but it is usable and I want to surface it now to solicit comments from those currently writing a lot of narratives, cc @emmahodcroft, @cassiawag & @sidneymbell. 

Currently it can be run locally (see [the readme](https://github.com/nextstrain/auspice/tree/auspice-ex-avibus/narratives-editor) for instructions) or via https://auspice-dev.herokuapp.com/ (which is currently very slow, see [2]). 

It allows you to drag & drop a markdown narrative onto the browser and then edit content in place while seeing the rendered narrative update. This is best explained via a screen-grab:

live-editing-example

This is a prototype. Please see [the readme](https://github.com/nextstrain/auspice/tree/auspice-ex-avibus/narratives-editor) for a list of items which still need implementing (I'm hoping others can help here). This PR starts us on the journey described in #957. 

[1] For one, it overwrites the heroku behavior so review apps won't work. See commit message of f1f2ed1 for more details.
[2] Mainly because everything's running in dev mode, so nothing's compressed and fetching large amounts of JSON + JS is slow.


cassiawag:
@jameshadfield -- this is so cool!! Very nice work on it. I am excited for using this with translations.

Most of the valuable functionality I see is already covered in your to-do list in the readme. In order for this to be *really* useful for translations, I think integration with GitHub repos is key.  Ideally, I could go to the app online, translate a narrative, and then have the GitHub repo automatically update with my translation. Translators vary in tech savvy-ness. Some solely edit on GitHub; some clone the repo and modify locally. Those who know how to use GitHub could copy and paste from their local repo to the online app, but this will be an barrier to uptake for others. Plus, moving the file to a new system is an extra step in the translations workflow. Some might prefer to just edit the markdown file locally and push up to GitHub for preview after as they currently do. 

Obviously, if not GitHub integrated, this will still be very useful for writing narratives. While writing, it's easier to screw up syntax, and live preview has more value.

eharkins:
@jameshadfield this is sweet! Happy to take on some of the to-do items, let's chat about it whenever works best for you. 

One risk associated with editing in the client (also exists of course when editing on the github UI) is that you can lose your work by leaving the page or refreshing. Not sure there are many ways around this, other than implementing some autosave feature (sounds complicated) or just warning users to save their work often.

tsibley:
@cassiawag One big risk with saving back to GitHub directly is blowing away other edits that happen in the intervening time between one person opening it and saving it. If that situation isn't handled, it's a fairly substantial usability hurdle. If it is handled, it's a fairly substantial technical hurdle.

emmahodcroft:
This is so cool James!!

My guess would be barring a big technical solution, we might just need to tell translators to stop and copy/paste slides into Github (or their local repo) whenever they finish a slide - to ensure their work is saved. For some, they might want to do a 'first pass' just translating, then they can drag and drop into this to ensure it all looks the way it should - if copying every slide back is too tedious?

But Cassia definitely better to comment on how the translations work!

cassiawag:
That's a good concern @trs. I think that's currently a concern with editing translations on GitHub, which groups that do edit on GitHub have mostly worked around by lots of communication about who's working on what when.

sidneymbell:
This is really fantastic, thanks so much for working on this @jameshadfield!! 

Most of my comments are in the `to-do` list already. 

- I'd echo the autosave issue -- easy to lose work. As a first pass, could we perhaps prompt the user for a local file path at the beginning, and automatically dump updates to that file at some regular interval? 

- Worth noting that any bugs that 'break' narratives in core auspice will become bigger issues in this context with non-technical users (e.g., links at line endings, map zooming, etc).  

- [Semantic] While I like the meaning, the naming may hurt discoverability of this awesome new feature.

tsibley:
> I'd echo the autosave issue -- easy to lose work. As a first pass, could we perhaps prompt the user for a local file path at the beginning, and automatically dump updates to that file at some regular interval?

Not really possible to save directly to a local file path from a web app without user interaction.

As an alternative, though, the editor could save contents to [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) and load it from there if present/resuming a session.

trvrb:
What about continually saving to a GitHub repo ala https://prose.io/?

sidneymbell:
Continually saving to github would work well for users who are editing
public narratives, but not for those who need to work with private
narratives (e.g., DPHs)

jameshadfield:
Thanks for all the comments :) 

I've cleaned up the code a bit (it's still very much a prototype) and converted it to a stand-alone app which imports auspice as a dependency. I've improved bits and pieces of the functionality, and fleshed out the to-do list which hopefully I (and other contributors) can tackle piece-by-piece.

You can find the code & future development at https://github.com/nextstrain/auctor.

PR #1119, opened 2020-05-10
Improve performance - Reduce apparent loading time

salvatore-fxpig:
The goal of this request is to reduce significantly the apparent loading time of  the data displays.

This is achieved through 2 distinct actions split over 2 commits:

1) Revisit and optimize the entropy calculation, which is a completely blocking function that can take more than 1s to compute. The optimization is mostly realized by putting in practice what is already suggested in one of the code comments, reduce the deep-clones of the state by reusing non-modified states. The new function also includes state storage in the nodes so that the state assignment is done only once and on following iterations only counting of (visible) nodes happens. Would appreciate is someone could help me confirm the following two assumptions about this:
1) The root node is always the same at every calculation
2) The state of a node is always the same at every calculation.
The above is achieved without making the code significantly more complex than it already is.

2) Heavily make use of async functions in the two biggest blocking functions of the call stack, that is to say, render phyloTree and the aforementioned entropy calculation. The former is converted into a gradual rendering function that renders branches and termination points in small batches, returning control to the DOM rendering between one batch and another. The former is converted in an completely non-blocking function that runs opportunistically rather than blocking the rendering.

Note: the entropy calculation function could be an interesting candidate for a Web Worker solution, but seeing the quite restrictive targets of this project in terms in compatibility + the huge size of the tree object, I don't believe the payoff would be worth the hassle.

I tested the time-to-display in from F5 to first paint of the phylo with a stopwatch on the ncov set, clocking @4.3s avg for the current master vs @3.2s avg for the updated version in this PR. However of those times around 2.2s are loading auspice + the dataset according to Chrome, so the time from data received to first rendering has been more or less halved (the effects maybe be more dramatic on the nexstrain.org repo, where there's some degree of caching so switching between different datasets should become significantly faster (hopefully).

### Related issue(s) 
https://github.com/nextstrain/auspice/issues/955

### Testing
Please someone check that all renders correctly, both the phylogeny tree and the entropy chart are touched! I did a general eyeball check and pressing-all-the-button-randomly check but someone with real understanding of what's going on in the chart could no doubt do better.


src/components/tree/phyloTree/renderers.js
```diff
@@ -86,37 +86,45 @@ export const drawVaccines = function drawVaccines() {
     .on("click", this.callbacks.onTipClick);
 };
 
+const BATCH_SIZE = 100;
```
mingir2, line 89: I think this line should be moved to globals.js

mingir2:
@salvatore-fxpig Very interesting! Now it looks like there's an animation for drawing the tree. 

I'm not sure if this is related to this PR, but I haven't been able to reproduce it on nextstrain.org. Perhaps, it related to the async calls?

Screen Shot 2020-05-13 at 2 32 41 PM
Screen Shot 2020-05-13 at 2 34 15 PM

Steps to reproduce the issue
1. Load auspice with one of the datasets.
2. Rapidly switch between tree layout options (ie, rectangular & radial)
3. Wait for it to draw the tree. Sometimes, it would draw an incorrect layout.

src/components/tree/phyloTree/renderers.js
```diff
@@ -103,6 +103,7 @@ export const drawTips = async function drawTips() {
 
   for (let i = 0, j = 0; i < this.nodes.length; i += BATCH_SIZE, j++) {
     this.groups.tips[j] = this.groups.tips[j] || this.groups.tips.append("g").attr("id", `tips-${i}`);
+    this.groups.tips[j].selectAll(".tip").remove();
```
salvatore-fxpig, line 106: By adding this line here it forces a redraw at the end. Should not impact initial draw significantly but should ensure that the final state of the animation is coherent. Also similar to what happens on master now (if one quickly switches between layouts, the animation "jumps" at the end)

salvatore-fxpig, line 106: By adding this line here it forces a redraw at the end. Should not impact initial draw significantly but should ensure that the final state of the animation is coherent. Also similar to what happens on master now (if one quickly switches between layouts, the animation "jumps" at the end)

src/components/tree/phyloTree/renderers.js
salvatore-fxpig: Done! :D

PR #1120, opened 2020-05-10
webpack updates

salvatore-fxpig:
This pull request changes some elements in webpack config, namely:
- Forces aliasing of `react` and react utilities to allow better compatibility with out-of-root-folder extensions
- Normalizes use of `react-hot-loader`
- Forces resolution of modules to pre-transpiled code where available (which is almost everywhere), hopefully alleviating the problem with non-transpiled code appearing in the final bundle and breaking compatibility with old browser
- Moves to `eval-source-map` because why not go with the best quality if it's only for debugging (note: I tried debugging performance using `cheap-module-source-map` and it was utterly impossible, everything was mapped to the bundles xD)

### Related issue(s)  
https://github.com/nextstrain/auspice/issues/902

### Testing
A general smoke test should suffice probably... !

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -14,16 +14,37 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
   /* which directories should be parsed by babel and other loaders? */
   const directoriesToTransform = [path.join(__dirname, 'src')];
 
+  // Pins all react stuff, and uses hot loader's dom (can be used safely in production)
+  // Format is either "libName" or "libName:libPath"
+  const coreDeps = [
+    "react",
+    "react-hot-loader",
+    "react-dom:@hot-loader/react-dom"
+  ];
+
+  // Actively searches for the "good" root starting from auspice dir and going backwards
+  let baseDir = __dirname;
+  let foundNodeModules = false;
+  while (!foundNodeModules) {
+    foundNodeModules = true;
+    for (const coreDep of coreDeps) {
+      const modulePath = path.join(baseDir, "node_modules", coreDep.split(":")[1] || coreDep);
+      foundNodeModules = foundNodeModules && fs.existsSync(modulePath);
+    }
+    baseDir = foundNodeModules ? baseDir : path.resolve(baseDir, "..");
+  }
+  const libDir = path.join(baseDir, "node_modules");
+
   /* webpack alias' used in code import / require statements */
   const aliasesToResolve = {
     "@extensions": '.', /* must provide a default, else it won't compile */
     "@auspice": path.join(__dirname, 'src'),
-    "@libraries": path.join(__dirname, 'node_modules'),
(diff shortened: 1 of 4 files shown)


salvatore-fxpig:
The check fail triggered my curiosity a bit about the packaging. 

There's a couple of "easy" optimizations in there that could save a couple of k (`react-select`, `d3-range`), then a couple more complex ones (chunking away `react-ga` and the `locales` folder.

If you have strong concerns for bundle size, have you considered keeping a separate vendor bundle ? It would not affect worst-case loading time at all, but it would improve average loading time significantly.

salvatore-fxpig:
As alternative to loading modules: ["main" ...] one could babelize node_modules, lots of people doing that now, takes a bit more to compile but should be safer from the point of view of adding new extensions in the future, and saves a bit of package size as well.

webpack.config.js
```diff
@@ -18,7 +18,11 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
   const aliasesToResolve = {
     "@extensions": '.', /* must provide a default, else it won't compile */
     "@auspice": path.join(__dirname, 'src'),
-    "@libraries": path.join(__dirname, 'node_modules')
+    "@libraries": path.join(__dirname, 'node_modules'),
+    // Pins all react stuff to auspice dir, and uses hot loader's dom (can be used safely in production)
+    "react": path.join(__dirname, 'node_modules/react'), // eslint-disable-line quote-props
```
jameshadfield, line 23: When auspice is a dependency of a project X, then `__dirname` does point to the correct location (`X/node_modules/auspice`) however `react` is not present within ``X/node_modules/auspice/node_modules` for reasons unknown to me. 

I think this approach is the right one in order to ensure that customisations use the same libraries as "core" auspice. My original idea was to force customisations to import code via `import react from "@libraries/react"` but this also had problems (perhaps the same as we're having now).

salvatore-fxpig: I've had a look and while it feels like a bit of a "nuclear" option, explicitly looking for the "root" node_modules folder in the webpack config seems to do the trick. 

As for the hot loader thing, could you verify if #1126 has the same problem ? I remember resolving this but it was fairly entangled with the react-i18n module IIRC so if it works in #1126 I'd rather not do it twice.

salvatore-fxpig: Seems by some lucky combination the hot loader actually works well without using the hot-loader dom nor the patch/import, so I pushed a micro commit that reverts those changes and removes the regression.

A more extensive, hopefully more stable solution using the "recommended" implementation of react-hot-loader is found in #1126 and requires a bit of fiddling with react-i18n.

jameshadfield:
Going to close this, as #1126 is much more complete (albeit with a few tweaks needed).

webpack.config.js
jameshadfield: #1126 still isn't perfect, but i'd rather pursue the stable solution, even if we do it in a subsequent PR.

PR #1122, opened 2020-05-12
Rename test snapshots (Windows compatibility)

salvatore-fxpig:
Remove the ":" from the image snapshot paths because they break git/GitHub on Windows. This would improve development experience for folks using Windows.

### Related issue(s)  
The ":" in the snapshot paths makes for a fairly nightmarish experience for those of us who are running on GitHub Windows or Windows Linux Subsystem. It forces to do all commits and push on git cli (feasible of course, but pretty annoying). 

### Testing
If the checks pass it should be enough.


jameshadfield:
Seems fine to me! @tihuan is this ok from your perspective?

Issue #1106, opened 2020-05-05
Reduce bundle sizes by more intelligent code-splitting 


jameshadfield:
**Context**  
Auspice is delivered via a number of (gzipped) bundles using [webpack's code-splitting functionality](https://webpack.js.org/guides/code-splitting/). The smaller these are the more performant auspice loading feels.

**Description**   
Our bundle sizes are creeping up due, in part, to less-than-ideal code splitting (see 90d91e94). You can see the breakdown of bundles via `auspice build --analyzeBundle` and `npx bundlesize`, the latter of which is run as a GitHub CI for PRs.

**Examples**  
All language translations are part of the main (first) bundle, adding ~13kb of (gzipped) assets. These should be split & fetched individually as needed. There are plenty of others, see `auspice build --analyzeBundle`.

PR #1126, opened 2020-05-14
More webpack updates - package size optimizations and long term caching

salvatore-fxpig:

The goal of this pull request is providing a stable and definitive solution to transpilation of extensions, using babel to transpile all the code (app and dependencies). This increases the build time to around 75s which is still largely tolerable, considering that rebuild time in dev mode is not affected.

 while at the same time tackling the problem of package size creep by implementing a number of optimizations:
- Splitting packages in vendor bundles and appbundle for long term caching.
- Splitting away locales
- Splitting away some heavy dependencies that are used only on specific paths, notably `papaparse`, `dompurify`, `marked`, `react-ga`
- Using the optimized (smaller) version of `react-select`, excluding non-used parts of the library
- Fixing the default `@extensions` path to avoid the webpack import being bamboozled
- Reducing the package size of `d3-scale` by selectively including the used modules
- Reducing the package size of `lodash` by applying `lodash-webpack-plugin`

### Related issue(s)  
#1106 
#902 

### Testing
I tested the build, with nextstrain.org extensions enabled (but running as standalone), and removing the forced include of the transpiled version of iso-639-1 from the extension code, and it displays in IE11 and Safari 5.1 (the latter with some errors, the same on  current master build).

The size of the core app bundle  (i.e. the bundle that is downloaded when navigating to localhost:4000/ and then picking e.g. dengue) is reduced to 390kB (from 400kB), of which around 200kB can be cached on long term (i.e. change a couple times per year) and another 50kB can be cached for mid term (i.e. change once every couple months).

To make the nexstrain repo fully compatible with the present PR, a minimal update of the code is required to make nextstrain.org serve index.html from the dist/ folder, I opened a separate (backwards-compatible) PR on nextstrain.org.
https://github.com/nextstrain/nextstrain.org/pull/166

Some more cross-browser testing would probably be a good idea.


salvatore-fxpig:
It must be noted that absolute package size reduction is not that impressive mainly because full babel transpilation + some arrangements required to ensure real stable long term hashes end up bloating the package a fair bit.

salvatore-fxpig:
If this gets merged it's very important that somebody from the core team goes through the core-vendors configuration to make sure that those modules are really stable, both internally (i.e. you're happy with using same version of the module for a long time) and externally (i.e. the module is not going to be switched out).

webpack.config.js
```diff
@@ -84,6 +98,60 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
       path.resolve(__dirname, "dist");
   utils.verbose(`Webpack writing output to: ${outputPath}`);
 
+  /**
+   * Here we put the libraries that are unlikely to change for a long time.
+   * Every change or update of any of these libraries will change the hash
+   * of the big vendor bundle, so we must be sure they're stable both internally
```
jameshadfield, line 107: When this needs changing, the bundle hash will change and so will the script src (in `dist/index.html`), correct? And this means that the client will treat it as completely different request, and thus there's no danger of using the old, now invalid bundle from the cache?

If this is true, is there benefit to adding a (large) `max-age` to the response for these bundles, since if their contents change the hash will also change?

webpack.config.js
```diff
@@ -50,18 +53,29 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
...
     algorithm: "gzip",
-    test: /\.js$|\.css$|\.html$/,
-    threshold: 10240,
-    minRatio: 0.8
+    test: /\.(js|css|html)$/,
+    threshold: 4096
+  });
+  const pluginHtml = new HtmlWebpackPlugin({
+    filename: 'index.html',
+    template: './index.html'
```
jameshadfield, line 62: Since the template is unusable by itself, would it be more typical to move this to `src/index.html`?

webpack.config.js
salvatore-fxpig: Yes all of the above is correct, I didn't check how you're caching .js files atm on the public site but basically the whole point of this setup is you set index.html to be served with no no cache whatsoever (something like `no-cache,no-store,must-revalidate`), then you set nice and long (multiple months) caching for the .js part, and the hashes in the names of the scripts take care of cache busting for you. 

And when you build the index.html will generate automatically with the correct hashes, modifying only those that should be modified.

src/components/controls/language.js
```diff
@@ -16,8 +16,18 @@ import { CHANGE_LANGUAGE } from "../../actions/types";
   };
 })
 class Language extends React.Component {
+  async ensureLanguageResources(lang) {
+    for (const ns of ["language", "sidebar", "translation"]) {
+      if (!i18n.hasResourceBundle(lang, ns)) {
+        const res = await import(/* webpackMode: "lazy-once" */ `../../locales/${lang}/${ns}.json`); // eslint-disable-line
```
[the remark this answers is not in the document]
salvatore-fxpig: So basically you have 3 scenarios:

1) Without an explicit locale config in webpack.config.js, and with no `webpackMode: "lazy-once"`, each locale will create its own micro-chunk. This may seem a good idea at first glance because most people are only ever going to use one locale, however what happens is that various chunk merging plugins tend to merge these micro-chunks in very unpredictable ways and you get the locales bundled up with other stuff, defeating the purpose. So I would advise against this.

2) With `lazy-once` and the explicit single-package locale config (as is now), the `lazy-once` is basically ignored, it only serves as a safeguard against scenario (1) above.

3) If you decide to split locales further, e.g.
```
          locales1: {
            test: /\/src\/locales\//,
            name: "locales-others",
            enforce: true,
            priority: 1,
            chunks: "all"
          },
          locales2: {
            test: /\/src\/locales\/(es|pt|fr)\//,
            name: "locales-es.pt.fr",
            enforce: true,
            priority: 2, // This priority must be higher than the "locales-others chunk
            chunks: "all"
          },
```,
in that case you should absolutely *remove* the `lazy-once` directive because otherwise when the locales get lazy loaded, all the separate packages will be loaded at once, thus nullifying the benefit of splitting them.

File: webpack.config.js (hunk 2 above)
Comment (salvatore-fxpig) [reply webpack.config.js:62]: Yep I think that could be a good idea 👍

**Comment by salvatore-fxpig** [comment]
I added the "new" node_modules resolution from #1120 in a slightly different flavor, virtually this implementation is more robust, in practice it's gonna be the same for 99.99% of practical cases.

**Comment by salvatore-fxpig** [comment]
In this latest commit I updated the bundling strategy with another tiny bit: using as entry point a micro-async entrypoint that only imports (asynchronously) the main, real entrypoint, and is hashed with [contenthash].

The reasoning for this is as follows:
If long term caching policies are applied, it will happen in some builds that one component chunk is updated (e.g., only the <Main> component is updated so its chunk alone is updated), while all the rest of the code and importantly all of the code contained in the entrypoint bundle is left untouched. 

Now, in webpack the URLs of the chunks are always stored in the entrypoint (the main, non-numbered bundle). This means that we have this situation:

1) If we use an algorithm that ignores the additional webpack code when calculating the hash (e.g. md5 `[chunkhash]`), the entrypoint bundle will keep the same hash, but people with a cached copy will then have an entrypoint bundle that points to a js chunk that does not exist anymore, and the site will break for them until they CTRL+F5
2) If we use for the entrypoint bundle an algorithm that considers everything in the bundle, including the (modified) webpack import code, like `[contenthash]`, then every time we change code anywhere in the codebase, _both_ the hash of the chunk that includes that code _and_ the entrypoint bundle hash will change, forcing everybody to download the entrypoint bundle (47kB gzipped) again.

The solution: having an ant-sized entrypoint bundle (<1.5kB) that basically just contains the updated URLs of the chunks at every build, and is `[contenthash]`ed so that everytime that any of the chunk hashes change, its hash changes, updating for everyone, so that caching is intact and overconsumption of bandwidth is minimal.

**Comment by salvatore-fxpig** [comment]
To be noted that the bundlesize test, which hardcodes the vendor bundle hashes, has been passing nicely through these latest modifications which hit both the entrypoint and the webpack configuration, so we can see that they're decently stable... !

Due to the hardcoding of the hashes in the bundlesize configuration, it means the bundlesize test will fail if for some reason the vendor bundles for long term hashing are unwittingly changed. 
Example: in a commit before, without realizing it, I had added back the 
```
resolve: {
      mainFields: ['browser', 'main', 'module'],
...
}
```
from the other PR.
The build was still successful, but obviously the whole code (and hashes) changed, and the bundlesize test started to fail, which helped me figure out what was wrong.

File: src/components/controls/language.js [hunk 4]
```diff
@@ -16,8 +16,18 @@ import { CHANGE_LANGUAGE } from "../../actions/types";
...
+        i18n.addResourceBundle(lang, ns, res.default);
+      }
+    }
+  }
 
-  componentWillMount() {
+  async componentWillMount() {
+    if (!this.props.language || this.props.language === "en") return;
```
jameshadfield, line 29: Not all translate keys return the same string value in english -- `t(x) === x` is mostly, but not always, true. I think the current implementation deliberately doesn't load the English local, instead utilizing the fact that i18n will fall back to the provided strings. This results in errors in only a few places, for instance clicking on the "Download data" modal produces 

image

I think an acceptable solution would be to let the English locale be bundled wherever webpack wants to put it (it's <2kb gz) but keep the other languages in a separate bundle which is only fetched when necessary (as done now).

webpack.config.js
jameshadfield: Great! While nextstrain.org uses a custom server, I know there are other groups using `auspice view` as a production server, so we should set these here. Playing around with it, it _seems_ as simple as 

```diff
- res.sendFile(path.join(auspiceBuild.baseDir, "dist/index.html"))
+res.sendFile(path.join(auspiceBuild.baseDir, "dist/index.html"), {headers: {"Cache-Control": "no-cache, no-store, must-revalidate"}});
```
and 
```diff
- app.use("/dist", expressStaticGzip(auspiceBuild.distDir));
+ app.use("/dist", expressStaticGzip(auspiceBuild.distDir, {maxAge: '365d'}));
```

But wanted to get your thoughts here, as the nextstrain.org server is similar to the auspice server here.

webpack.config.js
jameshadfield: The current change means that this doesn't build on nextstrain.org

```
This doesn't build when auspice is run as a project's dependency with extensions. For instance, on nextstrain.org we get
```
internal/fs/utils.js:230
    throw err;
    ^

Error: ENOENT: no such file or directory, open '/Users/naboo/github/nextstrain/nextstrain.org/node_modules/auspice/index.html'
    at Object.openSync (fs.js:457:3)
    at Object.readFileSync (fs.js:359:35)
    at Object.exportIndexDotHtml (/Users/naboo/github/nextstrain/nextstrain.org/node_modules/auspice/cli/utils.js:112:17)
    at /Users/naboo/github/nextstrain/nextstrain.org/node_modules/auspice/cli/build.js:57:15
 ...
```

I think this is relatively straightforward to fix by updating `exportIndexDotHtml()`

Comment (salvatore-fxpig) [reply webpack.config.js]: Actually with index.html being outputted in the /dist folder, exportIndexDotHtml() becomes completely obsolete. As far as I can see, so I think it can be killed.

File: src/components/controls/language.js (hunk 4 above)
Comment (salvatore-fxpig) [reply src/components/controls/language.js:29]: Uh, I had completely missed this. Yes I think your solution is best, I modified a tiny bit index.js and webpack.config.js to accommodate for it.

**Comment by salvatore-fxpig** [comment]
I must walk back on my initial claim that the modified `node_modules` searching function in webpack.config.js wouldn't make a difference in 99.9% of cases... It turns out that actually such a package conflict exists with the `master` version of nextstrain.org and as such the new searching function (which I further amended to solve a bug in it) turns out to be absolutely essential!

File: webpack.config.js (hunk 2 above)
Comment (salvatore-fxpig) [reply webpack.config.js]: In the interest of keeping the risk of breaking other moving parts at a minimum, it's still probably better to updated exportIndexDotHtml. This new updated version I pushed is compatible with nextstrain.org `master` as well.

**Comment by salvatore-fxpig** [comment]
> The resulting bundle sizes are pretty amazing and i'm really excited to get this out as it'll make a big difference.
> 
> The hot reloading ability of `auspice develop` seems to have gone with this PR (it was never perfect, but it's become worse here). The bundle is still being rebuild (the server prints messages to this effect, and reloading the page reflects changes) but the browser doesn't auto-update with changes. We can potentially fix this after merge if needs must.

I managed to make (most of) hot reload work again altho it's not exactly a best-in-class solution, it should suffice while we try to figure out why it does not work as expected in the first place.

File: webpack.config.js (hunk 1 above)
Comment (salvatore-fxpig) [reply webpack.config.js:128]: Sorry for forgetting to reply here, yep it should suffice as far as I can tell

**Comment by rleir** [comment]
@salvatore-fxpig Salvatore: I did a pull of master yesterday, and I must have got my webpack mixed up somehow. Now the build takes infinite time at 100% cpu, it puts out 
```
$ auspice build --verbose
[verbose]	Generating webpack config. Extensions? false. devMode: false
```
It is past 10 minutes now.  Oh, and I am doing this on Ubuntu. The mac-specific fsevents needs to be removed on Linux, maybe that is a factor. 
You know Webpack much better than I do. How would you debug a webpack infinite loop? 
thanks
Rick

rleir:
@salvatore-fxpig Salvatore: My bad, I had a newer version of node installed. I reverted to the specified version and the problem went away. 
Sorry for troubling you -- Rick

PR #1127, opened 2020-05-15
Add detailed JSON format descriptions from nextstrain.org docs

eharkins:
Moving the detailed file descriptions from the [nextstrain.org page](https://nextstrain.org/docs/bioinformatics/data-formats) into the auspice docs. That way auspice docs host the detailed info on file formats and it can be linked in / injected into nextstrain.org docs as necessary and they don't get out of sync.
 
### Related issue(s) 
https://github.com/nextstrain/nextstrain.org/pull/164
https://github.com/nextstrain/nextstrain.org/pull/168

### Testing
Tested links work locally with docusaurus.

### DIFF
--- a/docs-src/docs/introduction/how-to-run.md
+++ b/docs-src/docs/introduction/how-to-run.md
@@ -72,47 +72,47 @@ Right now, `auspice view` will automatically convert "v1" JSONs into "v2" JSONs,
 
 ## Input File Formats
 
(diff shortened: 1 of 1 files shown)


eharkins:
Added a similar commit to https://github.com/nextstrain/nextstrain.org/pull/168 to emphasize that v1 should only be used for backward compatibility

PR #1128, opened 2020-05-15
Svg style diet

rleir:
Memory consumption is reduced significantly.  Please see #955 

But this PR is experimental and is not ready for merging.

### Related issue(s)  

Fixes #  
Related to #955  Perfomance Audit

### Testing
* Shift-esc, check the memory footprint. 
* right click -> inspect and check the CSS
* check the console (auspice is running with --includeTiming)
* visual check

Yes I would like help testing this.


jameshadfield:
@rleir do you have some stats regarding how much memory is reduced for each of the major d3 style changes in this PR? Below you can see a screenshot comparing this PR (bottom) with master (top). We use branch thickness to convey the number of descendants of a branch, but I could imagine stripping this out once a tree gets big enough if performance is greatly increased.

image

rleir:
Hi James @jameshadfield 
> We use branch thickness to convey the number of descendants of a branch,

Yes, I want to preserve that. I have difficulty following the classes 'branch' and 'T' and 'S' through the code. Please help!

Your screenshot does not show the most recent patch, which saves memory but is less visually appealing.
How much memory is saved? I find this is difficult to measure due to the large allocations at startup followed by a large garbage collection. And the dev mode hot reloading might complicate this. The savings might be less than 1%. 

You must have talked about a scalability problem, look at how the number of tips grows over time. Perhaps the answer is, as you have done, to split the data by continents. But is that not a stop-gap solution? 
Other stop-gaps might be:
* Light-weight card data. Get the full data when needed by AJAX.
* Colours in the DOM. Use hex notation instead of RGB.
* Other CSS improvements to reduce element styling
* As Trevor suggested, less precision in the numbers
* Find any leaks
* Agregation of some sort. Zoom in to get correct details. I am being intentionally vague.

trvrb:
> We use branch thickness to convey the number of descendants of a branch, but I could imagine stripping this out once a tree gets big enough if performance is greatly increased.
Interesting thought. I do remember that branch thickness requires recalculating layout, which is expensive. But I do think that the branch thickness is particularly helpful when filtering to a subset of tips in a large tree. Though I guess you could experiment with this highlighting via opacity instead of branch thickness.

jameshadfield:
Closing in favor of #1142 which includes this 👍

PR #1129, opened 2020-05-15
Handle non-valid dataset paths

salvatore-fxpig:
Currently the dataset selector leads to invalid paths when used for example to switch to dengue, (because /dengue does not exist, /dengue/denv1 does, and /dengue/denv2 etc). 
On the current master build of auspice this is not handled, on nextstrain.org this is handled with a res.redirect on the getDataset endpoint.
Additionally the selector switches dataset even when same dataset is reselected (useless loading).

With this PR, the selector automatically loads fully valid paths, paths identical to the current path are not reloaded, and incomplete paths are handled by auspice's server directly.

### Testing
In dev mode, test for example 
http://localhost:4000/dengue
http://localhost:4000/zika/nothanksjeff
http://localhost:4000/VirusThatDontExist

Also try fiddling with Dataset Selector
image

### DIFF
--- a/cli/server/getDatasetHelpers.js
+++ b/cli/server/getDatasetHelpers.js
@@ -47,18 +47,25 @@ const interpretRequest = (req) => {
 
 /**
  * Given a request, does the dataset exist?
- * In the future, if there is no exact match but a partial one we
- * should extend this. E.g. `["flu"]` -> `["flu", "h3n2", "ha", "3y"]`
- * In that case, we should utilise `res.redirect` as we do in the nextstrain.org server
- * @throws
  */
-const extendDataPathsToMatchAvailable = (info, availableDatasets) => {
-  const requestStrToMatch = info.parts.join("/"); // TO DO
-  /* TODO currently there must be an _exact_ match in the available datasets */
-  if (!availableDatasets.map((d) => d.request).includes(requestStrToMatch)) {
-    throw new Error(`${requestStrToMatch} not in available datasets`);
+const extendDataPathsToMatchAvailable = (res, info, availableDatasets) => {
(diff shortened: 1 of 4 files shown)


cli/server/getDatasetHelpers.js
```diff
@@ -47,18 +47,25 @@ const interpretRequest = (req) => {
...
- * should extend this. E.g. `["flu"]` -> `["flu", "h3n2", "ha", "3y"]`
- * In that case, we should utilise `res.redirect` as we do in the nextstrain.org server
- * @throws
  */
-const extendDataPathsToMatchAvailable = (info, availableDatasets) => {
-  const requestStrToMatch = info.parts.join("/"); // TO DO
-  /* TODO currently there must be an _exact_ match in the available datasets */
-  if (!availableDatasets.map((d) => d.request).includes(requestStrToMatch)) {
-    throw new Error(`${requestStrToMatch} not in available datasets`);
+const extendDataPathsToMatchAvailable = (res, info, availableDatasets) => {
```
jameshadfield, line 51: This is great! Could you please rename the function to better reflect the functionality -- along the lines of https://github.com/nextstrain/auspice/commit/fc824fcc6736a8263c3d8f7d9d47a78d0b3db352 (can just cherry-pick that if you're happy with it)

salvatore-fxpig: The cherry-pick merged without any conflict so I kept it 👍

salvatore-fxpig: For the rest, the new commit should take care of the changes requested. 

I wrote in a slightly different, more compact form but if you prefer I can rewrite it with a for cycle, it may be better for readability.

Issue #738, opened 2019-07-16
genotype color-by not shown in sidebar dropdown on page load


jameshadfield:
When setting a genotype color-by via a URL (e.g. go to https://nextstrain.org/zika?c=gt-nuc_255), the colors are correct in the app, but the sidebar color-by dropdown hasn't updated to reflect this state (it shows the blank "Select..." placeholder). This issue is not present for non-genotype color-bys.

charlie-jones:
@jameshadfield I would be happy to work on this. It seems like the solution is to call gtGeneSelect() and gtPositionInput() and fill them in respectively; my proposal is to split the value  by "-" and "_" if the first two characters are "gt".

tsibley:
@charlie-jones There already exist functions to parse the serialized URL parameter `c`. Values from the "controls" Redux store flow into the ColorBy component via props. The issue here may be that the "controls" Redux store isn't being properly initialized from the URL param.

charlie-jones:
@tsibley I appreciate the advice, thank you! The initialization of the store was, in fact, the problem, so I will submit a PR after finishing up testing.

jameshadfield:
This issue is no longer -- perhaps closed by the above linked PR?

PR #1131, opened 2020-05-17
genotype color-by not shown in sidebar dropdown on page load

charlie-jones:
The goal of this pull request is to enable the color-by component to initialize from a URL parameter (c) when it denotes a genotype (for other categories the initialization has worked). From this PR, the redux store in color-by.js is initialized appropriately on page load. 

### Related issue(s)  
#738 

### Testing
In dev mode, the color-by genotype can be specified like this:
http://localhost:4000/ncov/north-america?c=gt-ORF1b_1000
Expected: 3 input fields are initialized with "genotype" (in color-by field), "ORF1b", and "1000". Any genotype with a position in the range should yield a similar result.

### DIFF
--- a/src/components/controls/color-by.js
+++ b/src/components/controls/color-by.js
@@ -32,9 +32,21 @@ class ColorBy extends React.Component {
       positionSelected: ""
     };
 
-    this.state = this.newState({
-      colorBySelected: props.colorBy
-    });
+    if (isColorByGenotype(props.colorBy)) {
+      const genotype = decodeColorByGenotype(props.colorBy);
+
+      if (genotype) {
+        this.state = this.newState({
+          colorBySelected: "gt",
+          geneSelected: genotype.gene,
+          positionSelected: genotype.positions.join(",")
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @charlie-jones -- tested & working well.

There's a small possible improvement that we could make in the future where currently loading positions which don't exist (i.e. pos 10,000 in a gene with only ~2700) show up in the sidebar when they shouldn't (e.g. try https://auspice-colorby-init-bay7w5rae.herokuapp.com/ncov/global?c=gt-ORF1b_314,10000). But the actual coloring of the tree isn't wrong, and neither is the legend, so I consider this a trivially minor bug which doesn't need fixing right now. Loading with a single invalid position results in the default color-by as expected. 

I'll look into why the smoke-tests are failing and then merge this in 👍

charlie-jones:
@jameshadfield On the improvement I agree 100%. And, thank you for developing this project; it is practical and especially relevant!

PR #1137, opened 2020-05-24
Precision diet2

rleir:
Hi James @jameshadfield 
It reduces the precision of svg path values as Trevor suggested.
Thanks -- Rick
The path values are full precision Javascript floats,
but when stored in the DOM in decimal text format they bloat the DOM.
This only matters when the number of paths is in 4 figures.
So, need to do a Math.round() on the values before constructing "M n,m" paths.
Likewise with stroke widths.
Note: I did not bother with the regression line, because there is only one of them.
Please test this on a Macbook, Apple does tricks with resolution!
### Related issue(s)  

Fixes #  
Related to #955 Perfomance Audit
See also #1136 PrecisionDiet

### Testing
* Shift-esc, check the memory footprint.
* right click -> inspect and check the CSS
* check the console (auspice is running with --includeTiming)
* visual check

Yes I would like help testing this.


jameshadfield:
Closing in favor of #1142 which includes this 👍

PR #1140, opened 2020-05-26
Add placeholder license for visualizations / graphics. 

sidneymbell:
We get many requests to reuse / alter / publish / distribute the visualizations from Nextstrain. A license makes these conversations much easier and more consistent.

For now, I've put in a CC-BY-SA license as a placeholder to attempt to mirror the GNU Affero General Public License v3.0 used for this repository. 

I'd suggest moving to a CC-BY if the team is comfortable with this.

### DIFF
--- a/src/components/download/downloadModal.js
+++ b/src/components/download/downloadModal.js
@@ -191,7 +191,7 @@ class DownloadModal extends React.Component {
         (<MetaIcon width={iconWidth} selected />), () => helpers.authorTSV(this.props.dispatch, filePrefix, this.props.tree)]);
     }
     buttons.push(
-      ["Screenshot (SVG)", "Screenshot of the current nextstrain display in SVG format; CC-BY-SA licensed.",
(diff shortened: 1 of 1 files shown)


sidneymbell:
Cool. I put it as a draft simply because I wasn't sure if we preferred
cc-by or cc-by-sa. Votes?

rneher:
What do you reckon the main difference between CC-BY and share-alike would be? Am I correct that media outlets then would have to make those articles/graphics open and available under CC-BY-SA?

tsibley:
@rneher Correct. The difference is the ShareAlike clause:

> ShareAlike — If you remix, transform, or build upon the material, you must distribute your contributions under the same license as the original. 

so any derivative graphics, figures, etc. would have to also be CC-BY-SA. This might have a chilling effect on usage, or it might not impact it at all.

trvrb:
I'd vote for CC-BY for graphics / screenshots, etc... I strongly suspect many media outlets would not be able use a CC-BY-SY license.

PR #1149, opened 2020-05-30
Add opt-in genome download feature

### LINKED ISSUE #1110: Download subset of genome sequences for selected tree nodes
**Context**  
Similar to download a subset of metadata, we want to use this to extract a subset of genome sequences for further analysis. This might not be helpful or allowed in the global nextstrain/auspice instance, but for our local one, it is legal and useful feature to have. 

**Description**   
This feature should work almost exactly like extract subset of metadata. 

**Possible solution**   

- `augur export` need to export and include genome sequences upon user choice (i.e. an `--export-genomes` flag)
- genomes are saved in the same directory as auspice's `datasetDir`.
- if `auspice` can find files endswith `-sequences.json` it present  the download subset of sequences button in the `Download Data` popup window. 

An observation: auspice removed handling of `sequences.JSON` at version 1.8.0. Probably this feature is related to the code of handling sequences.

biocyberman:
When a corresponding FASTA file is placed in the same directory with auspice's
JSON file, for example a `ncov_global.fasta` is placed in the same directory with
`ncov_global.json`: the genome database handling functionality will kick in and
generate a file-based NeDB database at server startup time. A "Selected
Genomes (Fasta)" download button also then appears in side the "Download Data"
popup window. For dataset that doesn't have a corresponding FASTA file, this
functionality is inactive.

### Related issue(s)  

Fixes #1110 

 

1. Apply this feature to the code base
2. Installation again as normal. 
2. Place a FASTA file `ncov_global.fasta` to `datasetDir` (i.e. ./data) for testing with `ncov_global.json`. Start or restart `asupice`. Fasta record IDs inside the FASTA file and tree tip names should match for auspice to fetch the correct genome from the database. 

>If you added or changed behavior in the codebase, did you update the tests, or do you need help with this?  

None.

### Note:
[Upgrading](https://github.com/biocyberman/auspice/blob/feature-download-genomes/package.json#L100)  `papaparse` to newer version is not required for this feature to function properly, but I upgraded that in response to [severe security warning](https://www.npmjs.com/advisories/1515) by `npm` during installation process.

### DIFF
--- a/cli/develop.js
+++ b/cli/develop.js
@@ -10,6 +10,7 @@ const version = require('../src/version').version;
 const chalk = require('chalk');
 const generateWebpackConfig = require("../webpack.config.js").default;
 const SUPPRESS = require('argparse').Const.SUPPRESS;
+const bodyParser = require('body-parser');
 
 const addParser = (parser) => {
   const description = `Launch auspice in development mode.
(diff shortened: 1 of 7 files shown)


biocyberman:
@jameshadfield May be you or your teammates could check this out so we can merge or have feedback? 
Thanks.

PR #1150, opened 2020-06-01
Have Legends Dynamically Change to only Show Visible Options

### LINKED ISSUE #1109: Have Legends Dynamically Change to only Show Visible Options
**Context**  
Currently for some builds, like nextstrain.org/ncov, the legends are currently so long that they are not very useful:
image
(Countries shown, but valid for many of the selections)

@MoiraZuber wrote a tweet for a recent push to nextstrain.org where she had only 'new' sequences selected via filters. She manually adjusted the legend so that it only showed entries for the nodes that were currently visible, with the filters selected. 
pic_1 (1)
('division' coloring on the nextstrain.org/ncov/north-america build shown)

This makes the legend suddenly incredibly useful again!

**Description**   
When filters are applied, only values that are present on the currently visible nodes should be shown on the legend. Note that the colours do not *change* - just the number of things shown in the legend is reduced to only be the tips visible.

However, for traits where ancestral reconstruction has taken place (like 'countries' on the main nextstrain.org/ncov build), and branches are coloured, it should be weighed up whether the traits on visible *branches* should also be included in the traits shown on the legend.  Happy to discuss the pros and cons of this with anyone interested in implementing - my gut feeling is that it would be worth including branches that lead to visible nodes (these are already identified in the visualisation by being thicker), but if difficult to implement, or on average introduces too many other traits to the legend, this might not be worth it.

mingir2:
Dynamically changes the visibility of legend values for categorical and ordinal scale types. 

Before

Screen Shot 2020-06-01 at 5 08 57 PM

After

Screen Shot 2020-06-01 at 5 08 33 PM

```/flu/seasonal/h3n2/ha/2y:flu/seasonal/h1n1pdm/ha/2y?dmax=2018-02-15&dmin=2015-11-05```

### Related issue(s)  

Closes #1109 

### Testing

I tried to fix potential issues related to having a second tree, but there should be more rigorous testing to make sure it doesn't fail.

### DIFF
--- a/src/util/colorScale.js
+++ b/src/util/colorScale.js
@@ -2,7 +2,7 @@ import { scaleLinear, scaleOrdinal } from "d3-scale";
 import { min, max, range as d3Range } from "d3-array";
 import { rgb } from "d3-color";
 import { interpolateHcl } from "d3-interpolate";
-import { genericDomain, colors, genotypeColors, isValueValid } from "./globals";
+import { genericDomain, colors, genotypeColors, isValueValid, NODE_VISIBLE } from "./globals";
 import { countTraitsAcrossTree } from "./treeCountingHelpers";
 import { getExtraVals } from "./colorHelpers";
 import { isColorByGenotype, decodeColorByGenotype } from "./getGenotype";
@@ -42,6 +42,28 @@ const getDiscreteValuesFromTree = (nodes, nodesToo, attr) => {
   }
   const domain = sortedDomain(Array.from(stateCount.keys()).filter((x) => isValueValid(x)), attr, stateCount);
   return domain;
+};
+
+export const createVisibleLegendValues = ({colorBy, scaleType, legendValues, legendBounds, nodes, visibility}) => {
+  console.log("createVisibleLegendValues", scaleType, legendValues, legendBounds, nodes, visibility);
+  // filter according to scaleType, e.g. continuous is different to categorical which is different to boolean
+  // filtering will involve looping over reduxState.tree.nodes and comparing with reduxState.tree.visibility
+
+  if (!visibility) {
+    console.warn("to debug - race condition on loading");
+    return legendValues.slice();
+  }
+
+  if (scaleType === "categorical") {
+    const legendValuesObserved = new Set(
+      nodes.filter((n, i) => (!n.hasChildren && visibility[i]===NODE_VISIBLE))
+        .map((n) => getTraitFromNode(n, colorBy))
+    );
+    return legendValues.filter((v) => legendValuesObserved.has(v));
+  }
+
+  // for testing non-categorical scales, i'm just filtering the list into odd entries
+  return legendValues.filter((x, i) => i%2);
 };
 
 const createDiscreteScale = (domain, type) => {
@@ -273,6 +295,9 @@ export const calcColorScale = (colorBy, controls, tree, treeToo, metadata) => {
     colorScale = () => unknownColor;
   }
 
+  // Would make more sense to calculate earlier in the function
+  const scaleType = genotype ? "categorical" : colorings[colorBy].type;
+
   return {
     scale: colorScale,
     continuous: continuous,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by jameshadfield** [comment]
Thanks @mingir2! I'll have a closer look tomorrow. The recalculation of visibility seems problematic in that (i) it's incomplete (doesn't take into account filters other than time) and (ii) is already being done on a page load, but apparently after the scales are initially calculated and (iii) it's an expensive function to rerun. I think that instead we could return-early from `createVisibleLegendValues` if visibility is not set and rerun it from `recomputeReduxStateFromJSONs...` once the visibility arrays have been calculated. 

P.S. Great work incorporating the second tree!

File: src/util/colorScale.js [hunk 1]
```diff
@@ -45,6 +46,46 @@ const getDiscreteValuesFromTree = (nodes, nodesToo, attr) => {
...
+ * Dynamically create legend values based on visibility for ordinal and categorical scale types.
+ */
+export const createVisibleLegendValues = (
+  {colorBy, type, legendValues, tree, treeToo = undefined, controls = undefined}
+) => {
+  let visibility = tree.visibility;
+  let visibilityToo = treeToo ? treeToo.visibility : undefined;
+
+  // Create visibility if undefined.
+  if (!visibility || !visibilityToo) {
```
jameshadfield, line 59: For single trees, `visibilityToo` is falsey and so this if block is run, which isn't what you want!

jameshadfield:
@mingir2 please see c3e9a5271d85a7abc2bf01bb4b435431e6af0392 which solves the race condition when loading in a way which avoids any recalculation of visibility within the function. 

Note that that WIP branch doesn't address any of the second tree code that you've nicely addressed here :)

mingir2:
@jameshadfield Thank you for the feedback. By your suggestion, I removed the visibility calculation from ```createVisibleLegendValues``` and merged the changes made in c3e9a5271d85a7abc2bf01bb4b435431e6af0392.

For some reason, changing filters doesn't result in immediate UI update for the legend values. It's consistently delayed by a single update. All relevant functions for calculating the legend values are executed as it should, at least the ones I'm aware of.

Also, reloading the page will result in complete legend values including non-visible ones. Changes made in ```src/actions/recomputeReduxState.js``` - ```createStateFromQueryOrJSONs``` doesn't seem to address this.

mingir2:
@jameshadfield Sorry for the delay, but the visible legend values delay is fixed now(pun intended). 

> For some reason, changing filters doesn't result in immediate UI update for the legend values. It's consistently delayed by a single update. All relevant functions for calculating the legend values are executed as it should, at least the ones I'm aware of.

The issue for this was that when `updateVisibleTipsAndBranchThicknesses` was called, `createVisibleLegendValues` was being called with an old version of `visibility` array. That's why it was exactly off by one.

> Also, reloading the page will result in complete legend values including non-visible ones. Changes made in src/actions/recomputeReduxState.js - createStateFromQueryOrJSONs doesn't seem to address this.

For this bug, there was a typo in the function parameter, so it was undefined and `createVisibleLegendValues` simply returned a `.slice()` of `legendvalues`.

emmahodcroft:
@mingir2 thank you for all this great work! @jameshadfield this would be a great feature to have for the Nextstrain tweets. I'd be happy to just do some 'use' testing, if we could have this deployed somewhere. Is that possible? Could I merge with a new branch here to do that, or best to just keep it local?

trvrb:
Thanks so much for the contribution @mingir2! I've created a review app at https://auspice-wip-dynamic-leg-zlqeh7.herokuapp.com/. Looks good based on simple testing (but you should look as well @emmahodcroft). I'll leave it up to @jameshadfield as to preferred merge strategy.

emmahodcroft:
Thank you so much Trevor! Unfortunately somehow I totally missed this until now... If we'd like more testing, could we set it up again? (Seems to be dead now). However, if all looks good, I'd be happy for @jameshadfield to merge if he feels it's ready!
Thanks all!

PR #1156, opened 2020-06-05
Keep sources of the root of the zoomed-subtree in sync. Closes #1139

### LINKED ISSUE #1139: Cannot zoom out in a stepwise fashion when a tree is loaded with a branch label in the URL query
**Current Behavior**  
When a tree is loaded with a branch label defined in the URL query, then we correctly display the subtree associated with that label. Clicking the root branch of the displayed subtree should zoom out in the tree, however it has no effect in this situation. 

**Expected behavior**  
Clicking on the root branch of the displayed subtree should zoom out in the tree.

**How to reproduce**  
1. Open https://nextstrain.org/staging/test/tree-labels-2020-05-26
2. Click on the branch labelled "string-value" to zoom into the subclade. 
3. Click on the root branch of the subtree (the left-most bit) & observe that we zoom out in the tree.
4. Load https://nextstrain.org/staging/test/tree-labels-2020-05-26?label=test:string-value and observe that we start with the same app-state as in step 2.
5. Click on the root branch of the subtree (the left-most bit). We expect to have the same effect as in step 3, however nothing happens.

Note that pressing "reset layout" does correctly zoom out to show the entire tree.

Auspice 2.15.0

jameshadfield:
We keep track of the root-node of the currently selected subtree within auspice via two pieces of state: `idxOfInViewRootNode` (redux state) and `zoomNode` (phylotree state). They were out-of-sync when loading a tree in a zoomed state, which resulted in the inability to zoom out via clicking on a basal branch. This commit fixes this bug. 

Long term, it would be better to only store this state in one place. This would involve some refactoring in order to expose the relevant redux state to callbacks such as `onBranchClick`.

### DIFF
--- a/src/components/tree/phyloTree/phyloTree.js
+++ b/src/components/tree/phyloTree/phyloTree.js
@@ -11,7 +11,7 @@ import * as confidence from "./confidence";
 import * as labels from "./labels";
 
 /* phylogenetic tree drawing function - the actual tree is rendered by the render prototype */
-const PhyloTree = function PhyloTree(reduxNodes, id) {
+const PhyloTree = function PhyloTree(reduxNodes, id, idxOfInViewRootNode) {
   this.grid = false;
   this.attributes = ['r', 'cx', 'cy', 'id', 'class', 'd'];
   this.params = createDefaultParams();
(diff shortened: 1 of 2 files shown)


jameshadfield:
> perhaps the dataset needs explicitly labeled branches to do that?

Correct, the dataset JSON itself defines branch labels (a common label key is "clade"). 

> Should I open another issue for the long term state consolidation here?

I feel as this should be noted somewhere, but not as an issue. This is an example of the broader topic of phylotree-state vs redux-state. Sometimes they are duplicated (like here), and sometimes they are separate but in complex ways. Any attempts to address this should be part of a higher-level direction decision.

PR #1163, opened 2020-06-10
Sample uncertainty

jameshadfield:
As part of thinking about ways we can express uncertainty I've been experimenting with resampling trait values. This PR allows you to resample the currently selected color-by or geo-resolution from the confidence data made available in dataset JSONs. The UI is via keyboard shortcuts as it's a experiment / proof-of-concept at this stage. 

`s` + `c` : resample color-by trait
`S` + `C`: reset color-by values to original state
`x`: animate the sampling of color-by values. Press a second time to end animation.
`s` + `r` : resample geo-res trait
`S` + `R`: reset geo-res values to original state
`e`: animate the sampling of geo-res values. Press a second time to end animation.

@trvrb @rneher is sampling of each node independently from their provided confidences valid? For discrete traits it's a simple sampling from a discrete distribution, for continuous values (e.g. date) I'm sampling uniformly from the provided interval. 

animate_map
_Zika dataset resampling the country (across internal nodes)_

animate_tree
_nCoV (global) dataset, resampling dates of internal nodes_

I'll say now that I don't think this is the best way for us to represent uncertainty (which is still unknown), but has been useful for me to interpret aspects of datasets and, if the appropriate UI can be achieved, may be useful for others.


trvrb:
Very cool idea! Quick thoughts:

Claus Wilke has argued that the most intuitive way to represent uncertainty for most audiences is exactly what you're doing here. Animate across realizations. https://docs.google.com/presentation/d/1zMuBSADaxdFnosOPWJNA10DaxGEheW6gDxqEPYAuado/edit#slide=id.p40

Separately, we could think about adding opacity to the transmission lines proportional to uncertainty (I guess this would need to be some function of uncertainty of parent and child states). Just doing the animation here for Zika made it clear that some transmission lines are much more supported than others.

>is sampling of each node independently from their provided confidences valid?

The provided CIs represent marginal distributions. If you draw a particular value for the one node it will change the distributions at other nodes. In this example from Zika:

zika

Marginal for these two branches is Thailand / French Polynesia. It's unclear when the transition is. But if you sample the parent as French Polynesia, the child should definitely be French Polynesia as well. If you sample independently you risk distorting inference. In this case you could easily sample the unsupported French Polynesia to Thailand to French Polynesia reconstruction.

Thus the animation will be appropriate for individual transmission lines but not their collection.

rneher:
How does one trigger resampling of timings in the tree?

regarding the marginal distributions: there are the probabilities at a node after summing over all states at all other nodes. using their maxima as transitions is not technically correct, but will in almost all cases correspond to the most-likely transition (within the model) and this is what we do now. guaranteed consistent would be the joint ML inference, but giving confidence for this is more tricky. 

Resampling the marginal distributions is indeed problematic. We could calculate the joint distribution of adjacent nodes in augur and add them as branch attributes.

PR #1167, opened 2020-06-12
Enable narratives to source multiple builds - Add Ons

salvatore-fxpig:
Just adding some magic sauce to #1164

Dunno if you've pulled in my fork, this may make it easier on your end (or not!)

### DIFF
--- a/cli/server/getDataset.js
+++ b/cli/server/getDataset.js
@@ -12,7 +12,9 @@ const setUpGetDatasetHandler = ({datasetsPath}) => {
       await helpers.sendJson(res, info);
     } catch (err) {
       console.trace(err);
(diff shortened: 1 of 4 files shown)


jameshadfield:
This looks really good to me (I haven't actually faked a slow connection to test the case where we change to a narrative slide who's cached json is a pending promise, but it's within an `async` function so it should work). I'm also a fan of avoiding `lodash` functions (or to any similar library) when we can accomplish the same functionality using `reduce`, `map` etc (or avoid it altogether as done here 😉 ). @eharkins I'll leave it to you to merge into your PR as you see fit.

eharkins:
Merged into #1164, thank you @salvatore-fxpig!

PR #1169, opened 2020-06-15
[bugfix] allow narratives to simultaneously zoom and change color

### LINKED ISSUE #1168: Branch Colors Don't Seem to Change in Narrative
**Current Behavior**  
When changing the color-by in a narrative, the branches remain the same as the original colouring of the tree when the narrative is loaded.

Here is a narrative loaded up with 'clades' colouring - but then it switches to 'division' and 'nucleotide' colouring - however the branches remain the same colour.

branch_color_narratives

Here, I loaded near the end when colouring is by 'nucleotide' - as we go backwards we can see the nucleotide colouring remains on the branches.
branch_color_narratives2

**Expected behavior**  
The branch colours should change along with the tip colours when changing color-by in the narratives.

**How to reproduce**  
Can be reproduced with any narrative where color-by changes. Can be see (not super obviously, but it's there) in slides 4 & 5 of this narrative:
https://nextstrain.org/groups/swiss/narratives/sit-rep-2020-06-07?n=4

**Your environment:**  
Windows 10, running in WSL, viewing auspice (just pulled today) locally. But also in the online version of auspice (see above link).

jameshadfield:
This fixes a bug where a narrative slide-change requests that the tree both zoom and change color.

This was introduced by 0639da85c4e1e1c343acbbe8d3231280ea8b68ca which, as part of the (now disabled) push for SVG gradient rendering of branches, moved the d3 call to update the stroke (branch color) into a separate code block, in order that it can behave differently for "T" and "S" parts of a branch. This code-block is not run when the tree zooms (that is the bug).

Here we partially revert that commit. This should be revisited when we re-enable SVG gradients.

Closes #1168

### DIFF
--- a/src/components/tree/phyloTree/change.js
+++ b/src/components/tree/phyloTree/change.js
@@ -60,14 +60,9 @@ const svgSetters = {
     // only allow stroke to be set on individual branches
     ".branch": {
       "stroke-width": (d) => d["stroke-width"] + "px", // style - as per drawBranches()
+      stroke: (d) => strokeForBranch(d), // TODO: revisit if we bring back SVG gradients
       cursor: (d) => d.visibility === NODE_VISIBLE ? "pointer" : "default",
       visibility: getBranchVisibility
-    },
-    ".branch.S": {
-      stroke: (d) => strokeForBranch(d, "S")
-    },
-    ".branch.T": {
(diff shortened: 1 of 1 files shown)


emmahodcroft:
Thanks James, this indeed fixes the bug locally!

PR #1170, opened 2020-06-16
Implement screenshot testing for narratives

jameshadfield:
See commit messages for details. (Note that the integration tests are still not being run automatically -- that's the next task to sort out).


eharkins:
Thinking about this as I am about to merge #1175 which fixes something specific to the URL query and so might benefit from narrative testing. In similar cases, I wonder if it makes sense to make it a requirement that PRs like this (or in general so long as they can be tested using narratives) add a slide to one of our test narratives that changes to / from the appropriate query, and writes one sentence in each slide about the expected behavior. This seems like a good middle ground between out current set up (no strict system of adding tests) vs requiring that PRs add automated tests. That way, at least we can more easily do manual narrative testing, and the test case is there if we want to go back and add automated tests for it. 

Sorry if this is duplicating something you have already said - if that's the case, take it as an endorsement of that and me volunteering to work on setting this up (perhaps by adding it into issue templates or PR templates if those exist or in another way)!

jameshadfield:
@eharkins completey agree with that sentiment! Unfortunately right now we haven't got the CI working well enough to mandate this I think (partly as I keep bumping it down the to-do list). Once things are working in a cross-OS way, then I think it'd be a great idea to enforce, or at least strongly recommend, such a test addition.

eharkins:
@jameshadfield it could just be for the purposes of manual testing at the moment, since even adding a narrative (or slide to an existing one) would make manual testing of PRs easier and just formalize what we are already doing to manually test PRs. Then we could make use of those slides down the road in CI once we have that working.

jameshadfield:
Yup! As long as we are aware that (a) the tests may fail locally in a false-negative way, and so looking at the image diffs is required and (b) failing tests (not currently part of the CI) are not a blocker for merge / release.

PR #1181, opened 2020-07-09
Send error messages in the response body, not the status line

### LINKED ISSUE #1180: Stop using response.statusMessage for error messages
**Current Behavior**
Auspice's CLI and documentation uses `res.statusMessage` (where `res` is an Express response object) to pass freeform error messages to the client/browser/user. However, [`statusMessage`](https://nodejs.org/api/http.html#http_response_statusmessage) is a low-level protocol field which controls the non-numeric part of the [HTTP status line](https://tools.ietf.org/html/rfc2616#section-6.1) (e.g. `200 Ok` and `404 Not Found`). It is a) not intended for arbitrary messages and b) not suitable for freeform text due to character restrictions.

Using `statusMessage` this way has caused nextstrain.org server crashes when the field is assigned an error containing a newline, as newlines are not allowed. Other restrictions may also be in effect.

**Expected behavior and solution**  
Error messages should still be conveyed to the client/browser/user, but via the response _body_, not the status line. The response can simply be a text/plain message. Both Auspice CLI code and documentation should be updated to reflect this.

**Additional context**  
image

tsibley:
The HTTP status line message is not intended for arbitrary messages and not suitable for freeform text due to character restrictions.

Using statusMessage this way has caused nextstrain.org server crashes when the field is assigned an error containing a newline, as newlines are not allowed.

Resolves #1180.


tsibley:
This PR should be merged and released before https://github.com/nextstrain/nextstrain.org/pull/202.

tsibley:
Thanks for merging and releasing, @jameshadfield!

PR #1182, opened 2020-07-14
[bugfix] allow `auspice view` to work with custom auspice clients

jameshadfield:
When running `auspice view` there is some logic to ask if we have a customised client bundle present, which we should serve, or whether we should serve the vanilla (default) auspice client.

PR #1126 introduced hashes into the (client) JS bundle names, which resulted in this logic missing when there was a customised client present, so we would always fall back to the vanilla auspice client.

Note that we didn't observe this in nextstrain.org as we don't use `auspice view` directly. This bug was identified while trying to upgrade [auspice.us](github.com/nextstrain/auspice.us).


jameshadfield:
Tested with auspice.us & nextstrain.org

PR #1186, opened 2020-07-16
wip: prototype node states graph

jameshadfield:


trvrb:
Brilliant idea! I had thought about this for commonly reconstructed node states like "host" (https://mbio.asm.org/content/mbio/9/3/e00909-18/F6.large.jpg), but very clever that this works for "clade" as well. 

The circular arrangement sometimes helps to pick out transition patterns: https://auspice-show-state-tran-mdvh0q.herokuapp.com/zika

It might make sense to limit the options to those that either have a geographic basis (with supplied lat/longs) or those that have internal node states labeled. Ie I'm not sure visualizations like https://auspice-show-state-tran-mdvh0q.herokuapp.com/testing/states (network layout by "country") are helpful.

In addition to circular arrangement you could imagine a force-directed layout option.

jameshadfield:
Thanks for the feedback @trvrb 

> It might make sense to limit the options to those that either have a geographic basis (with supplied lat/longs) or those that have internal node states labeled. 

This prototype requires the dataset JSON to supply the values as the geo-resolutions options, my idea being to rename these as "spatial resolutions" and display _either_ the map panel _or_ this states panel depending on if lat-longs were defined for that resolution. (Happy for other suggestions tho.) You could even imagine the ability to define the spatial location of your state values akin to how we define the spatial location of countries by lat/long values. 

> In addition to circular arrangement you could imagine a force-directed layout option.

💯 yes! and with ability to drag demes etc (further down the track of course)

trvrb:
>This prototype requires the dataset JSON to supply the values as the geo-resolutions options, my idea being to rename these as "spatial resolutions" and display either the map panel or this states panel depending on if lat-longs were defined for that resolution. (Happy for other suggestions tho.)

This makes sense to me. Good to have this be explicit. However, situations like https://auspice-show-state-tran-mdvh0q.herokuapp.com/WNV/NA make me think it would be useful to allow the network layout even if you have a trait with lat/long values. Something like lat/longs default to map view but allow a toggle to network view.

jameshadfield:
Update: Now working using (partial) force-directed layout which is interactive (i.e. you can drag demes around). Main outstanding tasks before this can be considered for merge into master are to update appropriately when redux state changes and to improve the physics of the force-directed layout.

trvrb:
So cool! I love this direction. Given that force directed is not always going to be the best layout (and will often be finicky), I'd think it valuable to allow either a circular layout or a force-directed layout (and have a UI for this either in the sidebar or as tabs in the top-right of the panel). In terms of merge strategy I'd be okay merging with a circular layout to avoid the finicky improvements to force-directed (but open to other strategies).

Words here are tricky, but I'd think it slightly more intuitive to call this "Show network layout view" rather than "Show state layout view".

For the force-directed layout parameters, it looks like it might be better to just freely zoom the bounds rather than constrain with forces pinning nodes within the bounding box, eg https://auspice-show-state-tran-mdvh0q.herokuapp.com/zika?showStateMap results in every node along the bounding box, but I'd assuming you could remove this constraint and just zoom the camera out.

rneher:
I like this as well! Essentially a 2-d histogram. 

But I agree with Trevor that the force directed layout is not always sensible. This for example makes all circles fly away:

https://auspice-show-state-tran-mdvh0q.herokuapp.com/zika?c=gt-NS1_349&p=grid&showStateMap&transmissions=hide

But I think you can adjust some center gravity or similar to keep these circles in view. But arrange in a circle and make them draggable is probably useful.

jameshadfield:
Update: It's getting closer to release-worthy I think. The force layout is working pretty well for most use cases, but when there are lots of transmissions it's unacceptably slow and when dragging demes it can be annoying that they snap back so much. All of this is tweakable, and there are plenty of performance improvements which can be implemented. Additionally, there are a few necessities such as handling browser resizes and more testing. Feedback on the current force-layout would be helpful for setting the best parameters. 


Here's WNV/NA

image

and zika

image

rneher:
zika is very pretty :)

trvrb:
I'd like to bump this prototype for development and inclusion in `master`. I just played around with things again and I still think it's a brilliant direction. Here's the current `clade_membership` view for `ncov/global`:

clades

If we make a "constellations" trait with states like `417N,484K,501Y` evolution of the virus between these eight genotypes would be described by connecting lines.

I really think this could help to outline genetic relationships between circulating variants.

It looks like the biggest issue is choosing robust parameters for the network view. One option here would be to give three values to click on in sidebar to toggle this between "strong", "medium" and "weak".

PR #1189, opened 2020-07-21
rewrite parsing of auspiceMainDisplayMarkdown;

eharkins:
See https://bedfordlab.slack.com/archives/C116R9PR7/p1595292741008800

This uses string.split instead of regex, which was causing browser to hang (possibly due to https://www.regular-expressions.info/catastrophic.html) despite technically valid paragraph missing both the opening string, "\`\`\`auspiceMainDisplayMarkdown", and the closing string, "```".

There are many other regexes in the narrative parsing function so this doesn't necessarily prevent the same or similar issues in the future. Maybe there is a way to wrap the whole function to handle any hanging regexes.

### Testing
Just tested that the narratives in the auspice repo are correctly parsed and that the browser doesn't hang for the example narrative from slack.

### DIFF
--- a/cli/server/parseNarrative.js
+++ b/cli/server/parseNarrative.js
@@ -120,15 +120,24 @@ function attributionLink(attribution, attributionLinkValue) {
  * Extract a code-block from the content, if it exists, tagged as auspiceMainDisplayMarkdown
  * Which we send to the client under a special key for parsing in the main display
  * This functionality is experimental, undocumented & subject to change.
- * The implementation relies on regex matching, which is not the best long term solution.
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @eharkins -- tested & confirm that it fixes the problematic narrative with no obvious changes to the others i've tested on. I enjoyed learning more about regex's through this!

Going forward, I consider it a high priority to merge #1172 which I'll endeavor to do in the next day or so. If you wouldn't mind checking that out with the knowledge you gained from this PR, that would be wonderful!

PR #1192, opened 2020-07-28
Improve redux state recalc on narrative slide changes. Closes #1191

### LINKED ISSUE #1191: Zoom into animation has broken
**Current Behavior**  
When zooming into a clade, the tree now snaps to the zoomed clade rather than doing a proper zoom animation. Interestingly, when "reset zoom" is run the zoom out animation occurs as intended. This bug in behavior can be ready seen in https://nextstrain.org/zika or https://nextstrain.org/groups/blab/sars-like-cov. I think it was missed in https://nextstrain.org/ncov/global, just because it's hard to see the animations in large trees due to lag.

**Expected behavior**  
When zooming into a clade, there should be animation going from original view to zoomed view.

jameshadfield:
This commit improves the creation of new state during narrative slide-changes which has a number of beneficial side-effects.

Previously we made a shallow-copy of the (redux) controls state when we changed slides, which resulted in props-comparison of nested state (i.e. objects within the controls state) being hard to reason with. This was problematic with the `temporalConfidence` object, which explains the approach employed in 0713e69b0707c92df806a575796e7205f08332ff to allow narrative slides to update CIs.

We now deep clone the controls state, allowing valid props comparison and allowing us to use this technique to decide when to render / remove CIs. (This is essentially a reversion of 0713e69b0707c92df806a575796e7205f08332ff, which was implemented as props comparisons were found to be unreliable).

This commit was motivated by, and fixes #1191. That bug was a result of the previous logic indicating that CIs should be removed on each branch-hover event. This meant that any subsequent tree-manipulation (such as clicking a branch to zoom) would be likely to be considered "in quick succession" to a previous manipulation and thus not employ an animation.

A test narrative is added.

@eharkins would you mind looking at this? It explains why the props-comparison wasn't working when you implemented the URL query for CIs!

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -1,4 +1,5 @@
 import queryString from "query-string";
+import { cloneDeep } from 'lodash';
 import { numericToCalendar, calendarToNumeric } from "../util/dateHelpers";
 import { reallySmallNumber, twoColumnBreakpoint, defaultColorBy, defaultGeoResolution, defaultDateRange, nucleotide_gene } from "../util/globals";
 import { calcBrowserDimensionsInitialState } from "../reducers/browserDimensions";
@@ -696,8 +697,9 @@ export const createStateFromQueryOrJSONs = ({
     controls["absoluteZoomMin"] = 0;
     controls["absoluteZoomMax"] = entropy.lengthSequence;
   } else if (oldState) {
-    /* revisit this - but it helps prevent bugs */
-    controls = {...oldState.controls};
+    /* creating deep copies avoids references to (nested) objects remaining the same which
+    can affect props comparisons. Due to the size of some of the state, we only do this selectively */
+    controls = cloneDeep(oldState.controls);
     entropy = {...oldState.entropy};
     tree = {...oldState.tree};
     treeToo = {...oldState.treeToo};
--- a/narratives/test_confidence-intervals.md
+++ b/narratives/test_confidence-intervals.md
@@ -1,0 +1,28 @@
+---
+title: Exploring the narrative functionality using the ongoing mumps epidemic in North America
+authors: "James Hadfield"
+authorLinks: "https://twitter.com/hamesjadfield"
+affiliations: "Fred Hutch, Seattle, USA"
+date: "August 2018"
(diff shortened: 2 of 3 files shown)


jameshadfield:
The CI is failing due to bundlesize, which should be fixed by #1172. I'll try to get that merged and then rebase this onto (new) master.

src/actions/recomputeReduxState.js
```diff
@@ -696,8 +697,9 @@ export const createStateFromQueryOrJSONs = ({
     controls["absoluteZoomMin"] = 0;
     controls["absoluteZoomMax"] = entropy.lengthSequence;
   } else if (oldState) {
-    /* revisit this - but it helps prevent bugs */
-    controls = {...oldState.controls};
+    /* creating deep copies avoids references to (nested) objects remaining the same which
+    can affect props comparisons. Due to the size of some of the state, we only do this selectively */
```
eharkins, line 701: This makes sense and as far as I can tell aligns with [the redux FAQ](https://redux.js.org/faq/performance#do-i-have-to-deep-clone-my-state-in-a-reducer-isnt-copying-my-state-going-to-be-slow), namely 

> Common Redux misconception: you need to deeply clone the state. Reality: if something inside doesn't change, keep its reference the same!

in that here it seems like we are abiding by that in general with shallow copies and specifically deep cloning `controls` due to the coincidence of calls to `createStateFromQueryOrJSONs` (with `oldState`) with changes to the controls state that will require props comparison. That is as opposed to some of the other slices of the state which are more tied to dataset and so don't rely on as many comparisons since dataset changes mostly come with a fresh redux state.

jameshadfield:
> that we were redundantly removing CIs even though they were already not being displayed (resulting in the events getting lumped together and throttled as you describe)

Exactly 👍 Without a working oldProps vs newProps comparison I don't think it's possible to accurately know when to remove CIs, so the approach 0713e69 was probably the best we could have done, but did have some unintentional (minor) side-effects. 

> That is as opposed to some of the other slices of the state which are more tied to dataset and so don't rely on as many comparisons since dataset changes mostly come with a fresh redux state.

Yes, but as this isn't formalised within Auspice this requires us to remember this and avoid props comparisons within the tree (etc) which could be misleading.

PR #1197, opened 2020-08-07
Fetch (optional) root-sequence JSON

jameshadfield:
Auspice has had a long-standing issue where choosing a genotype for a position where no mutations were observed resulted in an uninformative coloring of the tree. This is because we don't store the ancestral (root) sequence in the dataset JSON and thus rely on mutations to infer it.

Upon dataset load we now make a request for the "root-sequence" sidecar file. If this request is successful we use the data to color genotypes for which there are no mutations. The get-dataset script (which runs for auspice heroku deployments) has been modified to fetch vic & yam from the staging server, as corresponding root-sequence JSONs were present there but not on nextstrain-data. 

You can see this in action via https://auspice-root-seq-doeexpnizvfdt.herokuapp.com/flu/seasonal/vic/ha/3y?c=gt-HA2_120 -- notice how the tree starts off with a grey coloring and then once the root-sequence JSON request arrives the coloring updates to show the correct AA (T). Contrast this with https://auspice-root-seq-doeexpnizvfdt.herokuapp.com/flu/seasonal/h3n2/ha/3y?c=gt-HA2_120 for which the root-sequence JSON isn't available.

@emmahodcroft do you have a root-sequence JSON for a TB dataset? Presumably that will be a large JSON and it'd be good to use that for testing to see if there are any issues.


rneher:
this would be quite useful also for nextclade. we could then fetch the root sequence along with the tree and make it work for any of our builds.

jameshadfield:
Merging after rebasing onto master & retesting

PR #1207, opened 2020-09-15
[proof-of-principle] allow metadata to be fetched from an API

jameshadfield:


emmahodcroft:
I really like this idea!! I have also run into similar issues and it definitely gets old fast. Robert Dyrdak and I manually curated loads of age & gender data for EVD68 and having something like this (not to mention we could both have been editing it) would have been very nice!

Issue #1208, opened 2020-09-18
Auspice crashes when sequence name contains characters that are not allowed in CSS selectors


ivan-aksamentov:
**Current Behavior**  
Auspice crashes when sequence name contains characters that are not allowed in CSS selectors.

The following errors are shown in the console:
```
auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:1 Uncaught DOMException: Failed to execute 'querySelector' on 'Element': '#tip_some<unsafehtmlcharacters:hereclades' is not a valid selector.
    at SVGSVGElement.<anonymous> (https://auspice.us/dist/auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:1:84705)
    at J.select (https://auspice.us/dist/auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:1:125075)
    at a.Ve (https://auspice.us/dist/auspice.chunk.6.bundle.ecd503d56f440eb012ff.js:1:53779)
    at SVGCircleElement.<anonymous> (https://auspice.us/dist/auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:2:61827)
    at SVGSVGElement.<anonymous> (https://auspice.us/dist/auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:1:84705)
    at J.select (https://auspice.us/dist/auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:1:125075)
    at a.Be (https://auspice.us/dist/auspice.chunk.6.bundle.ecd503d56f440eb012ff.js:1:51700)
    at SVGCircleElement.<anonymous> (https://auspice.us/dist/auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:2:61827)
(anonymous) @ auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:1
select @ auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:1
Ve @ auspice.chunk.6.bundle.ecd503d56f440eb012ff.js:1
(anonymous) @ auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:2
(anonymous) @ auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:1
select @ auspice.chunk.other-vendors.bundle.7bee8f549fbad873d554.js:1
[...]

PR #1209, opened 2020-09-18
fix: ensure generated svg ids are escaped

ivan-aksamentov:
This ensures that runtime-generated HTML `id` properties are correctly sanitized, in particular, that invalid characters are removed or escaped.

Added dependencies:

`css.escape` as a polyfill for `CSS.escape()` (not available in IE)

There might be more places requiring sanitizaton.

### Related issue(s)  
Contributes to #1208 

### Testing
See linked issue.


jameshadfield (approved):
Thanks @ivan-aksamentov -- it's much nicer to switch from a regex to `css.escape` here

Issue #1049, opened 2020-04-08
Allow JSONs to define language for auspice


jameshadfield:
Auspice 2.12 allows one to set the language via a URL query, e.g. `?lang=ja` is Japanese (see the dropdown at the bottom of the sidebar for all the available languages). This is detailed [here](https://nextstrain.github.io/auspice/advanced-functionality/view-settings#url-query-options) in the docs.

This should also be settable via the `display_defaults` in the JSON (see [docs](https://nextstrain.github.io/auspice/advanced-functionality/view-settings#dataset-json-configurable-defaults)). For instance, a JSON with 
```json
"display_defaults": {
    "language": "ja"
}
```
Should display auspice in Japanese. This can be overridden by a URL query, and should be removed from the query and an error printed to the console if the language does not exist in auspice.

charlie-jones:
Hi, I can do a PR for this. My proposal is to use meta.display_defaults from recomputeReduxState.js to set the language in const defaults = {} from src/reducers/general.js.

jameshadfield:
That would be great!

charlie-jones:
Hey again -- so sorry for the delay. My steps for this were to import the JSON directly to general.js; that way, a function there can easily parse it to get a newly added language parameter (similarly to how recomputeReduxState.js works). I'm having trouble though finding the display_defaults data json (in augur - schema-export-v2.json) info that recomputeReduxState.js uses. Anyone know how to import json similarly for general.js? Thanks.

charlie-jones:
I'm going to move on from this issue. For others reviewing this issue -- feel free to refer to the steps above for inspiration and please try making a PR.

eharkins:
Hi @charlie-jones, thanks for working on this! I think I would start in the same place as you, i.e. 
> use meta.display_defaults from recomputeReduxState.js to set the language

However, I'm not sure I understand the following:
> My steps for this were to import the JSON directly to general.js; that way, a function there can easily parse it to get a newly added language parameter (similarly to how recomputeReduxState.js works)

I think instead of importing the JSON directly into the redux reducer, you should be able to make use of a pattern similar to existing ones for setting display defaults in the function `modifyStateViaMetadata`  in recomputeReduxState.js with the difference here being that most of those defaults are set on state.controls where as the language setting applies to state.general. 

When a new dataset is loaded, here is (my understanding of) how defaults are set in the redux state:
[...]

PR #1211, opened 2020-09-20
JSON metadata -- set language

charlie-jones:
I mainly followed the ideas from this comment on the issue (https://github.com/nextstrain/auspice/issues/1049#issuecomment-652703873)
I changed code in the general reducer and in recomputeReduxState. For the general reducer, I modeled everything after the controls reducer. (so the same format for getting displayDefaults metadata. The code for that is in recomputeReduxState)

### Related issue(s)  
Here: https://github.com/nextstrain/auspice/issues/1049

### Testing
However, I'm sending untested code because I don't know how to run the test JSON. (https://github.com/nextstrain/auspice/blob/master/examples/minimal_v2.json#L37).

Hope this helps! And, if I could get help on how to run the test JSON, I can work on it more next weekend.

### DIFF
--- a/examples/minimal_v2.json
+++ b/examples/minimal_v2.json
@@ -35,7 +35,7 @@
       "map"
     ],
     "display_defaults": {
-      "color_by": "country",
+      "color_by": "region",
       "language": "ja"
     },
     "geo_resolutions": [
--- a/examples/minimal_v2.json
+++ b/examples/minimal_v2.json
@@ -35,7 +35,8 @@
       "map"
     ],
     "display_defaults": {
-      "color_by": "country"
+      "color_by": "country",
+      "language": "ja"
     },
     "geo_resolutions": [
       {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "auspice",
-  "version": "2.17.4",
+  "version": "2.18.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -2,7 +2,7 @@ import queryString from "query-string";
 import * as types from "./types";
 import { getServerAddress } from "../util/globals";
 import { goTo404 } from "./navigation";
-import { createStateFromQueryOrJSONs, createTreeTooState, getNarrativePageFromQuery } from "./recomputeReduxState";
+import { createStateFromQueryOrJSONs, createTreeTooState, getNarrativePageFromQuery, createMetadataStateFromJSON } from "./recomputeReduxState";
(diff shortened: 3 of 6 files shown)


eharkins (changes requested):
Hi @charlie-jones, thanks for your work on this again! It seems like you are generally on the right track.  Here's how you can get setup to run auspice and test:

1. Install auspice locally: https://nextstrain.github.io/auspice/introduction/install#installing-from-source
2. run auspice with the example dataset directory `auspice develop --datasetDir examples` (https://nextstrain.github.io/auspice/introduction/how-to-run#how-to-get-an-example-dataset-up-and-running)
3. It would also be helpful (for you as you develop and for us when we merge) to `run npm run lint` to format the code. This will require having installed as above. You can also see where linting checks failed [here](https://github.com/nextstrain/auspice/pull/1211/checks?check_run_id=1141064434).

In terms of the implementation, I realize now that setting the display component (indirectly by getting the default state [here](https://github.com/nextstrain/auspice/pull/1211/files#diff-5aa3e1cc9cf1a24f7900459bafd76f60R747)) as a result of loading a JSON might get you stuck in a loop if it defaults to the datasetLoader which then calls loadJSONs again. Taking a step back from this, it seems like the `general` piece of the state should be treated differently from the `metadata`
 piece in that we should probably not start with a totally clean slate in this case. I can think of two ways to treat the `general` state specially and just set the `defaults`, and the `language` properties (and any other display defaults that we may track in the `general` state in the future):

1. Still use the `CLEAN_START` action as you have done, but in the reducer for the `general` state, only replace certain parts of the state when handling that action
2. Dispatch a different action just for the `general` state, in addition to the [`CLEAN_START` one which gets dispatched when the json is first loaded](https://github.com/nextstrain/auspice/blob/master/src/actions/loadData.js#L163) 

Hope this is helpful, I can try to be more specific if not. Appreciate you working on this!

charlie-jones:
Hi @eharkins, really appreciate the feedback! Thank you! I went with implementing #1. When running it on the main datasets (npm run view/dev), I noticed that the language field is empty, and there were on/off problems with changing the language through querying. On the minimal_v2 json with datasetDir and it runs smoothly and works. Any ideas about this? (anyone). I can work on this more, and I'd appreciate guidance. Thanks!

charlie-jones:
Hi all, I'm glad to say that I made some updates today and the language from json works now! On npm run view, npm run dev -- it functions normally, and does not use json data (because that's in augur?). On node.js auspice --view --datasetDir examples, the program reads from the JSON (currently set to Japanese). An important note - the program seems to need this CreateMetadataFromJSON import in the loadData class. During testing, without that import, the program only worked on "npm run dev" and running with --datasetDir examples. Hope this helped. Thanks!

Issue #1215, opened 2020-09-28
Smoke testing is currently broken on GitHub Actions


jameshadfield:
**Context**

We use smoke testing (`npm run smoke-test:ci`) to test, at a high level, that certain datasets are rendered as expected. For instance, the title of the "/ncov/global" dataset is "Genomic epidemiology of novel coronavirus - Global subsampling" and we smoke-test this by ensuring that the phrase "novel coronavirus" can be found when the page is rendered (as defined here https://github.com/nextstrain/auspice/blob/master/test/smoke-test/urls.txt#L4)

**Current Behavior**  

Running locally passes, as expected:

```sh
npm run get-data # get latest datasets
npm run smoke-test:ci # passes
```

However on GitHub Actions (as defined [here](https://github.com/nextstrain/auspice/blob/master/.github/workflows/ci.yaml#L21-L33)) the test fails with error:

```
FAIL test/smoke-test/urls.test.js (17.482s)
  smoke test rendering of paths in urls.txt
    ✕ it appropriately renders /ncov/global (1151ms)
    ✓ it appropriately renders /zika (7652ms)
    ✓ it appropriately renders /zika?c=region&l=clock&legend=open&m=div&r=region (2715ms)
    ✓ it appropriately renders /page-does-not-exist (1083ms)

  ● smoke test rendering of paths in urls.txt › it appropriately renders /ncov/global

    could not find text "novel coronavirus" on page /ncov/global

      42 |           expect(snapshotText).toEqual(expect.stringContaining(assertion));
      43 |         } catch (_err) {
    > 44 |           throw Error(`could not find text "${assertion}" on page ${testCase.path}`);
         |                 ^
      45 |         }
      46 |       }
      47 |     });

      at Object.it (test/smoke-test/urls.test.js:44:17)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 3 passed, 4 total
Snapshots:   0 total
Time:        17.551s
```

You can see the log of a failing GitHub Action [here](https://github.com/nextstrain/auspice/runs/1174367179).

It's perhaps relevant that when the test fails on a PR, the heroku review app which is spun-up renders the ncov/global dataset as expected, i.e. with "novel coronavirus" in the title! This matches local behavior.

**Expected behavior**  

Smoke tests on GitHub actions behave the same as running locally

**Possible solution**  

The snapshot testing was OS dependent (font rendering was different at the pixel level), so perhaps that's applicable here?

**Additional context**  

[...]

PR #1218, opened 2020-10-11
JSON Language display_defaults (Complete)

charlie-jones:

The goal of this PR is detailed in issue 1049 -- using display_defaults in the JSON to set the initial language. Once it's set, it can be changed normally by user clicking or making url query.

### Testing
For testing, I used "node auspice.js view --datasetDir examples", and minimal_v2.json sets the initial language to Japanese. I also tested it with "npm run view" and the program works normally, and the initial language is English there.

### DIFF
--- a/examples/minimal_v2.json
+++ b/examples/minimal_v2.json
@@ -35,7 +35,7 @@
       "map"
     ],
     "display_defaults": {
-      "color_by": "country",
+      "color_by": "region",
       "language": "ja"
     },
     "geo_resolutions": [
--- a/examples/minimal_v2.json
+++ b/examples/minimal_v2.json
@@ -35,7 +35,8 @@
       "map"
     ],
     "display_defaults": {
-      "color_by": "country"
+      "color_by": "country",
+      "language": "ja"
     },
     "geo_resolutions": [
       {
--- a/examples/minimal_v2.json
+++ b/examples/minimal_v2.json
@@ -36,7 +36,7 @@
     ],
     "display_defaults": {
       "color_by": "country",
-      "language": "ja"
+      "language": "en"
     },
     "geo_resolutions": [
       {
--- a/examples/minimal_v2.json
+++ b/examples/minimal_v2.json
@@ -36,11 +36,7 @@
     ],
     "display_defaults": {
       "color_by": "country",
-<<<<<<< HEAD
       "language": "en"
-=======
-      "language": "ja"
->>>>>>> 533a9d42fa571b619f378ce39ea5c886420b4d79
     },
     "geo_resolutions": [
       {
--- a/examples/minimal_v2.json
+++ b/examples/minimal_v2.json
@@ -36,7 +36,11 @@
     ],
     "display_defaults": {
       "color_by": "country",
+<<<<<<< HEAD
       "language": "en"
+=======
+      "language": "ja"
+>>>>>>> 533a9d42fa571b619f378ce39ea5c886420b4d79
     },
     "geo_resolutions": [
       {
--- a/examples/minimal_v2.json
+++ b/examples/minimal_v2.json
@@ -35,7 +35,7 @@
       "map"
     ],
     "display_defaults": {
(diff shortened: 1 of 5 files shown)


examples/minimal_v2.json
```diff
@@ -35,7 +35,8 @@
       "map"
     ],
     "display_defaults": {
-      "color_by": "country"
+      "color_by": "country",
+      "language": "ja"
```
eharkins, line 39: Can you change this back to leave out the language setting as it was before this PR, once you are done testing (if you make other changes) and ready to merge? I think we want to keep the example dataset as is. Thanks!

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "auspice",
-  "version": "2.17.4",
+  "version": "2.18.3",
```
eharkins, line 3: This change shouldn't be necessary. I think it will go away if you rebase your fork on nextstrain:master - would you mind doing that before we merge?

src/reducers/general.js
```diff
@@ -40,12 +40,16 @@ const general = (state = {
       return Object.assign({}, state, stateUpdate);
     case types.UPDATE_PATHNAME:
       return Object.assign({}, state, {
-        pathname: action.pathname
+        pathname: action.data
```
eharkins, line 43: Is this necessary to make the language feature work? All the `UDPATE_PATHNAME` actions i can find seem to use action.pathname, so I'm wondering where action.data will come from for this action.

src/reducers/general.js
```diff
@@ -40,12 +40,16 @@ const general = (state = {
...
       });
     case types.CHANGE_LANGUAGE:
       return Object.assign({}, state, {
         language: action.data
       });
+    case types.CLEAN_START:
```
eharkins, line 49: This approach generally makes sense to me. However, I think we might want to set the `state.general.defaults.language` according to the `displayDefaults` as well as setting the `state.general.language`, just since I think it would make sense to keep track of what the default was according to the JSON settings, even if we do have a query setting the language to something else. I would imagine it to look something like this:
```suggestion
    case types.CLEAN_START:
      const defaultLanguage = action.metadata.displayDefaults["language"] || defaults.language;
      return Object.assign({}, state, {
        defaults: Object.assign({}, state.defaults, {language: defaultLanguage}),
        language: query.lang ? query.lang : defaultLanguage
      });
```

eharkins (changes requested):
Thanks @charlie-jones for all your work on this, it's greatly appreciated! Sorry for not getting back to you earlier. This approach basically makes sense to me, so just a few changes and it should be ready to merge!

charlie-jones:
Thanks so much @eharkins ! Changes are implemented. You all have done so much amazing work on this!! I'm glad I could help in a small way.

src/reducers/general.js
eharkins: @charlie-jones this was marked as resolved but not changed back to `action.pathname`; is `action.data` here part of how you intend this feature to work? If so, I still wonder about the above question. Thanks!

charlie-jones: Very sorry about that, just changed it!

charlie-jones:
@eharkins any idea why smoke-test is failing here?

eharkins (changes requested):
@charlie-jones smoke-tests not passing is an existing problem (#1215). I think we can merge here if you remove that last commit. It's great that it seems to have fixed the smoke-tests in this case, but I think it's best to keep the two matters separate; if you want to submit a PR for fixing smoke-tests separately, that would definitely be appreciated! I still am ignorant of how they work, so I wouldn't be the best one to help you figure out if your solution is a reliable fix for smoke-tests.

charlie-jones:
No problem @eharkins , thanks for the help. Reverted to the previous version and the smoke tests actually work now (which is probably good/doesn't matter since it was on and off before anyway). Thanks again for all your feedback and help on this fix!

PR #1220, opened 2020-10-15
Migrate docs to docs.nextstrain.org

eharkins:
This moves auspice's:
- reference guide docs to https://docs.nextstrain.org/projects/auspice/en/migrate-docs/
- all other docs to https://docs.nextstrain.org/en/migrate/index.html

### Related issue(s)  
https://github.com/nextstrain/docs.nextstrain.org/pull/26
https://github.com/nextstrain/docs.nextstrain.org/issues/17

### Testing
TODO


docs/narratives/index.rst
```diff
@@ -0,0 +1,14 @@
...
+
+This page shouldn't be rendered & should jump to the first TOC entry instead. TODO.
+
+.. toctree::
+   :hidden:
+   :maxdepth: 0
+   :titlesonly:
+
+   introduction
+   how-to-write
```
eharkins, line 13: @jameshadfield these pages 404 (e.g. https://docs.nextstrain.org/projects/auspice/en/migrate-docs/narratives/how-to-write) now that we exclude them from the build. Should I go through and replace references to them with their URLs in the main RTD since probably users have not saved URLs like the one I mention here?

PR #1226, opened 2020-11-02
Use response status code not status message for 204 responses

jameshadfield:
API requests for `getDataset` which return 204 (No content) responses are a special case which should "cause Auspice to show its splash page listing the available datasets & narratives" (see https://nextstrain.github.io/auspice/server/api#charon-getdataset)

Auspice previously looked at the response statusMessage to check if responses were 204, however this wasn't appropriate (see https://github.com/nextstrain/nextstrain.org/commit/ff810503b0de6d03fc07e25d9f626d0e25f9e429 for more).

Here I switch the code to specifically check the response status code. This restores the intended behavior.

### DIFF
--- a/src/util/serverInteraction.js
+++ b/src/util/serverInteraction.js
@@ -1,7 +1,12 @@
+import {NoContentError} from "./exceptions";
+
 export const fetchWithErrorHandling = async (path) => {
   const res = await fetch(path);
 
   if (res.status !== 200) {
+    if (res.status === 204) {
+      throw new NoContentError();
+    }
     throw new Error(`${await res.text()} (${res.statusText})`);
   }
   return res;
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -8,15 +8,18 @@ import { fetchJSON, fetchWithErrorHandling } from "../util/serverInteraction";
 import { warningNotification, errorNotification } from "./notifications";
 import { hasExtension, getExtension } from "../util/extensions";
 import { parseMarkdownNarrativeFile } from "../util/parseNarrative";
+import { NoContentError } from "../util/exceptions";
 import { parseMarkdown } from "../util/parseMarkdown";
 import { updateColorByWithRootSequenceData } from "../actions/colors";
 
 /**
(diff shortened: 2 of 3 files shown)


jameshadfield:
This can be tested at https://dev.nextstrain.org/community/blab/ebola-narrative-ms -- notice how the 404 message doesn't appear

src/util/serverInteraction.js
```diff
@@ -1,7 +1,12 @@
+import {NoContentError} from "./exceptions";
+
 export const fetchWithErrorHandling = async (path) => {
   const res = await fetch(path);
 
   if (res.status !== 200) {
+    if (res.status === 204) {
+      throw new NoContentError();
```
kairstenfay, line 8: Using a custom error class instead of a hardcoded error message to evaluate the response seems like a big improvement. Thanks, James! My only comment is I wonder if it might be useful to pass a message to `NoContentError()`. I haven't tried this locally, but I wonder if it may produce more useful feedback to the user.

jameshadfield: Generally a fan of this approach but in this case I'm not sure what feedback should be provided to the user. A 204 response is used by auspice [to indicate that no dataset exists, but that no error message should be displayed](https://github.com/nextstrain/auspice/pull/1226/files#diff-7e6b4cc45e1472da7b3f8f6e7438315586ecbd5450bc7db2a5263a6fe2063e16R181-R182). Note the nextstrain.org server already sends 204 responses - [code here](https://github.com/nextstrain/nextstrain.org/blob/b1e09e57e91ed0c9343e1cd3104877ec3c5344a4/src/getDataset.js#L115-L124), they're just not currently handled appropriately by auspice! If you think it'd be sensible for auspice to log a console statement or similar then I'd be happy to implement.

kairstenfay (approved):

src/util/serverInteraction.js
jameshadfield: No problem -- there's lots of room for improvement in the interaction between the auspice client & the server (e.g. nextstrain.org's server) & error handling is definitely in there!

PR #1228, opened 2020-11-03
[WIP] Create redirect HTML files for switch to new docs platform

jameshadfield:
In November 2020 we moved the auspice documentation from github pages
to docs.nextstrain.org. In order to create redirects from the former
to the latter, we need to create simple HTML pages at each github-pages
page which instruct the client to redirect.

**This branch should not be merged into master**.
Instead, we should switch GitHub to source github-pages to this branch,
and then 

This switch should happen _after_ the `migrate-docs` branch has been merged into master
& thus https://docs.nextstrain.org/projects/auspice/en/latest (etc) are up and running.
This PR can then be closed without merging.

TODO: modify the auspice RTD URL & regenerate HTML after the redeployment of RTD from auspice master branch.

cc @eharkins


eharkins:
These redirects all look good to me!

jameshadfield:
GitHub pages is now running off this branch. Redirects are working as expected with the exception of https://nextstrain.github.io/auspice/ (an issue will be made for this).

PR #1229, opened 2020-11-12
Improve temporal axis labelling

jameshadfield:
_work in progress_

### DIFF
--- a/src/components/tree/phyloTree/grid.js
+++ b/src/components/tree/phyloTree/grid.js
@@ -115,7 +115,7 @@ const computeDivergenceGridPoints = (xmin, xmax, layout, minorTicks) => {
  */
 const calculateTemporalGridSeperation = (timeRange, pxAvailable) => {
   const [majorStep, minorStep] = [{unit: "DAY", n: 1}, {unit: "DAY", n: 0}];
-  const minPxBetweenMajorGrid = (pxAvailable < 1200 ? 200 : 300);
+  const minPxBetweenMajorGrid = (pxAvailable < 1000 ? 130 : 180);
   const timeBetweenMajorGrids = timeRange/(Math.floor(pxAvailable / minPxBetweenMajorGrid));
   const levels = {
     CENTURY: {t: 100, max: undefined},
--- a/src/components/tree/phyloTree/grid.js
+++ b/src/components/tree/phyloTree/grid.js
@@ -3,8 +3,8 @@ import { min, max } from "d3-array";
 import { transition } from "d3-transition";
 import { easeLinear } from "d3-ease";
 import { timerStart, timerEnd } from "../../../util/perf";
-import { months, animationInterpolationDuration } from "../../../util/globals";
-import { numericToCalendar } from "../../../util/dateHelpers";
+import { animationInterpolationDuration } from "../../../util/globals";
+import { numericToDateObject, calendarToNumeric, getPreviousDate, getNextDate, dateToString, prettifyDate } from "../../../util/dateHelpers";
 
 export const hideGrid = function hideGrid() {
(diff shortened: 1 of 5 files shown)


eharkins:
Looks like this is working nicely for the tree x-axis across pathogens on the review app! Have we already considered using javascript date utility libraries like https://momentjs.com/ or https://date-fns.org/ to help with the most common functions? We would still need some custom logic but might simplify overall by relying on utility functions from a library.

jameshadfield:
### What's changed

* Major and Minor grid line positioning and labeling of temporal axes in radial, clock and rectangular views of a phylogeny
* Tick positioning and labeling of x-axis of frequencies panel
* Tests added for date parsing functions
* Code is now easier to read and understand 🤞 

### Grid lines are now more helpful

As well as grid-lines now displaying the correct date (🐛  fixed), their layout is now much more interpretable based on the following logic:

1. Spacing between major grid lines (the ones with labels) always represents the same number of time-units. For instance, each space will represent 3 months, or 2 weeks, or 7 years etc. Note that this means the actual time may be slightly uneven (e.g. if each grid space represents a month, then different months have different numbers of days), but I believe this is intuitively understood. 
2. The interpretation of what time period minor grid lines represent should now be obvious (see screenshots below). If this is ambiguous, then please let me know as that implies it needs fixing!

### Examples

image

Dengue (one of the largest date ranges we have). Note the clipping of the left-most date label (a bug which exists on master as well)

image
image
image

Flu (12y) incl mobile & frequency views. The frequency panel has improved here. (A future addition is to make the time point which the mouse tooltip is associated with use the same code as the major/minor grid lines). cc @huddlej 

image

image

image

SARS-CoV-2, including zoomed in views. 

---

> Have we already considered using javascript date utility libraries like ...

I did a long time ago, but they were huge (in terms of gzipped bundle size) and so we didn't go there. Dates can be tricky, so this may be worth exploring as the bundle sizes seem to have come down a lot.

src/components/tree/phyloTree/grid.js
```diff
@@ -177,10 +102,112 @@ const computeXGridPoints = (xmin, xmax, layout, distanceMeasure, minorTicks, pxA
...
+  if (majorStep.n > 1 || majorStep.unit === "DAY") {
+    minorStep.unit = majorStep.unit;
+  } else {
+    minorStep.unit = levelsKeys[levelsKeys.indexOf(majorStep.unit)+1];
+  }
+  /* how many of those "units" should form the separation of the minor grids? */
+  const majorSpacing = majorStep.n * levels[majorStep.unit];
+  minorStep.n = Math.ceil(levels[minorStep.unit]/majorSpacing);
+
+  // XXX TODO remove following lines before merge of PR
```
jameshadfield, line 147: TODO - these console logs left in as they may be useful for review. Should be taken out before merge.

rneher (commented):
~~this wants to be `1/365`~~

trvrb (approved):

trvrb:
This commit tweaks display parameters to improve spacing on mobile. Previously there had been situations like:

old-1

that get resolved to

new-1

in this commit. Or situations like:

old-2

that get resolved to:

new-2

Basically, the wider px spacing was making it so that on small displays there weren't enough grid lines to be useful.

I experimented with this adjusted at various display sizes and I wasn't able to get overlapping grid text. This overlapping grid text is mostly an issue when displaying full `2015-Jan-12` style dates and limits the floor that the `minPxBetweenMajorGrid` variable can take.

jameshadfield:
Thanks Trevor & Richard for the comments.
P.S. Have fixed the `day = 1/265` typo 🤦

PR #1231, opened 2020-11-18
Improve appearance of sidebar headers

trvrb:
Using medium gray so it's less prominent on both Auspice's dark sidebar and Nextstrain's light sidebar. We only have a single text color in `globalStyles.js` and so I hard coded this to the component rather than borrowing something from `globalStyles.js`.


jameshadfield:
Thanks @trvrb -- also added a unrelated line but in the same file which fixes the vertical spacing on safari, where the `<div>` housing the heading had height of zero!

jameshadfield (approved):

trvrb:
Nice catch James!

PR #1236, opened 2020-11-22
Download Newick labeled by colorBy

trvrb:

I was just doing some downstream phylogeography analyses on top of https://nextstrain.org/ncov/north-america, where I needed to work from a labeled Newick for compatibility with [scripts](https://github.com/blab/global-migration/blob/master/geo/Rakefile) I had made for BEAST trees that include phylogeography.

I could have modified the Augur script [`json_tree_to_nexus.py`](https://github.com/nextstrain/augur/blob/master/scripts/json_tree_to_nexus.py) to collect a particular key from the JSON, but I thought that the more useful direction would be to build this labeling into Auspice download. This should surface this functionality to many more users.

BEAST trees that include traits (phylogeography or otherwise) use a format of `[&country="Thailand"]` to label nodes. This PR makes the Download Data modal download labeled Newicks with labeling based on current `colorBy`.

Some notes on implementation choices:
- `num_date` and `author` download a regular non-labeled Newick
- I've removed internal node names, e.g. `NODE_0000496` to fit with standard Newick format, I'm worried that a discordance between JSON from data.nextstrain.org and downloaded Newick will cause "gotcha" bugs
- I've escaped tip names with `'`
- I've removed white space from labels, eg changing `North America` to `NorthAmerica`. FigTree errors if whitespace is included here.
- The Download Data modal is smart about telling users that the Newick will include colorBy labeling

download

### Testing
I've tested this across a number of colorBys from different pathogens. It works for continuous trait values, for categorical trait values and sidesteps troublesome colorBys like `num_date` and `author`.

### DIFF
--- a/src/components/download/downloadModal.js
+++ b/src/components/download/downloadModal.js
@@ -12,6 +12,7 @@ import * as icons from "../framework/svg-icons";
 import { getAcknowledgments} from "../framework/footer";
 import { createSummary, getNumSelectedTips } from "../info/info";
 import { getFullAuthorInfoFromNode } from "../../util/treeMiscHelpers";
+import { isColorByGenotype } from "../../util/getGenotype";
 
 const RectangularTreeIcon = withTheme(icons.RectangularTree);
 const PanelsGridIcon = withTheme(icons.PanelsGrid);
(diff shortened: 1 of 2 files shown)


jameshadfield:
Closing in favor of #1245 which uses this approach (and code!) to allow Nexus trees with annotations 😄

PR #1237, opened 2020-11-30
[bugfix] handle unknown values in color scales with provided scales

jameshadfield:
Colorscales should be able to handle nodes with unknown values for the current `colorScale` (e.g. `getTraitFromNode(node, colorBy=undefined`). This was not being correctly handled in the case where we provided a scale in the dataset JSON. For certain cases (i.e. when traits had not been inferred for internal nodes) this resulted in internal nodes being coloured rather than grey as intended. (Note that which colour was chosen depended on the provided scale values.)

Before (note that the branches on the far left don't have values set for the current colorBy, i.e. `getTraitFromNode(node, colorBy=undefined`)

image

After:
image

### DIFF
--- a/src/util/colorHelpers.js
+++ b/src/util/colorHelpers.js
@@ -3,6 +3,7 @@ import { interpolateRgb } from "d3-interpolate";
 import scalePow from "d3-scale/src/pow";
 import { isColorByGenotype, decodeColorByGenotype } from "./getGenotype";
 import { getTraitFromNode } from "./treeMiscHelpers";
+import { isValueValid } from "./globals";
 
 /**
  * Average over the visible colours for a given location
@@ -48,7 +49,9 @@ export const getExtraVals = (nodes, nodesToo, colorBy, providedVals) => {
     nodesToo.forEach((n) => valsInTree.push(getTraitFromNode(n, colorBy)));
   }
   valsInTree = [...new Set(valsInTree)];
-  return valsInTree.filter((x) => providedVals.indexOf(x) === -1);
+  return valsInTree
+    .filter((x) => providedVals.indexOf(x) === -1)
+    .filter((x) => isValueValid(x));
 };
 
 
--- a/src/util/colorScale.js
+++ b/src/util/colorScale.js
@@ -11,7 +11,7 @@ import { setGenotype, orderOfGenotypeAppearance } from "./setGenotype";
 import { getTraitFromNode } from "./treeMiscHelpers";
 import { sortedDomain } from "./sortedDomain";
 
-const unknownColor = "#AAAAAA";
+export const unknownColor = "#AAAAAA";
 
 const getMinMaxFromTree = (nodes, nodesToo, attr) => {
   const arr = nodesToo ? nodes.concat(nodesToo) : nodes.slice();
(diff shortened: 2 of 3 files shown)


eharkins:
Code makes sense at a glance but I'm having trouble reproducing the bug. Do you have any concrete examples or ways to fake it locally?

jameshadfield:
@eharkins I was getting inconsistent results which were hard to debug (and so can't give you a simple JSON to test). This process resulted in me refactoring this code and writing tests for it. I had wanted to refactor this for a long time, as it was very hard to interpret at a glance. Re-review would be appreciated!

eharkins:
🌟   for this refactor, it seems like a big win for reading and debugging / testing this code! Running the tests you added worked for me. 

I still am not sure how to reproduce the original branch coloring bug that prompted your work here so can't comment on that specific issue. 

When clicking around, I noticed something that didn't match nextstrain.org in terms of color scales. However, it also **shows up on auspice master branch** so seems to not have been introduced by this PR - thought I would still mention it here as it seems relevant to this PR. I tried searching existing issues for this exact behavior but didn't find anything (sorry if i missed it and it's already documented!):

When loading a genotype color scale, e.g. https://nextstrain.org/ncov/north-america?c=gt-ORF10_3, I get the following 

Screen Shot 2020-12-18 at 2 47 33 PM

It seems like a bug given that nextstrain.org shows a non-empty value in the legend and on the tree.

Screen Shot 2020-12-18 at 2 49 06 PM

It seems distinct from #738 since it still happens when changing genotype position colorings (not just on page load).
Let me know if this is already a known issue, otherwise I can file a separate issue so that it doesn't block merging this.

eharkins (commented):
🌟   for this refactor, it seems like a big win for reading and debugging / testing this code! Running the tests you added worked for me. 

I still am not sure how to reproduce the original branch coloring bug that prompted your work here so can't comment on that specific issue. 

When clicking around, I noticed something that didn't match nextstrain.org in terms of color scales. However, it also **shows up on auspice master branch** so seems to not have been introduced by this PR - thought I would still mention it here as it seems relevant to this PR. I tried searching existing issues for this exact behavior but didn't find anything (sorry if i missed it and it's already documented!):

When loading a genotype color scale, e.g. https://nextstrain.org/ncov/north-america?c=gt-ORF10_3, I get the following 

Screen Shot 2020-12-18 at 2 47 33 PM

It seems like a bug given that nextstrain.org shows a non-empty value in the legend and on the tree.

Screen Shot 2020-12-18 at 2 49 06 PM

It seems distinct from #738 since it still happens when changing genotype position colorings (not just on page load).
Let me know if this is already a known issue, otherwise I can file a separate issue so that it doesn't block merging this.

jameshadfield:
Thanks @eharkins!

> When loading a genotype color scale, e.g. https://nextstrain.org/ncov/north-america?c=gt-ORF10_3, I get the following

This is because to colour a tree at a non-varying position requires the sidecar `_root-sequence.json` file. This is present on the S3 bucket, but not in your local `--datasetDir` folder which is why you're seeing different behavior (comparing the network requests between localhost & nextstrain.org is way to see this).

PR #1244, opened 2020-12-15
Drag drop metadata improvements

### LINKED ISSUE #1239: Filters should dynamically update when additional metadata is added
Currently the available filters are [set upon dataset load](https://github.com/nextstrain/auspice/blob/master/src/actions/recomputeReduxState.js#L197-L208) and are the values specified in the `JSON.filters` array as well as "strain". 

Upon loading additional metadata (via drag & drop functionality) we should add all color-bys (columns in the TSV/CSV) to the available filters. This may necessitate updating the `totalStateCounts` dict, as [this is used](https://github.com/nextstrain/auspice/blob/master/src/components/controls/filter.js#L46-L59) to collect valid values for each filter.

### LINKED ISSUE #1242: Adding metadata should automatically filter the tree
If metadata is dragged onto auspice, then we should filter the dataset to those samples included in the CSV/TSV file. 

As a shortcut: once #1293 has been closed, then adding a column to the CSV with the same value for all samples (e.g. `"additional": "yes"`) should allow one to filter the data by this trait.

jameshadfield:
A number of improvements to the drag-and-drop metadata functionality. See commit messages and documentation changes in this PR for more details. 

Closes #1242
Closes #1239


src/actions/filesDropped/metadata.js
```diff
@@ -1,124 +1,248 @@
...
+
+/* ---------------------- helper functions to parse data ---------------------------- */
+
+
+function processHeader(fields) {
+  const strainKey = fields[0];
+
+  /* There are a number of "special case" columns we currently ignore */
+  const fieldsToIgnore = new Set(["name", "div", "vaccine", "labels", "hidden", "mutations", "url", "authors", "accession", "traits", "children"]);
+  fieldsToIgnore.add("num_date").add("year").add("month").add("date"); /* TODO - implement date parsing */
```
jameshadfield, line 54: An obvious next step is to allow dates to be specified via additional metadata. Interpreting this as the `num_date` attr is slightly complex as the coloring & tree-metric are tied together in the code. A different approach would be to extend our color schemes to allow "temporal" types and then add temporal info in the CSV/TSV as simply an additional color-by.

emmahodcroft: So, this would mean sequences in a time-tree with inferred dates were allowed to have 'real' dates supplied? Would it change where they're plotted on X-axis?

emmahodcroft:
This all sounds super-cool James!! What a neat bunch of features. If you have a moment to supply a dummy CSV that shows an example how one can make use of these (how to organize them, etc) that would be fantastic! (No rush though!)

src/actions/filesDropped/metadata.js
jameshadfield: I'm leaning towards allowing metadata traits (in drag-n-drop data or in the main JSON) which are encoded as (e.g.) `YYYY-MM-DD` to be inferred as temporal types by auspice, and a colour scale generated according. A good example of this would be nCoV's "submitted date" field. These would be "just another color-by" and thus different to the `num_date` field which is used by auspice for the tree-time view. I don't see any easy way to use metadata to update / influence node (temporal) positioning, but am open to suggestions. (It's easy to see how such metadata could define the x-axis position of tips, but what do we do with internal nodes / tips not in the metadata?)

emmahodcroft: Sounds good! Yes, I agree, that's why I was curious - was wondering how you'd adjust the other parts of the tree to accommodate moving tips! Unfortunately I have no answers to offer, just curiosity :)

emmahodcroft:
If you can provide a dummy file showing how to format the drag-n-drop files for the new things that are accepted, I'm happy to work from that, modify, add, etc, and play around! (Sorry if I've missed this file already being somewhere!)

jameshadfield:
One way to test the functionality of this is via a modified version of the zika-tutorial at https://auspice-drag-drop-metad-2eccco.herokuapp.com/zika-sparse-metadata. Notably this build does not have the country included as a coloring or a geographic resolution.

The additional metadata TSV at https://nextstrain-scratch.s3.amazonaws.com/james/zika-extra-metadata.tsv:

```tsv
strain  country latitude        longitude       random  random__colour
1_0087_PF       French Polynesia        -17.6797        -149.4068       blue    #184ae8
1_0181_PF       French Polynesia        -17.6797        -149.4068       blue    #184ae8
1_0199_PF       French Polynesia        -17.6797        -149.4068       blue    #184ae8
Aedes_aegypti/USA/2016/FL05     USA     39.7837304      -100.4458824    green   #2dc234
BRA/2016/FC_6706        Brazil  -10.3333332     -53.1999999     red     #e7182a
Brazil/2015/ZBRA105     Brazil  -10.3333332     -53.1999999     blue    #184ae8
...
```

will add the following data to around 25 of the strains in the build:
* Country is added as a coloring, with auspice choosing the colours for these
* A coloring entitled "random", with a separate column defining the colours for each value
* An additional coloring (not in the TSV) is added by auspice identifying which strains were in the metadata file.
* Filterings in auspice are available for the above described colorings
* Lat/long for strains. This is actually the lat/long of each strain's country, so this PR will group them together appropriately.

emmahodcroft:
So is this the colouring that just defines what was in the TSV? :
image

This was a little confusing to me and I assumed it was a bug until I came back and re-read your points more carefully. I like the idea, but maybe you could show it on the drop-down as "Strains in XX.tsv" or similar?

Apart from that this seems to work great! Is it worth creating another random file for another build, or do you think testing this on the Zika tutorial is sufficient? I have some files for the LA builds which I could probably add random columns to - but maybe not until next week (sorry!).

cassiawag:
I tested this out on the WA trees (https://dev.nextstrain.org/groups/blab/ncov/wa/1y) w/ real metadata, and overall, these are fantastic new features that work well!

cassiawag:
I tested this out on the WA trees (https://dev.nextstrain.org/groups/blab/ncov/wa/1y) w/ real metadata, and overall, these are fantastic new features that work really well! Thanks James!

Specifically, I tested:
- specifying new colors

cassiawag:
I tested this out on the WA trees (https://dev.nextstrain.org/groups/blab/ncov/wa/1y) w/ real metadata, and overall, these are fantastic new features that work really well! 

Specifically, I tested:
- specifying new colors with hexcodes
- adding lat & long
- filtering by added metadata

Comments:
- All the features worked well and as expected on the real data, including on fields containing spaces!
- For this American speller, I would just encourage strong documenting that you must use `__colour` and not `__color` when adding hexcodes.
- Should clearly document that if you add latitude and longitude and choose the geographic resolution of your added metadata, this will only display sequences with lat & long in your metadata on the map. This behavior was initially unexpected for me.  But on thought, I think this makes sense for now as I'm not sure what geo resolution you would overlay added lat & longs on. In the future, maybe you could choose to overlay my_geo on any of the geo layers, like my_geo + division or my_geo + location or my_geo + region? I'm not sure what would be best; this would probably depend on how users interact with the map, and it's not a function I use very often. 
- I think the suggestion by @emmahodcroft to make the dropdown say "Strains in XXX.tsv" is a good one!

Thanks again for the awesome work, James!! These are super useful features.

emmahodcroft:
Very good point about spelling @cassiawag ! @jameshadfield would it be possible to let the code accept both `color` and `colour`? I know it's reflexive to type whichever you're used to 🙃 
Otherwise, (and know it pains me to write this), we may be better standardising to `color` as this is more common in programming languages, so users such as myself who prefer `colour` are probably more used to switching for 'programmy' things than US users switching to `colour` 😛

jameshadfield:
Thanks for the testing and comments @cassiawag & @emmahodcroft  -- I've updated the documentation, allowed American spellings (color), and changed the value to "Strains in XXX.tsv" which makes the filtering look much nicer! P.S. Microreact uses `__colour` which is why I initially chose it - but it makes sense to allow `__color` as well; if both are supplied then `__colour` takes preference.

PR #1245, opened 2020-12-17
Download data improvements

jameshadfield:
This PR makes a number of improvements to the download-data functionality in auspice. See individual commit messages for specifics. Testing on diverse datasets would be highly valuable at this point.

This PR introduces consistency by making downloaded data represent the selected data (before it was different depending on what was being downloaded). This means that if you have manipulated the data (e.g. zoomed the tree, filtered the dataset) then every downloaded file -- trees, metadata, entropy -- will reflect this. 

We also allow the abliity to download BEAST-style Nexus trees which can be read by FigTree. These are annotated with all of the colorings in the dataset (as well as genotype and div, as applicable). I failed to find a proper schema for the annotations block, which means there's likely some scenarios where exported trees can't be read by FigTree. I also imagine there's more annotations which can be exported, such as defining colours for annotation values. Please let me know of any you come across! This functionality allows nextstrain data to be used in FigTree -- for instance, here's a global nCoV tree filtered to England samples and zoomed to clade 20B in Auspice, exported, imported into FigTree, and annotated there by mapping tip colours to pangolin lineage, tip labels to num_date, node labels to num_date_CI, tip & node shapes to divergence. 

image

### DIFF
--- a/src/components/download/downloadButtons.js
+++ b/src/components/download/downloadButtons.js
@@ -1,6 +1,7 @@
 import React from "react";
 import { withTheme } from 'styled-components';
 import * as icons from "../framework/svg-icons";
+import { NODE_VISIBLE } from "../../util/globals";
 import { materialButton } from "../../globalStyles";
 import * as helpers from "./helperFunctions";
 import { getNumSelectedTips } from "../info/info";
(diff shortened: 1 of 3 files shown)


cassiawag:
I demoed this today on WA trees (https://dev.nextstrain.org/groups/blab/ncov/wa/1y), and overall it works super well! I thought it was very clear that I only downloaded data for the tips that were filtered and zoomed. It did take me awhile (several days) to figure out how to download divergence trees. This is probably because I'm used to the previous interface. I'd be a fan of including divergence tree options, even if the tree is set to time tree. Alternatively, you could be very explicit in the description. Something along the lines of: "The current view is displaying data for X/X tips in units of time. To download trees in units of divergence, please switch to divergence tree." 

Also, I didn't do much with the Nexus tree, so it might be worth having someone who uses Nexus trees frequently (@nicfel maybe?) try it out.

cassiawag (commented):
I demoed this today on WA trees (https://dev.nextstrain.org/groups/blab/ncov/wa/1y), and overall it works super well! 

I thought it was very clear that I only downloaded data for the tips that were filtered and zoomed. It did take me awhile (several days) to figure out how to download divergence trees. This is probably because I'm used to the previous interface. I'd be a fan of including divergence tree options, even if the tree is set to time tree. Alternatively, you could be very explicit in the description. Something along the lines of: "The current view is displaying data for X/X tips in units of time. To download trees in units of divergence, please switch to divergence tree."

Also, I didn't do much with the Nexus tree, so it might be worth having someone who uses Nexus trees frequently (@nicfel maybe?) try it out.

PR #1246, opened 2020-12-18
Allow tip labels to be user selected

### LINKED ISSUE #1201: Add "Label By" dropdown for labeling nodes by attributes other than `name`
**Context**  
We would like to be able to label nodes with attributes other than their `name`. Uses:

* A user has drag-and-dropped a file linking public identifiers to patient names onto auspice, and they would like to label the nodes by the patient names.
* Nodes have multiple name-like fields, such as the public GISAID name, a lab accession number, and an ID used by a partner institution. It would be useful to toggle between them.
* When an attribute has too many values for the color to be legible, as is the case with Admin Division in current ncov builds, labels with that attribute would be easy to read.
* A user who wants to see two attributes simultaneously could use one as a color and the other as the label. For example. facility name and worker/resident status. Currently, one has to rapidly toggle the color and squint to compare two sets of attributes at once.

**Description**   
An additional dropdown box called "Label By", under "Color By", in the navigation panel on the left, with the default set to `Name`. Attributes would appear in the drop-down. When selected, the nodes would be labelled by the value of the attribute instead of the current `name` field.

Open question: should these be specified separately in the JSON as a `labelings` field? For backwards compatibility, we could also use the union of `filters` and `colorings` OR simply the full set of node attributes if `labels` is not specified. Many name-like fields might not make sense as colorings or filters, so having an option to specify a might offer a better user experience.

jameshadfield:
Adds an additional dropdown box to allow users to change the attribute used as tip labels. This is similar in design to the branch label drop down.

Currently attributes are limited to the colorings. The function which collects these, `collectAvailableTipLabelOptions` could be easily extended to produce further valid attributes.

This should pave the way for smarter and more customisable logic around _when_ labels are displayed, and the dropdown should not be shown in the situation that labels aren't shown.

For a full description, see #1201
Closes #1201

@allenkj2 and @batson could you test this for me please? (I can't add you as reviewers due to GitHub collaboration settings)


allenkj2:
This is really great! The only oddity I'm seeing is with how dates are presented - when I select the 'sampling date,' I get the formatting below.
image

emmahodcroft:
This is super cool James!! 
I agree that we might want to see if we can format the date back to a YYYY-MM-DD format before displaying, if that's not too difficult. 

We might also want to filter out 'genotype' as a tip label. It would actually be pretty cool to enable this but I imagine it would be more work - we could always come back to that in future! For short term might be good to just leave off the list.

jameshadfield:
Tip label formats now updated to turn dates into pretty strings and genotype removed from the available options. (Genotype would make sense as a labeling when the coloring is also genotype, but implementing this would entail more complex logic which is best left for a subsequent PR.)

batson:
This is fantastic, @jameshadfield. Completely changes the navigation experience.

My only request now is that it go up on auspice.us, so the counties working with private data/trees can see the PII they're looking for!

Thanks again for the feature!

sidneymbell:
+1 -- I added a bug report for this somewhere or other :)

Issue #1132, opened 2020-05-17
Add button to zoom tree to visible tips


trvrb:
**Context**  
Currently, there are two primary ways to interact with the tree:
1. Click on clades to zoom in. This works pretty well much of the time but may require multiple clicks and with large trees it may be impossible to actually find the clade to click on. @batson summarizes this issue well in https://github.com/nextstrain/auspice/issues/1111.
2. Filter to a subset of tips via filters interface. This works well to highlight particular samples and show where they fit in the larger tree. However, once filtered it can be difficult to examine these in more detail. For example, https://nextstrain.org/ncov/north-america?c=division&f_location=Grand%20Princess%20cruise%20ship&r=division shows the 7 Grand Princess genomes but it's hard to see much other than that they cluster roughly together.

There is often a user desire to filter (or otherwise subset) to a portion of the available samples and "zoom" into this subset in the tree. 

**Description**   
I propose adding a button to the tree panel titled "ZOOM TO VISIBLE". Clicking this button acts to find the common ancestor of visible tips and initiate a clade zoom for this node.

![zoom](https://user-images.githubusercontent.com/1176109/82161319-93a39b80-9850-11ea-8327-9028c6bf822b.png)

However, there will situations where a filter subsets to a non-monophyletic (discontinuous) group of tips. In this case, I'd recommend that this "zoom" also compress y-values surrounding non-visible tips. @jameshadfield previously implemented this "accordion" style zoom in the branch [`js-y-vals`](https://github.com/nextstrain/auspice/tree/js-y-vals). Here's an example of this compression:

![accordian](https://user-images.githubusercontent.com/1176109/82161530-6eb02800-9852-11ea-987d-2bd85bdf3294.png)

Here, I'd still take the minimum and maximum y values as the min/max of the selected tips, but in between this min/max non-visible tips would be compressed.

The entire tree would be restored by clicking on "RESET LAYOUT" as is behavior for clade zoom.

This direction sets us up to have a box-selection option (say hold down option while dragging over the tree) that could select a user-specified region and followed with "ZOOM TO VISIBLE". I believe this would solve @batson's original issue, but would add significant flexibility by interacting through visible nodes rather than a direct tree operation.

What do folks think about this proprosal?

batson:
I like this proposal very much!

[...]

PR #1257, opened 2020-12-29
Zoom to selected

joverlee521:
Built on top of changes in #1248. Partly addresses #1132.

Adds a "Zoom to Selected" button that allows users to zoom into the selected tips, with TMRCA as the new root node. 
This is currently just a basic zoom, without the accordion style zoom described in #1132. Creating this PR because even the basic zoom adds a lot of usability while the accordion zoom is taking me longer to figure out.

### DIFF
--- a/src/actions/tree.js
+++ b/src/actions/tree.js
@@ -74,6 +74,7 @@ export const updateVisibleTipsAndBranchThicknesses = (
       branchThickness: data.branchThickness,
       branchThicknessVersion: data.branchThicknessVersion,
       idxOfInViewRootNode: rootIdxTree1,
+      idxOfFilteredRoot: data.idxOfFilteredRoot,
       cladeName: cladeSelected,
       selectedClade: cladeSelected,
       stateCountAttrs: Object.keys(controls.filters)
@@ -93,6 +94,7 @@ export const updateVisibleTipsAndBranchThicknesses = (
       dispatchObj.branchThicknessToo = dataToo.branchThickness;
       dispatchObj.branchThicknessVersionToo = dataToo.branchThicknessVersion;
       dispatchObj.idxOfInViewRootNodeToo = rootIdxTree2;
+      dispatchObj.idxOfFilteredRootToo = dataToo.idxOfFilteredRoot;
       /* tip selected is the same as the first tree - the reducer uses that */
     }
 
--- a/src/components/tree/tree.js
+++ b/src/components/tree/tree.js
@@ -97,14 +97,32 @@ class Tree extends React.Component {
   getStyles = () => {
     const activeResetTreeButton = this.props.tree.idxOfInViewRootNode !== 0 ||
       this.props.treeToo.idxOfInViewRootNode !== 0;
+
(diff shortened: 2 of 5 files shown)


trvrb:
@joverlee521 ---

This is fantastic. I've wanted this feature for a while and I think it adds a lot of expressiveness to visualization ability.

Just a few small things:
1. Can you add a small gap between the "zoom to selected" tab and the "reset layout" tab? This should mirror the slight spacing between the "entropy / events" tab and the "aa / nt" tab in the Diversity panel.
2. If I filter tips via the date slider, like so: https://auspice-zoom-to-selecte-xysadr.herokuapp.com/flu/seasonal/h3n2/ha/12y?c=region&dmax=2010-08-01, I don't get the option to zoom to these selected tips. Dates should work just like other filters.
3. The top of the y-axis doesn't appear to be zoomed correctly. Go to https://auspice-zoom-to-selecte-xysadr.herokuapp.com/zika?f_country=Brazil and then zoom to selected. You'll see that the entire top clade is kept in view green the green Brazil tip should be at the top of the page. This issue can be seen more strongly in the following: https://auspice-zoom-to-selecte-xysadr.herokuapp.com/zika?f_country=Usa

Finally, this can be saved for another PR, but it would be useful for Narratives and general sharing to expose "zoom to selected" in the URL. Kind of like `transmissions=hide`, etc...

jameshadfield:
Thanks @joverlee521 -- this is awesome and shows good understanding of the code flow in auspice! 

Currently this PR achieves zooming by leveraging the machinery built for click-on-branch-to-zoom., which is why the following two situations aren't covered:

> If I filter tips via the date slider... I don't get the option to zoom to these selected tips. Dates should work just like other filters.

> [when there's tips "above" the filtered ones within the clade] The top of the y-axis doesn't appear to be zoomed correctly 

Solving these will require a different approach, including modifying PhyloTree's `mapToScreen`. I think such changes will be "on top of" setting the root as done here, and will need a bit of thought as to the best way to implement them. I think these are best suited as a separate PR, allowing us to merge this with the current behavior (i.e. all that's needed right now is to space the buttons a bit more).

joverlee521:
Thanks for the feedback @trvrb and @jameshadfield! I've pushed up a commit to space out the buttons. 

> Solving these will require a different approach, including modifying PhyloTree's mapToScreen. I think such changes will be "on top of" setting the root as done here, and will need a bit of thought as to the best way to implement them. I think these are best suited as a separate PR, allowing us to merge this with the current behavior (i.e. all that's needed right now is to space the buttons a bit more).

Yes, the current PR only uses the existing mechanisms for zooming. I've been working locally to get a "new" type of zoom feature that will include the situations that @trvrb has asked for here, but it's been trickier than I had expected. I'm hoping to get it done this week, but I can definitely create a new PR for it.

jameshadfield (approved):
I'm going to merge now, as the functionality introduced here is great. I know @joverlee521's well on the way to achieving the extra functionality, which can be done on top of master and  make use of changes introduced here (e.g. `idxOfFilteredRoot`).

PR #1265, opened 2021-01-14
Filtering by genotype

jameshadfield:
_This comment was updated Jan 22nd_

### PR Status
Ready for review. I believe there are no bugs (see below) but testing is required and performance may not be good enough for release.

### Summary
This PR adds genotype filters to the side-bar dropdown allowing filtering of the dataset by any observed genotype, as long as it was at a variable site.
 
### Notes and example URLs

_Updated to relect the changes introduced in 3ca2c02_

* Example of N501Y + D614G + P681H which corresponds to 20I/501Y.v1 / B.1.1.17: https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/global?gt=S.501Y,614G,681H
* Example of N501Y + E484K, a constellation which is so far shared by 20J/501Y.V3 and 20H/501Y.V2: https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/global?gt=S.501Y,484K
* Example of a basal "mutation": https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/global?gt=S.982S
* Example of a mutation at a variable site but which is not observed in the data, which is silently removed at load: https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/global?gt=S.982S
* Example of a genotype state (nuc:502A) which does exist, but is at a non-variable site. The filter is silently removed: https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/global?c=gt-nuc_502&gt=nuc.502A
* "zoom-to-selected" hasn't been considered and doesn't work as it ideally should. This seems to require the same fix as that which  @joverlee521 is currently working on.
* ~Initial rendering of the filtering-dropdown is very slow. [This](https://codesandbox.io/s/lxv7omv65l?file=/index.js) seems to be one solution, which I may add in this PR or via another PR.~ See below for update.
* We don't yet allow OR filtering for the same position, thus one cannot filter by (e.g.) `S 501Y || S 501T` https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/global?gt=S.501Y,501T

cc @kairstenfay @joverlee521 @trvrb @eharkins @huddlej re: this morning's meeting, any testing would be super useful!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "auspice",
-  "version": "2.22.0",
+  "version": "2.22.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -3987,6 +3987,27 @@
         "babel-plugin-jest-hoist": "^25.2.6"
       }
     },
+    "babel-runtime": {
+      "version": "6.26.0",
+      "resolved": "https://registry.npmjs.org/babel-runtime/-/babel-runtime-6.26.0.tgz",
+      "integrity": "sha1-llxwWGaOgrVde/4E/yM3vItWR/4=",
+      "requires": {
+        "core-js": "^2.4.0",
+        "regenerator-runtime": "^0.11.0"
+      },
+      "dependencies": {
+        "core-js": {
+          "version": "2.6.12",
+          "resolved": "https://registry.npmjs.org/core-js/-/core-js-2.6.12.tgz",
+          "integrity": "sha512-Kb2wC0fvsWfQrgk8HU5lW6U/Lcs8+9aaYcy4ZFc6DDlo4nZ7n70dEgE5rtR0oG6ufKDUnrwfWL1mXR5ljDatrQ=="
+        },
+        "regenerator-runtime": {
+          "version": "0.11.1",
+          "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.11.1.tgz",
+          "integrity": "sha512-MguG95oij0fC3QV3URf4V2SDYGJhJnJGqvIIgdECeODCT98wSWDAJ94SSuVpYQUoTcGUIL6L4yNB7j1DFFHSBg=="
+        }
+      }
+    },
     "balanced-match": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
@@ -4836,6 +4857,11 @@
         "mimic-response": "^1.0.0"
       }
     },
+    "clsx": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/clsx/-/clsx-1.1.1.tgz",
+      "integrity": "sha512-6/bPho624p3S2pMyvP5kKBPXnI3ufHLObBFCfgx+LkeR5lg2XYy2hqZqUf45ypD8COn2bhgGJSUE+l5dhNBieA=="
+    },
     "co": {
       "version": "4.6.0",
       "resolved": "https://registry.npmjs.org/co/-/co-4.6.0.tgz",
@@ -5351,6 +5377,11 @@
           "dev": true
         }
       }
+    },
+    "csstype": {
+      "version": "3.0.6",
+      "resolved": "https://registry.npmjs.org/csstype/-/csstype-3.0.6.tgz",
+      "integrity": "sha512-+ZAmfyWMT7TiIlzdqJgjMb7S4f1beorDbWbsocyK4RaiqA5RTX3K14bnBWmmA9QEM0gRdsjyyrEmcyga8Zsxmw=="
     },
     "cwd": {
       "version": "0.10.0",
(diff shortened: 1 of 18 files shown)


trvrb:
Fantastic work here James! Observations during testing:
1. (I know you flagged this...) If I click on the "filter data" box here https://nextstrain.org/ncov/global it takes ~0.5 seconds to load a text cursor and provide list of clickable options, while if I click "filter data" box at  https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/global, it takes ~2 seconds to get to the same place.
2. Similarly, when deleting elements in the text box, getting back to zero text results in a ~2 second beachball.
3. It's a bit of a secret language that it needs to be `s:417`. I started with just `417` and got a bunch of samples with `417` in their name before mutations and then I tried `s 417` and nothing came up. I know that we label branches tooltip as "S: K417N...", but our clade spec calls for labels of `20C/S.417N` with a `.` rather than a `:`. We use the `.` notation in places like https://nextstrain.org/groups/neherlab/ncov/S.E484?c=gt-S_484.
4. I know this is special casing but I'd push "mutation" for `s` in particular above "sample". I wanted to just type `417` (or `501`) and something useful to happen. If I just type one of these I have to scroll through a bunch of options to find what I was looking for. This said, the `S.` entries could just be pushed to the top of the mutation list.
5. A little strange that the URL for genotype filtering is `gt=S:501Y` while other filters are of the form `f_division=Washington`. Maybe `f_gt` instead? Colorby is `c=gt-` vs `c=division`.
6. The genotype filter pillboxes in the header don't list number of tips with a genotype. We have `Washington (20)` vs `S:501Y`. That said I think it's going to be too many numbers if we had `S:501Y (429)`, so agree with this design decision. Side note: I like including sample counts in the footer text, but I'm not sure if it's as useful in the header.
7. I think that exposing this through magic "filter data" select box is definitely the right place to start, but that said, most people won't discover it this way. I'd imagine immediate use would most come from deep-linking into nextstrain.org by CoVariants, etc... to highlight constellations of interest. Surfacing genotypes as filters falls into our general issue of how to highlight this more via UI (another sidebar tab? or perhaps other options as we've discussed). Though keeping in universal search in the sidebar definitely makes total sense for now.

Generally, this will definitely be useful for flu as well: https://auspice-filter-by-genot-lf7x8o.herokuapp.com/flu/seasonal/h3n2/ha/2y?gt=HA1:159Y,198P. And I can see it working for variants of interest. `20C/S.542R` ie `CAL.20C` is: https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/north-america?f_clade_membership=20C&gt=S:452R. Or the "Ohio variant" is seen by searching for US sequences with 501Y: https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/north-america?f_country=USA&gt=S:501Y.

The performance issue is blocker for me for release (but not for merge). The magic text boxes really need to be snappy to be truly a good user experience.

trvrb:
Other thought here is that in written text it's most common to say `S 417N`, so this might be the best way to list these even if our clade labels use `.`. This is most likely to work with how people would naturally type things into the filter box. Same for `HA1 159Y` in flu.

jameshadfield:
Thanks @trvrb!

> The performance issue is blocker for me for release (but not for merge). The magic text boxes really need to be snappy to be truly a good user experience.

Absolutely, and thankfully this turned out to be easier than feared 😌 The performance has been greatly improved by f9944eb, and now seems good enough for production IMO. 

> It's a bit of a secret language that it needs to be s:417 ... 

Yeah. I've changed to using `.` in the URL query (to match prior art) and use a space in the rendering now, which is nicer but still somewhat secret. I think URL sharing will expose these to most users, but agree there's more to be done here.

> I know this is special casing but I'd push "mutation" for s in particular above "sample". I wanted to just type 417 (or 501) and something useful to happen. If I just type one of these I have to scroll through a bunch of options to find what I was looking for. This said, the S. entries could just be pushed to the top of the mutation list.

I've pushed mutations above strains, and sorted them, which help a bit. But one still needs to type the gene in most cases. I'll think about this a bit more. 

> A little strange that the URL for genotype filtering is gt=S:501Y while other filters are of the form f_division=Washington. Maybe f_gt instead? Colorby is c=gt- vs c=division.

So there is a reason for this, but perhaps it's a bit ridiculous. Note that filtering to strain(s) behaves similar -- they get `?s=` not `?f_strain=`. Since "gt" and "strain" have multiple meanings, this allows JSONs to define node-traits "gt" and "strain" without any issues (they would be filterable via `f_gt` and `f_strain`). 

> The genotype filter pillboxes in the header don't list number of tips with a genotype.

Yes. This is inconsistent but I think i'll leave it for another day. And as you say, we may choose to remove them from the header as they don't update when filtering, so are arguably more confusing than helpful. 

P.S. Bundlesizes are causing CI to fail, but I'll tackle this with #1266. 

P.P.S here are the updated URLs from your comment:
- https://auspice-filter-by-genot-lf7x8o.herokuapp.com/flu/seasonal/h3n2/ha/2y?gt=HA1.159Y,198P
- CAL.20C is: https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/north-america?f_clade_membership=20C&gt=S.452R
- "Ohio variant" is seen by searching for US sequences with 501Y: https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/north-america?f_country=USA&gt=S.501Y.

trvrb:
This all looks great. Thanks for making these changes. Performance is great now. The URL query is fine as it stands. I don't have any other suggestions for the PR.

huddlej:
This is really cool, @jameshadfield! After playing around with flu data (where I know what to look for), I really like the performance and UI experience, especially with the ability to type "mutation" or "HA1" to shortcut to specific sites.

You mentioned this, but I wish I could select multiple alleles for the same site and have them OR'd in the display. For example, all recent H3N2 strains have a HA1:186 mutation, but [some have 186S and others have 186D](https://auspice-filter-by-genot-lf7x8o.herokuapp.com/flu/seasonal/h3n2/ha/2y?gt=HA1.186D,186S). The number of results tells me there's an intersection happening behind the scenes, but the UI set logic tells me it should be a union. I don't think it's a blocking issue for a secret new feature, but I could see users getting caught by this when all the other filters within a group (like regions) are displayed additively.

image

Actually, as I play with this more, I see that even across different positions, the set UI doesn't indicate that mutation filters provide an intersection:

image

I guess for the OR'ing to work across positions, you'd have to treat each position (HA1:186, etc.) as its own filter category like region, country, etc.? But then you could still intersect across sites (categories intersect) and show multiple alleles per site (within categories union).

emmahodcroft:
This is so very cool James! I am so excited to have this in action 🌮 🎉  (that was supposed to be 2 'tada's, but I decided tacos are equally celebratory).

I didn't read John's comment thoroughly enough and immediately got into the same scenario as him. Though I did figure it out, it confused me initially and took me a few minutes to figure out why.  I'm not sure how to handle that - looking to highlight both 501Y and 501T, for example.  Perhaps as John says, making it a little clearer at the top would help this?

I also don't know if others would find it helpful or not - or if it would even be possible! - but the other thing I ran into was that I was trying out random mutations I spotted in the tree, and I'd immediately forget while typing which one was wild-type/reference & which was the mutation. I guess if you're looking at this more specifically you'd always know, but it did occur to me that maybe wild-type could be marked by an asterisk or something? But also maybe not worth the trouble!

trvrb:
>I guess if you're looking at this more specifically you'd always know, but it did occur to me that maybe wild-type could be marked by an asterisk or something? But also maybe not worth the trouble!

I'm worried that asterisk would be confused with stop codon. Immediate thought here would be to surface the count observed for different genotypes like `mutation S 501Y (169)` vs `mutation S 501N (3851)`. But this has the same issue of being too many numbers and I don't really like it.

Though the thing that this makes me realize is that `mutation S 501N` as currently displayed is incorrect. This is not a "mutation" it's a genotype because it's wildtype. Generally I would write the "N501Y mutation" where you state the change from one AA to another vs the "501Y genotype" that just states the state of the AA at a position. I'd change the dropdown text to `genotype S 501N` and `genotype S 501N`. This will match how we label color by as well, where this is also `Genotype`.

jameshadfield:
> This is not a "mutation" it's a genotype because it's wildtype.

Agreed - have changed the wording to "genotype" in 0ac6701

> I wish I could select multiple alleles for the same site and have them OR'd in the display.

Thanks @huddlej for this feedback -- it made me realise this was important to do right away rather than as a later improvement. I also took the oppertunity to improve the on-hover info boxes for filters in general. This is how it now looks:

image

emmahodcroft:
@jameshadfield The ability to filter to multiple AAs at the same position is *awesome*. 🚀

huddlej:
Thank you, @jameshadfield! Totally agree with @emmahodcroft here. This is amazing. I'm going to use this today for our upcoming flu call. :)

trvrb:
Amazing! I don't have further suggestions here. Looking forward to seeing merged.

trvrb:
Actually... I just found one small issue:

- If you navigate to https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/north-america?f_clade_membership=20I/501Y.V1 and click on "zoom to selected" it properly zooms to the clade of interest.
- while if you navigate to https://auspice-filter-by-genot-lf7x8o.herokuapp.com/ncov/north-america?f_clade_membership=20C&gt=S.452R and click "zoom to selected" it doesn't properly get the subclade bearing 452R.

PR #1272, opened 2021-01-22
Webpack hash

### LINKED ISSUE #1266: Ensure bundle hashes are appropriately computed
**Context**

We recently observed hard-to-debug bugs when deploying nextstrain.org which are caused by errors in how we compute bundle hashes during bundling (using webpack). These errors were (and sometimes still are) hard to pin down and reproduce. The core of the issue is that we encode the hash of each chunk in the filename so that chunks which haven't changed between different auspice versions (or nextstrain.org updates) look the same, which allows browsers to utilise cached assets. Any changes to chunk contents should result in a different filename, and thus the browser will not use the (previously-)cached asset. If the situation arrises where chunks are named the same, but have different content, then the browser may use the cached asset, thus resulting in an auspice client whose code is a combination of different auspice builds and unpredictable behavior will ensue. 

**Current Behavior**  

Sometimes, after rebuilding auspice (or nextstrain.org) you can get seemingly-random bugs, including complete app crashes, entropy-panel weirdness etc. These will go away if you reload the page with the cache disabled.

Inspecting the assests (`dist/*js`) of the previously build version (which has been loaded in a browser and cached) and the subsequent version (which is buggy when loading in the browser, with cache access enabled) indicates that there are JS bundle chunks with identical filenames but different contents. 

We use the `WebpackChunkHash` plugin to create hashes for chunks, which determines the chunk filename. This step apparently happens before minimisation/uglification steps (but this hasn't been confirmed for our webpack version/config).  Inspecting the bundles themselves (see previous paragraph), it seems that the differences do indeed stem from minification/uglification steps.

**Expected behavior**  

Changes in final bundle contents should result in a changed bundle filename, therfore not allowing a browser to use a cached resource.

**How to reproduce**  

> As this bug is related to webpack minisation & chunk creation, it cannot be reproduced using `auspice develop` or if we remove the minimisation steps from the webpack config.

These steps seem to work reproibly for me, but note that the bug is general and there will be many different paths to recreate it, which may appear different in terms of how the client crashes / behaves.

1. `cd auspice && mkdir ../auspice-bundle-hash-bug && mkdir ../auspice-bundle-hash-bug/a && mkdir ../auspice-bundle-hash-bug/b`
1. `git checkout 7ecbd6e3` (v 2.22.2)
1. `npm ci` (ensures the dep tree is correct) 
1. `auspice build && auspice view`
1. Load `http://localhost:4000/zika` with the cache disabled (in Firefox, this is a checkbox in the network tab of dev tools)
1. Confirm the app renders appropriately
1. Uncheck the "cache disabled" box. Keep the page open (shouldn't be necessary, but makes life simpler).
1. close the server process
1. `cp dist/*js ../auspice-bundle-hash-bug/a`
1. `git checkout 960dfea9` (`filter-by-genotype` branch, others could probably be used)
1. `auspice build && auspice view`
1. reload the open page, `http://localhost:4000/zika`, making sure the cache disabled box is _unchecked_. Note that the ~frequency~ entropy panel is broken.
1. Confirm that this is fixed by reloading the page with "cache disabled" ticked.
1. `cp dist/*js ../auspice-bundle-hash-bug/b`
1. `cd ../auspice-bundle-hash-bug`
1. `for i in $( ls a ); do if [[ -f b/$i ]]; then cmp --silent a/$i b/$i || echo "$i has different contents!"; fi; done;`
1. I get the following output (perhaps yours differs? The important thing is that there are chunks with the same filename which have different contents.)
```
auspice.chunk.12.bundle.1c2d5e3a950d2195bb40.js has different contents!
auspice.chunk.8.bundle.93172f6ce1a138aef270.js has different contents!
```

**Possible solution**  

Debug & fix webpack config.

Upgrade to webpack 5 so we can use [`optimization.realContentHash`](https://webpack.js.org/configuration/optimization/#optimizationrealcontenthash).

kairstenfay:
Drop third party webpack-chunk-hash library which appears to be abandonware and is causing unexpected behavior in our cached auspice files. 

### Related issue(s)  

Fixes #1266 

### Testing
Follow James's [handy list of steps to reproduce the initial bug](https://github.com/nextstrain/auspice/issues/1266#issue-786386345).
With this patch, the entropy panel is no longer broken when cache is not disabled.
In step 17, the bundles also have identical contents.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "auspice",
-  "version": "2.22.0",
+  "version": "2.22.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -17956,14 +17956,6 @@
         "mkdirp": "^0.5.1",
         "opener": "^1.5.1",
         "ws": "^6.0.0"
-      }
-    },
-    "webpack-chunk-hash": {
-      "version": "0.6.0",
-      "resolved": "https://registry.npmjs.org/webpack-chunk-hash/-/webpack-chunk-hash-0.6.0.tgz",
-      "integrity": "sha512-FsOg1RpW2nf3nYpGTy/Qs59RZ7gYG+sI4VrCE8TIBQYh/Kogi04xD39Pj9zUEeUcNx9HeTVPGSO3mtmpLeX9eQ==",
-      "requires": {
-        "@types/webpack": "^3.0.0 || ^4.0.0"
       }
     },
     "webpack-cli": {
--- a/package.json
+++ b/package.json
@@ -124,7 +124,6 @@
     "typeface-lato": "^0.0.75",
     "webpack": "^4.30.0",
     "webpack-bundle-analyzer": "^3.3.2",
-    "webpack-chunk-hash": "^0.6.0",
     "webpack-cli": "^3.1.2",
     "webpack-dev-middleware": "^3.1.3",
     "webpack-hot-middleware": "^2.24.3",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -6,7 +6,6 @@ const fs = require('fs');
 const utils = require('./cli/utils');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const { CleanWebpackPlugin } = require('clean-webpack-plugin');
-const WebpackChunkHash = require('webpack-chunk-hash');
 const LodashModuleReplacementPlugin = require('lodash-webpack-plugin');
 
 /* Webpack config generator */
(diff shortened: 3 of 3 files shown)


jameshadfield:
Looks good, thanks!

I tested by repeating the steps to reproduce the bug and cherry-picking this commit onto each of the 2 commits examined there (in steps 2 and 10), with the resulting bundles appearing correct. I'll re-test tomorrow using a different method to double check and merge if (as expected) it works.

kairstenfay:
> Looks good, thanks!
> 
> I tested by repeating the steps to reproduce the bug and cherry-picking this commit onto each of the 2 commits examined there (in steps 2 and 10), with the resulting bundles appearing correct. I'll re-test tomorrow using a different method to double check and merge if (as expected) it works.

Thanks, @jameshadfield !

PR #1276, opened 2021-01-26
Find smallest subtree with all visible data when using genotype filters

jameshadfield:
Previously we were not recomputing the MRCA of the the filtered nodes if genotype filters were applied, which resulted in the "zoom to selected" button behaving as if the genotype filters did not exist.

For non-genotype ("normal") filters, given a set `X` of visible nodes, we simultaneously find the MRCA  of `X` and add the paths from the MRCA to `X`.

For genotype filtering, we wish to find the MRCA[1] of `X` _without_ modifying `X`.  Note that this allows situations where `MRCA \not\in X`. This is introduced in this commit via the function `findFilteredMRCA` which uses a 3 step process:

1. Find the basal-most nodes of each (potentially non-monophyletic) visible cluster
2. Identify the paths from the root to the nodes in (1).
3. Find the first fork in this set of paths

image

_Filtering to clade 20C and genotype S 452 R. Left: "Zoom to selected" from PR #1265 wrongly identifies the MRCA of clade 20C and does not take into accound the genotype filtering. Right: This PR_

image

_Filtering to the homoplasic mutation 501Y. Left: In PR #1265, filtering to genotypes only did not allow one to "zoom to selected". Right: We can zoom into the subtree containing all samples with 501Y (which happens to be quite close to the root of the tree)_

[1] Is there a better word than MRCA? I'm not suggesting that this is the node where the genotypes originated, rather the node which contains the smallest subtree with all of the filtered nodes within it.

### DIFF
--- a/src/util/treeVisibilityHelpers.js
+++ b/src/util/treeVisibilityHelpers.js
@@ -178,7 +178,7 @@ const getFilteredAndIdxOfFilteredRoot = (tree, controls, inView) => {
     idxOfFilteredRoot = hideNodesAboveVisibleCommonAncestor(filtered, tree.nodes[0]);
   }
 
-  filtered = performGenotypeFilterMatch(filtered, controls.filters, tree.nodes);
+  ([filtered, idxOfFilteredRoot] = performGenotypeFilterMatch(filtered, controls.filters, tree.nodes) || [filtered, idxOfFilteredRoot]);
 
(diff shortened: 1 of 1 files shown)


trvrb:
This behavior looks great to me. Functions exactly as I'd expect it. I think that "MCRA" is appropriate. It's the MRCA of the set of tips in question. You can have the MRCA for an arbitrary set of tips.

Issue #1224, opened 2020-10-27
Normalized Frequency Panel Doesn't Respond to Date Sliders When Zoomed to Clade


emmahodcroft:
**Current Behavior**  
When zoomed in to some clusters (possibly particularly ones that have more recent common ancestors), with the "Normalize frequencies" option turned `on`, moving the date-sliders does not have any impact on the frequency panel shown.

It seems like if "Normalize frequencies" is `off` then this worked as expected. If you toggle between the two, the normalized frequencies seem to update. 

This may be related to triggering a refresh - if you resize the window after changing date sliders, the frequencies will update as expected.

**Expected behavior**  
As happens with "Normalize frequencies" turned `off`, the frequencies panel should update if the date sliders are dragged, even if "Normalize frequenices" is `on`. 

**How to reproduce**  
Steps to reproduce the current behavior:  
1. Go to https://nextstrain.org/ncov/europe
2. Zoom in to a recent clade with MRCA >~May
3. Drag date slider until it starts affecting the zoomed in clade view
4. See that frequencies view hasn't changed
5. Can toggle "Normalize frequencies" to off and repeat if want to see how it does adjust.

**Your environment: if browsing Nextstrain online**  
 - Operating system:  Windows 10
 - Browser:  Chrome

PR #1278, opened 2021-01-27
Enforce unnormalized frequencies when data is lacking

### LINKED ISSUE #1225: Frequency Panel Shows 'All Equal' Before Estimated Date of Ancestor on Zoomed-in Cluster
**Context**  
This is an improvement of the current frequency panel view, when "Normalize Frequencies" is turned `on`. 

**Description**   
Currently if you zoom in to a more recent cluster, then for any dates prior to the estimated most recent common ancestor (MRCA of that cluster, the frequency panel (with "Normalize frequencies" `on`) will show all available option as equally likely. This could be very misleading for some traits - shown here is a cluster that is entirely clade 20A, yet prior to the estimated MRCA of the cluster (early May), it is implied that all clades were equally likely - perhaps on the whole tree or perhaps as the ancestor for this cluster.

image

**To Reproduce**
Go to https://nextstrain.org/ncov/europe?c=clade_membership&d=tree,map,frequencies&f_region=Europe&p=grid and click on a clade that begins roughly after May or so. (Can also try other color-by options)

**Possible solution**   
One option would be to allow/enable x-axis zooming on the normalized frequencies panel, and only ever show 'starting from' the date of the MRCA of the currently visible tree section. 
Another option would be to just not show frequencies for dates prior to the MRCA of the currently visible tree section - instead starting the estimation of frequencies only from the date of that MRCA. (Showing white, black, grey, etc, before that time.)

**Notes**
@rneher adds this (from slack discussion):
> the normalization is tricky. The is no data to inform frequencies prior to the root of the clade. The KDE frequencies we use essentially send their Gaussian tail into the past. but this gets to be a very small number which causes underflow. we fix this by adding a small number to each category which results in uniform frequencies when there is no data.

@jameshadfield feel free to re-label as `bug` if more appropriate!

trvrb:

This commit forces `controls.normalizeFrequencies` to be false if there are any pivots where the total frequency is less than 0.1%.

This addresses two issues:
1. In the existing code, attempting to normalize situations where pivots have 0% total frequency results in bad looking "all equal" bands. This commit removes the capacity to get into this bad looking app state.
2. When filtering to a particular clade, we often want to switch to unnormalized frequencies anyway (as opposed to filtering to geography). This commit accomplishes this automatically because filtering to an emerging clade will generally result in pivots with <0.1% frequency.

### Related issue(s)  
Fixes #1225   

### Testing
Tested locally through `npm run dev`.

@jameshadfield --- I'm still not great at redux dataflow and there are places where I set `controls.normalizeFrequencies = false` that may need to be corrected. Everything seemed to work as it should in my testing, but that's not a guarantee that I'm not doing something wrong.

### DIFF
--- a/src/actions/frequencies.js
+++ b/src/actions/frequencies.js
@@ -5,9 +5,11 @@ import { computeMatrixFromRawData, checkIfNormalizableFromRawData, processFreque
 
 export const loadFrequencies = (json) => (dispatch, getState) => {
   const { tree, controls } = getState();
+  const { data, pivots, matrix, projection_pivot, normalizeFrequencies } = processFrequenciesJSON(json, tree, controls);
   dispatch({
     type: types.LOAD_FREQUENCIES,
-    frequencies: {loaded: true, version: 1, ...processFrequenciesJSON(json, tree, controls)}
+    frequencies: {loaded: true, version: 1, data, pivots, matrix, projection_pivot},
+    normalizeFrequencies
   });
 };
 
@@ -23,16 +25,8 @@ const updateFrequencyData = (dispatch, getState) => {
     return;
   }
 
-  const allowNormalization = checkIfNormalizableFromRawData(
-    frequencies.data,
-    frequencies.pivots,
-    tree.nodes,
-    tree.visibility
-  );
-
-  if (!allowNormalization) {
-    controls.normalizeFrequencies = false;
-  }
+  const normalizeFrequencies = controls.normalizeFrequencies &&
+    checkIfNormalizableFromRawData(frequencies.data, frequencies.pivots, tree.nodes, tree.visibility);
 
   const matrix = computeMatrixFromRawData(
     frequencies.data,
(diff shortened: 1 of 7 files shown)


trvrb:
Hi @jameshadfield I know you have a million priorities right now, if this is a quick review, I'd love some attention here. Filters like https://nextstrain.org/ncov/global?gt=S.501Y or https://nextstrain.org/ncov/global?f_clade_membership=20H/501Y.V2 are super common now, but the resulting rainbow band in the frequencies panel looks pretty broken.

jameshadfield:
This looks great! I added a few small commits to
* bbef8f8 display that frequencies cannot be normalized in the sidebar, rather than rendering a toggle that didn't work
* 0de2197 no changes to behaviour, but uses a more redux-y pattern of updating state via reducers
* 658f8dc fix a nasty :bug: where the frequency panel wouldn't update. Not technically part of this PR, but that's where I ended up doing the work.

Closes #1224
Closes #1225

jameshadfield (approved):

trvrb:
Fantastic! Thanks for the additional cleanup here @jameshadfield. And thanks for the redux help. This is ready to be merged from my perspective.

PR #1280, opened 2021-01-28
Show root-to-tip mutations in tip-clicked info box

### LINKED ISSUE #1259: Clicking on a tip should display full list of AA changes
In v20 and below, you could select a single tip and then change Diversity panel to "events" to show all amino acid changes in this tip:

In addressing #1240, hiding ancestral nodes means the Diversity panel no longer shows these events: 

@trvrb suggests the best solution to address this is might be adding a full list of amino acid changes to the info panel of the selected tip.

jameshadfield:
Initial version for testing and feedback, especially on datasets with lots of mutations!

cc @joverlee521 @emmahodcroft @trvrb 

closes #1259

### DIFF
--- a/src/components/tree/infoPanels/click.js
+++ b/src/components/tree/infoPanels/click.js
@@ -81,7 +81,7 @@ const MutationTable = ({mutations}) => {
     Object.entries(mutations)
       .sort(geneSortFn)
       .map(([gene, muts], index) => (
-        item(index === 0 ? "Mutations" : "", gene + ": " + [...muts].sort(mutSortFn).join(", "))
+        item(index === 0 ? "Mutations from root" : "", gene + ": " + [...muts].sort(mutSortFn).join(", ").substring(0, 200))
       ))
   );
 };
--- a/src/components/tree/infoPanels/click.js
+++ b/src/components/tree/infoPanels/click.js
@@ -78,24 +78,14 @@ const MutationTable = ({mutations}) => {
     return aa<bb ? -1 : 1;
   };
   return (
-    <table style={{marginTop: "10px"}}>
-      <tbody>
-        <tr>
-          <th style={{minWidth: "60px"}}>Gene</th>
-          <th>Mutations</th>
-        </tr>
-        {Object.entries(mutations)
-          .sort(geneSortFn)
-          .map(([gene, muts]) => (
-            <tr key={gene}>
-              <td>{gene}</td>
-              <td>{[...muts].sort(mutSortFn).join(", ")}</td>
-            </tr>
-          ))}
-      </tbody>
-    </table>
+    Object.entries(mutations)
+      .sort(geneSortFn)
+      .map(([gene, muts], index) => (
+        item(index === 0 ? "Mutations" : "", gene + ": " + [...muts].sort(mutSortFn).join(", "))
+      ))
   );
 };
+
 
 const AccessionAndUrl = ({node}) => {
   const accession = getAccessionFromNode(node);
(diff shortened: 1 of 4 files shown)


trvrb:
I've cleaned up the layout a bit in this commit:

mutations

trvrb:
@jameshadfield: There is one complication here that I think I know what to do, but want to flag because I suspect you're equipped to handle this quicker than I am. If you take a look at https://auspice-mutations-to-ti-bahczp.herokuapp.com/flu/seasonal/h3n2/ha/2y you'll see:

flu-muts

with `R142G` followed by `G142R` and `K171N` followed by `N171K`. You can see this is a real reversion on the tree here: https://auspice-mutations-to-ti-bahczp.herokuapp.com/flu/seasonal/h3n2/ha/2y?c=gt-HA1_171.

I think for these we need either to pick apart the mutations in the branch attrs to determine that site 171 is really `N171N` and so can be dropped or to do some actual comparison with the root node.

In current `collectMutations` there will also be an issue for sites of the form `X171Y`, followed by `Y171Z`.

And in general, we could call these "Mutations relative to root" to make what we're showing more clear.

emmahodcroft:
I really like this James - I can already see these lists being helpful! 

I think Trevor's points about how to handle reversions are good ones - and agree it's probably least confusing to show 'difference from root' (not show the mutation and reversion). Especially if someone's scanning quickly for one mutation, they might not notice the reversion!

src/components/tree/infoPanels/click.js
```diff
@@ -59,6 +60,33 @@ const Link = ({url, title, value}) => (
...
+  };
+  const mutSortFn = (a, b) => {
+    const [aa, bb] = [parseInt(a.slice(1, -1), 10), parseInt(b.slice(1, -1), 10)];
+    return aa<bb ? -1 : 1;
+  };
+  return (
+    Object.entries(mutations)
+      .sort(geneSortFn)
+      .map(([gene, muts], index) => (
+        item(index === 0 ? "Mutations from root" : "", gene + ": " + [...muts].sort(mutSortFn).join(", ").substring(0, 200))
```
joverlee521, line 84: Screen Shot 2021-01-29 at 4 57 44 PM
With this substring, the list of mutations gets randomly cut off when it's too long. 

If we want to limit the number mutations displayed, it might be better to slice the array first:
```javascript
const mutLimit = 30;
[...muts].sort(mutSortFn).slice(0, mutLimit).join(", ") + (muts.length > mutLimit && "...")
```

trvrb: That's definitely nicer.

emmahodcroft: If we are cutting off the display of mutations, we should also make this explicit in the box - so people don't think it's a complete list erroneously. (Apologies if this is integrated but doesn't show up in Jover's example above!)

jameshadfield: @trvrb given that #1259 asks to "display [the] full list of AA changes" why would we truncate this list? I changed back to displaying all AA muts in 4fee777 as I though this would be the expected behaviour - am I missing something?

trvrb: Try this out in dengue. It ends up being pages long. Good to truncate in this case.

trvrb: Got it. This makes sense. Thanks James. I just spent some more time with this PR and I think that the scrolly overlay panel is a fine solution. Having "mutations from root" at the end of the list of fields is helpful here. I'd agree that (2) is preferable to (1) for the moment.

trvrb:
Thanks for making these improvements @jameshadfield. I've just made a couple styling changes. Before commit 0a6fb8175d4fc1e33821002a73073d4a03988717, we had:

before

where "Mutations from root" on the left is not vertically lined up with the mutation list on the right. Also, the double-line "Mutations from root" was making the first `<tr>` wider than subsequent `<tr>`s.

This commit moves to a single `<tr>` for the entire mutation list and makes `infoPanelStyles.item` have `verticalAlign: "top"` and results in the following behavior:

after

This PR is now good to go from my perspective.

PR #1287, opened 2021-02-12
increase max number mutations to display in hover.js

babarlelephant:
increase max number mutations to display in hover.js

### Testing
Tested on time and divergence tree of https://nextstrain.org/charon/getDataset?prefix=/groups/neherlab/ncov/S.E484
and https://nextstrain.org/groups/blab/sars-like-cov

No exception raised. 
Away from long branches the normal behavior is preserved


babarlelephant:
The error suggests it wants me to upload the dist/*.js files too? I can do it if you want but my fix is only 3 lines

babarlelephant:
Thanks a lot to have solved this in **https://github.com/nextstrain/auspice/commit/1069d306ba96cb93378f710a5f64798f014bc870**

PR #1302, opened 2021-03-10
Legend no longer obscures branches/tips.

### LINKED ISSUE #1295: Legend masks branches and tips in upper left of tree panel
**Current Behavior**  
The legend is drawn over the upper left corner of the tree area, which sometimes has branches and tips that become less visible and unclickable because they're covered by the legend, even when the legend is minimized to just the label.  For example, in this image ([JSON file](https://hgwdev.gi.ucsc.edu/~angie/singleSubtreeAuspice_hgwdev_demo_angie_ab43_822a50.json)), all of the red sequences are obscured when the legend is not minimized:

image

and even when the legend is minimized there is still a sequence that can't be reached:

image

**Expected behavior**  

I expect to be able to click on any branch or tip in the tree.

**How to reproduce**  
1. Download [JSON file](https://hgwdev.gi.ucsc.edu/~angie/singleSubtreeAuspice_hgwdev_demo_angie_ab43_822a50.json)
2. Drag-drop onto https://auspice.us/  

(This also happens on nextstrain.org, e.g. https://nextstrain.org/fetch/hgwdev.gi.ucsc.edu/~angie/singleSubtreeAuspice_hgwdev_demo_angie_ab43_822a50.json)
  

**Possible solution**  
If the legend could be moved, perhaps to the right of the "Phylogeny" label (at least when it's minimized?), then it would not hide part of the tree.  Or if the legend could fit in its own vertical space when minimized, that would also keep it out of the tree.

 - Operating system:  macOS Catalina 10.15.7
 - Browser:  Chrome 88.0.4324.150

**Additional context**  
Having a wide web browser window on a large monitor ameliorates the problem by stretching the tree farther out to the right, but on a little laptop screen (like my PI's 🙂 ) it's worse.

If I'm missing an obvious way to avoid this problem please let me know!  Thanks.

jameshadfield:
Shifts the top of the tree down slightly so that tips and branches cannot be hidden behind the (closed) legend, which prevents interacting with them. This only happens in rectangular / unrooted trees, as radial / clock views almost never have tips rendered in the top-left corner.

image
_Note that the horizontal position of the tips in this screenshot has been changed for testing purposes_

This PR is a trade off -- we reduce the available display space for all (rectangular & unrooted) trees to prevent the (rare) cases of tips being rendered in the top-left corner and thus obscured by the legend. A more intelligent solution would be to make `mapToScreen()` aware of when tips are rendered behind the legend and only increase the margin in this case.

Closes #1295

### DIFF
--- a/src/components/tree/phyloTree/layouts.js
+++ b/src/components/tree/phyloTree/layouts.js
@@ -279,6 +279,10 @@ export const mapToScreen = function mapToScreen() {
     right: this.params.margins.right,
     top: this.params.margins.top,
     bottom: this.params.margins.bottom};
+  if (this.layout==="rect" || this.layout==="unrooted") {
+    // legend is 12px, but 6px is enough to prevent tips being obscured
+    tmpMargins.top += 6;
+  }
   const inViewTerminalNodes = this.nodes.filter((d) => d.terminal).filter((d) => d.inView);
   if (inViewTerminalNodes.length < this.params.tipLabelBreakL1) {
 


trvrb (approved):

trvrb:
This looks great to me. I think this can be merged. Testing suggests that 6px is a good amount of padding for this purpose.

Issue #1283, opened 2021-02-11
Narratives cannot display frequencies


jameshadfield:
Narratives cannot display frequencies panels in Auspice 2.33.0. I believe this _used_ to work in a previous version, but have not bisected to confirm. A test narrative showing this has been created on the [bug-narratives-frequencies](https://github.com/nextstrain/auspice/tree/bug-narratives-frequencies) branch -- please base any fix on that branch!

**How to reproduce**  

Steps to reproduce the current behavior:  
1. `get fetch`
2. `git checkout bug-narratives-frequencies`
3. `npm run get-data`
4. `auspice develop --narrativeDir narratives`
5. load http://localhost:4000/narratives/test/display-frequencies-panel

**Additional context**  

The frequencies sidecar JSON is not fetched when the narrative is loaded. When loading http://localhost:4000/ncov/global?d=tree,frequencies, the dataset the narrative is attempting to display and which correctly displays the frequencies when viewed by itself, the following `getDataset` requests are made:
```
http://localhost:4000/charon/getDataset?prefix=/ncov/global
http://localhost:4000/charon/getDataset?prefix=/ncov/global&type=tip-frequencies
http://localhost:4000/charon/getDataset?prefix=/ncov/global&type=root-sequence
```

However when loading the narrative, only the first of these is made. Auspice should be making all 3 requests in narratives mode.

P.S. I highly suspect that narratives won't be able to colour by non-varying bases, as the `root-sequence` JSON is not fetched either! I haven't created a narrative to test this, but will do at some point.

jameshadfield:
I've looked into this and the 🐛 arrived when we allowed multiple different datasets per narrative. When a narrative arrives, we call the function `fetchAndCacheNarrativeDatasets`, which (a) fetches the main dataset JSON and (b) fetches other dataset JSONs defined in the narrative and stores them in redux via `CACHE_JSONS`. This contrasts with loading a non-narrative page, where we use the function `fetchDataAndDispatch`, which not only fetches the main JSON but also sidecar JSONs, as required. 

Any solution should unify the dataset-fetching code into a single function (either async fn or async generator), which both narratives and non-narratives use. Additionally, the cache structure, currently simply `jsons[treeName]=jsonData`, will need to be expanded to `jsons[treeName] = {main: jsonData, "tip-frequencies": jsonData, ...}`.

jameshadfield:
Fixed in c190057

PR #1305, opened 2021-03-16
Unify data loading between narrative and non-narrative modes

jameshadfield:
This PR represents in-progress work to address issue #1283.

Dataset loading, as of v2.24.0, differs dramatically between narrative & non-narrative ("normal") mode. Specifically, the narrative mode failed to fetch sidecar files (tip-frequencies and root-sequence) which resulted in #1283. To some extent the differences in logic between the modes is unavoidable, however there are steps in common which this PR aims to isolate into functions, thus reducing duplicated logic which may diverge.

I may not have a chance to revisit this for a week or so, however there are a few places where 👀 would be welcome:

* Are there additions to `narratives/test_multiple-datasets.md` to capture examples not added via this PR?
* @eharkins this PR should correctly populate the cache with sidecar files (for datasets which differ from that of the starting page), however when changing pages those sidecar files aren't requested from the cache. Do you have ideas here?
* general comments on the abstractions started by 3c9d28c would be welcome


eharkins:
Rebased on master and merged #1312 to address 
> sidecar files aren't requested from the cache

Will continue reviewing and testing here.

src/actions/loadData.js
```diff
@@ -74,184 +75,118 @@ export const collectDatasetFetchUrls = (url) => {
...
- *
- * given a url to fetch, check if a second tree is defined.
- * e.g. `ha:na`. If so, then we want to make two fetches,
- * one for `ha` and one for `na`.
- *
- * @returns {Array} [0] {string} url, modified as needed to represent main tree
- *                  [1] {string | undefined} secondTreeUrl
- *                  [2] {string | undefined} string of old syntax
- */
-const collectDatasetFetchUrlsDeprecatedSyntax = (url) => {
```
eharkins, line 93: Does the existing `collectDatasetFetchUrls` handle the deprecated syntax or does this mean we are dropping the ability to parse the deprecated syntax here?

src/actions/loadData.js
```diff
@@ -74,184 +75,118 @@ export const collectDatasetFetchUrls = (url) => {
...
+    fetchJSON(apiCalls.tipFrequencies)
+      .then((data) => dispatch(loadFrequencies(data)))
+      .catch((err) => {
+        console.error("Failed to fetch frequencies", err.message);
+        dispatch(warningNotification({message: "Failed to fetch frequencies"}));
+      });
+  }
+  if (additionalRequests.rootSequence) {
+    fetchJSON(apiCalls.rootSequence)
+      .then((data) => dispatch({type: types.SET_ROOT_SEQUENCE, data}))
```
eharkins, line 161: Do we need to also dispatch `updateColorByWithRootSequenceData` here as in https://github.com/nextstrain/auspice/pull/1305/files#diff-7e6b4cc45e1472da7b3f8f6e7438315586ecbd5450bc7db2a5263a6fe2063e16L221 ?

src/actions/loadData.js
```diff
@@ -263,104 +198,225 @@ const narrativeFetchingErrorNotification = (err, failedTreeName, fallbackTreeNam
...
+ * A narrative (parsed into `blocks`) may define multiple datasets.
+ * This function returns the API endpoints needed to request these
+ * datasets (including sidecar files).
+ * Note: "initial": starting page, potentially different to frontmatter
+ */
+function createApiCallsForDatasetsInNarratives(blocks, query) {
+  let initialDatasetNames, frontmatterDatasetNames;
+  const datasetApiCalls = {};
+  const initialBlockIdx = getNarrativePageFromQuery(query, blocks);
+  collectDatasetFetchUrls(blocks[initialBlockIdx].dataset);
```
eharkins, line 408: I'm not sure I understand what calling this here does - I know it modifies the dataset url, but won't it also get called in the loop below in the same way? Does it need to happen twice for `blocks[initialBlockIdx].dataset`?

src/actions/navigation.js
```diff
@@ -27,17 +28,30 @@ export const chooseDisplayComponentFromURL = (url) => {
 
 /*
  * All the Fetch Promises are created before first render. When trying the cache we `await`.
- * If the Fetch is not finished, this will wait for it to end. Subsequent awaits will immeditaly return the result.
+ * If the Fetch is not finished, this will wait for it to end. Subsequent awaits will immediately return the result.
  * For the landing dataset, no problem either because await on a value just returns the value.
  */
 const tryCacheThenFetch = async (mainTreeName, secondTreeName, state) => {
```
eharkins, line 34: Should we rename this function to `awaitCachedDatasetPromise` or something more accurate? It doesn't fetch anything really if theres a cache miss, it just throws an error (since we shouldn't have cache misses if our cache-populating code is working the way I think it is).

eharkins (commented):
Hi @jameshadfield! Jumping back into the code, I left a few comments about specific lines. Here's my interpretation of what remains to be done:

- Use `narrativeFetchingErrorNotification`; https://github.com/nextstrain/auspice/pull/1305/files#diff-7e6b4cc45e1472da7b3f8f6e7438315586ecbd5450bc7db2a5263a6fe2063e16R290
- Handle hardcoded server calls in `parseUrlIntoAPICalls`; https://github.com/nextstrain/auspice/pull/1305/files#diff-7e6b4cc45e1472da7b3f8f6e7438315586ecbd5450bc7db2a5263a6fe2063e16R205
- It seems like you were intending on using google analytics exception tracking? Is this necessary for this PR or could it be in a separately-scoped one? https://github.com/nextstrain/auspice/pull/1305/files#diff-7e6b4cc45e1472da7b3f8f6e7438315586ecbd5450bc7db2a5263a6fe2063e16R14
- prevent duplicate dataset requests when collecting narrative datasets; https://github.com/nextstrain/auspice/pull/1305/files#diff-7e6b4cc45e1472da7b3f8f6e7438315586ecbd5450bc7db2a5263a6fe2063e16R334
- more testing in general of different narrative cases and specifically that #1312 is working. 

Please add to the list here if I missed some TODO items you had in mind :)

src/actions/loadData.js
jameshadfield: Yes we do! Now resolved.

src/actions/loadData.js
jameshadfield: Code updated, so I think this can be resolved?

src/actions/navigation.js
jameshadfield: Agreed - I've changed it to `updateNarrativeDataset` and expanded it to handle the dispatch itself

jameshadfield:
Thanks for the detailed review @eharkins - this prompted me to rewrite my original commit to further simply things. The concept of a "dataset" and its associated api calls, fetch results etc is now an object, so dataset views and narratives can both simply pass around instances of `Dataset`. I think this simplifies things quite a bit and prevents somewhat duplicated code.

I leant heavily on  http://localhost:4000/narratives/test/sidecar-files for testing, which identified numerous bugs (including stale sidecar data, broken tangletrees etc). I've also tested a large number of individual datasets, tangletrees and other narratives. The nextstrain.org review app should allow further testing still. 

@eharkins would you mind re-reviewing?

src/actions/loadData.js
eharkins: @jameshadfield it seems like in its current state this PR drops the ability to parse the deprecated second tree syntax, e.g. `/flu/seasonal/h3n2/ha:na/2y`, which works on the `master` branch. Is that intentional?

src/actions/navigation.js
```diff
@@ -26,18 +27,39 @@ export const chooseDisplayComponentFromURL = (url) => {
...
+        mainTreeName,
+        secondTreeName,
+        narrativeBlocks,
+        query,
+        dispatch
+      }),
+      pushState: true,
+      query
+    });
+    mainDataset.loadSidecars(dispatch);
```
eharkins, line 55: Makes sense in terms of frequencies to only fetch for the main dataset and not the second one since it seems like only the main dataset's frequencies can be displayed in a tanglegram view. What about for root sequences? Sorry, I'm not as familiar with how that file works / is displayed.

eharkins (commented):
This looks good. I still would like to test it more and get others to test it out as well before we merge but the code is much easier to understand in terms of logic and abstraction. A couple of abstraction notes: 

1. I know we talked about whether it would make sense to refactor `changePage` at this point but maybe that's best done in a separate PR. To maintain the best of both worlds, maybe we could make changePage only do case 3 where we actually change page. Then we could have a separate function called `changeDatasetState` or something like that which has all three options with appropriate option names like `queryUpdate`, `narrativeUpdate`, `changePage` to allow us some consistency if we do think these 3 cases belong under the same heading nominally.

2. Since you might need to be able to dispatch a clean state with two datasets, i think the abstraction level of the dataset object makes sense as is, since dispatching inside a prototype attached to the dataset object would become complicated in that case.

jameshadfield:
Thanks Eli - I agree that changes to `changePage` would make for a nicer (internal) API. I've re-tested and think this is good to go, so I'm going to merge and release to dev.nextstrain.org for a final testing pass.

PR #1312, opened 2021-03-27
narratives load sidecar files from cache;

eharkins:
As James and I sketched out verbally
over the phone, this tries to set up
then functions for promises in the dataset
cache which are initiated at the start
of loading a narrative with sidecar files.
This has been minimally tested with
the frequencies file and seems to work
on the surface.

@jameshadfield here's what we discussed, needs more testing.


trvrb:
Does this still need merging / review?

eharkins:
Thanks Trevor for reminding me about this. It needs to probably just get merged into #1305 since it is on top of that larger PR which lacks this small but important piece. Then further review and testing is definitely needed on #1305 before we merge, which I am working on now but would also appreciate your feedback.

PR #1313, opened 2021-03-29
Add support for "data_provenance" metadata

jameshadfield:
We now support data provenance to be defined in the dataset JSON
(see https://github.com/nextstrain/augur/pull/705) and all core
nCoV builds have been updated to include this here. This commit
parses and renders such information.

Note that a previous commit removed the "Build info" from the byline
for datasets displaying GISAID ([this commit](https://github.com/nextstrain/auspice/commit/18d5d21ac731d771c97bbefdc6ca20854fd1e2c2)), which I believe was an
oversight. This commit reinstates it.

Examples:

```json
 "data_provenance": [
      {
        "name": "GISAID"
      },
      {
        "name": "COG UK",
        "url": "https://www.cogconsortium.uk"
      },
      {},
      {
        "name": "source with invalid link",
        "url": "/bad.link"
      },
      {
        "name": "source without link"
      }
    ],
```

image

Rendering of /ncov/global with this PR:
image

For comparison, rendering of /ncov/global as of v2.24.1 (parent): 

image

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -648,6 +648,9 @@ const createMetadataStateFromJSON = (json) => {
   if (json.meta.genome_annotations) {
     metadata.genomeAnnotations = json.meta.genome_annotations;
   }
+  if (json.meta.data_provenance) {
+    metadata.dataProvenance = json.meta.data_provenance;
+  }
   if (json.meta.filters) {
     metadata.filters = json.meta.filters;
   }
--- a/src/components/info/byline.js
+++ b/src/components/info/byline.js
@@ -113,8 +113,10 @@ function renderDataProvenance(t, metadata) {
     .filter((source) => typeof source === "object")
     .filter((source) => Object.prototype.hasOwnProperty.call(source, "name"))
     .map((source) => {
-      if (source.name === "GISAID") { // SPECIAL CASE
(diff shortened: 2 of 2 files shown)


trvrb:
This looks great to me @jameshadfield. Thanks so much for implementing. The link parsing is a nice touch. This is good to go from my perspective.

huddlej (commented):
This worked for me, exploring the ncov builds in the Heroku deployment. We don't currently link out to GISAID from the button, but [there was a comment in the original issue about linking from the given image to GISAID's site](https://github.com/nextstrain/augur/issues/691#issuecomment-802180187). Is that functionality we still want? Should it be in a separate PR, so this most important functionaity can get this out as soon as possible?

emmahodcroft:
Good catch @huddlej . My expectation is that GISAID would likely prefer if that logo links to their website. But that might take a bit of fiddling, I agree that getting this live first and then working to incorporating the linking would be best, so that other builds can implement this quickly.

trvrb:
I didn't realize that the link out to gisaid.org was missing from the live version. We should make this addition. It won't be hard.

trvrb:
Okay. This commit adds a link to gisaid.org and will also catch `data_provenance.name` = `gisaid` (which seemed like it could be a common input). I _think_ this can be merged now.

jameshadfield:
Thanks all ⭐

PR #914, opened 2020-03-04
Handle display legend long name #899

dnprock:
Potential fix for issue #899.

PR #1320, opened 2021-03-30
Display longer legend names

### LINKED ISSUE #899: poor legend display for long names

### LINKED ISSUE #1306: Legend items are limited to 5 genotypes
**Current Behavior**  
When coloring by multiple genotypes ala https://nextstrain.org/ncov/global?c=gt-S_5,95,253,477,484,701, if using 6 or more sites, the legend is cutoff at 5 sites.

Here, the legend title lists "Genotype at S site 5,95,253,477,484,701" but each legend item only has 5 states, as in "L / T / D / S / E". All the legend items are present, it's just that the text is truncated.

**Expected behavior**  
It looks like `legend.js` just displays what's in `visibleLegendValues`. I'm not sure where the truncation is coming in. In any case, I'd think to solve this by adding an ellipsis when necessary to legend values.

jameshadfield:
Legend values were truncated via PR #914, which closed #899.

These can now be relaxed as testing indicates that legend values are no
longer being rendered behind other items. E.g.

image

Closes #1306

---

Some more testing would be appreciated to ensure we don't run into #899 again! (We should rewrite the legend to better handle all the edge cases, however that is for another day!)


jameshadfield:
This solution isn't good enough (see screenshot). I'll write some ideas for solutions on #1306. 

image

PR #1321, opened 2021-03-30
Zoom to selected is stored as a URL query

jameshadfield:
When clicking "Zoom to selected" we now add a URL query
`treeZoom=selected`, which is cleared upon the next zoom action.
This allows us to load a page to a zoomed in view of the tree.
This behavior is similar to when zooming to a branch which a clade
label, which results in a query such as `label=clade:A1`. In the
situation where both are applicable, the clade label takes
priority.

Note that due to the current implementation of zooming in Auspice,
"zoom to selected" will zoom to different subtrees. As an example,
load each of the following datasets and click "zoom to selected":
`/flu/seasonal/h3n2/ha/3y?f_clade_membership=A1b/137F,A1b/197R`,
`/flu/seasonal/h3n2/ha/3y?f_clade_membership=A1b/137F,A1b/197R&label=clade:A1b/131K`.
The resulting subtrees will be different for each case, however both
result in a URL query `treeZoom=selected`. This is a shortcoming
of the implementation used in this PR, however one that is not easily
solved.

### DIFF
--- a/narratives/test_simultaneous-tree-updates.md
+++ b/narratives/test_simultaneous-tree-updates.md
@@ -19,6 +19,10 @@ Check that both the branches and tips update.
 # [P3: Zoom into clade A1b _and_ change color](http://localhost:4000/flu/seasonal/h3n2/ha/3y?d=tree&label=clade:A1b)
 Check that the coloring of the branches and tips update as we zoom in.
 
+# [P3b: Zoom in slightly more via "zoom to selected"](http://localhost:4000/flu/seasonal/h3n2/ha/3y?d=tree&f_clade_membership=A1b/186D,A1b/197R,A1b/94N&treeZoom=selected)
+
+We're now recreating the functionality of filtering to clades A1b/186D, A1b/197R and A1b/94N then clicking "zoom to selected"
+
 # [P4: Lots of simultaneous changes](http://localhost:4000/flu/seasonal/h3n2/ha/3y?c=lbi&d=tree&dmin=2017-01-01&f_region=North%20America&label=clade:3c2.A&m=div)
 * Zoomed out to near the root (clade 3c2.A)
 * Changed the horizontal scale to divergence
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -3,7 +3,7 @@ import { cloneDeep } from 'lodash';
 import { numericToCalendar, calendarToNumeric } from "../util/dateHelpers";
 import { reallySmallNumber, twoColumnBreakpoint, defaultColorBy, defaultGeoResolution, defaultDateRange, nucleotide_gene, strainSymbol, genotypeSymbol } from "../util/globals";
 import { calcBrowserDimensionsInitialState } from "../reducers/browserDimensions";
-import { getIdxMatchingLabel, calculateVisiblityAndBranchThickness } from "../util/treeVisibilityHelpers";
+import { getIdxMatchingLabel, calculateVisiblityAndBranchThickness, getFilteredAndIdxOfFilteredRoot } from "../util/treeVisibilityHelpers";
 import { constructVisibleTipLookupBetweenTrees } from "../util/treeTangleHelpers";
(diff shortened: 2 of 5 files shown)


jameshadfield:
@joverlee521 would you be able to review this? I can think of a solution where we calculate something like `idxOfFilteredRootWhenTreeIsNotZoomed`, and then use this as [the comparison here](https://github.com/nextstrain/auspice/pull/1321/files#diff-5e3fb4f3fcfacf59fa9eb7563c5bf6d646310d7c32b0150a848aebd4e44f27a4R116), but I'm not sure if it's worth it, and we would want to ensure that calculation was efficient.

joverlee521:
@jameshadfield I might be missing something here, but why doesn't the query parameter`&label=clade:A1b/131K` stay in the the URL? I think that would help differentiate the two subtrees via the URL.

jameshadfield:
> but why doesn't the query parameter `&label=clade:A1b/131K` stay in the the URL? I think that would help differentiate the two subtrees via the URL.

Am I right in thinking you are performing the following steps:
1. Go to /flu/seasonal/h3n2/ha/3y?f_clade_membership=A1b/137F,A1b/197R&label=clade:A1b/131K
2. Click "zoom to selected"
3. `label=clade:A1b/131K` is now removed?

This behaviour isn't different to master: as we are no longer zoomed to the branch annotated as `clade:A1b/131K` it is removed from the URL. You raise an interesting point, that if both were in the query we could interpret it as "first zoom to A1b/131K then zoom to selected", however this won't generalise as in many cases the tree interaction isn't recorded in the URL. For instance, take this example:
1. Load https://auspice-bug-narratives--7d2cif.herokuapp.com/flu/seasonal/h3n2/ha/3y?f_clade_membership=A1b/137F,A1b/197R
2. Click a branch _without_ any clade labelling (see screenshot)
3. Click "zoom to selected"
4. Reload the page - notice that the view

(The reason I chose a branch with clade labelling originally was to make reproduction of the bug easier so one didn't have to click too much 😉 )

image

joverlee521:
Ahhh, I see the no-name branches do make this trickier. Like you said, I don't see this getting easily solved. 
Maybe we can just note this caveat in our documentation:  

The `treeZoom` URL parameter will only zoom based on the URL filters on initial load. 
Additional zoom filters through interactions with the tree that are not reflected in the URL will be lost on re-load of the page.

jameshadfield:
I rebased this and changed the code a little bit as things have moved on in the last 3 years. This URL query is not like any of the others as its behaviour depends on previous state which is not recorded in the URL query. When we are interacting with a tree and press "filter to selected" I think there are to distinct cases:

Case 1: All tips which match the selected filters are already displayed on screen when we click "zoom to selected". I think this _always_ works well with the URL query. For instance, load [H3N2 zoomed into clade 3C.2a1b.2a](nextstrain.org/seasonal-flu/h3n2/ha/3y?d=tree&label=clade:3C.2a1b.2a&p=full) (panel A), filter to clades 2a.3 & 2b (which are by definition within the subtree of clade 3C.2a1b.2a) (panel B), then click zoom to selected (panel C). This is loadable via the URL query `?treeZoom=selected`

image

Case 2: There are tips matching the selected filters which are off-screen (not in view). The URL query doesn't work as expected in these cases. Perform the same steps as in case 1, but instead filter to region=Europe (panel B) then click zoom to selected (panel C), which zooms in relative to panel B. When reloading the page we've lost the contextual information that we were zoomed in to clade 3C.2a1b.2a and so we zoom to show all European samples (panel B).

image


I think these should be differentiable in the code - I'll try to sort this out tomorrow.

jameshadfield:
I've fixed this up so that (from the previous comment) case 1 will add a `treeZoom=selected` URL query but case 2 will not. If anyone would like to test this out that would be much appreciated!

jameshadfield:
> Looking at this 3 years later, I wonder if the label query should be deprecated for the new treeZoom query so that there's only a single URL query that controls zoom.

Yeah I'd support his change, `treeZoom` is certainly more descriptive. I'll leave it for another PR however.

Thanks for the reviews over the years!

PR #1328, opened 2021-03-30
Apply clipping to first column of legend

### LINKED ISSUE #1306: Legend items are limited to 5 genotypes

jameshadfield:
We have had issues in the past with legend values from column 1 overflowing into column 2. For instance, issue #899 was fixed by PR #914 which implemented a maximum character limit for legend names. That solution could produce misleading views, such as those described in #1306.

The solution from this PR is to implement a clipping mask for column 1, avoiding the complication of limiting the string size. Column 2 already has similar behaviour because the SVG element of the legend itself performs the clipping.

Closes #1306


jameshadfield:
Tested using various datasets & colorings (including genotype) on nextstrain.org

PR #1344, opened 2021-05-10
Use filterOptions to modify filter search algorithm

### LINKED ISSUE #1271: Improve search algorithm in sidebar filtering UI
Currently (and unchanged in #1265) one must type an exact (case-insensitive) substring of the displayed filter options in order to have them show up as selectable. For example, given a filtering option `mutation HA1 136E` (buried in a large list), typing any the following examples do not identify the option: `mut 136E`, `mut HA1 136`, `136 HA1`.

This is also evident for options such as `Country → USA` where typing in `country USA` does not identify the option for filtering!

**Notes**   

- As part of this, we may want to upgrade the version of react-select we are using (we're pinned to 1.0.0-rc5, the latest is 3.2.0). 
- A solution here must take into account performance, as we frequently have lists of options of 10^4 or higher.
- During #1265 I investigated using https://github.com/bvaughn/react-select-fast-filter-options, which would easily expose the string matching algorithm, however this resulted in significantly decreased performance and thus wasn't persued.

didithilmy:
This PR addresses the issue outlined on https://github.com/nextstrain/auspice/issues/1271 without the need of upgrading `react-select` to a newer version.

### Related issue(s)  

Fixes #1271 

### Testing
I tested the changes manually by running `auspice develop` and trying out the filter using `ncov/global` data. There was no significant performance issues compared to the previous search algorithm.

image

### DIFF
--- a/src/components/controls/filter.js
+++ b/src/components/controls/filter.js
@@ -107,6 +107,7 @@ class FilterData extends React.Component {
     // to `loadOptions` we don't slow things down by comparing queries to a large number of options
     const options = this.makeOptions();
     const loadOptions = debounce((input, callback) => callback(null, {options}), DEBOUNCE_TIME);
+    const filterOption = (option, filter) => filter.toLowerCase().split(" ").every((word) => option.label.toLowerCase().includes(word));
     const styles = this.getStyles();
     const inUseFilters = this.summariseFilters();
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @didithilmy - tested this out locally and it works well without any (noticeable) performance penalty. Will include in the next auspice release!

Issue #1339, opened 2021-04-27
Add support for categorical colorbys to scatterplot


trvrb:
The scatterplot functionality has proved highly enabling to be able to view multiple aspects of data in a single view. One chief example at this moment is to compare attributes of emerging lineages, where a couple common views look like:

Color by emerging lineage, time on x axis, spike S1 mutations on y axis
https://nextstrain.org/ncov/global?branches=hide&c=emerging_lineage&l=scatter&scatterY=S1_mutations
<img width="1276" alt="s1" src="https://user-images.githubusercontent.com/1176109/116166265-7e66fc00-a6b2-11eb-93c5-27955cbe6aa2.png">

Color by emerging lineage, time on x axis, logistic growth on y axis
https://nextstrain.org/ncov/global?c=emerging_lineage&l=scatter&scatterY=logistic_growth
<img width="1276" alt="Screen Shot 2021-04-26 at 5 12 51 PM" src="https://user-images.githubusercontent.com/1176109/116166357-a7878c80-a6b2-11eb-9361-42a6a812a6a5.png">

Both these views help to surface which emerging lineages have highest S1 mutations (P.1) and which emerging lineages have highest rates of logistic growth (largely B.1.1.7), but other lineages get lost in the mix due to occlusion of tips.

For the desired comparison of S1 mutations and logistic growth across emerging lineages it would be preferable to have emerging lineage on the x axis and S1 mutations or logistic growth on the y axis.

In this case, rather than a regression line, I'd imagine a horizontal black line for each categorical variable demarcating its mean.

huddlej:
One other use case that occurred to me recently would be a kind of transmission network view where we could plot time on the x-axis, regions on the y-axis, color by clades, and look for clades whose branches traverse region boundaries (sort of like [a slope plot](https://vega.github.io/vega-lite/examples/line_slope.html)). For some subset of geographic locations, we could get a similar effect from the current functionality by plotting latitude or longitude on the y-axis.

jameshadfield:
This feature was released in v2.26.0

PR #1346, opened 2021-05-18
Allow non-continuous scatterplot variables

jameshadfield:
This implements a requested improvement to the original
scatterplot implementation. The implementation hinges on two changes:
(1) The collection of values for a given variable (e.g. x-var) need
to be computed and passed to PhyloTree to act as the scale's domain.
We reuse the colorScale machinery here, which could be optimised
(see todo messages in code), but this has the advantage that the
domain ordering matches the legend (unless user supplied).
(2) PhyloTree needed to be modified to use non-linear scales, in this
case `pointScale`.

This commit should be fully functional, however there are some
future improvements to be made:

(i) Grid text is obscured and unreadable when there are many entries
in the domain.
(ii) Genotypes and Boolean scales are not yet available.
(iii) Jitter should be added to nodes to avoid obfuscation.

@trvrb @huddlej would you mind trying this out? Thanks in advance.

image
image

### DIFF
--- a/src/actions/layout.js
+++ b/src/actions/layout.js
@@ -1,0 +1,37 @@
+import { CHANGE_LAYOUT } from "./types";
+import { validateScatterVariables, addScatterAxisInfo} from "../util/scatterplotHelpers";
+
+/**
+ * Redux Thunk to change a layout, including aspects of the scatterplot / clock layouts.
+ */
+export const changeLayout = ({layout, showBranches, showRegression, x, xLabel, y, yLabel}) => {
+  return (dispatch, getState) => {
+    if (window.NEXTSTRAIN && window.NEXTSTRAIN.animationTickReference) return;
+    const { controls, tree, metadata } = getState();
+
+    if (layout==="rect" || layout==="unrooted" || layout==="radial") {
+      dispatch({type: CHANGE_LAYOUT, layout, scatterVariables: controls.scatterVariables, canRenderBranchLabels: true});
+      return;
+    }
+
+    let scatterVariables = (layout==="clock" || layout==="scatter") ?
+      validateScatterVariables(controls, metadata, tree, layout==="clock") : // occurs when switching to this layout
+      controls.scatterVariables;
+
+    if (x && xLabel) scatterVariables = {...scatterVariables, ...addScatterAxisInfo({x, xLabel}, "x", controls, tree, metadata)};
+    if (y && yLabel) scatterVariables = {...scatterVariables, ...addScatterAxisInfo({y, yLabel}, "y", controls, tree, metadata)};
+    if (showBranches!==undefined) scatterVariables.showBranches = showBranches;
+    if (showRegression!==undefined) scatterVariables.showRegression = showRegression;
+    if (layout==="scatter" && (!scatterVariables.xContinuous || !scatterVariables.yContinuous)) {
+      scatterVariables.showRegression= false;
+    }
+
+    dispatch({
+      type: CHANGE_LAYOUT,
+      layout: layout || controls.layout,
+      scatterVariables: {...scatterVariables}, // ensures redux is aware of change
+      canRenderBranchLabels: scatterVariables.showBranches
+    });
+
+  };
+};
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -568,7 +568,7 @@ const checkAndCorrectErrorsInState = (state, metadata, query, tree, viewingNarra
   // todo: these should be JSON definable (via display_defaults)
   if (state.layout==="scatter" || state.layout==="clock") {
     state.scatterVariables = validateScatterVariables(
-      state.scatterVariables, metadata.colorings, state.distanceMeasure, state.colorBy, state.layout==="clock"
+      state, metadata, tree, state.layout==="clock"
     );
     if (query.scatterX && query.scatterX!==state.scatterVariables.x) delete query.scatterX;
     if (query.scatterY && query.scatterY!==state.scatterVariables.y) delete query.scatterY;
--- a/src/components/controls/choose-layout.js
+++ b/src/components/controls/choose-layout.js
@@ -27,6 +27,7 @@ export const RowContainer = styled.div`
     layout: state.controls.layout,
     scatterVariables: state.controls.scatterVariables,
     colorings: state.metadata.colorings,
+    colorBy: state.controls.colorBy,
     showTreeToo: state.controls.showTreeToo,
     branchLengthsToDisplay: state.controls.branchLengthsToDisplay
   };
@@ -37,7 +38,7 @@ class ChooseLayout extends React.Component {
     dispatch: PropTypes.func.isRequired
   }
   renderScatterplotAxesSelector() {
-    const options = collectAvailableScatterVariables(this.props.colorings);
+    const options = collectAvailableScatterVariables(this.props.colorings, this.props.colorBy);
     const selectedX = options.filter((o) => o.value===this.props.scatterVariables.x)[0];
     const selectedY = options.filter((o) => o.value===this.props.scatterVariables.y)[0];
     const miscSelectProps = {options, clearable: false, searchable: false, multi: false, valueKey: "label"};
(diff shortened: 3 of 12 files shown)


trvrb:
This is fantastic. A couple notes as I review:

1. I really like having default y be active color by. 

2. https://auspice-ordinal-scatter-39fzn6.herokuapp.com/ncov/global?branches=hide&l=scatter&scatterX=region&scatterY=emerging_lineage should really be a useful visualization of where in the world emerging lineages have been sampled. However, the lack of jitter makes this not as useful as it could be.

Medium priority would be adding jitter. I don't think blocking PR (which is accurate and useful as it stands), but I do think jitter would be a nice addition. Though I see why automating jitter will be a bit complex.

huddlej (approved):
This is so cool, @jameshadfield. I would be happy to use this now as it is, but I've provided more detailed comments below.

**I love that** I can compare categorical fields with each other and with quantitative fields! Just this functionality alone allows us to investigate some interesting questions for influenza that were previously more difficult with the tree view.

image
[Which clades have the most epitope mutations and what is their distribution?](https://auspice-ordinal-scatter-39fzn6.herokuapp.com/flu/seasonal/h3n2/ha/2y?branches=hide&l=scatter&scatterX=clade_membership&scatterY=ep) This same type of view could be helpful for inspecting which flu clades are underrepresented for total titer measurements.

image
[When were clade strains most recently submitted?](https://auspice-ordinal-scatter-39fzn6.herokuapp.com/flu/seasonal/h3n2/ha/2y?branches=hide&l=scatter&scatterX=clade_membership&scatterY=recency) The information is present in [the corresponding tree colored by submission date with branch labels](https://auspice-ordinal-scatter-39fzn6.herokuapp.com/flu/seasonal/h3n2/ha/2y?c=recency), but this view is much more succinct and unambiguous.

image
[Which clades are present in each region?](https://auspice-ordinal-scatter-39fzn6.herokuapp.com/flu/seasonal/h3n2/ha/2y?branches=hide&l=scatter&scatterX=clade_membership&scatterY=region) This is the kind of view where the ability to tune the opacity of points would produce a density plot where more overlapping points communicates a higher density. You could also imagine eventually supporting a radius encoding for that overlap to produce a punchcard plot.

image
[Which regions have strains that are most likely to seed the future and what clades are those strains from?](https://auspice-ordinal-scatter-39fzn6.herokuapp.com/flu/seasonal/h3n2/ha/2y?branches=hide&l=scatter&scatterX=region&scatterY=weighted_distance_to_future_by_ne_star-lbi)

image
[What is the LBI of the closest strains to the future for two specific regions with the highest LBI strains?](https://auspice-ordinal-scatter-39fzn6.herokuapp.com/flu/seasonal/h3n2/ha/2y?branches=hide&c=lbi&f_region=South%20Asia,Africa&l=scatter&p=grid&scatterX=region&scatterY=weighted_distance_to_future_by_ne_star-lbi) Enables a side-by-side comparison of forecast distributions between regions with the evidence from LBI to support those forecasts.

**I wish that I could**:
* set an alpha/opacity value for each point to produce a quick density plot especially for categorical vs categorical plots
* order categorical columns by any quantitative tip annotation. In addition to the UI and code complexity associated with this request, there’s also the issue of how to aggregate the quantitative values across each single categorical value. Would we order by the mean/median/max of the quantitative values?
* filter tips to specific values in a category and have all other other values in the category disappear from the axis (see the last image above). This could be a flu-specific issue related to the vaccine strain “x” symbols though.

**What if:**
* categorical vs. quantitative plots automatically sorted the categorical values by the mean/median of the quantitative field? This wouldn’t require any additional UI components unless you wanted to allow users to toggle between the default (maybe user-specific) and computed ordering.
* we allowed users to define the order for any category based on another tip annotation that had a name based on the category field's name? For example, users could annotate a quantitative field named `region-orderby` based on latitude, for example, and that field would be automatically used to order the `region` values if it was present. This wouldn't be as nice as a full UI to select any arbitrary field but it could be a useful half-step. ([This transmission network mockup](https://github.com/nextstrain/auspice/issues/1339#issuecomment-827925641) is what I had in mind for this specific request.)

jameshadfield:
Thanks for the detailed review @huddlej! I've added jitter & a few other improvements to the PR, so you can re-click your links above and see a newer rendering of them. 

Ordering is going to be complicated, so I'm going to avoid any changes in this PR!!!

> filter tips to specific values in a category and have all other other values in the category disappear from the axis (see the last image above).

This isn't flu-specific, I'll take a look and see how easy a fix is going to be. (Underneath it's the same logic as we use for displaying dynamically filtered legend items, so I'm hopeful the fix will be easy.)

P.S. There's a bug _somewhere_ which occurs in certain combinations of scatterplot variables and causes no nodes to be rendered, so this PR should at least wait until I've found that.

huddlej:
Nice! The jitter makes such a huge difference and basically gets at what I was thinking with reduced opacity/density plot view.

Is there a hard threshold on the viewport width for when jittering is activated? When I go to [this view of distance to future by region](https://auspice-ordinal-scatter-39fzn6.herokuapp.com/flu/seasonal/h3n2/ha/2y?branches=hide&c=lbi&f_region=South%20Asia,Africa&l=scatter&p=full&scatterX=region&scatterY=weighted_distance_to_future_by_ne_star-lbi) the points per region are jittered, but when I turn on [the grid display of the same scatterplot](https://auspice-ordinal-scatter-39fzn6.herokuapp.com/flu/seasonal/h3n2/ha/2y?branches=hide&c=lbi&f_region=South%20Asia,Africa&l=scatter&p=grid&scatterX=region&scatterY=weighted_distance_to_future_by_ne_star-lbi) the jitter disappears. I can imagine that it is hard to cram everything into such a small view, though. I can't really think of a better solution, if there is a hard threshold for width. 🤷🏻

trvrb:
Awesome! I think the jitter is a unexpectedly large improvement. I'm really happy with how things stand. And John, your flu are really interesting and things that I wouldn't have immediately thought to surface.

One small non-blocking issue. I saw from the commits that genotype was added as scatterplot option. But then I couldn't immediately figure out how to activate it. I eventually thought to select genotype as color by and toggle away from scatterplot layout and then toggle back to scatterplot layout. I think discovered the genotype option for x and y axes. I suspect this feature is basically hidden while it works like this.

emmahodcroft:
I'm really enjoying playing around with this @jameshadfield - and @huddlej your examples above helped to get me going: I hadn't realised all the options!

Anyway, unfortunately I'm also posting as I managed to find a bit of an odd bug. I originally had plotted by `X`: Genotype S: 501 (also as a colourby, of course), and `Y`: Clade.

I then was curious if it would auto-adjust to changing the colourby, so typed in `681`. It does auto adjust! However, it does not auto-recolour:

image
(All the columns should be the same colour)

Interestingly the map did recolour so it's just something about the scatter.
This isn't a big deal at all - and changing to another colourby and changing back resolves it - but well, figured I'd write it up! 🙃

jameshadfield:
Thanks all for the comments -- especially the two bugs and instructions on how to reproduce. 

> Is there a hard threshold on the viewport width for when jittering is activated? 

There's a threshold of 50px spacing between values on an axis as below this it got confusing which point belonged to which point, but 50px is somewhat arbitrary! In Emma's screenshot you can see it's jittered on the x-axis but not the y-axis.

> I saw from the commits that genotype was added as scatterplot option. 

Yeah, it's somewhat secret squirrel. The only way we have of selecting a specific genotype is via the colouring (which uses the cascading dropdown UI); so if the colouring is not genotype then I don't think we can select this as a scatter variable without exposing the more complex UI, which I wanted to avoid. P.S. If the colouring is genotype, then genotype shows up as an option in the scatterplot variables.

jameshadfield:
Bugs all fixed & (I believe) this is now good to go 😄 

@trvrb I changed the behaviour to improve visibility and consistency -  if you have selected a genotype colouring and then click scatterplot, the y-axis will be genotype. 

@huddlej I'm going to leave the dynamic domains (i.e. filtering updating the axes) for a future PR in order to get this out ASAP

trvrb:
Fantastic! Thank you James. This is good to go from my perspective. Really excited about this as it opens up a bunch of expressive possibilities.

PR #1347, opened 2021-05-19
GitHub Action to create nextstrain.org PR

jameshadfield:
This action will run on each auspice PR and create a corresponding
PR on nextstrain.org which includes a commit using the version
of auspice from this (auspice) PR. This functionality is extremely
useful for auspice development as it will allow us to use a Heroku
review app to test auspice in the context of nextstrain.org

There are a number of future improvements to implement:

* New auspice releases (tagged commits on `release` branch)
would ideally create a PR on nextstrain.org which could be merged
to update the version of auspice there. However the action will have
to wait until the npm registry has been updated, which from memory
can take up to ~30min.

* Other consumers of auspice (e.g. auspice.us) could be added to this
GitHub Action.

P.S. This action created https://github.com/nextstrain/nextstrain.org/pull/334

### DIFF
--- a/.github/workflows/make_prs_for_other_repos.yaml
+++ b/.github/workflows/make_prs_for_other_repos.yaml
@@ -1,0 +1,44 @@
+name: "Make PRs for Nextstrain projects which depend on Auspice"
+on:
+  pull_request:
+jobs:
+  make-pr-on-nextstrain-dot-org: # <job_id>
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: Checkout nextstrain.org repo
+        uses: actions/checkout@v2
+        with:
+          repository: nextstrain/nextstrain.org
+      - name: Install Auspice from PRs HEAD commit
+        if: ${{ github.event_name == 'pull_request' }}
+        # Note: $GITHUB_SHA is _not_ the same commit as the HEAD commit on the PR branch
+        # see https://github.community/t/github-sha-not-the-same-as-the-triggering-commit/18286/2
+        shell: bash
+        run: |
+          AUSPICE_COMMIT=$(cat $GITHUB_EVENT_PATH | jq -r .pull_request.head.sha)
+          echo "auspice_commit=$AUSPICE_COMMIT" >> $GITHUB_ENV
+          npm ci
+          npm install nextstrain/auspice#${AUSPICE_COMMIT}
+          git add package.json package-lock.json
+      - name: Create Pull Request for testing on nextstrain.org repo
+        if: ${{ github.event_name == 'pull_request' }}
+        id: cpr
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.JAMES_PAT }}
+          branch: "auspice-pr-${{ github.event.pull_request.number }}"
+          commit-message: "[testing only] upgrade auspice to ${{ env.auspice_commit }}"
+          title: 'Test auspice PR ${{ github.event.pull_request.number }}'
+          body: |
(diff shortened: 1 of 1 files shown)


jameshadfield:
@eharkins could you take a look at this? I believe it's all working as expected. My understanding is that after merge into master, PRs created from subsequently created branches will automatically trigger a PR on nextstrain.org. This won't happen if they are from forks.

eharkins (approved):
Nice work! This is cool. A couple of thoughts:
- since the auspice version will always be `nextstrain/auspice#${AUSPICE_COMMIT}`, you will never want to actually merge these auto-created PRs right? If you were hoping to merge them in some cases like when you have a new auspice release tag, is there a way you can capture the auspice commit HEAD that would reflect a tag in a format you would actually want to have in nextstrain.org's production package.json ?
- once you merge you could change [this message](https://github.com/nextstrain/auspice/pull/1347/commits/142754cca9151193a37484ebc70ec421f7989a21#diff-1b3e44c4d5f20ce7ceb31d18d89c96349246fde02e6b2d87fa1a8f2b96a30c52R38) to link to the actual workflow file, `.github/workflows/make_prs_for_other_repos.yaml`
- You're not currently using [`pull_request_target`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target), but in case you were considering using it in the future here, definitely be advised about [a known security risk](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) associated with it since you are using your PAT as a secret here.
- Lastly:
> My understanding is that after merge into master, PRs created from subsequently created branches will automatically trigger a PR on nextstrain.org. This won't happen if they are from forks.

Reading https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories:
> When you create a pull request from a forked repository to the base repository, GitHub sends the pull_request event to the base repository and no pull request events occur on the forked repository.

> Workflows don't run on forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository.

> When a first-time contributor submits a pull request to a public repository, a maintainer with write access must approve running workflows on the pull request. For more information, see "Approving workflow runs from public forks."

 it seems like it could be triggered if a pull request comes in from a fork if the PR is by someone who has contributed before?

jameshadfield:
Thanks @eharkins!

> you will never want to actually merge these auto-created PRs right? 

Correct (and one of the reasons they're marked as "draft")

> If you were hoping to merge them in some cases like when you have a new auspice release tag, is there a way you can capture the auspice commit HEAD that would reflect a tag in a format you would actually want to have in nextstrain.org's production package.json ?

Yes - I'd like to do this in the future. I'm not sure how exactly to achieve it, as to create such a PR we need the version to have been released on npm, which happens some time _after_ the tag is created on GitHub. Perhaps there's some API to trigger GitHub actions from npm when each version is released, but I haven't looked into this. 

> once you merge you could change this message to link to the actual workflow file

Good idea!

> You're not currently using pull_request_target, but in case you were considering using it in the future here, definitely be advised about a known security risk associated with it since you are using your PAT as a secret here.

Ouch! That's a big gotcha. I do get confused about PR branches vs ephemeral merges of PRs into the master branch (e.g. how this workflow can't use `${GITHUB_SHA}`). Thanks for the link!

> it seems like it could be triggered if a pull request comes in from a fork if the PR is by someone who has contributed before?

Yes -- that seems to be true: ["After a contributor has at least one pull request merged into a project's repository, any future pull requests from that contributor's fork will automatically run workflows."](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) but then the docs state that ["With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository"](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) which implies that the action will run but won't be able to create the PR on nextstrain.org.

PR #1352, opened 2021-05-29
Always show regression toggle for clock layout

jameshadfield:
Fixes a bug where the ability to toggle regression lines was hidden for clock views. (The ability to hide this toggle is only intended for scatter layouts, where we should not expose the toggle unless both axes are showing continuous variables.)


trvrb:
Looks great. Thanks for the fix James!

trvrb (approved):

PR #1353, opened 2021-06-04
Adjust grayscale color ramp

trvrb:
The existing grayscale color ramp (used for values absent in an explicitly specified color scale) had values that were too dark and threw off the overall color balance. This commit narrows the grayscale color ramp to be more in line with pastel color ramp.

I mainly did this because I didn't like the aesthetics of the new clades coloring. Here it is on current Auspice `master`:

master

And here is this PR:

PR

### DIFF
--- a/src/components/frequencies/functions.js
+++ b/src/components/frequencies/functions.js
@@ -203,7 +203,7 @@ export const removeStream = (svg) => {
 };
 
 const generateColorScaleD3 = (categories, colorScale) => (d, i) =>
-  categories[i] === unassigned_label ? "rgb(190, 190, 190)" : rgb(colorScale.scale(categories[i])).toString();
+  categories[i] === unassigned_label ? "#ADB1B3" : rgb(colorScale.scale(categories[i])).toString();
 
 function handleMouseOver() {
   select(this).attr("opacity", 1);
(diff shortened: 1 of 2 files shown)


jameshadfield (approved):
Happy for these changes to go through ✅ 

In the future we may want to better distinguish between "extra values" (i.e. defined traits which weren't assigned a colour by the dataset) and "unknown values":

auspice_greys

P.S. It's not this simple, as auspice doesn't render hex values truthfully for the tree. We [make branch colours closer to grey](https://github.com/nextstrain/auspice/blob/a0e9080dcd61249f24f6c822dfcf6478e4e07ac2/src/util/colorHelpers.js#L93-L111), [brighten terminal nodes](https://github.com/nextstrain/auspice/blob/a0e9080dcd61249f24f6c822dfcf6478e4e07ac2/src/components/tree/reactD3Interface/initialRender.js#L48) and brighten legend swatches by [a different (!) amount](https://github.com/nextstrain/auspice/blob/797090f8092ffe1291b58efd113d2c5def8b092a/src/components/tree/legend/legend.js#L169). I believe colours in the frequencies panel & the map are rendered truthfully, but I may be wrong. Here's a side-by-side of current auspice (where we lighten things) and a version where we don't:

image

trvrb:
Distinguishing between Auspice assigned colors for traits that don't have an explicit listing and traits that are "unassigned" is a good argument to keep `unknownColor` at` #AAAAAA`.

We could have a situation that involves both kinds of grays and I can see it being useful to distinguish.

Issue #1354, opened 2021-06-05
edge case with 37 pango_lineages


AngieHinrichs:
**Current Behavior**  
Most leaves of this tree have a pango_lineage, but currently it's displaying with no colors:

https://nextstrain.org/fetch/hgwdev.gi.ucsc.edu/~angie/thirtysevenpangos.json?c=pango_lineage&label=nuc%20mutations:G11222T

The JS console says
```
auspice.chunk.4.bundle.c364ac792223cbbc9d29.js:formatted:7449 Error creating color scales. Details:
 Cannot read property 'slice' of undefined
```
The minified code where the exception arises is
```
        "ordinal" !== n && "categorical" !== n ||
         (t = e.length <= s.i.length ? s.i[e.length].slice() : s.i[s.i.length - 1].slice());
```
-- specifically both e.length and s.i.length are 37, so s.i[e.length] is just off the end of the array and is undefined, so no slice.

I believe that corresponds to this in [colorScale.js](https://github.com/nextstrain/auspice/blob/d1801b1667a8b20c27c7616bcb45c3be7f9734c5/src/util/colorScale.js#L337) [I know I can run a dev version of auspice to avoid minimized code but it was quicker to just search for "ordinal" and "categorical"] :
```
function createDiscreteScale(domain, type) {
  // note: colors[n] has n colors
  let colorList;
  if (type==="ordinal" || type==="categorical") {
    /* TODO: use different colours! */
    colorList = domain.length <= colors.length ?
      colors[domain.length].slice() :
      colors[colors.length - 1].slice();
  }
```

**Expected behavior**  
A Pango rainbow!

**Possible solution**  
I think domain.length needs to be strictly less than colors.length.

Thanks!

jameshadfield:
I think you've solved it! 37 is our unlucky number it seems 🤦 

I think this change will fix it, but i need to do a bit more testing
```diff
diff --git a/src/util/colorScale.js b/src/util/colorScale.js
index 697e1e37..770ccc32 100644
--- a/src/util/colorScale.js
+++ b/src/util/colorScale.js
@@ -339,7 +339,7 @@ function createDiscreteScale(domain, type) {
   let colorList;
   if (type==="ordinal" || type==="categorical") {
     /* TODO: use different colours! */
-    colorList = domain.length <= colors.length ?
+    colorList = domain.length < colors.length ?
       colors[domain.length].slice() :
       colors[colors.length - 1].slice();
   }
```

[...]

PR #1355, opened 2021-06-08
[bugfix] generate colour scales for 37 values

jameshadfield:
Our colour-scale generation logic contained a bug where we couldn't
have domains of length 37 (off-by-one error).

Closes https://github.com/nextstrain/auspice/issues/1354


trvrb:
@jameshadfield --- Looks like this is safe to merge?

PR #1364, opened 2021-06-30
Styling adjustments to footer text

trvrb:

This is a simple update to the CSS styling of the page footer specified in `footer.js`. This is motivated by display of long Markdown description in `open` builds. Here is resulting display prior to this update:

before

And here is the resulting display after this update:

after

I don't think the main paragraph text was ever meant to be slightly bold. It has `font-weight: 300` specified under `FooterStyles`. However, because it wasn't specified under `p` the main paragraph text was getting set as `font-weight: 400`.

I borrowed the font sizing in the section headers from bedford.io, where I preferred to give more of a dynamic range compared to default CSS. In this case, I thought the `#####` to be much too large without providing the styling.

### Testing

Tested locally.

### DIFF
--- a/src/components/framework/footer.js
+++ b/src/components/framework/footer.js
@@ -20,39 +20,43 @@ const FooterStyles = styled.div`
   color: rgb(136, 136, 136);
   line-height: 1.4;
 
+  p {
+    font-weight: 300;
+  }
+
   h1 {
     font-weight: 700;
-    font-size: 2.2em;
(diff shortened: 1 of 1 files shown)


trvrb:
Nextstrain.org deployment looks like it worked as expected: https://nextstrain-s-auspice-pr-6i41bz.herokuapp.com/ncov/global?dmin=2021-05-17&r=division

PR #1371, opened 2021-07-09
Allow hiding of all filters in a category

### LINKED ISSUE #1358: Allow 'turn off' of all filters, not just individual ones
**Context**  
When trying to inspect a set of filtered tips, one cannot zoom in so easily if only one filtered tip is visible. The best thing to do is turn off all filters, then click on the branch you want, then turn filters back on. However, for multiple tips, this takes ages to turn them on and off when switching views. 

turnOffFilterTips

**Possible Solution**   
Perhaps alongside the 'delete all' icon below the 'Filter Data' on the menu, there could be a 'hide all' option? So that one could hide and unhide all the sample filters at once, to allow for more easy zooming in and out?

image

**Examples**  
With any set of multiple tips on filter, try to zoom in and out to look at them each more closely in turn, without accidentally deleting any along the way :) 
https://nextstrain.org/ncov/europe?c=S1_mutations&s=France/PAC-IHU-3126i/2020,India/MH-NIV-8404/2020,Mongolia/7908/2020,Peru/LIM-INS-768/2020

jameshadfield:
Expands the functionality of the sidebar filter badges to allow inactivation of all (active) filters in a category.

The choice was made to keep a sidebar badge "active" if any of its constituent filters was active. For example, if there were a mixture of 3 active and 2 inactive country filters (displayed at the top of the page), then the sidebar badge will be active and show "3 x country", with the option to inactivate those 3 filters, or remove all 5.

Closes #1358

### DIFF
--- a/src/components/controls/filter.js
+++ b/src/components/controls/filter.js
@@ -98,7 +98,10 @@ class FilterData extends React.Component {
       return {
         filterName,
         displayName: filterBadgeDisplayName(n, this.getFilterTitle(filterName)),
-        remove: () => {this.props.dispatch(applyFilter("set", filterName, []));}
+        anyFiltersActive: () => this.props.activeFilters[filterName].filter((f) => f.active).length>0,
+        remove: () => {this.props.dispatch(applyFilter("set", filterName, []));},
+        activate: () => {this.props.dispatch(applyFilter("add", filterName, this.props.activeFilters[filterName].map((f) => f.value)));},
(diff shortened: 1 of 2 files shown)


emmahodcroft:
This is super-awesome James!! I think this solves the problem perfectly, and it will amazing to be able to turn all on/off with one click to aid zooming around on the tree. I also think it pairs perfectly with #1372!

Issue #1357, opened 2021-06-10
Clicking on filtered tips, then away, means they're removed from filter


emmahodcroft:
**Current Behavior**  
If a tip is selected under a filter, then you click on that tip to see more information, when you click away to close the box, the filter is removed.
![tipClickDisappear](https://user-images.githubusercontent.com/14290674/121502935-e498ab80-c9e0-11eb-9c26-edd3ec0496aa.gif)
This is particularly annoying if you are trying to explore a set of sequences, but have to keep re-adding them, or never click on the tips (particularly if you can't see the information you want because the mouse-over box is cut-off 🙃 #1356 ).
I run into this a lot when trying to evaluate the 'suspect' sequences proposed by Moira & Eli for potential exclude, but other times as well.

Happens with any filtered set of tips but here's a link for the gif shown (may have changed since then as these are live builds):
https://nextstrain.org/ncov/europe?c=S1_mutations&l=clock&s=India/MH-NIV-8404/2020,Mongolia/7908/2020,Peru/LIM-INS-768/2020,France/PAC-IHU-3126i/2020

**Expected behavior**  / **Possible Solution**
This is a tricky one as I know that clicking a random tip also applies a 'filter' of a sort, and we want that to disappear once box is clicked away from. 
Perhaps a variable could be set as to whether a tip was in a filter _prior_ to the box being clicked, so it could remain filtered once box is clicked away from?

**Your environment: if browsing Nextstrain online**  
 - Operating system:  Windows 10 Pro
 - Browser:  Chrome

trvrb:
I see what you mean Emma. My guess here is that activating the `infoPanel.click` component shouldn't also filter the dataset. Just clicking on a tip should bring up modal panel but shouldn't filter.

PR #1372, opened 2021-07-09
Inactivate tips when no longer focused

jameshadfield:
Selecting a tip (and the corresponding info panel modal) is intrinsically linked to setting a single active "sample" filter. When clicking away from this modal the filter was removed.

This caused issues in the situation where "a tip is selected under a filter, then you click on that tip to see more information, when you click away to close the box, the filter is removed."

The solution here is to inactivate the sample filters when removing the info panel. This has the benefit (bug?) of keeping a history of clicked tips in the form of inactive filters. (Note that all strain filters can be removed by a single click in the sidebar filter badge.)

Closes https://github.com/nextstrain/auspice/issues/1357

@emmahodcroft could you test this out? There are alternative (harder) solutions where we keep track of whether a clicked tip was previously an inactive filter or not if we want to pursue that. However this PR, in conjunction with #1371, could potentially make @eharkins & Moira's job easier via the following workflow:

1. Click on every outlier tip. Each click will open a tip-info panel, so click outside of that to remove it each time.
2. Notice that all the clicked tips are now inactive filters
3. Click on the "eye" icon in the filter badge in the sidebar for "samples", which will make all these inactive sample filters active.

### DIFF
--- a/src/components/tree/reactD3Interface/callbacks.js
+++ b/src/components/tree/reactD3Interface/callbacks.js
@@ -125,5 +125,5 @@ export const clearSelectedTip = function clearSelectedTip(d) {
     .attr("r", (dd) => dd["r"]);
   this.setState({selectedTip: null, hovered: null});
   /* restore the tip visibility! */
-  this.props.dispatch(applyFilter("remove", strainSymbol, [d.n.name]));
+  this.props.dispatch(applyFilter("inactivate", strainSymbol, [d.n.name]));
 };


emmahodcroft:
I've just played around with this and I think it's a nice solution. It's easy to clear them out if you aren't interested in them, or to just remove one if it turns out you are 'making a list' but you don't want to include that one. Plus, in combo with #1371  it'll be super easy to then flick them on and off to zoom & investigate further. 

However, happy to wait and see what Eli & Moira think since they're on the 'opposite' end of the checking than I usually am!

PR #1373, opened 2021-07-11
Zoom to visible nodes

### LINKED ISSUE #1368: Improve "Zoom to Selected" bounding box
**Context**  
The implementation of a "Zoom to Selected" button on the tree has been really useful. I'll often filter to say clade 21A and then be able to just click to zoom in properly.

However, there are a number of edge cases where "Zoom to Selected" isn't very helpful. For a simple example, if we filter to early samples (https://nextstrain.org/ncov/global?dmax=2020-02-01) there's no way to zoom into these (the button is deactivated). Or if we filter to clade 19A viruses (https://nextstrain.org/ncov/global?f_clade_membership=19A) there's also no way to zoom in.

**Description**   
We could at some point contemplate a fancier "accordion" view which would help to emphasize situations where multiple blocks of tips are selected, but before this we should get basic bounding box working (the accordion view would require updating node y-values and will necessarily be more involved).

Instead, I'd recommend when clicking "Zoom to Selected" to calculate the maximum x and y-values and the minimum x and y-value out of all visible nodes. Use these values to define the bounding box to zoom to.

cc @joverlee521 who implemented the initial version of zoom to selected in #1257. This PR contains discussion of the implementation challenge of this bounding box design.

trvrb:

This commit re-implements the "zoom to selected" function in the tree panel to emphasize visible nodes by expanding their "yValues" to take up 80% of the vertical span of the panel. Notes on implementation details:

- I mirrored redux dataflow of `distanceMeasure` and `layout` to create a new redux variable of `treeZoom`. This is defaults to `even` but is updated to `zoom` when clicking the "zoom to selected" tab. Further clicks increment the redux variable to `zoom-2`, `zoom-3`, etc... and clicking "reset layout" restores it to `even`.
- A PhyloTree redraw is triggered when redux treeZoom variable is updated. This allows filters to change, etc... without triggering immediate changes to layout, but then clicking "zoom to selected" will redraw layout to emphasize currently selected nodes.
- phylotree.layouts contains the actual logic in the `calcYValues` function. This dynamically sets `node.n.yValue` based on `node.visibility`, so that calls to other layout functions like `rectangularLayout` will have updated `node.n.yValue` from which to construct `node.y`.

This is mostly working at this point, though there are a few issues (like initial render with `?z=zoom` in URL does not work as it should due to `node.visibility` not existing in the initial calls to layout). I'm hoping first for user interface feedback (hence the draft PR).

Primary use case is to:
1. Filter to a trait value of interest (whether monophyletic or not)
2. Click "zoom to selected" to emphasize nodes with trait.

The biggest wins are being able to zoom to basal tips in the tree. For example, you can zoom to early SARS-CoV-2 samples or to clade 19A where this wasn't previously possible.

### Related issue(s)  
Fixes #1368


huddlej:
I haven't had a chance to do a full review yet, but at first glance, this is a really cool feature! It seems pretty helpful for situations like H3N2 where [we have recurrent mutations like 186D](https://auspice-dynamic-yvalues-mqlvu0.herokuapp.com/flu/seasonal/h3n2/ha/2y?gt=HA1.186D) and we want to focus on these.

emmahodcroft:
I haven't tested this out super-extensively yet, but I tried out zooming to 19B (currently can be quite difficult!) and also picking out a few sequences at the 'edges' of clades (which can also be extremely difficult to get a good look at, if they don't cluster but branch sequentially away from the clade). Both worked beautifully and gave _super useful_ views that I don't think you can get with current auspice!

joverlee521:
It's really cool that the zoom works well with the [date filter](https://auspice-dynamic-yvalues-mqlvu0.herokuapp.com/flu/seasonal/h3n2/ha/2y?dmax=2016-08-24&dmin=2015-10-14)! 🤩 

I noticed that the interaction between zooming via branches and the zoom button can be a little clunky. 
1. Zoom in via a branch:
Screen Shot 2021-07-13 at 5 13 32 PM

2. Then clicking on the zoom button brings up previously hidden branches on the bottom, making the x-axis hard to read:
Screen Shot 2021-07-13 at 5 13 55 PM

3. Zooming back out via clicking on base branches leaves the tree in a strange configuration:
Screen Shot 2021-07-13 at 5 15 12 PM

jameshadfield:
The general behaviour is fantastic. I’m really happy to see the y-value-zoom functionality resurrected in this (much improved) form! Here’s a longer review of the functionality (I haven’t dug into the code at all yet, but I will).

### Zoom to selected
My main issue is the confusing UI due to us mixing the y-zoom behaviour with the “zoom to selected button”. 

(Overview of how this button currently works: it zooms into the MRCA of the selected tips - using the terminology of Auspice we change the in-view root node to be the MRCA of the tips which are both visible and in-view. If this node is the same as the current in-view root node, then the button isn’t selectable. Note that this MRCA node may not be visible, and thus you can’t always achieve this behaviour by clicking on the branch to zoom.)

Importantly, “zoom to selected” _isn’t_ a flag or a state, it’s more like an action which is conditional on other state; this is why it’s so hard to store it as a URL query (see https://github.com/nextstrain/auspice/pull/1321).   

With this PR, it’s really easy to get into intermediate states. For instance  [filter to Oceania](https://auspice-dynamic-yvalues-mqlvu0.herokuapp.com/ncov/global?f_region=Oceania), then zoom-to-selected, then inactivate the filter. The y-zoom is still focused on Oceania, but all the tips are visible again.  There are plenty of other ones too (e.g. Jover’s screenshots), and it’s not clear when you’re in one of these intermediate states.

An alternative UI approach would be a **“Focused zoom” toggle**.  With this on, any change in `visibility` (zooming, filtering etc) would trigger a recalculation of y values. This would look really nice when changing filters or animating (if it could be made fast enough).  I think this UI would avoid all the intermediate states currently available in this PR, and make it extremely clear what mode we are currently in. It’s somewhat analogous to separating out the horizontal and vertical zooming. 

### Y value padding

In certain situations, (e.g. [South America + 614D](https://auspice-dynamic-yvalues-mqlvu0.herokuapp.com/ncov/global?branchLabel=none&f_region=South%20America&gt=S.614D)) the y-value calculations aren’t right. It looks like we’re setting some padding above selected tips, but not below them, which means the squashed branches are directly behind selected tips.

image


### Do we still need the (existing) “zoom-to-selected” behaviour?

Perhaps not. I haven’t fully thought through all the use cases here, but I think we could remove this. We can still zoom into clades by clicking on branches to explore certain parts of the tree, which is essentially what the button does.

### Zooming into early samples

> The biggest wins are being able to zoom to basal tips in the tree.

This is much improved in this PR, but there are still some cases it doesn’t solve, as the upper bound of the time domain is always the latest date of the in-view tips. For instance, zooming into clade 1B of flu/seasonal/vic/ha/12y could be better:

image

(Perfect is the enemy of good, and I don’t think we need to address this in this PR.)

victorlin:
Not able to build a new review app due to old NPM version. Rebase seems mostly trivial so I'm going to do that and hopefully get a working review app. Original commit is e4b02c510e2124c61b09aeb98e08685402539792 if we want to revert back.

victorlin:
PR is working now. I've updated the description with a link to a nextstrain.org preview instance and some remaining tasks. Also added some broader thoughts in the linked issue: https://github.com/nextstrain/auspice/issues/1368#issuecomment-2354019338

jameshadfield:
I'm really excited about this functionality, as well as the next steps of x-zoom! Here's some comments after using the UI for ~30min, but I haven't looked at the code.

**URL query**
Let's disable the URL query for the first implementation. We've generally followed a pattern of testing a feature out for a while before adding the query - especially so for new functionality like this. Stepping back a bit, this relates to my comment above and also referenced in [this related issue](https://github.com/nextstrain/auspice/issues/1368#issuecomment-2354019338) regarding a future where we drop the "zoom to selected" and just have focus. Or maybe vice-versa!

**Y-value calculations**
The y-value calculations are still not quite right - I think this is unchanged from my comment above. Using a filter with few samples makes this very obvious. Here's a screenshot from a subtree of Oropouche filtered to n=2 Columbia samples:

image

**Focus on selected button UI:**
1. The element looks to be clickable when a tree (with no filters) is first loaded, but it does nothing
2. How can we "disable" focus -- the element still appears to be clickable and says "focus on ..." and clicking it does nothing. Clicking "reset layout" restores the original view, but this wasn't obvious to me.
3. Is this supposed to act like a toggle or a button? I think we're aiming for the former¹. Assuming so, as we zoom the tree (clicking on branches / "zoom to selected" button) the y-values need to recalculate as different samples enter the viewport but they currently don't. Here's a screenshot of the above oropouche example where the focus was applied with n=2 samples selected and then I've zoomed out so now n=6 are selected but the new 4 are all squashed down the bottom (LHS) - the correct behaviour is on the RHS:
image 2
4. Assuming it is a toggle, the element should have a little on/off toggle or similar to convey this.


5. Inactivating / disabling a filter doesn't re-calculate the y-value (when "focused") -- this is the case I referred to in my original review about "intermediate states". 


---

¹ Adding a "focus on selected" element by itself is clarifying, especially for a v1 design.

trvrb:
Cool! Thanks so much for picking this up @victorlin. I have a couple comments (similar to James's above) based on UI testing.

1. I don't like adding things as options to the top right of a panel unless it's really something that should be an immediate click away (like "reset"). I'd suggest keeping this as a real toggle in the sidebar. It would act in a similar way as "Show confidence intervals" as changing the continuous behavior of the tree. I'd place this toggle at the end below "Explode Tree By" and I'd give it a similar experimental icon. For text I'd suggest "Focus tree on selected nodes".

2. Currently, the y-values aren't getting updated after new nodes are selected. To demonstrate go to https://nextstrain-s-nextstrain-l8gysc.herokuapp.com/zika?f_country=Colombia and then "Toggle focus" this updates y-values to focus on Colombia tips. However, then if you remove the Colombia filter and/or apply a USVI filter the y-values don't update. In "focus mode" the y-values should always reflect active tips.

I think that overall I recommend a "toggle" behavior and placing the toggle in the sidebar. I could see an argument for making it a button rather than a toggle ("focus on selected") that's essentially the current behavior. In this case I think it should really be possible to combine this button with the "zoom to selected" button. However, removing a filter with zoomed y-values results in a funky looking tree, like so:

Screenshot 2024-10-07 at 3 20 08 PM

while a toggle would prevent these funky trees from being accessible.

victorlin:
@jameshadfield @trvrb thanks for the feedback! Switched to toggle behavior in fa421d562bcbf84499de0c538af66845ab6dd96d. PR description has been updated with a list of the clear bugs to tackle and open design decisions.

I've also done a bit of extra refactoring and will make a base PR with changes that can be considered before the rest of this PR is ironed out.

victorlin:
I've pushed updates to drop the in-panel button and add a toggle in the sidebar. Auto-focus has also been addressed.

To combine this with the zoom feature, I'd propose moving "Zoom to Selected" to a sidebar toggle named "Auto-zoom to selected". This toggle, similar to the focus toggle, will update with any filter changes. We can keep the "zoom out" and "reset layout" in-panel buttons for manual adjustments, where any interaction with these buttons will switch off auto-zoom.

victorlin:
Everything but the Y calculation bug have been addressed. I'll look more into that tomorrow.

src/components/tree/phyloTree/layouts.js
```diff
@@ -288,6 +289,57 @@ export const setDistance = function setDistance(distanceAttribute) {
...
 
+/**
+ * given nodes add y values (node.displayOrder) to every node
+ * Nodes are the phyloTree nodes (i.e. node.n is the redux node)
+ * Nodes must have parent child links established (via createChildrenAndParents)
+ * PhyloTree can subsequently use this information. Accessed by prototypes
+ * rectangularLayout, radialLayout, createChildrenAndParents
+ * side effects: node.displayOrder and node.displayOrderRange (i.e. in the phyloTree node)
+ */
+export const calcYValues = (nodes, focus) => {
```
jameshadfield, line 300: Brief comments (after looking at the code for perhaps the first time since I wrote the original prototype!):

1. We should be able to get rid of `setDisplayOrder` as it's essentially doing the same thing as `calcYValues` and just use one function throughout the code.
2. `setDisplayOrder` specifically considers the case of an exploded tree which made me wonder what happens if we explode a tree and then toggle on focus - it's a bug! We should test the similar case of a JSON with n>1 trees (subtrees). In either case, for v1 functionality we can disable the focus functionality when we've exploded a tree / have multiple subtrees in the JSON.
3. Now that this is called frequently (rather than simply on dataset load) we can probably make it faster if needed. I see `setTreeFocus` has timers in it - we should build Auspice with `--includeTiming` and see how fast/slow this currently is.

src/components/controls/toggle-focus.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { FaInfoCircle } from "react-icons/fa";
+import { Dispatch } from "@reduxjs/toolkit";
+import Toggle from "./toggle";
+import { SidebarIconContainer, StyledTooltip } from "./styles";
+import { TOGGLE_TREE_FOCUS } from "../../actions/types";
+import { RootState } from "../../store";
+
+
+function ToggleFocus({ tooltip, treeFocus, layout, dispatch, mobileDisplay }: {
+  tooltip: Element;
```
genehack, line 13: I _think_ if you spell this `React.Element`, the type check warning in `controls/controls.tsx` may be silenced? 

```suggestion
  tooltip: React.Element;
```

src/components/controls/toggle-focus.tsx
```diff
@@ -0,0 +1,54 @@
...
+  treeFocus: boolean;
+  layout: "rect" | "radial" | "unrooted" | "clock" | "scatter";
+  dispatch: Dispatch;
+  mobileDisplay: boolean;
+}) {
+  const validLayouts = new Set(["rect", "radial"]);
+  if (!validLayouts.has(layout)) return <></>;
```
genehack, line 20: maybe deserves a little commentary around why the list of valid layouts here is a subset of the allowable types in the method signature?

src/components/tree/phyloTree/layouts.js
```diff
@@ -288,6 +289,57 @@ export const setDistance = function setDistance(distanceAttribute) {
...
+    node.displayOrderRange = [children[0].shell.displayOrder, children[children.length - 1].shell.displayOrder];
+  };
+  recurse(nodes[0]);
+};
+
+/**
+ * Recalculates y values based on focus setting
+ * @param treeFocus -- whether to focus on filtered nodes
+ */
+export const setTreeFocus = function setTreeFocus(treeFocus) {
```
genehack, line 338: is this not simpler as: 

```suggestion
export function setTreeFocus(treeFocus) {
```

src/components/tree/phyloTree/layouts.js
victorlin: 1. I chose to drop `calcYValues` and add focus functionality to `setDisplayOrder`: 9a2bdb451f3c09046ab0341bb946d7080953ab33
2. Added to checklist:

    > Disable toggle and turn off focus when using exploded tree / multiple trees in JSON

3. I did this locally and through some spot checking only saw `Timer setTreeFocus (#XX) took 0ms. Average: 0ms.`. Maybe I'm not checking in the right places though. Maybe we can tweak the review app to build with `--includeTiming`?

src/components/controls/toggle-focus.tsx
victorlin: Fixed: https://github.com/nextstrain/auspice/blob/4ad6b43bb1476903308afaabef474cf31cd711e3/src/components/controls/toggle-focus.tsx#L13

src/components/tree/phyloTree/layouts.js
victorlin: Keeping as-is to be consistent with the rest of the file. Once this is merged, I'm all for changing the entire file (or more of the codebase?) altogether.

src/components/controls/toggle-focus.tsx
victorlin: If the concern is that the misalignment is made apparent by the contrasting literal string lists, I'm hoping to change the method signature to `layout: Layout` in the near future by having `Layout` defined somewhere better, for example here in #1864: https://github.com/nextstrain/auspice/blob/b933874b8094ece368655e35a8a08fa3681249b6/src/components/tree/phyloTree/phyloTree.ts#L14

In the meantime, how about adding this as a comment above `const validLayouts`?

> The focus function is only available to layouts that have the concept of a unitless y-axis.

genehack: > If the concern is that the misalignment is made apparent by the contrasting literal string lists, I'm hoping to change the method signature to `layout: Layout` in the near future by having `Layout` defined somewhere better

It's not the contrasting literal string lists as much as the "if this only works on a subset of layout types why allow all layout types to be passed in?"; IOW even with `layout: Layout` in the method sig, I'd want to know why there's a filtering. 

> In the meantime, how about adding this as a comment above `const validLayouts`?
> 
> > The focus function is only available to layouts that have the concept of a unitless y-axis.

I would say add it, and leave it; it's a good bit of "why" context for the code.

src/components/tree/phyloTree/layouts.js
jameshadfield: That looks great! Yes ideally we'd allow it, I just wanted to convey it wasn't essential for the v1.

victorlin: Re: timing (3)

I've updated https://github.com/nextstrain/nextstrain.org/pull/1025 with https://github.com/nextstrain/nextstrain.org/pull/1025/commits/0671e90b0cdfa55d0ee5b01d0b78bc933cedb121 so the review app has timing messages.

The function is now `setDisplayOrder`. Playing with various date range filters on [this n=3343](https://nextstrain-s-nextstrain-l8gysc.herokuapp.com/ncov/gisaid/global/all-time@2024-10-14) dataset, the function takes between 0-3ms for me (M1 mac, Chromium browser):

image

src/components/tree/phyloTree/helpers.js
```diff
@@ -78,27 +81,76 @@ function _getSpaceBetweenSubtrees(numSubtrees, numTips) {
...
+
+  /**
+   * Keep track of whether the previous node was selected
+   */
+  let previousWasVisible;
+
+  if (focus) {
+    const numVisible = nodes.filter((d) => !d.hasChildren && d.visibility === NODE_VISIBLE).length;
+    const yPerFocused = (0.8 * nodes.length) / numVisible;
+    const yPerUnfocused = (0.2 * nodes.length) / (nodes.length - numVisible);
```
victorlin, line 131: This calculation translates to 80% space on focused nodes + 20% space for unfocused nodes.

The current description of the focus feature says "nodes that do not match the filter will occupy less vertical space on the tree". With the current implementation, there is a visual bug in the case of a large (>80%) selection: selected nodes will appear to shrink which is counter-intuitive.

https://github.com/user-attachments/assets/4be417e0-7fe8-47b7-84e9-6642ef5ad4d4

I'll look into adjusting the calculation to not be a fixed 80/20 and instead consider `numVisible` and `nodes.length`.

victorlin: Fixed: https://github.com/nextstrain/auspice/blob/72008c5a9d6935e69e761e3dd5eb01264d9cbd15/src/components/tree/phyloTree/helpers.js#L129-L133

src/components/tree/phyloTree/change.js
victorlin, line 1: Another bug: date filter change doesn't auto-focus.

https://github.com/user-attachments/assets/3c84ce24-db74-4e5c-9133-6f8997793249

This seems trickier since date filter changes aren't reflected in Redux state change:

https://github.com/nextstrain/auspice/blob/72008c5a9d6935e69e761e3dd5eb01264d9cbd15/src/reducers/tree.js#L42

victorlin: Nevermind, fix was easy: ba86c184984c58c9ba515a66588e09e2b32384ce

src/components/tree/phyloTree/layouts.js
jameshadfield: Those times are fine - they make it clear it's not something we need to optimise at the moment. Thanks for looking into it!

victorlin: Thanks for the pointer! I'll still plan to keep 330e87d43ab459822ae1584036e4c788ee5369d9 in the PR.

src/components/tree/phyloTree/helpers.js
```diff
@@ -33,19 +34,20 @@ export const applyToChildren = (phyloNode, func) => {
...
   const children = node.n.children; // (redux) tree node
   if (children && children.length) {
     for (let i = children.length - 1; i >= 0; i--) {
-      yCounter = setDisplayOrderRecursively(children[i].shell, yCounter);
+      yCounter = setDisplayOrderRecursively(children[i].shell, getDisplayOrder, yCounter);
     }
   } else {
-    node.displayOrder = (node.n.fullTipCount===0 || yCounter===undefined) ? yCounter : ++yCounter;
-    node.displayOrderRange = [yCounter, yCounter];
+    node.displayOrder = (node.n.fullTipCount===0 || yCounter===undefined) ? yCounter : yCounter + getDisplayOrder(node);
```
jameshadfield, line 50: Previously, `yCounter` referred to the previous tip's display order, and we incremented it by one every time we visited a new terminal node. But now things are a bit stranger (IMO):

* `yCounter` is not incremented except for when we jump into a new subtree (via `yCounter+=spaceBetweenSubtrees;`)
* `getDisplayOrder` is a closure and returns the display order for this node, itself keeping track of the previous tip's display order etc.
* so `yCounter` is more like an offset value that changes across subtrees.

I _think_ we can get rid of `getDisplayOrder` entirely and inline it within the existing two functions, using argument passing instead of closures. `yCounter` would then continue to represent the previous tip's value. 

Note that there are a few strange bugs around trees with many subtrees because `yCounter` gets large (lots of `yCounter+=spaceBetweenSubtrees;` calls) and thus the `yPerFocused` and `yPerUnfocused` values are smaller than they should be. I think it'd be clarifying to compute all these values within `setDisplayOrder` so we can know ahead of time the range of display order values we'll generate.

I'm happy to test out a prototype of what I'm thinking (tomorrow), if that works?

src/components/tree/phyloTree/helpers.js
```diff
@@ -99,6 +103,51 @@ export const setDisplayOrder = (nodes) => {
...
+  let displayOrder = 0;
+
+  /**
+   * Keep track of whether the previous node was selected
+   */
+  let previousWasVisible;
+
+  if (focus) {
+    const numVisible = nodes.filter((d) => !d.hasChildren && d.visibility === NODE_VISIBLE).length;
+    const yProportionFocused = Math.max(0.8, numVisible / nodes.length);
```
jameshadfield, line 130: Mentioned above, and not a big deal for the v1 of this feature, but when we have subtrees this no longer represents the proportion of vertical space the focused tips will take up as the space between subtrees hasn't been accounted for. The more subtrees there are the more this inter-subtree space will start to dominate.

jameshadfield, line 130: [minor - not blocking] When we only have one (or a few) tips selected, 80% is a bit much. I played with
```js
    const focusMaxProportion = Math.min(0.8, 0.2 * numVisible)
    const yProportionFocused = Math.max(focusMaxProportion, numVisible / nodes.length);
```
which felt better, but there's probably no perfect solution here.

src/components/controls/miscInfoText.js
```diff
@@ -64,3 +64,11 @@ export const ExplodeTreeInfo = (
     It works best when the trait doesn&apos;t change value too frequently.
   </>
 );
+
+export const ToggleFocusInfo = (
```
jameshadfield, line 68: [commenting on this line to enable a thread, it's not related to the info content]

Animation + focus gets a little crazy. I don't think it's worth resolving this at the moment - it's labelled "experimental" after all - but wanted to note it here.

jameshadfield (changes requested):
Lots of comments - most minor, the only major one is around how display order is actually computed. 

The functionality is generally working very well and will be a really nice improvement to Auspice.

src/components/tree/phyloTree/helpers.js
jameshadfield: > I'm happy to test out a prototype of what I'm thinking

Spent 15 minutes and I do think this'll be simpler. Dealing with symmetric padding around focused tips is still a TODO in my version however. https://github.com/nextstrain/auspice/commit/ec13e23b3657f499fee2b254f2854b27a4b9f6b3

src/components/tree/phyloTree/helpers.js
```diff
@@ -78,27 +81,78 @@ function _getSpaceBetweenSubtrees(numSubtrees, numTips) {
...
+   */
+  let displayOrder = 0;
+
+  /**
+   * Keep track of whether the previous node was selected
+   */
+  let previousWasVisible;
+
+  if (focus) {
+    const numVisible = nodes.filter((d) => !d.hasChildren && d.visibility === NODE_VISIBLE).length;
```
jameshadfield, line 129: 🐛  
```diff
-     const numVisible = nodes.filter((d) => !d.hasChildren && d.visibility === NODE_VISIBLE).length;
+    const numVisible = nodes.filter((d) => !d.n.hasChildren && d.visibility === NODE_VISIBLE).length;
```

But even simpler, just use `nodes[0].n.tipCount`

src/components/tree/phyloTree/helpers.js
jameshadfield: And then spent another bit of time getting the symmetry working... I'll stop now.

6074280535c6c9a104a7d1b5d3ef850641fe3716 (on top of https://github.com/nextstrain/auspice/commit/ec13e23b3657f499fee2b254f2854b27a4b9f6b3 above) accomplishes it using a closure to capture `previousWasVisible`, just as you did.

https://github.com/nextstrain/auspice/commit/80fe5386d8c5a445117e2f9310514ee2d538fdae uses a stack to traverse the tree rather than a recursive function. It's probably better and faster in the long run, but probably has bugs and so I wouldn't use it in this PR.

src/components/tree/phyloTree/helpers.js
victorlin: With 512b530980b89dcf37749adcedbd1ae2548d789a there is a slight difference in behavior when all tips are selected and toggling focus, but I'm thinking overall it's better.

src/components/controls/toggle-focus.tsx
```diff
@@ -0,0 +1,55 @@
...
+  mobileDisplay: boolean;
+}) {
+  // Focus functionality is only available to layouts that have the concept of a unitless y-axis
+  const validLayouts = new Set(["rect", "radial"]);
+  if (!validLayouts.has(layout)) return <></>;
+
+  const label = (
+    <div style={{ display: "flex", alignItems: "center" }}>
+      <span>Focus on Selected</span>
+      <ImLab style={{ margin: "0 5px" }} />
```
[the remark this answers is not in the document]
jameshadfield: Thanks!

src/components/tree/phyloTree/helpers.js
jameshadfield: Thanks!

src/components/tree/phyloTree/helpers.js
jameshadfield: > I plan to circle back to it after this PR!

Nice one. Feel free to effect `/tipCount/visibleTipCount/` at that stage if you'd like.

src/components/tree/reactD3Interface/change.js
```diff
@@ -47,6 +55,20 @@ export const changePhyloTreeViaPropsComparison = (mainTree, phylotree, oldProps,
...
+  }
+
+  /* enable/disable focus */
+  if (oldProps.focus !== newProps.focus) {
+    args.newFocus = newProps.focus;
+    args.updateLayout = true;
+  }
+  /* re-focus on changes */
+  else if (oldProps.focus == true &&
+           newProps.focus == true &&
```
jameshadfield, line 68: Both these should use strict equality (`x === true`). In practice the supplied values are going to be `undefined`, `false` and `true`, so it shouldn't make a difference, but I'd like strict equalities used everywhere.

src/components/tree/reactD3Interface/change.js
```diff
@@ -47,6 +55,20 @@ export const changePhyloTreeViaPropsComparison = (mainTree, phylotree, oldProps,
...
+           (zoomChange || dateRangeChange || filterChange)) {
+    args.newFocus = true;
+    args.updateLayout = true;
```
jameshadfield, line 71: The naming of "newFocus" isn't quite right I think - sometimes it represents a change in focus state, but in the other 2 cases it simply represents the existing (unchanged) state as we need to pass this value to the `setDisplayOrder` function downstream. I'd suggest "focusState" or something.

(I know this is minor, but this section of code is relatively complex and when we re-read it months down the track little things like this become very helpful.)

src/components/tree/phyloTree/helpers.js
```diff
@@ -77,26 +83,61 @@ function _getSpaceBetweenSubtrees(numSubtrees, numTips) {
...
+  let incrementer = (_node) => 1;
+
+  if (focus) {
+    const numVisible = nodes[0].n.tipCount;
+
+    let yProportionFocused = 0.8;
+    // Adjust for a small number of visible tips (n<4)
+    yProportionFocused = Math.min(yProportionFocused, numVisible / 5);
+    // Adjust for a large number of visible tips (>80% of all tips)
+    yProportionFocused = Math.max(yProportionFocused, numVisible / nodes.length);
```
jameshadfield, line 108: `nodes` includes internal nodes whereas we just want the proportion of tips, so this should be `numVisible / nodes[0].n.fullTipCount`

jameshadfield (changes requested):
Just a few minor (<5min) changes left now, and also remember to drop the last 2 commits!

src/components/tree/phyloTree/helpers.js
jameshadfield: Looks good - you're right about the slight difference in behaviour, but this highlights a subtle issue where we're confusing `nodes.length` with total number of tips. Fixing this improves other things as well!

```diff
- const yPerFocused = (yProportionFocused * nodes.length) / numVisible;
- const yPerUnfocused = ((1 - yProportionFocused) * nodes.length) / (nodes.length - numVisible);
+ const yPerFocused = (yProportionFocused * nodes[0].n.fullTipCount) / numVisible;
+ const yPerUnfocused = ((1 - yProportionFocused) * nodes[0].n.fullTipCount) / (nodes[0].n.fullTipCount - numVisible);
```

jameshadfield: Nice one! Resolving this, but note [the above requested change to this code](https://github.com/nextstrain/auspice/pull/1373#discussion_r1807940275)

src/components/tree/phyloTree/helpers.js
victorlin: Good catch! Fixed and used a new variable for clarity: https://github.com/nextstrain/auspice/blob/126b895d6cc0623b6a1b5534873291f6d8977bd3/src/components/tree/phyloTree/helpers.js#L102-L112

src/components/tree/phyloTree/helpers.js
victorlin: Good catch! Fixed and used a new variable for clarity: https://github.com/nextstrain/auspice/blob/126b895d6cc0623b6a1b5534873291f6d8977bd3/src/components/tree/phyloTree/helpers.js#L102-L112

src/components/tree/reactD3Interface/change.js
victorlin: https://github.com/nextstrain/auspice/pull/1872

jameshadfield (approved):
Thanks for all the changes @victorlin - code looks good from my end, although the final three 🚧 commits need to be squashed with others etc. Parent PR #1373 also looks good. I'll let you merge both PRs at your convenience, then we can release a new version.

Issue #1356, opened 2021-06-10
Mouseover tip labels sometimes cut off from view


emmahodcroft:
**Current Behavior**  
Depending on tip position and tree view, sometimes mouseover tip labels are cutoff so you can't see the entire box. Since these mouseover boxes contain some info that's not in the tip-click boxes (like divergence), it makes this information not-accessible. Also if trying to see a bunch of tips, can be time-consuming to click them all. 

Understandable that this won't be possible at all zooms and screen resolutions, but below is at 100% browser zoom in 1920x1200 resolution, full screen (laptop), and it seems like if box was moved up a bit, at least a lot more of it could be visible:
![image](https://user-images.githubusercontent.com/14290674/121501612-aea6f780-c9df-11eb-9c93-a4e261c98518.png)
https://nextstrain.org/ncov/europe?c=S1_mutations&l=clock&s=India/MH-NIV-8404/2020,Mongolia/7908/2020,Peru/LIM-INS-768/2020,France/PAC-IHU-3126i/2020

**Expected behavior**  
Maybe reduce the amount trying to link tool-tip mouseover-box to 'start' at the node being moused over, and let it be at a higher point if more of the box would be visible?

However, I know tool-tip boxes can be a real pain to get right. Still thought it was worth flagging as I keep running into it!! Ensuring all info in mouse-over boxes is also in tip-click boxes would also help.

**Your environment: if browsing Nextstrain online**  
 - Operating system:  Windows 10 Pro
 - Browser:  Chrome

trvrb:
I agree that this is annoying. Two thoughts:

1. We should definitely make sure that everything listed in `infoPanel.hover` is available in `infoPanel.click`. Easy to add "divergence" here. Were there other attributes missing? I think it's just branch-specific nuc and AA mutations? I'm not actually sure if these should be in `infoPanel.click`.

2. It will often be impossible to fit a hover tooltip just given height of Phylogeny panel relative to height of tooltip. But that said, maybe an easy improvement would be to make the hover tooltips slightly wider to fit more content into the space? I'm rarely getting an issue of occlusion on left-side of panel or right-side of panel.

emmahodcroft:
[...]

PR #1375, opened 2021-07-13
Dynamically set infoPanel.hover width

trvrb:

This sizes up width of infoPanel.hover, eg "tooltip", when the Tree panel is wider. This helps with two issues:

1. Longer strain names don't wrap into as many lines and thus look better
2. This often decreases height of tooltip and thus helps with vertical overflow

Specific relationship between panel width and tooltip width was set via experimentation, but the panel can roughly support a tooltip half its width.

### Related issue(s)  
Partial fix for #1356

### Testing
Tested locally.

### DIFF
--- a/src/components/tree/infoPanels/hover.js
+++ b/src/components/tree/infoPanels/hover.js
@@ -281,9 +281,22 @@ const AttributionInfo = ({node}) => {
 };
 
 const Container = ({node, panelDims, children}) => {
+
   const xOffset = 10;
   const yOffset = 10;
-  const width = 200;
+
+  let width = 200;
+  if (panelDims.width < 420) {
+    width = 200;
+  } else if (panelDims.width < 460) {
+    width = 220;
+  } else if (panelDims.width < 500) {
+    width = 240;
+  } else if (panelDims.width < 540) {
+    width = 260;
+  } else {
+    width = 280;
+  }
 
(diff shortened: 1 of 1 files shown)


emmahodcroft:
I think this is a great idea, and I do think it is an improvement, but at least in testing out on full-screen laptop, this increase in what was visible wasn't quite as much as I might have expected:

Current:
image


On this PR:
image

(The cut-off here is what gets cut off from being visible in the box, not a cut from screenshotting)

jameshadfield:
This looks like a solid improvement so I'm going to merge - as above, this is only a partial fix of obscured tip-info boxes.

The vertical overflow (e.g. see @emmahodcroft's screenshots) is more complicated because the info-panel `<div>` is contained within (and clipped by) the tree card (`<div id="PhylogenyCard"`>) and thus, even if there is real estate visible above/below the tree, the info box cannot use it. I think it would be worth stepping back and looking how we deal with hover boxes so we can improve these more generally (added #1356 to my to-do list 😉).

jameshadfield (approved):

trvrb:
Thanks for the feedback and thanks for merging. I didn't want to make any wider than 280 pixels as the wider tooltips cover up useful bits of the tree and because the tooltip contains so many shorter elements with explicit new lines, the wider the tooltip gets the more it consists of empty gray space. I felt that 280 was a good compromise here.

PR #1380, opened 2021-07-16
Update TravisCI config

jameshadfield:
The npm Auspice 2.29.0 release had integrity shasum errors:

```sh
# reported by npm publish running on TravisCI (node.js 10)                               
npm notice integrity:     sha512-zeNaSDrnIgvH6[...]PA0JfybEzP8Ig==
# reported by npm view auspice@2.29.0 after publishing
.integrity: sha512-27B/yNAPP1/vmSoqdBb3g1EMXJW/FLGFWSlTSBl7dDLksHAabqWOYP14aDWF5p4jUuvUNsT20O6G4b3pRcxRAg==
```

Looking through the TravisCI logs it became clear we were running `npm publish` for both versions of Node.js we ran TravisCI on, which resulted in one failure each time we released such as

```
npm ERR! 403 403 Forbidden - PUT https://registry.npmjs.org/auspice - You cannot publish over the previously published versions: 2.29.0.
```

This PR updates the Node.js versions we run CI on (v10 is no longer maintained), and only runs `npm publish` when building under v14. It is not clear whether this is related to the integrity shasum errors, or if they were stochastic (or some other 🐛 ).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ install:
 script:
 - npm run build
 after_success:
-- if [[ "$TRAVIS_PULL_REQUEST" == "false" && "$TRAVIS_BRANCH" == "release" ]]; then
+- if [[ "$TRAVIS_PULL_REQUEST" == "false" && "$TRAVIS_BRANCH" == "release" && "$TRAVIS_NODE_VERSION" == "14" ]]; then
     npm run rebuild-docker-image;
     npm config set //registry.npmjs.org/:_authToken="$NPM_TOKEN";
     npm publish;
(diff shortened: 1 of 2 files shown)


eharkins:
I would be surprised if there wasn't some stochastic issue involved if this has been happening all along where we try to publish to npm for multiple versions of node and it only just now caused the issue with 2.29.0.

I also noticed this error with respect to updating the docker image: https://travis-ci.com/github/nextstrain/auspice/jobs/524607375#L576 that occured in one of the builds for 2.29.0 but can't say whether it contributed to this issue - it seems relatively unrelated.

jameshadfield:
> I would be surprised if there wasn't some stochastic issue involved if this has been happening all along

That would be my hunch too, but I wanted to take the chance to improve the Travis config here. 

Version 2.29.1 was successfully released using the config introduced in this PR.

PR #1382, opened 2021-07-24
Include attribution link with gisaid_epi_isl

trvrb:

This updates behavior to match what we were doing for `genbank_accession`. Here, we special case both `genbank_accession` and `gisaid_epi_isl` and construct direct DB links when these are present rather than using the more generic `accession` / `url` metadata pair.

This update allows users to lookup attribution info for individual samples even if workflow metadata did not include this attribution data.

### Testing
Tested locally.

### DIFF
--- a/src/components/tree/infoPanels/click.js
+++ b/src/components/tree/infoPanels/click.js
@@ -88,23 +88,39 @@ const MutationTable = ({mutations}) => {
 
 const AccessionAndUrl = ({node}) => {
   /* If `gisaid_epi_isl` or `genbank_accession` exist as node attrs, these preempt normal use of `accession` and `url`.
-  These special values were introduced during the 2019 nCoV outbreak. */
+  These special values were introduced during the SARS-CoV-2 pandemic. */
   const gisaid_epi_isl = getTraitFromNode(node, "gisaid_epi_isl");
   const genbank_accession = getTraitFromNode(node, "genbank_accession");
+  let gisaid_epi_isl_url = null;
+  let genbank_accession_url = null;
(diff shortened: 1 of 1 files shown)


jameshadfield (approved):
Tested using a variety of (core) datasets to check the different formats data may appear in (partly to remind myself!)

| Fields | example | behavior |
| --- | --- | --- |
| (none) | H7N9 | Nothing displayed |
| `url` | locally tested | "Strain URL" displayed with link to supplied `url` |
| `accession` | H3N2 | Displayed without any links |
| `accession`, `url` | Historical nCoV dataset | display the accession which is a link to the specified URL |
| `genbank_accession`, `url` | ncov/open | "Genbank Accession" displayed, with link constructed from the accession number. The provided `url` information is ignored. |
| `gisaid_epi_isl`, `url` | ncov/gisaid | "GISAID EPI ISL" displayed, with a link to the epicov JSON. The provided `url` information is ignored. | 
| `genbank_accession`, `gisaid_epi_isl`, `url` | ncov/gisaid | "Genbank accession" and "GISAID EPI ISL" displayed, each with a link as above. The provided `url` information is ignored. | 

@trvrb if this is as you expect it to be, please merge & I can release in this week's auspice release.

trvrb:
Thanks for outlining James. We may want to rethink how we do accessions at some point to make it easier to have multiple accessions with matched URLs for a sample, but I think this current priority system is good for the moment. `genbank_accession` and `gisaid_epi_isl` take precedence over more generic `accession` / `url` inclusion.

PR #1383, opened 2021-07-28
Clip tree to improve visibility of axes

jameshadfield:
This (essentially) clips branches, tips and grids to the range of the
{x,y} scales in order to prevent branches / tips from obscuring the axes

This works well on the scatterplots I've thrown at it, but a bit more testing is needed. I haven't tested the slowdown (if any) which I'll do before merging.

The display of y-axis tick labels is another area of improvement to make for scatterplots


jameshadfield:
Merging after further testing

PR #1386, opened 2021-08-28
Drag-and-drop Excel

joverlee521:
Support drag-and-drop of metadata in Excel format. 
Currently only reads in data from the first sheet in the Excel workbook.

### DIFF
--- a/src/actions/filesDropped/metadata.js
+++ b/src/actions/filesDropped/metadata.js
@@ -1,7 +1,7 @@
 import { rgb } from "d3-color";
 import { errorNotification, successNotification, warningNotification } from "../notifications";
 import { ADD_EXTRA_METADATA } from "../types";
-import { parseCsvTsv } from "./parseCsvTsv";
+import { parseCsv } from "./parseCsv";
 
 
 const handleMetadata = async (dispatch, getState, file) => {
@@ -18,7 +18,7 @@ const handleMetadata = async (dispatch, getState, file) => {
 
       /* All accepted file formats have been converted to CSV string by xlsx */
       /* Use papaparse to parse & interrogate the CSV string */
-      const {errors, data, meta} = await parseCsvTsv(sheetAsCsv);
+      const {errors, data, meta} = await parseCsv(sheetAsCsv);
       if (errors.length) {
         console.error(errors);
         throw new Error(errors.map((e) => e.message).join(", "));
--- a/src/actions/filesDropped/metadata.js
+++ b/src/actions/filesDropped/metadata.js
@@ -6,38 +6,49 @@ import { parseCsvTsv } from "./parseCsvTsv";
 
 const handleMetadata = async (dispatch, getState, file) => {
   const fileName = file.name;
+  const reader = new FileReader();
+  reader.onload = async (event) => {
+    try {
+      const XLSX = (await import("xlsx/xlsx.mini")).default;
+      /* Convert accepted dropped file to CSV string */
+      /* If dropped file is Excel workbook, only reads in the data from the first sheet */
+      const workbook = XLSX.read(event.target.result, { type: 'binary' });
+      const firstSheet = workbook.Sheets[workbook.SheetNames[0]];
+      const sheetAsCsv = XLSX.utils.sheet_to_csv(firstSheet);
 
-  try {
-    /* Parse & interrogate the CSV file */
-    const {errors, data, meta} = await parseCsvTsv(file);
-    if (errors.length) {
-      console.error(errors);
-      throw new Error(errors.map((e) => e.message).join(", "));
+      /* All accepted file formats have been converted to CSV string by xlsx */
+      /* Use papaparse to parse & interrogate the CSV string */
+      const {errors, data, meta} = await parseCsvTsv(sheetAsCsv);
+      if (errors.length) {
+        console.error(errors);
+        throw new Error(errors.map((e) => e.message).join(", "));
+      }
+      const {coloringInfo, strainKey, latLongKeys, ignoredFields} = processHeader(meta.fields);
+      const rows = {};
+      data.forEach((d) => {rows[d[strainKey]]=d;});
+
+      /* For each coloring, extract values defined in each row etc */
+      const newNodeAttrs = {};
+      const newColorings = processColorings(newNodeAttrs, coloringInfo, rows, fileName); // modifies `newNodeAttrs`
+      const newGeoResolution = latLongKeys ? processLatLongs(newNodeAttrs, latLongKeys, rows, fileName) : undefined;
+      /* Fix errors in data & dispatch warnings here, as we cannot dispatch in the reducers */
+      const ok = checkDataForErrors(dispatch, getState, newNodeAttrs, newColorings, ignoredFields, fileName);
+      if (!ok) return undefined;
+
+      dispatch({type: ADD_EXTRA_METADATA, newColorings, newGeoResolution, newNodeAttrs});
+      return dispatch(successNotification({
+        message: `Adding metadata from ${fileName}`,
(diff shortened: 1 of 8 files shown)


src/actions/filesDropped/parseCsv.js
```diff
@@ -26,7 +21,7 @@ export const parseCsvTsv = async (file) => {
       },
       encoding: "UTF-8",
       comments: "#",
-      delimiter: (csv_file_types.includes(file.type)) ? "," : "\t",
+      delimiter: ",",
```
ivan-aksamentov, line 24: I only glanced over the PR and it looks good overall. But this place made me wonder how is TSV handled now. Is it still working? It is unclear from code changes immediately where it migrated. Is it now handled by the `xlsx` rather than `papaparse`?

src/actions/filesDropped/metadata.js
```diff
@@ -1,43 +1,53 @@
...
+      const rows = {};
+      data.forEach((d) => {rows[d[strainKey]]=d;});
+
+      /* For each coloring, extract values defined in each row etc */
+      const newNodeAttrs = {};
+      const newColorings = processColorings(newNodeAttrs, coloringInfo, rows, fileName); // modifies `newNodeAttrs`
+      const newGeoResolution = latLongKeys ? processLatLongs(newNodeAttrs, latLongKeys, rows, fileName) : undefined;
+      /* Fix errors in data & dispatch warnings here, as we cannot dispatch in the reducers */
+      const ok = checkDataForErrors(dispatch, getState, newNodeAttrs, newColorings, ignoredFields, fileName);
+      if (!ok) return undefined;
```
ivan-aksamentov, line 35: This is not related to the PR, but I noticed that in this function some errors are manifested as exceptions (including the ones from libraries) and then dispatched to the store, but this one uses an error code and returns `undefined`.

Might be just different kind of errors, like ones we want to manifest vs ones to be silent. We might also bubble the exception handling up on the call stack, and decide how different errors are handled upstream.

src/actions/filesDropped/parseCsv.js
joverlee521: Ah, sorry this isn't clear. Everything gets converted to a CSV string by `xlsx` and then `papaparse` processes that CSV string.

src/actions/filesDropped/metadata.js
ivan-aksamentov: I see. Just noted that this particular bit looks inconsistent, I thought might be a bug in case this error gets erroneously silenced through return.

> Do you mean to hand these messages back up through custom errors that can be caught by the outer try/catch?

Yes, I usually try to do something like that. But probably not worth the trouble if already done otherwise.

src/actions/filesDropped/parseCsv.js
joverlee521: @huddlej I can definitely add a comment! Whatever makes this clearer! 
Would it make more sense to have the comment here in `parseCsv.js` or in `metadata.js` where the function gets called?

src/actions/filesDropped/metadata.js
```diff
@@ -1,43 +1,53 @@
...
-import { parseCsvTsv } from "./parseCsvTsv";
+import { parseCsv } from "./parseCsv";
 
 
 const handleMetadata = async (dispatch, getState, file) => {
   const fileName = file.name;
+  const reader = new FileReader();
+  reader.onload = async (event) => {
+    try {
+      const XLSX = (await import("xlsx/xlsx.mini")).default;
```
jameshadfield, line 12: I’ll note here that we’re lazily importing two libraries for drag-and-drop parsing, and each gets bundled separately:  `xlsx.mini.js` (67kb gzipped) and `papaparse` (7kb gzipped), and this results in two separate browser fetches. Ideally these would be bundled together, but AFAIK this requires  them as a cache group to webpack, rather than a JS solution (e.g. `Promise.all`). 

P.S. You can interrogate the bundles via  `auspice build --analyzeBundle`

src/actions/filesDropped/metadata.js
```diff
@@ -1,43 +1,53 @@
...
+      /* Convert tabular file to CSV string */
+      /* Only reads in the data from the first sheet of Excel files */
+      const workbook = XLSX.read(event.target.result, { type: 'binary' });
+      const firstSheet = workbook.Sheets[workbook.SheetNames[0]];
+      const sheetAsCsv = XLSX.utils.sheet_to_csv(firstSheet);
```
jameshadfield, line 17: Is there a need to go through CSV and then parse with `papaparse`? The XLSX library seems to correctly split my test CSV/TSVs into cells, and then we could use `XLSX.utils.sheet_to_json(firstSheet)`. Are there cases where special characters etc need this roundtrip - I would think these fail to be read correctly into cells initially.

jameshadfield (changes requested):
Thanks Jover, and 💯  for updating the docs contemporaneously. 

I tested on a few different CSV, TSV and Excel files (the latter exported by MacOS numbers) and it works well. I was hoping the library would also handle `.numbers` files, but it doesn't and it's a small user pool compared with `.xlsx` so not necessary. 

See comment above re: dropping papaparse.

src/actions/filesDropped/metadata.js
joverlee521: I thought about dropping `papaparse` but I couldn't find equivalent options in `xlsx` for `comments` and `dynamicTyping`. If these aren't important to keep, then we can drop `papaparse`.

jameshadfield: Probably worth keeping it for those benefits. It's only 7kb, and only fetched when this functionality is needed.

src/actions/filesDropped/metadata.js
joverlee521: @jameshadfield Would this involve adding a new entry to `cacheGroups` in `webpack.config.js` that only contains `xlsx` and `papaparse`?

joverlee521:
```diff
+  const dragAndDrop = [
+    "papaparse",
+    "xlsx"
+  ];
+
   /**
    * It's better to keep small libraries out of the vendor bundles because
    * the more libraries we include, the more small changes or updates
@@ -215,6 +222,12 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
             enforce: true,
             chunks: "all"
           },
+          dragAndDrop: {
+            test: new RegExp("[\\\\/]node_modules[\\\\/](" + dragAndDrop.join("|") + ")[\\\\/]"),
+            name: "drag-and-drop",
+            enforce: true,
+            chunks: "all"
+          },
```
When I add the above to `webpack.config.js`, I can see both `papaparse` and `xlsx` are in one drag-and-drop bundle. When I build and test locally, I see the browser fetches the drag-and-drop bundle but it also fetches a numbered bundle that only contains the following:
```js
(window.webpackJsonp=window.webpackJsonp||[]).push([[13],{751:function(n,w){}}]);
```
I am very confused why this bundle is created and fetched when it seems to contain nothing 
Am I missing something? 🤔

src/actions/filesDropped/parseCsv.js
huddlej: @joverlee521 A comment in `metadata.js` would probably be most useful. Once you get to `parseCsv.js`, as a code reader, you probably already know what's going on.

src/actions/filesDropped/metadata.js
jameshadfield: I can't answer this without another deep dive into webpack, and given the options available I would advise leaving the config unmodified - i.e. we fetch xlsx and papaparse separately. We could make an issue with the above comments and tag it "revisit sometime".

src/actions/filesDropped/parseCsv.js
joverlee521: I re-pushed with updated comments

jameshadfield (approved):

Issue #1348, opened 2021-05-19
Add back button to allow gradual zoom back out from clade in addition to "reset layout"


corneliusroemer:
**Context**  
In auspice, one can iteratively zoom into clades, but the only way to get back out is to click `RESET LAYOUT` which isn't terribly user friendly when exploring trees.

![image](https://user-images.githubusercontent.com/25161793/118861633-010f5f80-b8dd-11eb-832a-067b109a557e.png)

**Description**   
It would be great if there was an additional button `zoom out one step`

rneher:
this exists, but isn't very discoverable...

If you click on the root of the clade you currently zoomed to, it will zoom out one branch.

rneher:
A zoom out button might be an easy and intuitive fix.

corneliusroemer:
Yes, I don't think it would harm.

The existing option of moving up one parent at a time is good, but tedious if you went in multiple steps. Moving out one level at a time is slightly different than going back to the previous view. There's a use case for both.

trvrb:
I'd think the best solution here would be to properly work with the browser back/forward buttons and have this be a feature for generic app state. Ie you change state of Auspice to zoom tree, or filter or whatever and can easily get back to previous state by clicking back button. I think this would be much more intuitive than trying to add a universal back button to the app somewhere.

joverlee521:
Resolved by #1392 . If desired, we can make new issue for a "back to previous state" feature that is separate from the zoom out.

corneliusroemer:
Amazing! Yes I think it'd still be cool to have a "back to previous state" feature in addition.

Issue #1390, opened 2021-09-09
Clicking root doesn't zoom out


eharkins:
**Current Behavior**  
When clicking on a branch in auspice, you zoom into view a clade or specific part of the tree. When @BryanTegomoh tried recently to click on the root of the clade in view (hover shows "_click to zoom out to parent clade_"), nothing changed / the tree did not zoom out.

**Expected behavior**  
Clicking on the root of a zoomed in auspice view (where it says "_click to zoom out to parent clade_" as you hover) should zoom out to the parent clade.

**How to reproduce**  
Still need to get in touch with Bryan about reproducing this. I think it's related to his browser which is different from mine.

**Possible solution**  
Adding some animation or other type of support for the Mircrosoft Edge browser to allow this feature to work in that browser.

**Your environment: if running Nextstrain locally**  
 - Operating system:   Windows
 - Browser:  Microsoft Edge
 - Version (e.g. `auspice 2.7.0`):  unknown

PR #1392, opened 2021-09-14
Improve ability to zoom out in the tree

jameshadfield:
See commit messages for details.

### DIFF
--- a/src/components/tree/phyloTree/layouts.js
+++ b/src/components/tree/phyloTree/layouts.js
@@ -360,6 +360,10 @@ export const mapToScreen = function mapToScreen() {
       maxX = minimumXAxisSpan - minX;
       spanX = minimumXAxisSpan;
     }
+    /* In rectangular mode, if the tree has been zoomed, leave some room to display the (clade's) root branch */
+    if (this.layout==="rect" && this.zoomNode.n.arrayIdx!==0) {
+      minX -= (maxX-minX)/20; // 5%
+    }
     xDomain = [minX, maxX];
   } else {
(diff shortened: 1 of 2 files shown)


joverlee521 (commented):
This works great! There's the added bonus of being able to zoom out after clicking the "zoom to selected" button as well! Previously this was not possible since the root branch was "inactivated" with the filter.

I updated the icon to be clear that the button is a "zoom out" button. I also added zoom out for the second tree. This means the zoom out button affects both trees, so if the user wants to zoom out of a specific tree they would have to use the old method of clicking the root branch. This pattern makes sense to me since the user already has to click on each tree individually in order to zoom in.

joverlee521:
This should close #1348 and maybe #1390?

trvrb:
This is great! Based on current behavior, I'd be supportive of merging. A couple brief thoughts however:
1. I had pictured the icon appearing over the root node when mousing over the tree. However, I really like having the icon in the row at the top, it makes it discoverable and not too in the way.
2. It slightly bugs me that the map zoom out is just a minus icon, while the tree zoom out is the magnifying glass minus icon. However, I would think just minus would be confusing and "zoom out" would take up too much space.

emmahodcroft:
After playing with this a bit, I really like it! I agree with Trevor that I'd pictured the button differently but I actually find that position very neat and tidy and accessible.

jameshadfield:
Thanks for the feedback and the improvements @joverlee521 - I think the icons can still be improved a bit in the future, but merging this now as it's immediately useful.

Issue #1395, opened 2021-09-16
Inconsistent parsing of age metadata in dragged on csv


eharkins:
**Current Behavior**  
Dragging on a csv of metadata onto auspice, with a column named "Age Group", with values such as "Age 0-4", those values showed up in the color legend as date-like values instead of their original values, seemingly inconsistently. Some values were preserved while others were converted to date-like values with `/` characters.

**Expected behavior**  
Auspice uses the value of the metadata without trying to parse it into a date.

**How to reproduce**  
This was observed in a zoom meeting with @BryanTegomoh, who sent over the example data that caused the issue. I will try to reproduce locally soon.

**Possible solution**  
There are two functions, `processHeader` and `processColorings` which I would think would be doing this parsing here https://github.com/nextstrain/auspice/blob/master/src/actions/filesDropped/metadata.js#L26-L32 unless it's happening earlier on when the file is read in by the parseCsv function that uses the "paparse" package https://github.com/nextstrain/auspice/blob/master/src/actions/filesDropped/parseCsv.js.

**Your environment: if running Nextstrain locally**  
 (Bryan's environment)
 - Operating system: Windows
 - Browser:  Edge
 - Version (e.g. `auspice 2.7.0`):  

**Additional context**  
The behavior of parsing csv files dragged and dropped onto auspice could have been changed in #1386 in a way that might be relevant here.

joverlee521:
This sounds like this could be a weird parsing issue with xlsx here: https://github.com/nextstrain/auspice/blob/master/src/actions/filesDropped/metadata.js#L15

We can tinker with `XLSX.read` function's many [parsing options](https://github.com/SheetJS/sheetjs#parsing-options). 
The `raw` option sounds the most promising as the docs note:
> Excel aggressively tries to interpret values from CSV and other plain text. This leads to surprising behavior! The raw option suppresses value parsing.

genehack:
Given the referenced closed PR, I'm going to close this issue too.

PR #1401, opened 2021-09-18
XLSX.read use raw option to avoid parsing issue;

eharkins:
addresses #1395 using @joverlee521's suggestion: https://github.com/nextstrain/auspice/issues/1395#issuecomment-922033712. More testing needed.


joverlee521 (approved):
Thank you for getting this done @eharkins! This seems to work as expected with some bogus data I created to test with the zika build ([test-drag.csv](https://github.com/nextstrain/auspice/files/7206073/test-drag.csv)) 

I will note that real dates are no longer parsed into uniform formats and keep their raw values (as expected). I think this is fine and it should be on the user to format the data how they want it to be displayed.

Issue #1036, opened 2020-04-01
Add node `name` to the deeplink URL


sidneymbell:
**Context**  
As I'm writing the weekly situation reports, it would be a *huge* help to be able to deep link to a view of the tree that's zoomed into any arbitrary node, rather than having to define a specific clade label to highlight some portion of the data. I currently waste a lot of time figuring out how to hack around this :)

**Description**   
Currently, we use deep linking to enable sharing of a specific view of the data. I'd like to add one additional field to this to enable sharing a view that's zoomed to any specific node in the tree. This change doesn't need to be in the UI, just the URL.

**Example**   
I don't know how the zooming is currently implemented in the codebase, but I'd imagine one option could be something like:

1 - I click on a node to zoom into it  
2 - Under the hood, that node's `name` attribute (unique, already included in the JSON) is added to state. Ex: `{'node_zoom': 'NODE_0002547'}`  
3 - That node ID is added to the URL, something like `nextstrain.org/ncov?node_zoom=0002547`
4 - I then share that URL, and this replicates the app's zoom state for the next viewer.

sidneymbell:
I'm guessing this is at least theoretically possible, given that zooming is available via 
`https://nextstrain.org/ncov/2020-04-01?label=clade:A2a`, etc, although this might actually be acting as a filter / would maybe require a flag assigned to each tip. Not sure if this would be a perf issue.

jameshadfield:
As I said on slack, I really want to avoid this, but I do have an alternative. Comments below adapted from slack messages:

I designed zoom not to be able to zoom into an internal node via the node name as it changes between builds -- NODE_0002547 will represent different clades. This is dangerous, as the URL hasn't changed & these get shared. (I realise the narrative datasets we are using for nCoV sit-reps will not change, but that's not typical.)

A potential auspice solution would be something along the lines of allowing clade zooming via `?label=mrca(<tipName1>,<tipName2>)`.

A non auspice solution would be to manually add a [branch label](https://github.com/nextstrain/augur/blob/master/augur/data/schema-export-v2.json#L399) to the nodes you want to zoom into for each narrative. Or write a script to label every single internal node for these datasets with a label of "internal_node"  which will then be usable via `?label=internal_node:X` (be aware this would increase the JSON size, but perhaps not too much with compression). You can choose the label key & value (i.e. it doesn't have to be "clade"), and you could hide it from view.

sidneymbell:
[...]

PR #1403, opened 2021-09-21
[tree] button to go back to previous zoom

jameshadfield:
The ability to return to the previous zoom level is useful when
exploring trees in detail, especially when combined with the
magnifying glass (which zooms out).

A hover info box has been added to convey the functionality of this
icon.

The implementation involving multiple trees is conceptually more
complicated, as I don't want us to return to the previous zoom state
in each tree. (This is similar to the magnifiying glass icon, which
zooms out in both trees.) The best solution is probably tree-specific
UI. For now, this functionality is only surfaced if we are viewing a
single tree.

### DIFF
--- a/src/actions/tree.js
+++ b/src/actions/tree.js
@@ -49,7 +49,7 @@ export const applyInViewNodesToTree = (idx, tree) => {
  * @return {function} a function to be handled by redux (thunk)
  */
 export const updateVisibleTipsAndBranchThicknesses = (
-  {root = [undefined, undefined], cladeSelected = undefined} = {}
+  {root = [undefined, undefined], cladeSelected = undefined, revertTreeZoom = false} = {}
 ) => {
   return (dispatch, getState) => {
     const { tree, treeToo, controls, frequencies } = getState();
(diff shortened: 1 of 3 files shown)


trvrb:
Hmm... very interesting idea. However, I have to admit, I'd prefer it if the browser back button could just be a general purpose UI to go back to previous visualization state (whether it's tree zoom or tree layout or whatever). Although, given that tree zoom is not usually encoded in URL, I see how this is difficult.

I do think that the previous idea of tagging branches in a dataset specific fashion (https://github.com/nextstrain/auspice/issues/1036) could interact well here. However, in the interim I'd support this small "go back" icon.

Also, the hover tooltip is really helpful. It would be nice if this existed for the other top-right panel buttons.

corneliusroemer:
This is really cool! Love it.

One thing I'm not sure about: should _zoom out_ via lens count as a step that is reversed by the back button? Since it's easy to undo zoom out by just clicking into the branch and it feels that zooming out goes in the same direction as the back button I'd potentially prefer go back to the state where I was before I clicked into a specific branch. Does that make sense? It's a small detail, not very important but maybe others have the same thought.

Also: since it's sometimes quite hard to find the right branch to click into, would it make sense to add a forward button that goes back in to where we started off before clicking the back button? Again, less important than back button but maybe something to think about.

src/actions/tree.js
```diff
@@ -111,6 +124,15 @@ export const updateVisibleTipsAndBranchThicknesses = (
       visibilityToo: dispatchObj.visibilityToo
     });
 
+    /* update the history stack for zoom levels (if we are zooming) */
+    if (tree.idxOfFilteredRoot !== dispatchObj.idxOfInViewRootNode) {
```
joverlee521, line 128: Should this line be 
```js
if (tree.idxOfInViewRootNode !== dispatchObj.idxOfInViewRootNode) {
```
Currently the revert button doesn't always work after clicking the "zoom to selected" button and I believe this is the cause.

joverlee521 (commented):
This is great! I agree that this function seems like it should work separately for each tree if there are two trees. I think the biggest question there is where those buttons should go on the UI 🙃 

I could also see a use for the suggested "forward" button, but maybe we can come back to that when zooming is added to the URL. 

If I'm not misunderstanding the history stack, I think I found a small bug. See comment below.

src/actions/tree.js
jameshadfield: Yes! It should. Thanks :)

PR #1408, opened 2021-09-27
narratives: Set max-width on images in the main (right-hand) panel

tsibley:
This prevents wide or high-resolution images from exceeding the panel's
width, which is important for visibility, especially since the main
panel doesn't scroll horizontally.


jameshadfield (approved):
Thanks @tsibley!

PR #1412, opened 2021-10-06
Add options to edit min/max dates

### LINKED ISSUE #1406: Date sliders have minimum span making reasonably narrow time slice impossible
I would sometimes like to zoom into the very most recent sample dates. Unfortunately, the date range slider does not allow me to get in closer than maybe 5-10% of the whole time span. For SC2 trees this is now around 6 weeks, which is quite a lot of time. I'd like to be able to zoom into the most recent 2 weeks. To do this, I currently have to edit the URL to change the date manually. But that's not great since it makes a browser reload necessary.

I can't slide the left hand slider any further to the right than this:
image

I don't know what the best way would be to give the user more control. I presume you chose the minimum slider width on purpose because otherwise it becomes difficult to slide open again. One could add a text field?

victorlin:

Currently, the date range is only customizable by either interfacing with the slider (has limitations) or editing the URL directly (hacky, forces page reload).

This pull request adds options to change min/max dates individually.

Usage:

1. Click on the calendar icon to replace the date label with a browser-native HTML5 date picker.
2. Select or type a new value. Save the value one of 3 ways:
    1. Click the save icon.
    2. Click outside of the browser date picker.
    2. Hit `Enter`.
3. Changing a value via date picker will update the slider + URL params, and vice versa.

Safari:

safari

Brave (Chromium):

brave

Notes:

- Invalid values are disabled by setting [`min`/`max`](https://www.w3schools.com/jsref/dom_obj_date.asp) properties.
- The date picker does not utilize `theme` – [customizability is limited](https://stackoverflow.com/q/14946091) and the style varies by browser.

### Related issue(s)
Fixes #1406  

### Testing

- [ ] add tests


corneliusroemer:
@jameshadfield Could testing simply be done on the end product? I'd be curious to check out what @victorlin came up with but  couldn't immediately find a CI deployment. Did I miss something?

huddlej (commented):
Thank you for contributing this, @victorlin! I tested the date pickers in a local Auspice installation and they generally worked as expected. I think the functionality in this PR addresses the original issue sufficiently to merge now (other code style/implementation reviews from @jameshadfield notwithstanding).

A couple of minor thoughts occurred to me while using the pickers:

 - I wish I could click on the text of the current min/max date itself to activate the date field edit mode. I found myself repeatedly clicking or doubleclicking on the text before remembering I had to click the little calendar icon.
 - I was able to enter invalid dates (like a date later than the max date) by typing it into the date field. The date picker's calendar correctly prevented me from picking a later date though.

Manually entered date in 2020 shouldn't be possible

src/components/controls/date-picker.js
```diff
@@ -0,0 +1,62 @@
...
+        type="date"
+        defaultValue={value}
+        min={minDate}
+        max={maxDate}
+        onKeyUp={onKeyUp}
+        onBlur={onBlur}
+      />
+    );
+    const saveIcon = <IconSpan onClick={() => setEditing(false)}><FiSave /></IconSpan>;
+    content = right ? <>{dateInput}{saveIcon}</> : <>{saveIcon}{dateInput}</>;
```
jameshadfield, line 49: The should be tweaked (or perhaps some other element) so that when editing it doesn't appear on the line below the icon when there's enough space for them to be on the same line. I can change this by removing `max-width: 50%` from `<FloatDiv>` but I'm unsure if this is the best without diving deeper.

For instance: 
image

jameshadfield (changes requested):
This looks great! I have some minor syling requests, and also echo @huddlej's desire for the input to be edible when clicking on the date text itself (with `cursor: pointer` on the text also). 

Re: invalid dates. As you say, this is possible via the URL so it's not a blocker for this PR, however we should fix this in later work.

src/components/controls/date-picker.js
```diff
@@ -0,0 +1,62 @@
...
+      processNewValue(e.target.value);
+    }
+  }
+  function onBlur(e) {
+    processNewValue(e.target.value);
+  }
+  let content;
+  if (editing) {
+    const dateInput = (
+      <input
```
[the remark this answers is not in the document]
victorlin: [wontfix] agreed, there are definitely some limitations with the native date picker but it gets the job done.

src/components/controls/date-picker.js
victorlin: One of the limitations of the native date picker is that [you can't do much styling](https://stackoverflow.com/questions/14946091/are-there-any-style-options-for-the-html5-date-picker) on the input box as it varies per browser. Seems like, at least for some browsers, the parent div containing icon+input is larger than half the width of the sidebar, and something must be pushed down to a second "row". Here's without `max-width: 50%`:

2021-10-19 at 11 52 35 PM

victorlin: For the record, imposing a `max-width` on the `input` gets ugly:

2021-10-19 at 11 58 27 PM

victorlin:
Let's squash and merge this once approved, since the commit history is a bit verbose.

jameshadfield (approved):
Approved (finally 😬 ) pending removal of the merge commit and (optional) squashing some commits together

victorlin:
I just realized this doesn't work with negative dates. The HTML date picker (at least on a couple browsers I've tested) won't let you pick earlier than `0001-01-01`. Passing negative numeric dates by URL works fine though ([example](https://nextstrain.org/tb/global?dmin=-3000&l=rect)).

Would this be appropriate to document as a feature gap somewhere, or should we reconsider implementation?

jameshadfield:
Choosing precise BCE dates is a bit ridiculous, in my opinion, so I don't think it's a problem. Perhaps we could hide the datepicker when the date is BCE?

victorlin:
> Perhaps we could hide the datepicker when the date is BCE?

Good idea, I'll make a new PR for that.

huddlej:
I bet this will make @BryanTegomoh's day when it gets officially released! Thank you for pushing this through, @victorlin and @jameshadfield!

PR #1428, opened 2021-11-10
Limit scope of narratives search in URL

huddlej:

Changes the substring to search a given URL for narratives content to include a leading and trailing forward slash. This change limits the scope of the search in the URL to match paths like https://nextstrain.org/narratives/ and
https://nextstrain.org/community/narratives/ but to avoid trying to load as narratives datasets that happen to have the word "narratives" as a substring of their names (e.g., "nextstrain-narratives-demo_ncov_2020-03-04.json").

The following steps recreate the original issue from a clone of the Auspice repository:

```
curl "https://data.nextstrain.org/ncov_2020-03-04.json" | gzip -c -d > data/nextstrain-narratives-demo_ncov_2020-03-04.json

auspice view --datasetDir data/
```

### Testing

Tested locally with a build named `nextstrain-narratives-demo_ncov_2020-03-04.json` and with a narrative that was working before and after this change to the URL search.

### DIFF
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -52,7 +52,7 @@ export const loadJSONs = ({url = window.location.pathname, search = window.locat
       dispatch({type: types.DATA_INVALID}); // "resets" state
     }
     const query = queryString.parse(search);
-    if (url.includes("narratives")) {
+    if (url.includes("/narratives/")) {
       loadNarrative(dispatch, url, query);
     } else {
       warnDeprecatedQuerySyntax(dispatch, query);


huddlej:
Additional context for this is that I'm developing a training session on how to create Nextstrain narratives including a section on how to upload your own build JSONs to GitHub and source them from the GitHub-hosted narrative. My training materials all use a repository named "nextstrain-narratives-demo" (to match the eventual tutorial name), so the resulting auspice JSON files need to be named like [nextstrain-narratives-demo_ncov_2020-03-04.json](https://github.com/huddlej/nextstrain-narratives-demo/blob/main/auspice/nextstrain-narratives-demo_ncov_2020-03-04.json) to get properly loaded from the community interface.

Without the fix in this PR, loading the JSON as it is named above produces the following error in the developer console:

image

I'm not sure if the solution in this PR is correct, but I do think searching the full URL for "narrative" is a problem.

tsibley (approved):
> I'm not sure if the solution in this PR is correct, but I do think searching the full URL for "narrative" is a problem.

Yes, 100% a problem. This change is a clear improvement and will reduce the occurrence of this bug, though not fully eliminate it (as I'm sure you're aware).

To eliminate it, we'd need a mechanism other than the URL path to switch Auspice between dataset and narrative mode. One way, for example, would be producing two entrypoints (HTML files) that a server like nextstrain.org can switch between serving. This seems preferable to other ways (e.g. serving a JS-readable HTTP cookie) because it is compatible with other server modes, like in the context of static site generation.

Relatedly, we also have [other latent bugs due to similar URL-matching](https://github.com/nextstrain/auspice/blob/2677b768f60913bb9b81a4e28799a636192c2850/src/components/framework/footer-descriptions.js#L205-L241).

tsibley:
BTW, you can see this fix integrated into nextstrain.org and verify it works at: https://nextstrain-s-auspice-pr-qlvm52.herokuapp.com/community/huddlej/nextstrain-narratives-demo/ncov/2020-03-04/

jameshadfield (approved):
Thanks John!

PR #1434, opened 2021-11-17
Replace TravisCI with GitHub Actions

victorlin:
Context: for simplicity, we are migrating [.travis.yml](https://github.com/nextstrain/auspice/blob/220e012f20d4ee22a128d77e9bd3eb7785728503/.travis.yml) to the current GitHub workflow.

- [x] Add build job to GitHub workflow
- [x] Standardize node setup across jobs:
    - Use v2 of `actions/setup-node`
    - Use versions=12,14,16 since [10 has already reached End-of-Life status](https://github.com/nodejs/Release#end-of-life-releases).
- [x] `npm publish` on `release` branch runs only (can't test this until we do a new release - 2.32.2?)
    - better yet, run on [GitHub release](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release) if those will be created anyways. This way we can drop the need for a `release` branch. Maybe in another PR.
- [x] Update [status badge in README](https://github.com/nextstrain/auspice/blob/220e012f20d4ee22a128d77e9bd3eb7785728503/README.md?plain=1#L1-L2)

References:

- [GitHub Actions | Publishing packages to the npm registry](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry)
- [NPM | Creating and viewing access tokens](https://docs.npmjs.com/creating-and-viewing-access-tokens)


victorlin:
Note that TravisCI should be disabled once this is merged, otherwise there will be two instances running `npm publish` during the next release.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,20 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+          registry-url: https://registry.npmjs.org/
+      - run: node --version
+      - run: npm install
```
jameshadfield, line 16: I would expect `npm ci` to be better here, as you've used in your additions to `.github/workflows/ci.yaml`. If you agree, could you also change all the existing  `npm install` commands there to `npm ci`? 

I notice that [the GitHub](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry) docs use `npm install` here, so there may be a reason behind this.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - run: npm run rebuild-docker-image
+      - run: npm publish
```
jameshadfield, line 18: What's your advice for preventing releases where the tests fail?

The only "test" TravisCI ran before release was `npm run build`, and this will happen here as well since building is part of our `prepare` script. We already run some tests in `.github/workflows/ci.yaml`, but this publish action isn't dependent on that CI passing.

jameshadfield (changes requested):
hanks @victorlin!

> Note that TravisCI should be disabled once this is merged, otherwise there will be two instances running npm publish during the next release.

I think this PR should remove the Travis config (which, as far as I understand things, will stop Travis from running). 

Finally, can we update the "Deploying to npm" section in the README to indicate the new approach?

.github/workflows/publish.yaml
victorlin: You're right, I think `npm ci` is the move for anything running on build servers. Based on [the official doc](https://docs.npmjs.com/cli/v8/commands/npm-ci), I think this is the main difference/benefit for us:

> If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.

I've replaced `npm install` with `npm ci` in both YAML files: https://github.com/nextstrain/auspice/pull/1434/commits/cc8fa32bf059340946bcfc769600a9d9a6f662be

.github/workflows/publish.yaml
victorlin: Good point, I think the solution is to trigger on [`workflow_run`](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_run) of the `ci` workflow. Implemented here: https://github.com/nextstrain/auspice/pull/1434/commits/595fd23cbe7ccba049b16a26d4487e7119a66938

victorlin:
@jameshadfield

> I think this PR should remove the Travis config (which, as far as I understand things, will stop Travis from running).

Done: https://github.com/nextstrain/auspice/pull/1434/commits/1355a84a30114521b1f4e3feca4ba6cd4cb66a4e

> Finally, can we update the "Deploying to npm" section in the README to indicate the new approach?

Done (nothing in README but I think this is the right place): https://github.com/nextstrain/auspice/pull/1434/commits/8282bdf6dd61551b96da14809dd5c1a2822504cb

.github/workflows/publish.yaml
victorlin: So reading a bit deeper, this stands out:

> This event will only trigger a workflow run if the workflow file is on the default branch.

I'm not sure if it means this won't run properly on `release` since it's not the default branch, but in our use cases `release` and `master` (default) will always be even at runtime, meaning the workflow file does exist on the default branch... only way to find out would be to do a test release (2.32.2?)

If this doesn't work, I'd suggest getting rid of `release` and instead triggering on the [GitHub release](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release) event which can be based on `master`.

tsibley: For `workflow_run` events, I believe their state is only ever `requested` or `completed`, with the latter including both successful and failed runs. That's not quite what we want here.

However, instead of having "publish" be a separate workflow, I suggest moving it into the "ci" workflow as a final job that conditions on the success of the test jobs and the current branch being `release`.

victorlin: @tsibley good point. Implemented here: https://github.com/nextstrain/auspice/pull/1434/commits/64268970677f990ebefd0f57e1510cefb7390414

Reference: [Contexts - GitHub Docs](https://docs.github.com/en/actions/learn-github-actions/contexts#determining-when-to-use-contexts)

There's even a nice visual on the [GitHub workflow summary page](https://github.com/nextstrain/auspice/actions/runs/1478161125):

auspice CI graph

jameshadfield (approved):
Looks good to me. I'd propose we merge now and then make an auspice release early next week. (I think the date-sliders PR is good to go, but I need to double check, so that could be the impetus for a new release.)

PR #1376, opened 2021-07-15
update packages via npm audit

jameshadfield:
Only remaining warnings are related to development mode or
`auspice build` deps, not client-facing packages.

PR created to test heroku building here & on nextstrain.org. Will merge if they build successfully.

PR #1436, opened 2021-11-23
Update npm packages and compatible node versions

victorlin:
Upgrading and pruning npm packages while retaining functionality.

## Detail

1. Removed some unused packages: 3c0be6cb4e55dd2b57ac4a42ca3bcd81058497a9, c6ba3a2aa946b9bacdb57488ea4578ff30de881b
2. `npm audit fix` updated several packages: 95b009d08f3b2d41e1b70e44bdbf6caed7ec3349
3. That somehow failed to resolve package dependencies. Fixed by regenerating package-lock.json: 21aff3a8d84a2f79dfff808360bcaf87f2e2ec91
4. This then surfaced issues with `babel-eslint` being out of date. Fixed by replacing `babel-eslint` with `@babel/eslint-parser`: ef915607488dd563a0a5987a7f8053ba2a5929d5
5. This then surfaced a few previously uncaught linting issues. Fixed/suppressed in ef915607488dd563a0a5987a7f8053ba2a5929d5, f0af38d643463effdd17674a327f89ac4853c1cb, and 5154823694ff9b1cf8a329ba89d56916a5b1d47a.
6. Manually updated PapaParse 4.3.5 -> 5.3.1 to fix one "high vulnerability": 1e1b91357eacb55b1f84973d9bb50cb0a54ae856
7. Other things non-npm:
    1. Support node engine versions 12, 14, 16: 645fbdf8de893053a4518c471a9206e898bd77a3

## Caveats

- `npm audit` still shows some (but less) vulnerabilities. These are probably fine for now since [`npm audit` is an ongoing pain](https://overreacted.io/npm-audit-broken-by-design/).
    - Not upgrading `webpack` v4 -> v5. I got errors and it seems nontrivial. [Official guide](https://webpack.js.org/migrate/5/)
        - Notes for uncommitted progress:
            - Replace `node.fs: 'empty'` with `resolve.fallback.fs: false`
            - Update `compression-webpack-plugin` 3.x -> 9.x:
                - Use `[base]` per [version 6.0.0 breaking change](https://github.com/webpack-contrib/compression-webpack-plugin/blob/master/CHANGELOG.md#600-2020-09-14)


jameshadfield (changes requested):
I think these changes are desirable and we should pursue adding them to auspice. However when building using a customised version for auspice -- e.g. [this nextstrain.org review app](https://github.com/nextstrain/nextstrain.org/pull/437) auspice doesn't work. The following is printed to the console:

```
 ChunkLoadError: Loading chunk 5 failed.
(error: https://nextstrain-s-auspice-pr-iybmbe.herokuapp.com/dist/auspice.chunk.5.bundle.56a2bd4dfbf23336129e.js)
```

https://github.com/nextstrain/auspice/pull/1376 attempted to do something similar and ran into the same (or similar) issue.

victorlin:
Pushing and force-removing a commit to trigger new workflow for nextstrain.org PR ([cannot re-run current one](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs) since it is more than 30 days old)d

victorlin:
Force-pushing `git rebase -i master`

victorlin:
@jameshadfield there was an issue with that review app using node=13 which might have complicated things with this PR. It's now using node=14 after nextstrain/nextstrain.org#455. Can you check again?

I also rebased onto `master` and fixed a newly surfaced lint error with the date picker.

jameshadfield (approved):
Yes - the review app from https://github.com/nextstrain/nextstrain.org/pull/437 is now working, and I've build that PR locally as well. 

Do you understand why node13 wouldn't correctly build nextstrain.org + this version of auspice?

victorlin:
Awesome!

> Do you understand why node13 wouldn't correctly build nextstrain.org + this version of auspice?

I wasn't able to repro your `ChunkLoadError` but saw something different on the TravisCI page. Unfortunately we can't see exact CI error on the [review PR checks page](https://github.com/nextstrain/nextstrain.org/pull/437/checks) since it's been force-pushed over.

More broadly, node13 reached EOL 1.5 years ago so maybe something came up that was incompatible. I would assume that it could still use node12 though, not sure what the issue was.

PR #1439, opened 2021-12-15
Remove scrollbar in PanelsContainer when no scrolling is possible

### LINKED ISSUE #1438: Scrollbar still showing in panels not needed
**Current Behavior**  
 When then window is larger than the content, the scrollbar area will still be showing. This is very evident when the page is embedded in an `iframe`, creating a not-so-nice look of having two scrollbars next to each other when it's not really needed.

image

**Expected behavior**  
The scrollbar is hidden when the content cannot be scrolled.

image

**How to reproduce**  
1. Go to <https://www.gisaid.org/phylodynamics/global/nextstrain/>
2. Open up developer tools in the browser
3. Navigate to the `iframe` component (in `#normalcontent`)
4. Append `&d=tree` or `&d=frequencies` to see view that needs no scroll. That is, change the attribute of `iframe src` to [`https://nextstrain.org/ncov/global?branchLabel=none&c=GISAID_clade&onlyPanels&p=full&sidebar=closed&d=tree`](https://nextstrain.org/ncov/global?branchLabel=none&c=GISAID_clade&onlyPanels&p=full&sidebar=closed&d=tree)
5. See the error as in the first screenshot

**Possible solution**  
Change the stylesheet [`src/components/main/styles.js`](https://github.com/nextstrain/auspice/blob/master/src/components/main/styles.js) from `overflow-y: scroll` to `overflow-y: auto`.

artt:
Removes scrollbar in the `PanelsContainer` component when the window's height is enough to show the entire content.

### Related issue(s)
Fixes #1438  

### Testing
Build from this fork, go to `/ncov/gisaid/global?branchLabel=none&c=GISAID_clade&d=tree&p=full&sidebar=closed` with adequate window height (> 700px should be fine) and the scrollbar should be hidden. The scrollbar should be shown when scrolling is needed.

### DIFF
--- a/src/components/main/styles.js
+++ b/src/components/main/styles.js
@@ -11,7 +11,7 @@ export const PanelsContainer = styled.div`
   height: ${(props) => props.height+"px"};
   width: ${(props) => props.width+"px"};
   overflow-x: hidden;
-  overflow-y: scroll;
+  overflow-y: auto;
   left: ${(props) => props.left+"px"};
 `;
 


jameshadfield:
Scrollbars are something that I'd like to improve in Auspice - they are [automagically hidden on MacOS by default](https://www.nytimes.com/2017/02/02/technology/personaltech/solving-the-case-of-the-disappearing-scroll-bars.html) which is where I do most of my development. 

I've tested this change (by restricting the panels to "entropy" only and also using narratives, as well as reading [these docs](https://css-tricks.com/almanac/properties/o/overflow/)) and can't find any regressions with it, so thankyou @artt! It's also a good reminder to me to fix the scroll bars in a couple of other places.

PR #1442, opened 2021-12-22
Multitree

jameshadfield:
This PR implements two major concepts:
1. The ability for multiple trees to be defined via `dataset.tree -> Array<Tree>`
2. The ability to cut a tree by a user-definable trait and display these as multiple (sub-trees). AKA Exploded trees, see screen grab 👇 
3. Introduces the concept of defining major / minor parents of nodes (see below).

In addition it includes the removal of a lot of technical debt within Auspice's code; as such extensive testing will be required. See commit messages for more information, however I'll use this PR description to foster discussion.

https://user-images.githubusercontent.com/8350992/147037611-df800607-af9e-4218-a9a7-8eec1a5cd5be.mov

### To-dos in this PR (Updated 2022-01-30)
(Not all of these need to be done in this PR, mind.)

- [ ] add "beta" in the dropdown UI for the first release
- [ ] store exploded trait in URL state
- [x] consider how to treat undefined (internal) traits, e.g. currently nodes (X -> undefined -> Y) would not be exploded. _Update: we break on any trait change, including to/from `undefined`; a subtree with no tips (e.g. one internal node with an undefined trait) is no longer displayed._
- [x] multiple trees (i.e. two different tree datasets) are not considered here & have not been tested (much). _Update: I've disabled the ability to use both features at once._
- [ ] remove test dataset commit & some console logs
- [ ] add connecting lines to indicate major/minor parents (see below)
- [ ] add more space between subtrees (in rectangular, unrooted & radial layouts)
- [ ] unrooted layout / clock / scatter layout need attention here (I haven't tested much, but they're definitely misleading at the moment)
- [x] consider the case of orphaned branches (i.e. those whose children have all become subtrees). Currently these are drawn without a tip, but they do have a tip label (actually the node name, e.g. "NODE_1234"). _Update: these are no longer rendered._
- [x] branch thicknesses aren't quite right
- [ ] Filtering may need to change & is different depending on the order it is applied (i.e. explode first vs explode last). This is due to the traversals operating on different trees (as noticed by @huddlej, see below)
- [ ] Zoom-to-selected depends on the MRCA of selected tips, and thus is tips are selected in multiple subtrees the MRCA is `__ROOT` and thus we can't zoom to selected!  (as noticed by @huddlej, see below)
- [ ] branch names are sometimes dissapearing

### To-dos in subsequent PRs (Updated 2022-01-30)

- we should limit the explode options to those colourings which are defined on internal nodes / would not result in the currently displayed subtrees (currently a warning is shown)
- genotype should be available to explode on

### Testing (updated 2022-01-30)
I've tested this on a bunch of (local) datasets, although I've only created two multi-tree JSONs.

* Use the review app -- https://auspice-multitree-h51yyzgffb90.herokuapp.com/ -- to test a variety of datasets
* A multi-tree datset  has been created from the seattleflu's h3n2/1y dataset. https://auspice-multitree-h51yyzgffb90.herokuapp.com//test-multiple-subtrees
* A small toy dataset may be useful as well: https://auspice-multitree-h51yyzgffb90.herokuapp.com/test-simple-exploded-tree

### Descriptions of major/minor parents

Commit 1d5f725945e46d40e5109ffb6644cd9d269f8e52 introduces the concept of defining major / minor parents of nodes. This may include where in another tree this (sub-)tree originated (referred to here as “major”) or a recombination donor (“minor”). In the future, we may use solid/dashed lines to represent the connection to these parent nodes, using   @mlewinsohn, @miparedes, Russell and @cassiawag's [recombination project](https://cse512-21s.github.io/FP-Pathogen-Phylogenies/) as inspiration. This commit only implements this concept for exploded trees (as we need to know how long the root stem should be). I'd like to make this  user definable via a `branch_attrs.parent` annotation, the details of which haven't been fully worked out. The following situations are possible:

(i) Single tree, not exploded: Minor parents (via  `branch_attrs.parent`) may be used to visualise recombination events (see above). Major parents are not applicable.
(ii) Multi tree, not exploded: Major parents (via  `branch_attrs.parent`) can indicate the placement of the subtree (in another tree). Minor parents may also be applicable here.
(iii) Single tree, exploded:  `branch_attrs.parent` is ignored. We set the major parent of each subtree to be the branch it originated from.
(iv) Multi tree, exploded: Same as (iii).

It'd be great to discuss this further before implementing it in a user-configurable fashion. 

cc @evogytis 
🙏 many thanks to @frogsquire who's work in #1105 was really helpful here


huddlej:
This is so cool, @jameshadfield! I don't know if I'm using this new feature correctly yet, but I noticed a couple of odd behaviors in an H3N2 tree:

 - [explode by region and then filter to region (e.g., Africa) still shows branches for node not assigned to that region](https://nextstrain-s-auspice-pr-ca983t.herokuapp.com/flu/seasonal/h3n2/ha/2y?c=region&f_region=Africa)
 - [explode and color by clade and then filter to two adjacent clades in the tree does not support zooming to selected](https://nextstrain-s-auspice-pr-ca983t.herokuapp.com/flu/seasonal/h3n2/ha/2y?f_clade_membership=3C.2a1b.2a.2,3C.2a1b.2a.1) (maybe this is expected behavior though?)

tsibley:
@jameshadfield I couldn't find an example dataset file with multiple trees in the links above (I think most are currently inactive/dead?), but I'm curious if this would mean a new major version of the dataset schema, i.e. to dataset v3?

jameshadfield:
> Just a heads up that the review app does not seem to be working in any of the test links provided.

> I couldn't find an example dataset file with multiple trees in the links above 

Yeah, the heroku review app died after some amount of time (a week?) and when it gets recreated the URL changes so those links were dead. I've updated them in the above message.

> but I'm curious if this would mean a new major version of the dataset schema, i.e. to dataset v3?

I don't think so, however the schema will have to be extended to allow both a single tree and an array of trees, as well as allowing some extra `branch_attrs` if that capability gets implemented in this PR.

jameshadfield:
This is now (finally) ready for some release, however I've added an "experimental" label to the dropdown as there are this pushes auspice in directions which expose some pre-existing limitations. There may also be some bugs / papercuts as this PR touches a lot of the different parts of the code. 

* "Zoom to selected" fundamentally doesn't play nicely when viewing subtrees, due to the way we calculate the links between in-view tips. We'll revisit this concept shortly with the accordion zoom & other ways of zooming the tree.
* Connecting lines (e.g. between the subtree root and it's parent in the unexploded tree) are for the next PR, however the tree layout knows about this link as it's used to compute the subtree stem length.
* Augur needs a corresponding PR to allow the tree to be an array, however this is not blocking. 
* URL state is not implemented, as I wanted to wait for more testing here
* Subtrees are separated in Radial & Unrooted views, however this separation is expressed as a change in angle so if each subtree root has a similar origin divergence/time, then there will be little to no separation. Plenty of work can be done here (see commit messages for more). 

P.S. The final commit is to be dropped before merge.

PR #1448, opened 2022-01-26
Add a "finePrint" extension hook

tsibley:
Provides a bit of Markdown to render in the fine print at the bottom of
pages, including the splash page and dataset pages.

This will make it much easier to folks customizing Auspice to add their
own attributions, links to source code, etc.

### Testing
Tested locally by building Auspice with and without configs that define `finePrint`.


joverlee521 (approved):
Easy win for a simple custom fine print!

Issue #1444, opened 2022-01-17
Add reversions/back-mutations as within-Auspice-computed branch label


corneliusroemer:
**Context**  
Reference backfilling is a big problem in SARS-CoV-2 sequences. All the information one needs to identify reversions back to reference is included in the auspice.json. This would for example allow me to quickly check that a Nextclade reference tree doesn't contain any reversions.

**Description**   
As a user, I would like to be able to see nucleotide reversions (either only to reference, or to any previous state) be highlightable on the tree. For example as a branch label, like we do with clades or sometimes Spike mutations.

**Examples**  
[Usher](https://genome.ucsc.edu/cgi-bin/hgPhyloPlace) already implements this feature, they must do it in the backend, so there's clearly some interest in this feature beyond me.
![image](https://user-images.githubusercontent.com/25161793/149780974-27c2b2cd-f6e8-45a4-aab6-1a50811ae16c.png)

**Possible solution**   
I could write a custom Python script that post-processes an auspice.json to add this as a branch annotation. But it's silly to do this with a script when it could be implemented within auspice.json for all trees, for all users.

jameshadfield:
Thanks @corneliusroemer  — I completely agree and think this feature will immensely help with interpreting trees, especially Omicron. I’m going to expand this issue slightly to encompass changes we've discussed regarding display of mutations more generally. 

**Current situation for mutation display**
Currently dataset JSONs report mutations on a branch per-nucleotide and per-gene. This data typically comes from `augur ancestral` and `augur translate`, respectively, although for nCoV we are using nextclade for the AA changes. Whether Ns are included is influenced by parameters to those augur commands. The JSON structure looks like so:

```json
"branch_attrs": {
    "mutations": {
        "nuc": [ "T1N", "T2N", ...],
        "S": ["T716I"]
```

The tooltips used in auspice behave as follows:

| tooltip            | mutations shown           | Ns 			   | gaps	(deletions) | insertions |
| ---                | ---                       | ---             | ---              | ---        |
| branch / tip hover | subset listed, Ns ignored | count shown     | treated as mut¹ | N/A² |
| branch shift+click | all shown, incl gaps      | treated as muts | treated as mut¹ | N/A²  |
| tip click          | all listed w.r.t. root³   | click to copy list | treated as mut¹ | N/A²  |

¹ No grouping is performed, e.g. if we have deletions of pos a, a+1, a+2 then we report three "mutations" (to `-`). 
[...]

PR #1449, opened 2022-01-28
categorise mutations

jameshadfield:

See #1444 for context & commit messages for further info. Or just look at the examples 😉 

### Testing
Unit tests added, as well as interactive testing

### Examples

**Branch Hover**
I've grouped nucleotide mutations on branch hover, as I find it super useful for exploring the tree to see if mutations are homoplasic or reversions to the root (i.e. potentially backfilled). In the interest of keeping the size of the info-box manageable, I've kept the AA mutations unchanged. 

image

---

**Branch (shift+)click**
Branches group mutations into unique vs homoplasies vs gaps vs Ns. They also report which are reversions to the root. Gaps and Ns are grouped into intervals.
image
image

---

**Tip hover**
When hovering over tips we no longer report the mutations on the branch leading to the tip, rather we report a summary of changes w.r.t the root. If you want to see the mutations on the branch leading to the tip, then you can still hover on the branch to see them.

image

---

**Tip click**
When clicking on a tip we list those changes between the root and this sample:
image

### Questions

* Should we restrict homoplasies to those which occur on non-terminal branches? 
    * If so, if viewing a terminal branch with mutation X, which occurs on another non-terminal branch Y, would we classify this as a homoplasy? (we wouldn't when viewing branch Y)
* Branch on-hover for AA mutations is unchanged - is there interest in grouping these? If so, how can we do it without using too much space?

### Further work

I purposefully didn't spend a huge amount of time tweaking the aesthetics of the changes here, as I plan to incorporate the nextclade / CoVariants badges shortly.

### DIFF
--- a/src/components/tree/infoPanels/hover.js
+++ b/src/components/tree/infoPanels/hover.js
@@ -4,9 +4,10 @@ import { numericToCalendar } from "../../../util/dateHelpers";
 import { getTipColorAttribute } from "../../../util/colorHelpers";
 import { isColorByGenotype, decodeColorByGenotype } from "../../../util/getGenotype";
 import { getTraitFromNode, getDivFromNode, getVaccineFromNode,
-  getFullAuthorInfoFromNode, getTipChanges } from "../../../util/treeMiscHelpers";
+  getFullAuthorInfoFromNode, getTipChanges, getBranchMutations } from "../../../util/treeMiscHelpers";
 import { isValueValid } from "../../../util/globals";
 import { formatDivergence, getIdxOfInViewRootNode } from "../phyloTree/helpers";
+import { parseIntervalsOfNsOrGaps } from "./MutationTable";
 
 const InfoLine = ({name, value, padBelow=false}) => {
   const renderValues = () => {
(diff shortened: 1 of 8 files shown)


src/components/tree/infoPanels/hover.js
```diff
@@ -137,46 +139,74 @@ const ColorBy = ({node, colorBy, colorByConfidence, colorScale, colorings}) => {
...
-    const nUnknownMutations = mutations.nuc.filter((mut) => isMutUnknown(mut)).length;
-    if (nUnknownMutations) {
-      elements.push(<InfoLine name='Ns:' value={nUnknownMutations} key="Ns"/>);
+    if (categorisedMutations.nuc.gaps.length) {
+      const value = `${parseIntervalsOfNsOrGaps(categorisedMutations.nuc.gaps).length} regions, ${categorisedMutations.nuc.gaps.length}bp.`;
+      elements.push(<InfoLine name='Gaps:' value={value} key="nuc_gaps"/>);
+    }
+    if (categorisedMutations.nuc.ns.length) {
+      const value = `${parseIntervalsOfNsOrGaps(categorisedMutations.nuc.ns).length} regions, ${categorisedMutations.nuc.ns.length}bp.`;
+      elements.push(<InfoLine name='Gaps:' value={value} key="nuc_ns"/>);
```
jameshadfield, line 209: Typo, should be `name='Ns:'`

PR #1452, opened 2022-02-03
Measurements Panel

joverlee521:
Please see commits for details. There's a lot, so commit by commit code review is recommended. 

A couple improvements/features that have been discussed that I want to address in the future:
1. Revamp panels' layout for grid view. (see Slack discussions [1](https://bedfordlab.slack.com/archives/C2G8C5LUD/p1643246971008600), [2](https://bedfordlab.slack.com/archives/C2G8C5LUD/p1643396881000989))
2. URL parameters for the measurements panel options
3. Ordering of subplots by field or calculation across fields (see [Slack discussion](https://bedfordlab.slack.com/archives/C2G8C5LUD/p1643305652009300?thread_ts=1643246971.008600&cid=C2G8C5LUD))
3. 2 x group-bys (see [Slack discussion](https://bedfordlab.slack.com/archives/C2G8C5LUD/p1642068645003300))


joverlee521:
Temporarily edited `scripts/get-data.sh` to get the measurements JSONs from [my testing repo](https://github.com/joverlee521/nextstrain-testing) for testing with the auspice review app. Will drop commit before merging PR.

src/components/measurements/measurementsD3.js
```diff
@@ -223,7 +230,57 @@ export const drawMeasurementsSVG = (ref, svgData) => {
...
+};
+
+export const drawMeansForColorBy = (ref, svgData, treeStrainColors) => {
+  const { xScale, groupedMeasurements } = svgData;
+  const svg = select(ref);
+  // Re move all current color by means
+  svg.selectAll(`.${classes.colorMean}`).remove();
+  // Calc and draw color by means for each group
+  groupedMeasurements.forEach(([groupingValue, measurements]) => {
+    // For each color-by attribute, create an array of measurement values and keep track of color
```
joverlee521, line 256: While playing around with the review app, I realized this doesn't work well for color-by's with a continuous scale.  I think I have to group by the legend values rather than the specific strain attribute values.

jameshadfield: Nice find. The legend is the only (?) place which performs this binning, but I've wanted to use this binning to create a set of normal auspice filters, which I think is exactly the approach you're proposing for the group-by here. Currently we ignore continuous traits for the filters, and that approach would be perfectly acceptable for this PR if you don't have time to use the legend binning.

src/components/measurements/index.js
```diff
@@ -59,10 +63,27 @@ const treeStrainPropertySelector = (state) => {
...
+      // for visible strains will be displayed
+      if (isVisible(tree.visibility[index])) {
+        /*
+        * If the color scale is continuous, we want to group by the legend value
+        * instead of the specific strain attribute in order to combine all values
+        * within the legend bounds into a single group.
+        */
+        let attribute = getTipColorAttribute(node, colorScale);
+        if (colorScale.continuous) {
+          colorScale.visibleLegendValues.forEach((legendValue) => {
```
jameshadfield, line 76: >  This does mean we have to loop through all of the visible legend values for each visible strain in the tree to determine the matching legend value, which is definitely not ideal.

Two potential speed-ups here are to (a) use a `for` loop and `break` when you have a legend match and (b) add an `else` statement and only set `attribute = getTipColorAttribute(node, colorScale);` within that. You'd have to time things to really work out what gain you get here, but at least (a) seems like it would be beneficial.

tsibley: You could also use the short-circuiting `Array.find()` method, e.g. something like

```
const legendValue = colorScale.visibleLegendValues
  .find(legendValue => determineLegendMatch(legendValue, node, colorScale));

if (legendValue) attribute = legendValue;
```

tsibley:
Minor, but I noticed that the fonts in the axes didn't match the rest of the site, so I quickly tried and pushed 3b3a5c5 to fix it. Please feel free to adjust/tweak as desired!

Before/after:
image
image

src/components/main/index.js
```diff
@@ -145,8 +159,35 @@ class Main extends React.Component {
...
+          }
+          {this.props.panelsToDisplay.includes("measurements") && this.props.measurementsLoaded ?
+            <Suspense fallback={null}>
+              <Measurements
+                width={this.inGrid() ? grid.width : full.width}
+                height={this.inGrid() ? grid.height : full.height}
+                key={keyName+"_measurements"}
+                showLegend={this.shouldShowMeasurementsLegend()}
+              />
+            </Suspense> :
```
tsibley, line 178: I noticed that the measurements panel takes a couple seconds to pop in, so I first see the tree adjacent to the map and then suddenly the map jumps and is replaced by the measurements panel. I think this is due to the condition on `this.props.measurementsLoaded` here and/or the `null` Suspense fallback? Are there alternatives that would be less disruptive during page load?

cli/server/getAvailable.js
```diff
@@ -22,7 +22,8 @@ const getAvailableDatasets = async (localDataPath) => {
       !file.endsWith("_meta.json") &&
       !file.endsWith("_tip-frequencies.json") &&
       !file.endsWith("_root-sequence.json") &&
-      !file.endsWith("_seq.json")
+      !file.endsWith("_seq.json") &&
+      !file.endsWith("_measurements.json")
```
tsibley, line 26: There are a few other places in other repos which will probably want this same change. Should grep, but off the top of my head, at least the Nextstrain CLI and nextstrain.org repos.

src/actions/measurements.js
```diff
@@ -0,0 +1,121 @@
...
+      if (!collection.fields.has(fieldKey)) {
+        collection.fields.set(fieldKey, {title: fieldKey});
+      }
+    });
+
+    // Add jitter and stable id for each measurement
+    collection.measurements.forEach((measurement, index) => {
+      // Generate Gaussian jitter with a Box-Muller transform
+      measurement["measurementJitter"] = Math.sqrt(-2*Math.log(Math.random()))*Math.cos(2*Math.PI*Math.random());
+      measurement["measurementId"] = index;
```
tsibley, line 156: Should add these two reserved keys to the schema (in the Augur repo) as forbidden properties. Alternatively, prefix these here with a reserved char (e.g. `__` to form `__measurementId`) and then forbid any properties starting with that (e.g. `^(?!__).+$`) in the schema.

src/actions/measurements.js
```diff
@@ -0,0 +1,241 @@
...
+     * First add fields from JSON to keep user's ordering
+     * Then loop over measurements to add any remaining fields
+     */
+    const collectionFieldsArray = collection.fields;
+    collection.fields = new Map();
+    if (collectionFieldsArray && collectionFieldsArray.length > 0) {
+      collectionFieldsArray.forEach(({key, title}) => {
+        collection.fields.set(key, {title: title || key});
+      });
+    }
```
tsibley, line 121: I think something like the following is, at a glance, a bit more clear about what's going on here, but YMMV.

```js
    collection.fields = new Map(
      (collection.fields || [])
        .map(({key, title}) => [key, {title: title || key}])
    );
```

tsibley (commented):
Read through commits up to and including ef7ecc2. Checkpointing my comments for the weekend. Will look more next week!

cli/server/getAvailable.js
tsibley: (Not saying you have to do this search, just noting it should get done.)

src/components/measurements/measurementsD3.js
```diff
@@ -0,0 +1,338 @@
...
+  overallMean: "measurementsOverallMean",
+  colorMean: "measurementsColorMean",
+  mean: "mean",
+  standardDeviation: "standardDeviation"
+};
+
+export const svgContainerDOMId = "measurementsSVGContainer";
+const getMeasurementDOMId = (measurement) => `meaurement_${measurement.measurementId}`;
+const domIdRegex = new RegExp("[^\\w\\-_]+", "gi");
+const getSubplotDOMId = (groupingValue) => `measurment_subplot_${groupingValue.replace(domIdRegex, "_")}`;
```
tsibley, line 44: Replacements like this one that collapse a) all forbidden chars into a single replacement char and/or b) runs of forbidden chars into a single replacement char can produce colliding values in edge cases. Since grouping values here are user-provided and not fixed values, it'd be good to make this more robust.

One way is to generate completely synthetic ids for each `groupingValue`, e.g. `measurement_subplot_1`, as is done for measurements. This requires a bit more bookkeeping.

Another way is to choose some CSS selector safe encoding, like base64, instead of replacing chars (base64 in JS [requires a little dancing](https://gist.github.com/tsibley/618f119d1cd3bf3da885170403ebfba7)), e.g. `measurement_subplot_Zm9v4pmlYmFy`.

Stepping back though, it's worth nothing that DOM attributes, including id, support arbitrary strings. So the issue is primarily with supporting arbitrary ids for the CSS selector syntax further down:

```js
const svg = select(ref);
// …
const subplot = svg.select(`#${getSubplotDOMId(groupingValue)}`);
```

If the ids aren't needed for other purposes, then instead of using a CSS selector, this could change to:

```
const subplot = ref.getElementById(getSubplotDOMId(groupingValue));
```

and have `getSubplotDOMId()` stop doing any replacement.

(Also, minor typo of no consequence: `measurment_subplot_` → `measurement_subplot_`)

tsibley (commented):
Ok, back at it this week and read through the remainder of the commits. This looks pretty clean and well put together! Added just one more additional comment. I didn't dig super deep into the particulars of the new panel or Auspice integration, but focused instead on a higher level reading through of the code.

None of my comments are blocking/required changes, though I think there are some worthwhile suggested improvements.

joverlee521:
Jotting down some comments made during the Nextstrain meeting on Feb 9th:
- [ ] more spread out distribution of the y-axis jitter
- [ ] overall mean shouldn't be hidden when data is displayed in means
- [ ] hover panel size should be more responsive
- [ ] hover panel should follow tree UI pattern of bolding field names

src/actions/measurements.js
```diff
@@ -0,0 +1,241 @@
...
+        if ((collection.filters.has(field)) || !collectionFiltersArray) {
+          const filterObject = collection.filters.get(field) || { values: new Set()};
+          filterObject.values.add(fieldValue);
+          collection.filters.set(field, filterObject);
+        }
+      });
+
+      // Add jitter and stable id for each measurement to help visualization
+      // Generate Gaussian jitter with a Box-Muller transform
+      measurement["measurementJitter"] = Math.sqrt(-2*Math.log(Math.random()))*Math.cos(2*Math.PI*Math.random());
```
jameshadfield, line 155: @joverlee521 [here's the jitter code currently in use by the scatterplot layout](https://github.com/nextstrain/auspice/blob/e48e4326240740bc9a1053f1f71849e89e597d85/src/components/tree/phyloTree/layouts.js#L507-L527), which could potentially be refactored for the measurements. It would be nice for both approaches to use the same algorithm at least. 

I precomputed random values as this turned out to be a speed bottleneck.

src/actions/measurements.js
jameshadfield: or use symbols ;)

src/components/measurements/measurementsD3.js
jameshadfield: Resolved by 666d8b994f6fc83483cd672c451a681d5ebf458c

src/components/measurements/measurementsD3.js
```diff
@@ -283,23 +285,36 @@ export const changeMeasurementsDisplay = (ref, display) => {
...
       select(elements[i]).transition()
         .duration("200")
         .attr("r", layout.circleRadius);
       // sets hover data state to null to hide the hover panel display
-      handleHoverMeasurement(null);
+      handleHover(null);
     });
+
+  // Handle hovers of mean and standard deviations
+  svg.selectAll(`.${classes.mean}, .${classes.standardDeviation}`)
```
jameshadfield, line 324: [no changes needed] For what it's worth, CSS selection can be a decent botteneck. A way around this is to store references to the SVG groups (as we do elsewhere in auspice) and then `svgGroup.selectAll('*')`. Without actually testing I couldn't say if it would noticeably change things (I'd guess not in this case, there's not that many elements per row and each row is its own SVG).

src/components/controls/filter.js
```diff
@@ -85,10 +89,26 @@ class FilterData extends React.Component {
           });
         });
     }
+    if (this.props.measurementsOn && this.props.measurementsFiltersMap && this.props.measurementsFieldsMap) {
+      this.props.measurementsFiltersMap.forEach(({values}, filterField) => {
+        const { title } = this.props.measurementsFieldsMap.get(filterField);
+        values.forEach((value) => {
+          options.push({
+            _type: "measurements", // custom field to differentiate measurements filters
+            label: `${title} → ${value}`,
```
jameshadfield, line 98: I think we need some UI hint that these are for the measurement panel only, especially as the `title` is arbitrary and I can imagine collisions (e.g. tree nodes with "source" and measurement panel with "source"). Not a whole lot of visual space to play with here tho 😅

src/actions/measurements.js
```diff
@@ -130,3 +179,21 @@ export const changeMeasurementsCollection = (newCollectionKey) => (dispatch, get
...
   });
 };
+
+/*
+ * The filter actions below will create a copy of `controls.measurementsFilters`
+ * then clone the nested Map to avoid changing the redux state in place.
+ * Tried to use lodash.cloneDeep(), but it did not work for the nested Map
+ * - Jover, 19 January 2022
+ */
+export const toggleSingleFilter = (field, value, active) => (dispatch, getState) => {
```
jameshadfield, line 189: Could this be renamed to `applyMeasurementFilter`?

src/reducers/controls.js
```diff
@@ -87,7 +88,12 @@ export const getDefaultControlsState = () => {
     mapLegendOpen: undefined,
     showOnlyPanels: false,
     showTransmissionLines: true,
-    normalizeFrequencies: true
+    normalizeFrequencies: true,
+    measurementsGroupBy: undefined,
+    measurementsDisplay: "mean",
```
jameshadfield, line 93: I found myself always toggling to raw data, for what it's worth. Perhaps a question for @huddlej?

src/components/measurements/measurementsD3.js
```diff
@@ -0,0 +1,338 @@
...
+      .range([layout.subplotHeight, 0])
+      .nice()
+  );
+};
+
+/**
+ * Uses D3.groups() to aggregate measurements into a nested array of groups
+ * The groups are sorted with the most number of measurements first.
+ *
+ * If groupByFilters Map is provided, sort the gruops by the order of the keys.
```
jameshadfield, line 83: Minor typo

jameshadfield (changes requested):
Awesome work Jover — looking really good. Most of the in-line comments are minor ones or questions.

You place everything into the SVG and then use display properties to hide things appropriately. This is pretty clean actually - I wonder if there’s a speed penalty but it doesn’t seem major on the surface. However, the DOM is recreated each time the `svgData` changes (e.g. filtering, time slicing). This has benefits, in that the ordering of rows is continually  updated for the selected data and empty rows don’t exist. The downsides are that the performance suffers a bit, and it’s hard to keep track of things as you scrub time because the order changes. I would prefer the order to be fixed (as is the case with the arcs on the map & legend entries), and this would look much better while animating as well. For empty rows I’m less sure of what to do. Perhaps more a design decision than a coding one, but I imagine it would have some big changes to the flow of `<Measurements>`. 


Could you make an issue with the todo items you won't cover in this PR (e.g. URL state)? 

P.S. I fully agree with your comment about filters: “We may want to update our `react-select` package to use their “group”functionality to visually separate tree and measurement filter options within the dropdown menu.” Might be quite a bit of work as we use a lot of `<Select>` elements.

src/components/main/index.js
joverlee521: Yeah, this behavior is not my favorite either. 

We can include the measurements panel on load but display the Nextstrain spinner while the sidecar data is loading to be less disruptive. Would it be disruptive to then remove the panel if there was an error loading the data? Or do we let the empty panel (maybe an error message) sit there and leave it to the user to toggle off the panel?

src/actions/measurements.js
joverlee521: I had not made these reserved properties since I was not sure if their values should be generated within auspice (i.e. should they be generated by augur), but feedback from the Nextstrain meeting made it clear these values should be within auspice only since they are only relevant for the visualization. 

I'll change them to symbols (gotta use the new shiny thing I just learned about 😉 )

src/actions/measurements.js
```diff
@@ -0,0 +1,121 @@
...
+  if (!collections || collections.length === 0) {
+    throw new Error("Measurements JSON does not have collections");
+  }
+
+  collections.forEach((collection) => {
+    // Convert provided fields array to Map for easier access and to keep ordering
+    const collectionFieldsArray = collection.fields;
+    collection.fields = new Map();
+    if (collectionFieldsArray && collectionFieldsArray.length > 0) {
+      collectionFieldsArray.forEach(({key, title}) => collection.fields.set(key, {title: title || key}));
```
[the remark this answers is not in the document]
joverlee521: Ah yeah, I debated on this data structure because it can seem overly complicated. 

My reasoning is that `collection.fields` and `collection.filters` are Maps that utilize both the keys and the values as data stores. For example, I use `collection.filters` to check whether a field should be a filter and I think the current structure makes more sense:
```js
// current
collection.filters.has(field)

// suggestion
collection.filterValues.has(field)
```

src/components/measurements/index.js
```diff
@@ -1,9 +1,95 @@
...
+  const xScale = useMemo(() => createXScale(width, measurements), [width, measurements]);
+  const yScale = useMemo(() => createYScale(measurements), [measurements]);
+  // Memoize all data needed for basic SVG to avoid extra re-drawings
+  const svgData = useDeepCompareMemo({ xScale, yScale, x_axis_label, threshold, groupedMeasurements});
+
+  // Draw SVG from scratch
+  useEffect(() => {
+    clearMeasurementsSVG(d3Ref.current);
+    drawMeasurementsSVG(d3Ref.current, svgData);
+  }, [svgData]);
```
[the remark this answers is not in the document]
joverlee521: Yes-ish. `xScale` and `yScale` are already memoized with `useMemo`. `x_axis_label` and `threshold` are string and number values that should pass the shallow reference equality. The only thing we would have to additionally memoize is `groupedMeasurements` to specify each dependency individually. 

The major win for memoizing them together as `svgData` is we only have one dependency to pass to all of the other effects that also need to run if the SVG is redrawn. (It would be a very annoying bug to figure out if we forgot one of the five dependencies in one of the other effects).

src/components/measurements/measurementsD3.js
```diff
@@ -0,0 +1,172 @@
...
+  const {xScale, yScale, x_axis_label, threshold, groupedMeasurements} = svgData;
+
+  // Do not draw SVG if there are no measurements
+  if (groupedMeasurements && groupedMeasurements.length === 0) return;
+
+  const svg = select(ref);
+  const svgWidth = svg.node().getBoundingClientRect().width;
+
+  // The number of groups is the number of subplots, which determines the final SVG height
+  const totalSubplotHeight = (layout.subplotHeight * groupedMeasurements.length);
```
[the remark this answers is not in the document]
joverlee521: Hmm, I hadn't thought about this. On one hand, it would be a great way to save space within the viz. On the other hand, this also visually devalues the groups that have less measurements. Maybe that's something we want?

src/components/main/index.js
tsibley: I may not have full context here, but: if the measurements panel has to be toggled on (by the dataset's Auspice config? or by the user viewing the dataset?), then I'd say it's friendlier to display it with an error (or some indication of an error, at least) if the data can't be loaded rather than quietly not show it. Otherwise it may seem as if the config/toggle isn't working?

src/actions/measurements.js
tsibley: Symbols are a much better solution!

src/components/measurements/measurementsD3.js
```diff
@@ -0,0 +1,172 @@
...
+      .attr("x", layout.leftPadding + ((svgWidth - layout.leftPadding - layout.rightPadding)) / 2)
+      .attr("y", layout.bottomPadding * 2 / 3)
+      .attr("text-anchor", "middle")
+      .attr("fill", "currentColor")
+      .text(x_axis_label);
+
+  // Create a subplot for each grouping
+  let prevSubplotBottom = layout.topPadding;
+  groupedMeasurements.forEach(([groupingValue, measurements], index) => {
+    // Make each subplot it's own SVG to re-use the same subplot yScale
```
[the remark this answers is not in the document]
joverlee521: I have not found any downsides to this approach in my brief search. 
(For reference, [this post](https://www.sarasoueidan.com/blog/nesting-svgs/) is what I read for nesting SVGs.)

src/components/measurements/measurementsD3.js
```diff
@@ -170,3 +170,9 @@ export const drawMeasurementsSVG = (ref, svgData) => {
         .attr("r", layout.circleRadius);
   });
 };
+
+export const colorMeasurementsSVG = (ref, treeStrainColors) => {
+  const svg = select(ref);
+  svg.selectAll(`.${classes.rawMeasurements}`)
+    .style("fill", (d) => treeStrainColors[d.strain]);
```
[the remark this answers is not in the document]
joverlee521: I agree that lightened without any stroke looks too bright. I think lightened w/ stroke can look a little busy when so many colors are are clashing within a subplot, but I do like that it delineates each individual measurement. I can update to lightened + stroke to match the tree.

src/components/measurements/measurementsD3.js
```diff
@@ -176,3 +177,22 @@ export const colorMeasurementsSVG = (ref, treeStrainColors) => {
   svg.selectAll(`.${classes.rawMeasurements}`)
     .style("fill", (d) => treeStrainColors[d.strain]);
 };
+
+export const setHoverTransition = (ref, handleHoverMeasurement) => {
+  const svg = select(ref);
+  svg.selectAll(`.${classes.rawMeasurements}`)
```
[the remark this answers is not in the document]
joverlee521: For me, this was just a separation of concerns. I did not realize that the CSS selection can be a bottleneck! With that in mind, I think it's better to add the handlers during the drawing of the SVG.

src/reducers/controls.js
joverlee521: Ah, this was from [feedback](https://bedfordlab.slack.com/archives/C03KWDET9/p1636058549018400?thread_ts=1636057276.018300&cid=C03KWDET9) for the prototype app so it might be very specific to flu titer data.

src/components/controls/filter.js
joverlee521: Yeah, I think upgrading `react-select` to use option groups is the best thing to do here...

For now, would it be confusing for measurements options to be `(M) ${title} → ${value}`? I can add a note within `FilterInfo`.

joverlee521:
> You place everything into the SVG and then use display properties to hide things appropriately. This is pretty clean actually - I wonder if there’s a speed penalty but it doesn’t seem major on the surface. 

During my local testing, toggling the display property was a lot faster than removing & redrawing SVG elements.

> However, the DOM is recreated each time the svgData changes (e.g. filtering, time slicing). This has benefits, in that the ordering of rows is continually updated for the selected data and empty rows don’t exist. The downsides are that the performance suffers a bit, and it’s hard to keep track of things as you scrub time because the order changes. I would prefer the order to be fixed (as is the case with the arcs on the map & legend entries), and this would look much better while animating as well. For empty rows I’m less sure of what to do. Perhaps more a design decision than a coding one, but I imagine it would have some big changes to the flow of <Measurements>.

I definitely get the performance issue and I hadn't even thought about the animation...The reason I went with the redrawing method was to allow for user-based ordering of rows and filtering to only a subset of rows. (These were requested by the CDC in the prototype for flu titer data). I think leaving empty rows is not a great user experience, especially if users have to scroll all the way to the bottom of the panel to see their selection of data.

src/components/main/index.js
joverlee521: Screen Shot 2022-02-16 at 3 27 46 PM
Currently, when the sidecar JSON isn't able to load property, a warning notification pops up. There's no disruption in the rest of the page because the measurements panel itself was never rendered due to the condition on `this.props.measurementsLoaded`. 

However, given the expectation that the sidecar JSON should load properly more often than not, I see that we should reverse the disruptive behavior. The panel should be rendered on page load (with a fallback spinner waiting for data), and then removed if the data doesn't load properly. There would be a disruption but I think an expected one since there's a warning notification as well.

package.json
```diff
@@ -108,7 +108,7 @@
     "react-hot-loader": "^4.8.4",
     "react-i18next": "^11.3.3",
     "react-icons": "^3.9.0",
-    "react-redux": "^5.1.0",
+    "react-redux": "^7.2.6",
```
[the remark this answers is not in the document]
joverlee521: Documenting here that there were only two major breaking changes that I could find in [their release notes](https://github.com/reduxjs/react-redux/releases):
> The `withRef` option to connect has been replaced with `forwardRef`. 

I could not find any references to `withRef` within `src/`.  

> Passing `store` as a prop to a connected component is no longer supported

However this was then restored in [v7.0.1](https://github.com/reduxjs/react-redux/releases/tag/v7.0.1) so no-op for us 😄

src/components/measurements/measurementsD3.js
joverlee521: Resolved in https://github.com/nextstrain/auspice/pull/1452/commits/8614feb1339acbf25e9431bbd78f7b1c51ff0273

src/actions/measurements.js
joverlee521: Resolved in https://github.com/nextstrain/auspice/pull/1452/commits/12dc1ba8c002dc5e280b7a56fc869b5fbe94ab8e

src/actions/measurements.js
joverlee521: Resolved in https://github.com/nextstrain/auspice/pull/1452/commits/a2d414d67c60e3c469355816cff5761507d02e8d

src/components/measurements/index.js
joverlee521: Thanks both! Resolved in 29fed88

jameshadfield:
> The reason I went with the redrawing method was to allow for user-based ordering of rows...

I think of the ordering similar to the way we order arcs in the map: the order is fixed before any filters are applied [1].

>  filtering to only a subset of rows.

This is independent of the ordering, and in my mind essentially is the same as producing rows with no data points, i.e.👇 

> I think leaving empty rows is not a great user experience

Agreed. Let's drop them if there are no points to plot. This will mean some rows pop in and out of existence as we scrub through the data, but that for plots which are displayed they stay in the same order. 

[1] This should work with user-defined ordering, or the default, or any other we add in the future.

joverlee521:
> I think of the ordering similar to the way we order arcs in the map: the order is fixed before any filters are applied [1].

Ah okay, this means we keep the ordering stable for all other filtering _except_ when the user specifically filters on the group-by field?

Right now the default ordering is by the most measurements first, so the row with the most number of measurements _after_ filtering is displayed on top. In order to keep ordering stable, this means we will display the row with the most number of _total_ measurements at the top regardless of how filtering affects the measurement counts per row.

src/components/measurements/measurementsD3.js
joverlee521: Resolved in 5921b96. I went with the synthetic ids since I had to do the extra bookkeeping to keep stable ordering of subplots anyways.

src/actions/measurements.js
joverlee521: Addressed in bb2b19a, I went with a slightly simpler jitter generation.

src/components/measurements/measurementsD3.js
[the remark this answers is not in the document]
joverlee521: Resolved in 4f92b47

src/components/measurements/measurementsD3.js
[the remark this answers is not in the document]
joverlee521: Updated in 11f2605

joverlee521:
Rebased to fix merge conflicts in the `package-lock.json`. 

Added stable default ordering for subplots and added ability to use JSON defined ordering. 
[Updated the testing JSON](https://github.com/joverlee521/nextstrain-testing/commit/771816c0a95fe53333cdc7834974d6eea650ed6c) to include customized orders for groupings, so `Raw Cell Passaged HI assay measurements` should display `A/England/195/2009` as the first row in the default ordering even though it does _not_ have the most measurements.

src/reducers/controls.js
jameshadfield: That's a good reason to keep as is than!

src/components/controls/filter.js
jameshadfield: Yes I think that's an acceptable solution for now -- it's a good trade off between not adding characters (as overflowing labels are not correctly displayed) but indicating something is different. The user-base for measurements is expected to be close collaborators for at least the time before we upgrade react-select!

src/components/measurements/index.js
```diff
@@ -0,0 +1,279 @@
...
+};
+
+const Measurements = ({height, width, showLegend}) => {
+  const measurementsLoaded = useSelector((state) => state.measurements.loaded);
+  const measurementsError = useSelector((state) => state.measurements.error);
+
+  const [title, setTitle] = useState("Measurements");
+
+  return (
+    <Card title={title}>
```
jameshadfield, line 259: I think the title CSS (in `<Card>`) needs to be changed as when in grid mode the panels can be displayed on top of each other due to long title names

image

src/components/measurements/index.js
```diff
@@ -0,0 +1,279 @@
...
+        // Handle single measurement data
+        // Filter out internal auspice fields (i.e. measurementsJitter and measurementsId)
+        const displayFields = Object.keys(data).filter((field) => fields.has(field));
+        // Order fields for display
+        const fieldOrder = [...fields.keys()];
+        const orderedFields = orderBy(displayFields, (field) => fieldOrder.indexOf(field));
+        orderedFields.forEach((field) => {
+          newData.set(fields.get(field).title, data[field]);
+        });
+      } else if (dataType === "mean") {
```
jameshadfield, line 171: reminder to add a title to the on-hover panel (essentially the colour-by value, I think). We probably want to add the row-name (reference / group-by name) I think?

jameshadfield (changes requested):
Haven't finished my re-review but adding these comments now as I'm at a stopping point.

src/components/measurements/measurementsD3.js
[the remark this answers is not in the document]
jameshadfield: (Noting here outcome from discussion last week) this suggestion is a UI/UX idea to try one day, probably in the python version which will be quicker. It's not blocking this PR.

jameshadfield (commented):
Finished re-reviewing. Looks great Jover, excited to see this released ✨ The stable ordering is great and really highlights drift:

https://user-images.githubusercontent.com/8350992/155905822-8bb05bc3-c4e6-4a08-82e7-10b00f2c675c.mov

P.S. Don't forget to remove 51d009c068a0c7a085c742f8a28bf7551d421da2

src/components/measurements/index.js
joverlee521: Added as todo in #1463

src/components/measurements/index.js
joverlee521: Added as todo in #1463.

src/components/tree/reactD3Interface/change.js
```diff
@@ -24,7 +23,7 @@ export const changePhyloTreeViaPropsComparison = (mainTree, phylotree, oldProps,
     args.changeColorBy = true;
     args.branchStroke = calcBranchStrokeCols(newTreeRedux, newProps.colorByConfidence, newProps.colorBy);
     args.tipStroke = newTreeRedux.nodeColors;
-    args.fill = newTreeRedux.nodeColors.map((col) => rgb(col).brighter([0.65]).toString());
+    args.fill = newTreeRedux.nodeColors.map((col) => getBrighterColor(col));
```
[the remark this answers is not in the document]
joverlee521: Updated in 89de0de

joverlee521:
Rebased onto master and dropped the testing commit. 
Also added 3db7a96 to fix the filter ordering issue that came up in lab meeting.

Issue #1457, opened 2022-02-10
Auto-preview to nextstrain.org fails on PRs from forks and dependabot


victorlin:
I've been testing out using forks to create PRs (#1456; #1412 fork necessary from before joining the team).

For PRs from forks, the Github Action workflow used to create PRs to nextstrain.org for preview ([src](https://github.com/nextstrain/auspice/blob/master/.github/workflows/make_prs_for_other_repos.yaml)) fails with a generic error message ([example](https://github.com/nextstrain/auspice/runs/5147686266?check_suite_focus=true)):

```
  fatal: could not read Username for 'https://github.com/': No such device or address
  Error: The process '/usr/bin/git' failed with exit code 128
```

This is due to, for PRs from forks, workflows triggered by the `pull_request` event having read-only permissions and thus no access to `secrets.NEXTSTRAIN_BOT_PAT` which results in this error. The dependabot issue is discussed here: peter-evans/create-pull-request#871

In order for a workflow to have write access, it should use `pull_request_target` which comes with some [security caveats](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). I'm having a hard time wrapping my brain around if this could be a security issue for the existing usage (PR from auspice fork -> PR to nextstrain.org with forked auspice code -> preview app).

It would be ideal if we could use `pull_request_target` then require manual approval for first-time contributors [like this](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks), but unfortunately this doesn't apply with `pull_request_target`.

Another thing to note is that PRs from forks require manual creation of Heroku Review Apps ([doc](https://devcenter.heroku.com/articles/github-integration-review-apps#automatic-creation)):

> For security and billing reasons, Heroku does not automatically create review apps for pull requests to public repos that are sent from forks. You can still create review apps for these pull requests manually.

Overall, this is not something that needs to be fixed in the grand scheme of things, since we can still test with manual creation of Heroku Review Apps. Just wanted to document a bit of research.

tsibley:
> I'm having a hard time wrapping my brain around if this could be a security issue for the existing usage (PR from auspice fork -> PR to nextstrain.org with forked auspice code -> preview app).

There are a lot of security concerns here. Some things that come to mind.

[...]

PR #1468, opened 2022-02-15
Show date picker only for non-numeric dates

victorlin:

*follow-up to https://github.com/nextstrain/auspice/pull/1412#issuecomment-1039696219*

The HTML date picker added in #1412 only works for `YYYY-MM-DD` dates with `0001-01-01` being the earliest pickable date. However, Nextstrain supports analyses that predate `0001-01-01` in the form of numeric dates (e.g. https://nextstrain.org/tb/global).

This PR adds an intermediate component `DateLabel` which only shows `DatePicker` if the date value is not numeric. For numeric dates, it shows a read-only text span.

date-label

### Testing

Tested behavior locally using `tb/global` dataset.


jameshadfield:
Thanks @victorlin! Reviewing this has alerted me to some _weird_ behaviour around auspice date handling which this PR has alerted me to. I think best to tackle this here, although it increases the scope a lot so feel free to disagree!

Separately, could you create PRs from a branch off nextstrain rather than a fork? Makes my life a little easier 😉 

**Dates are stored as both floats and strings**

We store dates via the following properties (on `reduxState.controls`)

* currently selected minimum date: `dateMin` `dateMinNumeric`
* currently selected maximum date: `dateMax` and `dateMaxNumeric`
* dataset min date (of any node): `absoluteDateMin` and `absoluteDateMinNumeric`
* dataset max date (i.e. most recent tip): `absoluteDateMax` and `absoluteDateMaxNumeric`.

This duplication isn't desirable and is a good way to let things get out-of-sync.

I'd recommend we reduce the above down to 4 properties, and since the dates in the dataset JSON are numeric I'd store everthing in redux as numeric dates rather than calendar dates. This also avoids the complication that calendar dates are sometimes numeric and sometimes YYYY-MM-DD (`numericToCalendar` returns the rounded year for BCE dates).

We can then use the  `numericToCalendar` helper function in the few places where we actually need a calendar string, for instance inside `DatePicker` or on-hover info boxes. 

URL queries should remain calendar dates, and these currently work with BCE dates (e.g. `?dmin=-1063-02-26` is accurately parsed). The syntax `?dmin=-1063` (i.e. only specifying the year) also works, and this URL format is what I suggest the app uses whenever the date shown in the sidebar is shown as simply the year.

**Dates from 0-100CE are changed to 1900-2000**

This is a 🐛 which I don't know the origin of. It can be seen when you drag the date slider to a date in this range. Loading via a URL such as `?dmin=0063-02-2` works, so I believe the bug is within the date slider code?

**How much precision should we show?**

For many of our use cases it makes sense to show dates accurate to the day (e.g. most nCoV sequences have the actual sampling dates attached). However for "old" dates it's not so clear cut this is desirable. As an obvious example, and the original remit of this PR, we'll never know BCE dates with the precision of a day. I'd arbitrarily suggest the year 1900CE as where we switch from displaying simply a year in the date-slider to showing YYYY-DD-MM, with the latter having a datepicker. Open to suggestions here.

For node-matching purposes for dates prior to this threshold, if you shange the date slider to (e.g.) 1234, because the dates attached to nodes are arbritrarily precise, I'd store this value as 1234.0 (for `minDateNumeric`) or 1234.999 (for `maxDateNumeric`).

victorlin:
@jameshadfield haha, I've spent a lot of time lately thinking about date formats in Augur, and there's some common ground here. Notably, there's no clear mapping from numeric to calendar date when it's negative – another reason to store internally as numeric and only display as calendar date when appropriate.

> I'd arbitrarily suggest the year 1900CE as where we switch from displaying simply a year in the date-slider to showing YYYY-DD-MM, with the latter having a datepicker.

I don't see any harm in having a date picker all the way back to `0001-01-01`, since that's also the cutoff for when we should stop supporting calendar dates (see point above). Having 1900CE as another cutoff would be extra logic to think about and explain to any confused users.

> could you create PRs from a branch off nextstrain rather than a fork?

Yeah, I was being curious about the experience of PRs from forks. Curiosity satisfied (#1457) so for future PRs I'll go back to nextstrain repo branches to have proper collaboration and nextstrain.org review apps.

If you agree with the small change in this PR, how about we merge this then create a separate issue to further discuss date handling? That way I can start a new PR from a branch we can both collaborate on.

jameshadfield:
> If you agree with the small change in this PR, how about we merge this then create a separate issue to further discuss date handling? That way I can start a new PR from a branch we can both collaborate on.

Sounds great. In case others notice the ❌ , all the relevant CI checks have passed so no problem to merge.

PR #497, opened 2018-02-09
Frequencies

jameshadfield:
This PR implements some sort of frequency stream graph (the devil is in the details). cc @huddlej @trvrb @rneher 

### Availability:
* flu datasets live via https://auspice-dev.herokuapp.com/flu (3y, 6y & 12y)
* locally via:
```bash
git checkout frequencies
bash scripts/get_data.sh
python scripts/convert_augur_frequency_json.py # creates _pivots.json from _frequencies.json
npm run start:local
```

### How are frequencies calculated
The `global_clade:X` entries in the frequencies JSON are extracted into a separate smaller JSON (called `_pivots.json`). For each tip in the tree *(not internal branches)* the frequencies (i.e. the array with n entries, where n is the number of pivots) are binned by the `colorBy` value (or the bounds if the scale is continuous). Non-visible tips (e.g. via filtering) are not included. For each of these bins, at each pivot point, the frequencies are summed and potentially normalized (see below). This allows the data to reflect the selected `colorBy` as well as any enabled filters.

### The frequencies panel
* Displayed as a stream graph below entropy.
* The streams represent the combined frequencies across pivots for each colorBy (i.e. each entry in the tree legend).
* A tool tip on hover shows you more information. 
* The data can be normalized (the values of each pivot point sum to 1) via the toggle (top right).
* When forecasted frequencies are available these can easily be displayed.

### Known bugs / things to do
* ~genotype doesn't work~
* browser resize breaks things
* styling need improving
* tips are visible outside of the selected date range if the branch intersects the visible range (#473)
* tips continue to have some visibility long after their collection date (potential augur bug, not sure). I've filtered out any values `< 0.0002`
* rounding errors in the frequency JSON mean that totals for all tips don't sum to 1.
* having a `colorBy` of clade / named clades (or some ability to select this) would result in these being different streams in the graph.
* no legend
* ~sometimes the stacking order is wrong~

### Screenshots

* 12y coloured by date, all tips selected
image

* 12y coloured by antigenic advance, all tips selected
image

* same data but restricted to asia
image

* 12y coloured by region, all tips selected
image

PR #1471, opened 2022-02-15
Stop storing an unused "weight" property from tip frequencies

tsibley:
This property was introduced with the original frequencies work¹ as an
anticipated need², but it was never used.  Omit it for now to avoid
carrying around unnecessary baggage; it can be added back in the future
easily if its time comes.

I uncovered this while authoring a JSON Schema for the tip-frequencies
format.³

¹ In PR https://github.com/nextstrain/auspice/pull/497 as a7bda1e.
² https://github.com/nextstrain/augur/pull/83#issuecomment-367405686
³ https://github.com/nextstrain/augur/pull/852

### DIFF
--- a/src/util/processFrequencies.js
+++ b/src/util/processFrequencies.js
@@ -128,8 +128,7 @@ export const processFrequenciesJSON = (rawJSON, tree, controls) => {
       }
       data.push({
         idx: n.arrayIdx,
-        values: rawJSON[n.name].frequencies,
-        weight: rawJSON[n.name].weight
+        values: rawJSON[n.name].frequencies
       });
     });
 


trvrb:
I do think it's highly relevant to work to properly include "weights" even if we haven't done so up to this point. For example, we could be weighting tips based on case counts relative to sequencing intensity to get a better picture of frequencies.

That said, if it's not implemented, it's not implemented and can be safely removed.

We had been "weighting" regions in the seasonal flu build, but just recently turned this off: https://github.com/nextstrain/seasonal-flu/commit/40f850ba4c03cd3eb2dc12b85c5136b2d66ff239

However, I think that even including `--weights` in `augur frequencies` was not producing a `tip-frequencies.json` with a `weights` attribute and instead was directly modifying the area under the curve of each tip kernel. @huddlej: Does this sound correct?

In this case, maybe we just continue this direction and plan rely on this built-in weighting in Augur rather than doing something on the fly in Auspice?

tsibley:
+1 for handling weights one way or another. @huddlej and @jameshadfield likely have the most input here.

> However, I think that even including `--weights` in `augur frequencies` was not producing a `tip-frequencies.json` with a `weights` attribute

Yep, that's my understanding having recently read through this code and looked at output files.

When I ran my tip-frequencies schema in https://github.com/nextstrain/augur/pull/852 against all `s3://nextstrain-data/*frequencies.json`, only one file fell out as not validating because it had this `weight` attribute for each node:

https://data.nextstrain.org/flu_seasonal_h1n1pdm_na_pandemic_tip-frequencies.json

This file was last modified on 12 May 2018 and all the weights are 1.0.

> That said, if it's not implemented, it's not implemented and can be safely removed.

AFAICT with some careful grepping and looking, it's not used. However, I would appreciate @jameshadfield's confirmation that I didn't miss some spot.

huddlej:
Just confirming all the comments so far; the weighting in `augur frequencies` changes the kernels of the KDEs internally. We never used the weights in the JSON.

jameshadfield (approved):
Yes, my memory is that this was included to future proof auspice but never actually used, and the code seems to back this up!

PR #1472, opened 2022-02-17
narratives: Load YAML frontmatter safely

tsibley:
Prevents deserializing potentially unsafe data types from user input.

yaml-front-matter uses js-yaml 3.x, which is unsafe by default, hence
the "safe" mode. js-yaml 4.x is safe by default, but not used here.

### DIFF
--- a/src/util/parseNarrative.js
+++ b/src/util/parseNarrative.js
@@ -11,12 +11,12 @@
 
 /* eslint no-param-reassign: off */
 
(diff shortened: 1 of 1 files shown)


tsibley:
See also https://github.com/nextstrain/nextstrain.org/pull/473.

jameshadfield (approved):
Looks good and tested on a number of existing narratives with no observed changes (as expected!)

PR #1482, opened 2022-03-09
Upgrade react select

### LINKED ISSUE #1467: Upgrade `react-select` package
We are still using v1 of `react-select` which is [no longer maintained](https://github.com/JedWatson/react-select/issues/3981). We should upgrade the package to use new features that have been added, such as option groups. The latest version is v5, but v2 was a complete rewrite so we should follow [their v2 upgrade guide](https://react-select.com/upgrade-to-v2) to make the necessary changes.

According to @jameshadfield: 
> Might be quite a bit of work as we use a lot of `<Select>` elements.

joverlee521:
Upgrades our `react-select` package to version 5.2.2 and replaces the `react-virtualized-select` package with our own virtualized menu list built with `react-virtualized` components. 

Note that the upgrade of `react-select` did include style changes. I did not dig into the style changes, but please let me know if we want the styles to match the old version.

Resolves #1467 and ##1453.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3054,27 +3054,6 @@
         "babel-preset-current-node-syntax": "^0.1.2"
       }
     },
-    "babel-runtime": {
-      "version": "6.26.0",
-      "resolved": "https://registry.npmjs.org/babel-runtime/-/babel-runtime-6.26.0.tgz",
-      "integrity": "sha1-llxwWGaOgrVde/4E/yM3vItWR/4=",
-      "requires": {
-        "core-js": "^2.4.0",
-        "regenerator-runtime": "^0.11.0"
-      },
-      "dependencies": {
-        "core-js": {
-          "version": "2.6.12",
-          "resolved": "https://registry.npmjs.org/core-js/-/core-js-2.6.12.tgz",
-          "integrity": "sha512-Kb2wC0fvsWfQrgk8HU5lW6U/Lcs8+9aaYcy4ZFc6DDlo4nZ7n70dEgE5rtR0oG6ufKDUnrwfWL1mXR5ljDatrQ=="
-        },
-        "regenerator-runtime": {
-          "version": "0.11.1",
-          "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.11.1.tgz",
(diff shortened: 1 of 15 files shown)


joverlee521:
A good test of the dynamic options is searching for "submitting lab" in the [seasonal flu build of the test app](https://auspice-upgrade-react-s-amd7fu.herokuapp.com/flu/seasonal/h3n2/ha/2y).

I'm currently looking into the bundle size issue.

joverlee521:
The [test PR in the nextstrain.org repo](https://github.com/nextstrain/nextstrain.org/pull/491) is failing because I did not realize that the `react-select` and `react-virtualized-select` packages are used in nextstrain.org as well. I'll work on updating the package in the nextstrain.org repo.

joverlee521:
The latest changes can now be tested with the SARS-CoV-2 builds that include the originating/submitting labs at https://nextstrain-s-upgrade-re-8ecxpi.herokuapp.com/staging/ncov/gisaid/trial/add_authors/global

jameshadfield:
Thanks @joverlee521! Haven't dived into the code yet but here's feedback comparing nextstrain-s-upgrade-re-8ecxpi.herokuapp.com against nextstrain.org

* Suggestion text (e.g. "Type filter query here...") is now `rgb(128, 128, 128);` but should be `#aaa` (which is `rgb(170, 170, 170);`)
* Cell size looks great! 
* Aesthetically I prefer the filled in down icon compared to the new vertical line & arrow outline, but if this is hard to change in react select I wouldn't be too concerned.
* Performance seems good (for the filter dropdown, which is the biggest we have)

emmahodcroft:
I tried this out this morning and it looks good to me! Even for the very long lists, as James said, the performance seemed good and not too laggy. I think the names wrap well and look really nice and tidy!
I had some fun exploring some of the interesting lab names we have in there... I guess with free-text fields you're always going to have some entries that are... creative! 🙃

joverlee521:
@jameshadfield The latest commit ([561c1fd](https://github.com/nextstrain/auspice/pull/1482/commits/561c1fda162f4732c470a7d7db02f56b38c8fed0)) centralizes the `Select` component into our own `CustomSelect` component that makes it easier to make broad style changes across all drop downs.

I used this to make the styles tweaks of changing the placeholder font color and replacing the drop down arrow. I also updated the dropdown menu to not have a top margin so there's not a weird gap between the input box and the drop down.

jameshadfield (approved):
Looks good @joverlee521. Thanks for addressing the styling -- I remember trying to combine styled-components and react-select before they had the `components` prop and it wasn't pretty! This is much better.

Issue #1483, opened 2022-03-15
Regression statistics are not updated when in view samples change


trvrb:
### Current Behavior

The scatterplot view provides a handy option to "show regression". For example, with https://nextstrain.org/ncov/gisaid/global?branches=hide&l=scatter&regression=show&scatterY=div we get an R^2 of 0.845 of sampling date vs divergence.

However, if you [filter to a subset of dates](https://nextstrain.org/ncov/gisaid/global?branches=hide&dmin=2022-01-11&l=scatter&regression=show&scatterY=div) or if you [filter to a particular clade](https://nextstrain.org/ncov/gisaid/global?branches=hide&f_clade_membership=21J%20%28Delta%29&l=scatter&regression=show&scatterY=div) the regression statistics remain the same with R^2 of 0.845.

### Expected behavior

I expect regression statistics to update to match in view tips. This is keeping with general Auspice ethos of having everything in the visualization reflect in view tips (diversity, etc...).

I'm assuming this may need debouncing just like how diversity works in order keep overhead of recalculating statistics down (or it may be cheap enough not to worry about it).

PR #1484, opened 2022-03-16
Regression line only considers visible nodes

jameshadfield:
See #1483 for context. A couple of before (left) vs after (right) screenshots:

 
image
image

### DIFF
--- a/src/components/tree/phyloTree/change.js
+++ b/src/components/tree/phyloTree/change.js
@@ -333,6 +333,11 @@ export const change = function change({
   if (changeColorBy) {
     this.updateColorBy();
   }
+  // recalculate existing regression if needed
+  if (changeVisibility && this.regression) {
+    elemsToUpdate.add(".regression");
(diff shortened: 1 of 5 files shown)


trvrb:
Awesome! Thanks for the quick work here James. I noticed this originally when I was attempting to explain to someone that you should still less clock correlation with a dataset that spans a narrower time window. With PR, this works as expected, where you can go from [R^2 of 0.82 for the full ~2 year dataset](https://auspice-fix-regression-dcxebqb.herokuapp.com/ncov/gisaid/global?branches=hide&l=scatter&regression=show&scatterY=div) to an [R^2 of 0.33 for dataset spanning ~6 months](https://auspice-fix-regression-dcxebqb.herokuapp.com/ncov/gisaid/global?branches=hide&dmax=2021-07-06&dmin=2021-01-24&l=scatter&regression=show&scatterY=div).

Really cool to be able to drag the date slider and watch the regression line update live in response.

Everything seems to be working well in my testing.

trvrb (approved):

Issue #1243, opened 2020-12-10
Deselecting a single strain doesn't remove the info panel overlay


trvrb:
**Current Behavior**  

With new filters behavior (implemented in #1200) clicking on a single tip in the tree adds this single tip as a "filter" and covers tree with gray info panel overlay, like so:

<img width="992" alt="tip-selected" src="https://user-images.githubusercontent.com/1176109/101833280-98ca7f80-3aed-11eb-99d0-ab1124d6d9dd.png">

This behavior is as intended. Once in this state you can click outside the overlay panel and both the overlay panel and the strain filter will be dismissed. This is also as intended.

However, if you click on the trash can icon to remove the filter you're left with the full tree displayed but the overlay panel remains. It takes a second click outside the panel to dismiss it. 

<img width="992" alt="dismissed" src="https://user-images.githubusercontent.com/1176109/101833511-f6f76280-3aed-11eb-83e9-478c5e18e6f7.png">

**Expected behavior**  

Instead of this behavior, clicking the trash can icon should remove both strain filter and the overlay panel.

victorlin:
@jameshadfield I started looking into this. I'm a bit confused about the different trees.

This is what I've gathered:

1. Clicking outside the overlay panel restores the view by removing the modal (L149) and inactivating the filter (L152): https://github.com/nextstrain/auspice/blob/d8b3d2ae700978a8ef6ebb817b52e994936b1a09/src/components/tree/reactD3Interface/callbacks.js#L149-L153
2. Clicking the trash can icon only removes the filter using the same `applyFilter` function. It doesn't reset `selectedNode` which is the reason for this bug. https://github.com/nextstrain/auspice/blob/d8b3d2ae700978a8ef6ebb817b52e994936b1a09/src/components/info/filtersSummary.js#L54

And respectively,

1. The external click, through [a callback binding](https://github.com/nextstrain/auspice/blob/d8b3d2ae700978a8ef6ebb817b52e994936b1a09/src/components/tree/tree.js#L37), removes the modal via `selectedNode` on this `Tree` class: https://github.com/nextstrain/auspice/blob/d8b3d2ae700978a8ef6ebb817b52e994936b1a09/src/components/tree/tree.js#L23
2. The icon click is connected to this `Tree` reducer: https://github.com/nextstrain/auspice/blob/d8b3d2ae700978a8ef6ebb817b52e994936b1a09/src/reducers/tree.js#L32

[...]

PR #1485, opened 2022-03-16
[tree] escape key dismisses any info modal

### LINKED ISSUE #1479: ENH: Make Esc dismiss sample details popup
### Context
I often (shift)-click on samples or branches to see details.

Right now, to dismiss the popup, I need to click outside the popup, requiring a mouse movement which is a little cumbersome.

I've repeatedly found myself hitting `esc` in the expectation that it would dismiss the popup, which it didn't do.

### Description
It would be nice if `esc` dismissed the details popup, just how clicking outside does at the moment.

I'm talking about this popup:
image

jameshadfield:
This implements what we think is an intuitive UI, and also helps with
the bug described in #1243 where removing/disabling a strain-filter
results in the tree visualisation falling out-of-sync with the modal.

See https://github.com/nextstrain/auspice/issues/1479 for context.

Closes #1479

### DIFF
--- a/src/components/tree/tree.js
+++ b/src/components/tree/tree.js
@@ -41,6 +41,12 @@ class Tree extends React.Component {
         root: [0, 0]
       }));
     };
+    /* pressing the escape key should dismiss an info modal (if one exists) */
+    this.handlekeydownEvent = (event) => {
+      if (event.key==="Escape" && this.state.selectedNode?.node) {
+        this.clearSelectedNode(this.state.selectedNode);
+      }
+    };
   }
   setUpAndRenderTreeToo(props, newState) {
     /* this.setState(newState) will be run sometime after this returns */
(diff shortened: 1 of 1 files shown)


src/components/tree/tree.js
```diff
@@ -41,6 +41,12 @@ class Tree extends React.Component {
         root: [0, 0]
       }));
     };
+    /* pressing the escape key should dismiss an info modal (if one exists) */
+    this.handlekeydownEvent = (event) => {
+      if (event.keyCode===27 && this.state.selectedNode) { // key 27 is esc
```
joverlee521, line 46: FYI, `keyCode` has been [marked as deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). 

The alternative is to use `key`, which is equal to "Escape" or "Esc" in older browsers (see [key values](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values)):

joverlee521, line 46: FYI, `keyCode` has been [marked as deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). 

The alternative is to use `key`, which is equal to "Escape" or "Esc" in older browsers (see [key values](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values)):

jameshadfield: Did not know, thanks Jover! Fixed & force-pushed.

PR #1486, opened 2022-03-16
Swap to "experimental" icon for exploded trees sidebar option

trvrb:

This swaps the previous use of "[experimental]" in the "Explode tree by" sidebar option to a cleaner looking icon. I chose a wrench, thinking that this best represents work in progress.

Screen Shot 2022-03-16 at 1 58 59 PM

### Testing
I've tested locally, but it should be a small enough change that it should be safe.

### DIFF
--- a/src/components/controls/choose-explode-attr.js
+++ b/src/components/controls/choose-explode-attr.js
@@ -1,10 +1,10 @@
 import React from "react";
 import { connect } from "react-redux";
 import { withTranslation } from 'react-i18next';
-import { FaWrench } from "react-icons/fa";
-
+import { ImLab } from "react-icons/im";
+import { FaInfoCircle } from "react-icons/fa";
 import { explodeTree } from "../../actions/tree";
-import { SidebarSubtitle } from "./styles";
+import { SidebarSubtitleFlex, StyledTooltip, SidebarIconContainer } from "./styles";
 import { controlsWidth } from "../../util/globals";
 import CustomSelect from "./customSelect";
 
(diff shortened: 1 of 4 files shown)


jameshadfield:
Happy to have an icon here, but for me the wrench doesn't convey "experimental", how about one of:

`<GiStandingPotion/>`
image

`<AiTwotoneExperiment/>`
image


`<ImLab/>`
image



Also, we should have some text pop up on-hover - I can add a commit for this shortly.

trvrb:
Thanks James! I understand that "wrench" isn't conveying "experimental" to you. I guess I was thinking of it most as "in progress" or "under construction". I'm having a bit of difficulty with the beaker as it conveys "science" to me rather than "experimental". But that said, after looking a bit I don't have a better suggestion. Out of the three `<ImLab/>` would seem to work best at small font size.

For tooltip, I'd think to just use the regular `FaInfoCircle` on the right hand side of the sidebar and note in the tooltip the "Experimental:" warning.

jameshadfield:
Changed to the following, which I prefer, but either icon is preferable to the current version! I'll merge shortly unless others have objections / want to move back to the wrench 😉 

image

jameshadfield:
(Force pushed to fix the typo in the on-hover text.)

trvrb:
Thanks @jameshadfield! I'm happy with the beaker, but I do have a preference here for consistency in how the explanatory tooltips behave. In the current PR, if you mouseover any of the text in "Explode Tree By" or the beaker icon you get that explanatory tooltip with "This functionality is experimental and should be treated with caution! Exploding a tree by trait X means that for each branch where the trait changes value..."

This is different from the explanatory tooltip for "Tree Options", etc... where you get nothing when you mouseover the text, but you get the tooltip when you mouseover the (i). I really like having the visual cue that you can mouseover to get explanatory text. I'd recommend either:

1. Just a single tooltip with the existing text that appears on hover of an (i) icon on the right
2. A "This functionality is experimental and should be treated with caution!" tooltip when mousing over the just the beaker (and not the text) and a separate "Exploding a tree by trait X means that for each branch where the trait changes value..." tooltip when mousing over an (i) icon on the right

I lean towards option 2.

jameshadfield:
Thanks @trvrb -- implemented option 1 as I felt people may not hover over the experimental icon. I didn't originally do this as I was worried about having too many "info" icons, but even if we built them for all of the options it's not actually too bad (and if we do get here, we can tweak the styles further):

image

trvrb:
Looks great. Thanks James!

tsibley:
Alternative options if the beaker is too sciencey given our domain:

image
image

These speak more to construction than the beaker to me, but the beaker does mean "experimental" to me given its common use that way in other software products.

PR #1493, opened 2022-03-22
Fix: narrative dataset filters

### LINKED ISSUE #1492: Narrative: No filter options in "Explore the data"
Navigate to a slide within a narrative, click on "Explore the Data Yourself" and the filter dropdown displays "No Options". If filters are defined in the dataset URL, then only the selected filters are available in the dropdown. The filter options in the footer are also missing, indicating this bug is most likely an issue stemming from the redux store and not the filter dropdown itself. 

First noticed the bug in @huddlej 's [community narrative for the flu meeting](https://nextstrain.org/community/narratives/huddlej/seasonal-flu-reports/2022-03-21?n=0), but I can reproduce the bug in other narratives (e.g. [intro to narratives](https://nextstrain.org/narratives/intro-to-narratives)).

joverlee521:
When users change the slide in a narrative, the controls state gets
reset to its default state which updates the filters state to an empty
object. This leads to the bug described in #1492.

~This commit extracts the filters creation from `modifyStateViaMetadata`
into its own function so that it can be reused to recreate the filters
when users change the slide within a narrative.~

Resolves #1492.

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -786,6 +786,7 @@ export const createStateFromQueryOrJSONs = ({
     metadata = {...oldState.metadata};
     frequencies = {...oldState.frequencies};
     controls = restoreQueryableStateToDefaults(controls);
+    controls = modifyStateViaMetadata(controls, metadata);
   }
 
   /* For the creation of state, we want to parse out URL query parameters
--- a/src/reducers/controls.js
+++ b/src/reducers/controls.js
@@ -22,6 +22,7 @@ export const getDefaultControlsState = () => {
     layout: defaultLayout,
     geoResolution: defaultGeoResolution,
     filters: {},
+    filtersInFooter: [],
     colorBy: defaultColorBy,
     selectedBranchLabel: "none",
     showTransmissionLines: true
@@ -67,7 +68,8 @@ export const getDefaultControlsState = () => {
     canRenderBranchLabels: true,
     analysisSlider: false,
     geoResolution: defaults.geoResolution,
-    filters: {},
+    filters: defaults.filters,
+    filtersInFooter: defaults.filtersInFooter,
     showDownload: false,
     quickdraw: false, // if true, components may skip expensive computes.
     mapAnimationDurationInMilliseconds: 30000, // in milliseconds


jameshadfield:
Nice find on the :bug: @joverlee521!

Notes looking at the codepaths involved here as part of review:

1. Loading datasets / changing narrative slides happens via a call to `changePage`, which has 3 cases ([description](https://github.com/nextstrain/auspice/blob/master/src/actions/navigation.js#L70-L72)):
    * The initial load is "case3" and doesn't result in the 🐛 
    * Moving to a narrative slide which changes the dataset is "case 2" and doesn't result in the 🐛 
    * Moving to a narrative slide where the dataset doesn't change is "case 1" and we have the 🐛 
2. "case1" triggers a call to `createStateFromQueryOrJSONs` which supplies existing redux state and a (URL) query and asks for the state to update accordingly
3. This results in the controls state being calculated via:
```js
controls = cloneDeep(oldState.controls);
controls = restoreQueryableStateToDefaults(controls); # only ever called here
controls = modifyStateViaURLQuery(controls, query);
```
4. Running `restoreQueryableStateToDefaults` will set `controls.filters = {}` ([src](https://github.com/nextstrain/auspice/blob/master/src/actions/recomputeReduxState.js#L166-L176)) but it won't affect `controls.filtersInFooter` as that's not part of `controls.defaults` (probably by mistake, but oh well!). Logging the state change in this function when we change slides shows:
```
controls.filters was {"clade_membership":[],"region":[]...} and is now {}
controls.filtersInFooter was [clade_membership,region,...] and is now [clade_membership,region...]
```
5. An empty `controls.filters` means that there are no available filters, so the dropdown doesn't work!

The root cause of the bug is in `restoreQueryableStateToDefaults`. The "default" in this case is not correct, as it is restoring to the "default app state before we have any data" rather than the "default state as determined by the dataset JSON".  One way to fix this is to mimic what we do when we first load JSONs, which is a 1-line change [here](https://github.com/nextstrain/auspice/blob/master/src/actions/recomputeReduxState.js#L788) to:

```js
controls = restoreQueryableStateToDefaults(controls);
controls = modifyStateViaMetadata(controls, metadata); # add this line
```

However a better fix would be to rewrite `restoreQueryableStateToDefaults`. This is along the lines of what you're doing in this PR, but it's the code inside `restoreQueryableStateToDefaults` which should be changed, rather than modifying it afterwards.

joverlee521:
Thanks for writing out the codepath notes @jameshadfield, it's really helpful to see it all! 

I'm in favor of just doing the 1-line change of adding `modifyStateViaMetadata`. This ensures that we restore all dataset default controls without worrying about missing something within `restoreQueryableStateToDefaults`.

jameshadfield (approved):
Thanks @joverlee521!

PR #1494, opened 2022-03-23
Scatterplots can display node order

jameshadfield:
See commits for details and [this internal slack thread](https://bedfordlab.slack.com/archives/CSKMU6YUC/p1647966053814009)

### Screenshots:

Node ordering vs time essentially recreates the rectangular tree

image

Node ordering vs antigenic advance

image

vs S1 mutations

image

Comparing with categorical traits which haven't "evolved" across the tree isn't so useful

image

vs logistic growth produces what looks like subtrees, but this is because most nodes don't have this annotated, so we are only drawing a subset of nodes, and branches only appear when both parent+child have values

image

Exploded trees aren't quite right, as the thin branches here shouldn't be rendered (they're not a real branch, it's an artefact of how we store subtrees). I'll take a look at this now.

image

### DIFF
--- a/src/components/tree/phyloTree/helpers.js
+++ b/src/components/tree/phyloTree/helpers.js
@@ -167,6 +167,14 @@ export function guessAreMutationsPerSite(scale) {
 }
 
 /**
+ * Is the node a subtree root node? (implies that we have either exploded trees or
+ * the dataset has multiple subtrees to display)
+ * @param {ReduxTreeNode} n
+ * @returns {bool}
+ */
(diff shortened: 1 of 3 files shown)


jameshadfield:
Subtree 🐛 fixed by 

Now each subtree appears on its own (in the node-ordering dimension), as it should

image

trvrb:
This is so cool James. Seems like such an obvious feature in hindsight. I really like how easily you can see distinct quantitative traits of different clades.

PR #1501, opened 2022-04-10
Show/hide non-AA branch labels according to tip count

trvrb:

This is a simple change that causes non "aa" branch labels to use current behavior of "aa" branch labels. This causes label visibility to be dynamically determined based on the size of the subtending clade relative to total in view tips.

There have been a lot of "clade" labels that appear in the middle of the tree and are not meaningful.

Previous behavior:

previous-1

Updated behavior:

updated-1

I think much cleaner. Zooming appropriately expands visibility.

### Testing

I've tested locally and with Heroku review app with a variety of SARS-CoV-2 trees and flu trees.

### DIFF
--- a/src/components/tree/phyloTree/labels.js
+++ b/src/components/tree/phyloTree/labels.js
@@ -78,11 +78,7 @@ const branchLabelFontWeight = (key) => {
  */
 const createBranchLabelVisibility = (key, layout, totalTipsInView) => (d) => {
   if (d.visibility !== NODE_VISIBLE) return "hidden";
-  if (key!=="aa") return "visible";
-  const magicTipFractionToShowBranchLabel = 0.05;
-  if (layout !== "rect") {
-    return "hidden";
-  }
+  const magicTipFractionToShowBranchLabel = 0.03;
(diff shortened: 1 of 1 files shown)


jameshadfield (commented):
Looks good, thanks @trvrb!

One change which might be unintentional is that aa labels are now shown on non-rectangular trees, whereas previously they were all hidden as long labels don't look particularly good:

image

trvrb:
Thanks @jameshadfield! This was indeed intentional as I could imagine situations where I want to annotate circular tree, unrooted tree or clock tree with AA mutations. It's often a bit messy, but you have to work at getting the app into this state (switching away from rectangular tree and switching to "aa" branch labeling).

Just as examples of what I think is valuable to surface:

flu

unrooted tree for flu Vic.

zika

radial tree for Zika.

That said long labels for ncov are indeed messy for non-rectangular trees.

If you you'd prefer to hide aa labels for non-rectangular trees, I'd be okay with it. Whatever your preference.

jameshadfield:
Happy to leave them for non-rectangular trees (users can always toggle them off, after all) I just wanted to ensure it was a deliberate change!

PR #1503, opened 2022-04-28
Show labels for subtree root nodes

jameshadfield:
This commit is related to the recent PR #1501 which hid (non-aa) labels
depending on their in-view tip-fraction. Here if the label is on the
root of a subtree then we always show it (unless the label is "aa" as
these can be very long).

Without this change, when exploding by "emerging lineage" only BA.1.1 would be labelled in the following screenshot (when not exploded, then only BA.1.1 is labelled, as per #1501)

image
image


trvrb:
Good call. Definitely agree with this addition.

PR #1504, opened 2022-05-04
add getDarkerColor() to colorHelpers to help ensure tip and legend fill colors match for categorical attributes

tomkinsc:
This addresses an issue where the fill colors in the legend for categorical metadata fields do not match the fill colors of tips in the tree (but the stroke colors do match). As an example, here are two screenshots inspecting fill colors from the [live nextstrain.org deployment of auspice](https://nextstrain.org/ncov/gisaid/global/6m?d=tree&f_submitting_lab=Infectious%20Disease%20Program,%20Broad%20Institute%20of%20Harvard%20and%20MIT&p=full) as of 2022-05-03:

tree_tip_color_inspection
legend_color_inspection

A little difficult to pinpoint, and it may yet be a quirk of my system (macOS 11.6 w/ Chrome 100.0.4896.75), but the issue could stem from imprecision in the use of Math.pow() in the [`d3-color` `brighter()` function](https://github.com/d3/d3-color/blob/main/src/color.js#L248-L251), used by auspice in a few places. I was interested in making sure the fill colors match between the legend and the tips (for ease of "select-all->replace" actions in Inkscape or Illustrator), so rather than brighten the fill, I set the fill to the color returned from the categorical color map and use the `darken()` function from `d3-color` to darken the stroke (which also uses Math.pow(), but it's less of a problem if the stroke colors do not match exactly). This change is implemented as the new function `getDarkerColor()` in `colorHelpers.js`, which is then used to adjust the stroke color in the files where `getBrighterColor()` was previously applied to the fill.

add `getDarkerColor()` as the inverse of `getBrighterColor()` (and remove the latter); set legend item fill and tree tip fill colors to be the same (returned from the color map), and adjust the stroke of each via `getDarkerColor() ` to show a contrasting border, rather than by setting the fill via getBrighterColor(). This helps ensure the fill color of the legend matches the fill of associated tips

### Related issue(s)
n/a

### Testing
Inspection of legend fill and stroke colors vs tip fill and stroke colors

### DIFF
--- a/src/components/measurements/measurementsD3.js
+++ b/src/components/measurements/measurementsD3.js
@@ -5,7 +5,7 @@ import { select, event as d3event } from "d3-selection";
 import { symbol, symbolDiamond } from "d3-shape";
 import { orderBy } from "lodash";
 import { measurementIdSymbol, measurementJitterSymbol } from "../../util/globals";
-import { getBrighterColor } from "../../util/colorHelpers";
+import { getDarkerColor } from "../../util/colorHelpers";
 
 /* C O N S T A N T S */
 export const layout = {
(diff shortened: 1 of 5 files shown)


jameshadfield:
Hi @tomkinsc -- thanks for identifying the :bug: I think this has gone unnoticed for many moons!

I dug into this a bit, and it was caused by brightening the colours by different amounts in the legend vs nodes -- I created #1506 to make these consistent - would you mind taking a look at that?

PR #1505, opened 2022-05-04
Call new GitHub Actions workflow to rebuild images

victorlin:

See https://github.com/nextstrain/docker-base/pull/41

Also, move docker rebuild to a separate job.

### Related issue(s)

_N/A_

### Testing

Unable to test until nextstrain/docker-base#41 is merged and a new Auspice version is released.

- [ ] Remove `TRAVIS_AUTH_TOKEN` from [org Actions secrets](https://github.com/organizations/nextstrain/settings/secrets/actions) if this is the last repo that's using it (see [query](https://github.com/search?q=TRAVIS_AUTH_TOKEN+org%3Anextstrain&type=code))


tsibley (approved):

.github/workflows/ci.yaml
```diff
@@ -81,7 +81,11 @@ jobs:
           registry-url: https://registry.npmjs.org/
       - run: node --version
       - run: npm ci
-      - run: npm run rebuild-docker-image
       - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+  deploy:
```
tsibley, line 87: Would slightly prefer another name here for this job, as it doesn't "deploy" Auspice. Perhaps `publish-image` or `rebuild-image`?

victorlin: Agreed, changed to `rebuild-docker-image`. I used `deploy` to be consistent with [Augur](https://github.com/nextstrain/augur/blob/9cd6eb489290536f8b9ad65c6d6c27843f19ca15/.github/workflows/ci.yaml#L52), but will update the name there in another PR.

PR #1506, opened 2022-05-04
consistent legend and tip (fill) colours

jameshadfield:
This resolves a (very very old!) bug described by @tomkinsc in
https://github.com/nextstrain/auspice/pull/1504 where the RGB values of
the legend fill were slightly different to the corresponding tip-fill.

The legend lightened by 0.35 whereas the tips lightened by 0.65. This
difference wasn't obvious (I don't think anyone noticed it with human
eyes) but it certainly resulted in different RGB values:

legend: `rgb(hex).brighter([0.35]).toString()`
tip:    `getBrighterColor(hex) -> rgb(color).brighter([0.65]).toString()`

This commit modifies the legend to use the helper function so all
brightening is performed in one place.

### DIFF
--- a/src/components/tree/legend/legend.js
+++ b/src/components/tree/legend/legend.js
@@ -4,6 +4,7 @@ import { rgb } from "d3-color";
 import LegendItem from "./item";
 import { headerFont, darkGrey } from "../../../globalStyles";
 import { fastTransitionDuration, months } from "../../../util/globals";
+import { getBrighterColor } from "../../../util/colorHelpers";
 import { numericToCalendar } from "../../../util/dateHelpers";
 import { isColorByGenotype, decodeColorByGenotype } from "../../../util/getGenotype";
 import { TOGGLE_LEGEND } from "../../../actions/types";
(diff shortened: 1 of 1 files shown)


jameshadfield:
One aesthetic change here is that now the legends look different (slightly worse, in my opinion, but only slightly). Current master (LHS) vs this PR (RHS):

image

I experimented with changing the lightening amount to 0.35 so the legends are unchanged, but then the tips look worse (again, in my opinion). Flagging this here to see if anyone feels strongly either way? I do think it's better that they have consistent RGB values, and I chose a nicer looking tree over a nicer looking legend 😉

emmahodcroft:
This is something that's driven me a little crazy a few times, when trying to manually adjust things in Inkscape - as you have to get a little creative in what you dropper-sample.

I agree that having consistency here is just easier in all the ways. I wonder if one thing that makes it 'worse' is that the outline looks much more visible in the 'new' version of the legend? (Presumably the same on the tree?) It seems like there's a much bigger difference between outline & fill than in the old version. Even though the colours themselves have also changed, I do wonder if maybe adjusting the outline so that it blends a bit better with the new colours (as it seems to do in the 'old' version), might mean that it looks less jarring to the eye?

tomkinsc:
Great! Thank you for the quick attention on this @jameshadfield. Exactly, @emmahodcroft, trying and failing to select tips matching categories in the legend via Inkscape `Select Same`->`Fill Color` is what prompted #1504 and this cleaner fix. 

I don't have a strong opinion on the aesthetics, though I wonder if outlines may not be needed in the legend since the whitespace delineates the boxes just fine, though I realize the tips also have an outline and the legend should probably match them...

jameshadfield:
I'm going to merge & release as-is. We can address the legend aesthetics in a later PR if we decide the change is detrimental. (If we do change the legend stroke colour by itself, then that'll just get us back to square one when people want to edit the SVG etc.)

Issue #1512, opened 2022-05-23
Truncated display on main monkeypox tree when zooming out


theosanderson:
### Current Behavior
On the monkeypox tree it is possible to get a truncated display like 

![image](https://user-images.githubusercontent.com/19732295/169917486-7fae8223-64ec-4b4f-be48-4a179017cafe.png)

### Expected behavior
One does not get this issue

### How to reproduce
1. Seems to happen more when the pixel-width of the tree is small (narrow window)
2. Zoom into the outbreak node
3. Move back, perhaps 2 nodes up until this happens

### Your environment: if browsing Nextstrain online
 - Operating system: macOS
 - Browser: Chrome
 - Version Auspice v2.37.1

### Additional context
The console spams messages like 
![image](https://user-images.githubusercontent.com/19732295/169917681-6f805fb9-ed24-44d3-87d1-b5125305946f.png) referring to NaNs in a <path

Apologies if this is a known issue

victorlin:
Hey @theosanderson, thanks for describing this! Yes this was brought up internally, and it's an Auspice issue so I plan to file an issue there.

In the meantime, the workaround is to make a slight adjustment the window size while it is buggy. This will make things appear correctly until the page is reloaded.

theosanderson:
Thanks, please feel free to close/edit this issue as desired!

victorlin:
@theosanderson whoops, updated my comment – didn't realize this was already the Auspice repo!

(Nice to see you outside of the the Serratus PRs 🙂)

jameshadfield:
#1513 should fix this, I think. I could never find an easily reproducible way to trigger the :bug:, but it's related to the clipping mask becoming out of sync with the tree scales. I'll push that version of auspice up to dev.nextstrain.org where it should be easier to test (and add a message here when I've done so).

PR #1513, opened 2022-05-25
Fix clip mask bug

jameshadfield:
The clip-mask uses the scale ranges and thus is prone to getting out-
of-sync, as described in https://github.com/nextstrain/auspice/issues/1512.
This commit should mitigate such bugs at the expense of regenerating
the mask more often.

Closes https://github.com/nextstrain/auspice/issues/1512

### DIFF
--- a/src/components/tree/phyloTree/change.js
+++ b/src/components/tree/phyloTree/change.js
@@ -384,10 +384,6 @@ export const change = function change({
     elemsToUpdate.add('.tipLabel'); /* will trigger d3 commands as required */
   }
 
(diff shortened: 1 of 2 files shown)


victorlin (approved):
Tested with [the example dataset I linked on Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1653463856585559?thread_ts=1653126530.954459&cid=C01LCTT7JNN), bug seems to be fixed.

jameshadfield:
Thanks @victorlin! Will hopefully merge #1514 shortly then release a new version

PR #1518, opened 2022-06-01
Make GitHub release when releasing new version

jameshadfield:
This follows the example set by Nextclade (@ivan-aksamentov) and should
result in https://github.com/nextstrain/auspice/releases staying
up-to-date. Note that this does not replace `npm` as the typical
way to install new auspice versions.

P.S. Slack discussion [here](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1653653815905769)

### DIFF
--- a/releaseNewVersion.sh
+++ b/releaseNewVersion.sh
@@ -2,9 +2,9 @@
 
 # VARIABLES
 read -r -d '' purposeMsg <<'EOF'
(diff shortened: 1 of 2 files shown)


victorlin:
Just adding a link to nextstrain/augur#957 which is trying to accomplish the same thing over there.

PR #1520, opened 2022-06-03
Upgrade webpack v4 → v5

victorlin:

[webpack 4 does not support Node.js v16 or above](https://github.com/webpack/webpack/issues/14532) without the [`--openssl-legacy-provider` option](https://nodejs.org/api/cli.html#--openssl-legacy-provider).

### Related issue(s)

Support for Node.js 18 (#1553) relies on this webpack upgrade.

### Pre-merge tasks

- Merge https://github.com/nextstrain/nextstrain.org/pull/634

### Testing

- [x] Checks pass
- [x] `auspice develop` and hot-reloading works ([comment](https://github.com/nextstrain/auspice/pull/1520#discussion_r977097904))
- [x] nextstrain.org `auspice-client` works after pre-merge tasks ([comment](https://github.com/nextstrain/auspice/pull/1520#discussion_r985331370))
- [x] Manually click around Heroku preview app for this PR
- [x] Create downstream preview PRs
    - https://github.com/nextstrain/nextstrain.org/pull/607
    - https://github.com/nextstrain/auspice.us/pull/43
- [ ] Manually click around downstream preview apps

### Other notable discussion threads

- Use a workaround for code-splitting ([ref](https://github.com/nextstrain/auspice/pull/1520#discussion_r966519420))
- Use a workaround for `auspice develop` ([ref](https://github.com/nextstrain/auspice/pull/1520#discussion_r987382659))


webpack.config.js
```diff
@@ -74,7 +74,7 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
   });
   /* gzip everything - https://github.com/webpack-contrib/compression-webpack-plugin */
   const pluginCompress = new CompressionPlugin({
-    filename: "[path].gz[query]",
+    filename: "[path][base].gz",
     algorithm: "gzip",
```
victorlin, line 78: Something weird is going on with code splitting not happening on `optimization.splitChunks.cacheGroups.coreVendors`.

The output follows `filename` instead of `chunkFilename` here:

https://github.com/nextstrain/auspice/blob/1d33b1e667320ba2c272054ba28a330f4c2d2c83/webpack.config.js#L189-L190

So the file can't be found by bundlesize ([link](https://github.com/nextstrain/auspice/runs/8260498558?check_suite_focus=true#step:5:5)):

```
 ERROR  There is no matching file for ./dist/auspice.chunk.core-vendors.bundle.*.js in /home/runner/work/auspice/auspice 
```

I tried adjusting `minChunks` and `minSize` but still can't get code splitting to happen. Putting this investigation off for another day.

victorlin: Seeme relevant: https://github.com/webpack/webpack/issues/9297

victorlin: Still don't fully understand this. Using a workaround: ae8ce43bc818a3dfb304eb55092c175b7ad4955d

package.json
```diff
@@ -7,8 +7,8 @@
   "repository": "github:nextstrain/auspice",
   "homepage": "https://www.npmjs.com/package/auspice",
   "engines": {
-    "node": "^12 || ^14 || ^16",
-    "npm": ">=6.9"
+    "node": "^14 || ^16 || ^18",
```
jameshadfield, line 10: I'm running node v18.9.0 / npm 8.9.2 and thus using the new lockfile format 👇 , so we're going to run into the [same bug we ran into for nextstrain.org](https://github.com/nextstrain/nextstrain.org/pull/587) I think? I'm not sure of a good path to mitigate this, other than having the developers + dependebot use the same node versions.

```
npm WARN old lockfile 
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile 
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile 
```

package.json
```diff
@@ -113,16 +113,17 @@
     "style-loader": "^0.13.2",
     "styled-components": "^4.0.3",
     "typeface-lato": "^0.0.75",
-    "webpack": "^4.30.0",
-    "webpack-bundle-analyzer": "^3.3.2",
-    "webpack-dev-middleware": "^3.1.3",
-    "webpack-hot-middleware": "^2.24.3",
+    "webpack": "^5.74.0",
+    "webpack-bundle-analyzer": "^4.6.1",
+    "webpack-dev-middleware": "^5.3.3",
```
jameshadfield, line 118: I'm getting the following error when in development mode - do you as well?
```
$ auspice develop
/Users/enigma/projects/nextstrain/auspice/node_modules/webpack-dev-middleware/node_modules/schema-utils/dist/validate.js:115
    throw new _ValidationError.default(errors, schema, configuration);
    ^

ValidationError: Invalid options object. Dev Middleware has been initialized using an options object that does not match the API schema.
 - options has an unknown property 'logLevel'. These properties are valid:
 ...
 ```

File: webpack.config.js (hunk 1 above)
Comment (jameshadfield) [reply webpack.config.js:78]: It's been a long time since I delved into the depths of webpack (and that was v4) so I don't understand what's happening here either!

The idea behind the coreVendors chunk is that it has a longer cache timeout, and thus browsers don't need to fetch it as often. This presumes that it's more stable between auspice releases than other chunks (which bundle up our code). 

Given our cadence of auspice releases and the actual sizes involved (the largest chunk is ~200kb uncompressed) I don't know if this development complexity is warranted anymore, but you seem to have got it working correctly again so we should be good for another few years!!!

**Review by jameshadfield — COMMENTED** [review]
Thanks for this work Victor -- nice to keep these things updated periodically!

It _looks_ like it's working with `auspice build && auspice view` but not `auspice develop`. Have you tried building nextstrain.org and/or auspice.us with it? The extension architecture was quite tricky to get integrated with webpack & I could imagine won't work out-of-the-box with webpack v5. 


When it comes time to merge, we should make sure we spend a bit of time testing on nextstrain canary to see if there are any strange bugs, especially cache-related ones.

File: webpack.config.js (hunk 1 above)
Comment (jameshadfield) [reply webpack.config.js:78]: It's been a long time since I delved into the depths of webpack (and that was v4) so I don't understand what's happening here either!

The idea behind the coreVendors chunk is that it has a longer cache timeout, and thus browsers don't need to fetch it as often. This presumes that it's more stable between auspice releases than other chunks (which bundle up our code).

Given our cadence of auspice releases and the actual sizes involved (the largest chunk is ~200kb uncompressed) I don't know if this development complexity is warranted anymore, but you seem to have got it working correctly again so we should be good for another few years!!!

File: package.json (hunk 3 above)
Comment (victorlin) [reply package.json:118]: Yes I see this too. Looks like something with the Webpack update. Thanks for checking, maybe it'd be good to check that `auspice develop` runs properly in CI as well.

File: package.json (hunk 2 above)
Comment (victorlin) [reply package.json:10]: > having the developers + dependebot use the same node versions

I think this is the ideal solution here, so we can avoid unnecessary diffs like https://github.com/nextstrain/nextstrain.org/pull/587. Based on https://github.com/dependabot/dependabot-core/issues/2643#issuecomment-775474764, it seems like dependabot is smart enough to not mess with the lockfile version so that shouldn't be an issue. We should also should consider what we use in production (e.g. [nextstrain.org is on Node 14](https://github.com/nextstrain/nextstrain.org/blob/adae78a29e34a38aba9bc91bf33d32ba0070976f/package.json#L11), which is why I choose to use it locally).

But it's also nice to have developers test on newer versions like you've been doing with Node v18. Ideally, errors with newer versions should be caught with CI, but there are still gaps such as https://github.com/nextstrain/auspice/pull/1520#discussion_r977097904. I'd say it's fine for you or other developers to use Node v18 as long as you switch back to Node 14 to make any `package-lock.json` updates.

File: package.json (hunk 3 above)
Comment (victorlin) [reply package.json]: The arises from behavior differences starting from [webpack-dev-middleware version 4.0.0](https://github.com/webpack/webpack-dev-middleware/blob/master/CHANGELOG.md?plain=1#L190).

I replaced ffe9375762677e2fa3ab53e51288be8d25783d6d with a different, more conservative upgrading approach which should fix this: 922581e022c6439be8186b26e99902f4bb935e01

File: package.json (hunk 2 above)
Comment (jameshadfield) [reply package.json:10]: Ideally we'd run the server on node v18 as it's slightly faster (due to V8 upgrade), but I don't know if this would be a measurable difference in our case (if we were to measure such things).

P.S. You're correct that [heroku uses the nodejs version listed in the package.json](https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version).

Comment (victorlin) [reply package.json:10]: I created a PR to update the server to v18, which requires merging this PR first: https://github.com/nextstrain/nextstrain.org/pull/595

**Comment by jameshadfield** [comment]
👍 thanks Victor -- I'll re-review this before the end of the week

File: package.json (hunk 3 above)
Comment (jameshadfield) [reply package.json]: I'm still not able to run `auspice develop` -- the terminal now looks ok but the client won't load. Console error: `Uncaught (in promise) ReferenceError: module is not defined`.

Furthermore the console error is coming from `auspice.chunk.core-vendors.bundle.js line 7145 > eval:8:3` which indicates that the development mode is actually running from the (minimised) bundles rather than using source maps to link errors back to the actual src code.

File: package.json [hunk 4]
```diff
@@ -113,16 +113,17 @@
     "style-loader": "^0.13.2",
     "styled-components": "^4.0.3",
     "typeface-lato": "^0.0.75",
-    "webpack": "^4.30.0",
-    "webpack-bundle-analyzer": "^3.3.2",
-    "webpack-dev-middleware": "^3.1.3",
-    "webpack-hot-middleware": "^2.24.3",
+    "webpack": "^5.74.0",
```
jameshadfield, line 116: [using this code line simply in order to start a thread]

I tested the extension architecture by building nextstrain.org with this version of auspice (see below). I remember the extension-architecture was difficulat as it involved lots of webpack configuration which (at the time) didn't have the best docs. I'd be happy to help debug this stuff with you, although you're probably the webpack expert after this PR :)

P.S. I haven't tested auspice.us, but that should be done.


```
## working directory: nextstrain.org
$ npm install nextstrain/auspice#9c1df8bcc7f42733077343218fb0e8e5003c704c
$ cd auspice-client/
$ ../node_modules/.bin/auspice build --verbose --extend ./customisations/config.json
[verbose]       Generating webpack config. Extensions? true. devMode: false
[verbose]       Webpack writing output to: /Users/enigma/projects/nextstrain/nextstrain.org/auspice-client/dist
Running webpack compiler
[verbose]       Generating Babel Config
[BABEL] Note: The code generator has deoptimised the styling of /Users/enigma/projects/nextstrain/nextstrain.org/node_modules/react-icons/fa/index.esm.js as it exceeds the max of 500KB.
assets by status 2.82 MiB [cached] 44 assets
Entrypoint main = auspice.core-vendors.bundle.d772f646f5d6159b105d.js auspice.main.bundle.efb7ad8e25040a23a1df.js 3 auxiliary assets
orphan modules 1.37 MiB [orphan] 518 modules
runtime modules 8.81 KiB 14 modules
javascript modules 7.12 MiB
  modules by path ./ 2.26 MiB
    modules by path ./node_modules/core-js/ 683 KiB 701 modules
    modules by path ./src/ 1.57 MiB
      cacheable modules 1.57 MiB 89 modules
      modules by path ./src/locales/ 382 bytes 2 modules
    ./node_modules/regenerator-runtime/runtime.js 24.3 KiB [built] [code generated]
  modules by path ../ 4.86 MiB 359 modules
  fs (ignored) 15 bytes [built] [code generated]
json modules 57.8 KiB 37 modules

ERROR in ../../auspice-client/customisations/languageSelector.js 33:0-44
Module not found: Error: Can't resolve 'iso-639-1/build/index' in '/Users/enigma/projects/nextstrain/nextstrain.org/auspice-client/customisations'
Did you mean 'index.js'?
BREAKING CHANGE: The request 'iso-639-1/build/index' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ../../auspice-client/customisations/ sync ^\.\/.*$ ./languageSelector.js ./languageSelector
 @ ./src/util/extensions.js 13:24-71
 @ ./src/util/googleAnalytics.js 10:0-58 19:16-28 44:31-43
 @ ./src/index.js 18:0-77 45:0-35
 @ ./src/indexAsync.js

ERROR in ../../auspice-client/customisations/navbar.js 38:0-50
Module not found: Error: Can't resolve './languageSelector' in '/Users/enigma/projects/nextstrain/nextstrain.org/auspice-client/customisations'
Did you mean 'languageSelector.js'?
BREAKING CHANGE: The request './languageSelector' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ../../auspice-client/customisations/ sync ^\.\/.*$ ./navbar.js ./navbar
 @ ./src/util/extensions.js 13:24-71
 @ ./src/util/googleAnalytics.js 10:0-58 19:16-28 44:31-43
 @ ./src/index.js 18:0-77 45:0-35
 @ ./src/indexAsync.js

ERROR in ../../auspice-client/customisations/splash.js 25:0-69
Module not found: Error: Can't resolve 'auspice/src/components/markdownDisplay' in '/Users/enigma/projects/nextstrain/nextstrain.org/auspice-client/customisations'
Did you mean 'index.js'?
BREAKING CHANGE: The request 'auspice/src/components/markdownDisplay' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ../../auspice-client/customisations/ sync ^\.\/.*$ ./splash.js ./splash
 @ ./src/util/extensions.js 13:24-71
 @ ./src/util/googleAnalytics.js 10:0-58 19:16-28 44:31-43
 @ ./src/index.js 18:0-77 45:0-35
 @ ./src/indexAsync.js

ERROR in ../../auspice-client/customisations/splash.js 26:0-30
Module not found: Error: Can't resolve './navbar' in '/Users/enigma/projects/nextstrain/nextstrain.org/auspice-client/customisations'
Did you mean 'navbar.js'?
BREAKING CHANGE: The request './navbar' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ../../auspice-client/customisations/ sync ^\.\/.*$ ./splash.js ./splash
 @ ./src/util/extensions.js 13:24-71
 @ ./src/util/googleAnalytics.js 10:0-58 19:16-28 44:31-43
 @ ./src/index.js 18:0-77 45:0-35
 @ ./src/indexAsync.js

4 errors have detailed information that is not shown.
Use 'stats.errorDetails: true' resp. '--stats-error-details' to show it.

webpack 5.74.0 compiled with 4 errors in 31508 ms
[error] Webpack built with errors. Exiting.
```

jameshadfield (changes requested):
Thanks for the efforts here Victor! Building auspice itself seems to work well - I couldn't detect any regressions using the app - but dev mode + extensions are still broken.

.github/workflows/ci.yaml
```diff
@@ -10,10 +10,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [12, 14, 16]
+        node: [14, 16, 18]
```
victorlin, line 13: Deprecation warnings from `npm ci` on Nodejs 18:

```
npm WARN deprecated source-map-resolve@0.5.3: See https://github.com/lydell/source-map-resolve#deprecated
npm WARN deprecated source-map-url@0.4.1: See https://github.com/lydell/source-map-url#deprecated
npm WARN deprecated sane@4.1.0: some dependency vulnerabilities fixed, support for node < 10 dropped, and newer ECMAScript syntax/features added
npm WARN deprecated request-promise-native@1.0.9: request-promise-native has been deprecated because it extends the now deprecated request package, see https://github.com/request/request/issues/3142
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated superagent@3.8.3: Please upgrade to v7.0.2+ of superagent.  We have fixed numerous issues with streams, form-data, attach(), filesystem errors not bubbling up (ENOENT on attach()), and all tests are now passing.  See the releases tab for more information at <https://github.com/visionmedia/superagent/releases>.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated puppeteer@2.1.1: Version no longer supported. Upgrade to @latest
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated formidable@1.2.6: Please upgrade to latest, formidable@v2 or formidable@v3! Check these notes: https://bit.ly/2ZEqIau
npm WARN deprecated iltorb@2.4.5: The zlib module provides APIs for brotli compression/decompression starting with Node.js v10.16.0, please use it over iltorb
npm WARN deprecated @hapi/topo@3.1.6: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/address@2.1.4: Moved to 'npm install @sideway/address'
npm WARN deprecated @hapi/bourne@1.3.2: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/hoek@8.5.1: This version has been deprecated and is no longer supported or maintained
npm WARN deprecated @hapi/joi@15.1.1: Switch to 'npm install joi'
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated core-js@3.19.1: core-js@<3.23.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. Please, upgrade your dependencies to the actual version of core-js.
npm WARN deprecated core-js@2.6.12: core-js@<3.23.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. Please, upgrade your dependencies to the actual version of core-js.
```

package.json
victorlin: Ok I also see this with both NodeJS 14 and 18. Thanks for flagging!

victorlin:
Split the changes into 3 PRs:

- Update webpack v4 → v5 (this PR)
- #1559 (this should be good to merge)
- #1560 (this relies on the webpack update)

babel.config.js
```diff
@@ -27,7 +27,7 @@ module.exports = function babelConfig(api) {
     "lodash"
   ];
   if (api.env("development")) {
-    plugins.push("react-hot-loader/babel");
+    plugins.push(["react-hot-loader/babel", { safetyNet: false }]);
```
victorlin, line 30: This fixes [the console error](https://github.com/nextstrain/auspice/pull/1520#discussion_r985327588) in `auspice develop`, but it seems weird and I don't understand why it's necessary only after the webpack upgrades (see 6b9fddf0e187ee5bf9f2bd9537eb0eee164156d2).

package.json
victorlin: @jameshadfield that console error should be fixed now, see https://github.com/nextstrain/auspice/pull/1520#discussion_r987382659. And _I think_ the development mode is using the actual source code now.

package.json
victorlin: Note that after [the PR splitting](https://github.com/nextstrain/auspice/pull/1520#issuecomment-1267359650), I'm using Node.js 14 for manual testing on this PR since it is now scoped to just webpack version changes.

package.json
victorlin: This works on my machine now, but I've created some PRs to test this in a more robust way: https://github.com/nextstrain/auspice/pull/1565, https://github.com/nextstrain/nextstrain.org/pull/599

jameshadfield: This also works on my machine - both the source maps and the hot reloading are working as expected 😄

package.json
jameshadfield: Nice one! Both the source maps and the hot reloading are now working as expected 😄

package.json
jameshadfield: I can still reproduce my errors above via the following (now using the latest commit)

```
npm install nextstrain/auspice#6b9fddf0e187ee5bf9f2bd9537eb0eee164156d2
cd auspice-client
../node_modules/.bin/auspice build --verbose --extend ./customisations/config.json
```

victorlin: You're right, the same errors show up in the [preview PR build](https://github.com/nextstrain/nextstrain.org/actions/runs/3527897039/jobs/5917442431).

My first thought was that this is related to https://github.com/nextstrain/nextstrain.org/issues/616, but the presence of successful preview PR builds that came after the ESM conversion ([example](https://github.com/nextstrain/nextstrain.org/actions/runs/3432471538/jobs/5721781208)) rules that out.

A search online shows that it is indeed related to the webpack 5 upgrade: https://github.com/webpack/webpack/issues/11467

I'm going to investigate this some more tomorrow.

victorlin: This is a simple fix on the auspice-client side: https://github.com/nextstrain/nextstrain.org/pull/634

I've noted in this PR's description that the above PR should be merged first.

babel.config.js
jameshadfield: 🤷 JS imports / require is messy business

victorlin: 😬

jameshadfield: Hot reloading is working well now 👍

webpack.config.js
```diff
@@ -186,15 +185,16 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
...
+      filename: `auspice.[name].bundle${!devMode ? ".[contenthash]" : ""}.js`,
       chunkFilename: `auspice.chunk.[name].bundle${!devMode ? ".[chunkhash]" : ""}.js`,
       publicPath: "/dist/"
     },
     resolve: {
-      alias: aliasesToResolve
-    },
-    node: {
-      fs: 'empty'
+      alias: aliasesToResolve,
```
jameshadfield, line 193: Something (perhaps this line?) is resulting in auspice.us not building coreectly. I tested this in the auspice.us repo by running `npm ci && npm install nextstrain/auspice#7be662dc4554319a3d3f02fc5d75cdf8ea5366d8` 

1. `npm run build` works, although it has a couple of deprecation warnings (I _think_ these are new)

```
WARNING in ../../auspice_client_customisation/splash.js 115:164-171
Should not import the named export 'version' (imported as 'version') from default-exporting module (only default export is available soon)
...
WARNING in ../../auspice_client_customisation/splash.js 115:194-214
Should not import the named export 'dependencies'.'auspice' (imported as 'dependencies') from default-exporting module (only default export is available soon)
```

2. Loading the app (`npm run start`) shows that the extensions didn't work -- the splash page is that of default auspice not auspice.us. I'm surprised there weren't build errors.

3. I tried to mimic https://github.com/nextstrain/nextstrain.org/pull/634, but auspice.us uses the `@auspice` webpack alias so it's not quite the same. Didn't work. Here's what I tried:

```diff
diff --git a/auspice_client_customisation/handleDroppedFiles.js b/auspice_client_customisation/handleDroppedFiles.js
index 96a93fc..b8a5e3e 100644
--- a/auspice_client_customisation/handleDroppedFiles.js
+++ b/auspice_client_customisation/handleDroppedFiles.js
@@ -1,10 +1,10 @@
-import { createStateFromQueryOrJSONs } from "@auspice/actions/recomputeReduxState";
-import { errorNotification, warningNotification } from "@auspice/actions/notifications";
-import { Dataset, addEndOfNarrativeBlock, getDatasetNamesFromUrl } from "@auspice/actions/loadData";
-import { parseMarkdownNarrativeFile } from "@auspice/util/parseNarrative";
-import { parseMarkdown } from "@auspice/util/parseMarkdown";
-import { isAcceptedFileType as isAuspiceAcceptedFileType } from "@auspice/actions/filesDropped/constants";
-import newickToAuspiceJson from "./parseNewick";
+import { createStateFromQueryOrJSONs } from "@auspice/actions/recomputeReduxState.js";
+import { errorNotification, warningNotification } from "@auspice/actions/notifications.js";
+import { Dataset, addEndOfNarrativeBlock, getDatasetNamesFromUrl } from "@auspice/actions/loadData.js";
+import { parseMarkdownNarrativeFile } from "@auspice/util/parseNarrative.js";
+import { parseMarkdown } from "@auspice/util/parseMarkdown.js";
+import { isAcceptedFileType as isAuspiceAcceptedFileType } from "@auspice/actions/filesDropped/constants.js";
+import newickToAuspiceJson from "./parseNewick.js";
 
 /* The following requires knowledge of how auspice works, is undocumented, and is liable to change since auspice
 doesn't officially expose these functions */
diff --git a/auspice_client_customisation/splash.js b/auspice_client_customisation/splash.js
index b06c118..73cf222 100644
--- a/auspice_client_customisation/splash.js
+++ b/auspice_client_customisation/splash.js
@@ -1,6 +1,6 @@
 import React from "react"; // eslint-disable-line
-import { handleDroppedFiles } from "./handleDroppedFiles";
-import { P, Bold, Title, NextstrainTitle, CenterContent, Line, GitHub } from './styles';
+import { handleDroppedFiles } from "./handleDroppedFiles.js";
+import { P, Bold, Title, NextstrainTitle, CenterContent, Line, GitHub } from './styles.js';
 import { version, dependencies } from "../package.json";
```

jameshadfield (changes requested):
Thanks victor - this stuff is a huge amount of work and involves so many testing iterations!

It's working well for me for auspice on it's own, but see above re: auspice.us. The only aspect I didn't test explicitly was the `cacheGroups` chunk behavior (as they have longer cache max-ages, although I can't seem to find where we're setting this anymore 🤷 ); there's no obvious changes to these chunk groups here so it should be ok.

webpack.config.js
victorlin: Fixed by adding 5 characters 🤦 (see [force-push](https://github.com/nextstrain/auspice/compare/33b5cbd7eb9a1c1ad73b10071b8c1c79640deeb9..1d2825d87b29476b4abcd6a242276a87a1b88399)).

After an afternoon of tinkering and many red herrings (which at least helped me understand Webpack a little better), I figured this out. Turns out that, without that change, `auspice view` thinks that it is running from Auspice's root directory, not auspice.us, and therefore serves `auspice/dist` instead of `auspice.us/dist`.

Now I'm asking myself, how did it work for nextstrain.org? My guess is that nextstrain.org doesn't use `auspice view` and instead serves the `auspice-client` files directly ([here](https://github.com/nextstrain/nextstrain.org/blob/8ff7529de28e81f6e56e6dbc08a8cfe4974af436/src/endpoints/static.js#L17-L18)). Does that sound right?

victorlin: Also, good to note that this was already observable in the auspice.us review apps https://github.com/nextstrain/auspice.us/pull/43 and https://github.com/nextstrain/auspice.us/pull/44:

image

So once those re-deploy with the new changes, I'd expect to see them working properly.

victorlin: Re: deprecation warnings, I've created https://github.com/nextstrain/auspice.us/pull/45 to address that.

jameshadfield: > My guess is that nextstrain.org doesn't use auspice view and instead serves the auspice-client files directly

Correct!

I'll re-test this today. Future to-do:  improve the message "Serving auspice version ${version}" as if that indicated it was using the non-auspice.us version it probably would have quickened the search for this bug.

victorlin: > Future to-do

Done with #1608.

This is the output of `npm run start` from 33b5cbd7eb9a1c1ad73b10071b8c1c79640deeb9 (pre-fix):

```
---------------------------------------------------
Auspice server now running at http://localhost:4000
Serving auspice version 2.41.0
Custom server handlers provided.
---------------------------------------------------
```

This is the output after applying changes from #1608 (it's an error so it does not actually run the server):

```
ERROR: Auspice build files not found under /Users/vlin/Dropbox/tmp/auspice.us. Did you run `auspice build` in this directory?
```

cli/view.js
```diff
@@ -82,7 +82,7 @@ const getAuspiceBuild = () => {
     cwd !== sourceDir &&
     fs.existsSync(path.join(cwd, "index.html")) &&
     fs.existsSync(path.join(cwd, "dist")) &&
-    fs.readdirSync(path.join(cwd, "dist")).filter((fn) => fn.match(/^auspice.bundle.[a-z0-9]+.js$/)).length === 1
+    fs.readdirSync(path.join(cwd, "dist")).filter((fn) => fn.match(/^auspice.main.bundle.[a-z0-9]+.js$/)).length === 1
```
tsibley, line 85: I noticed in passing that those `.` should be escaped in the regex since they're meant to be literals.

victorlin: Good catch, thanks!

webpack.config.js
jameshadfield: 1d2825d fixes this 🎉

cli/view.js
jameshadfield: @victorlin - I'm going to fix this now (it was originally my oversight) as this'll allow multiple PRs to be merged

jameshadfield: Actually scrap that, let's fix this over in https://github.com/nextstrain/auspice/pull/1611

PR #1531, opened 2022-06-28
fix: prevent crash on unexpected gisaid_epi_isl

joverlee521:
In cases where the `gisaid_epi_isl` value does not fit the expected
format with two underscores ("EPI_ISL_\<number\>"), Auspice crashes when
you click on a tip with an error:
```
TypeError: Cannot read properties of undefined (reading 'length')
```
This commit adds a check to verify the `gisaid_epi_isl` value matches
our expected format with the regex. The regex specifies that the
accession number must be at least 5 digits so that it can be sliced
to create the proper URL.

### Testing
I ran into this crash when I was testing if we can add seasonal flu GISAID accessions to the acknowledgement's table by changing the `accession` attribute to `gisaid_epi_isl`. The seasonal flu accessions are the sequence segment accessions instead of the virus isolate accessions, so it did not fit the expected format for `gisaid_epi_isl`. If you go to the https://next.nextstrain.org/staging/with-gisaid-accession/flu/seasonal/h3n2/ha/2y and click on a tip, the app crashes. 

This change prevents the crash in my local testing with the same JSON and properly adds the default `https://gisaid.org` URL.

### DIFF
--- a/src/components/tree/infoPanels/click.js
+++ b/src/components/tree/infoPanels/click.js
@@ -60,11 +60,12 @@ const AccessionAndUrl = ({node}) => {
   let gisaid_epi_isl_url = null;
   let genbank_accession_url = null;
   if (isValueValid(gisaid_epi_isl)) {
-    const gisaid_epi_isl_number = gisaid_epi_isl.split("_")[2];
-    // slice has to count from the end of the string rather than the beginning to deal with EPI ISLs of different lengths, eg
-    // https://www.epicov.org/acknowledgement/67/98/EPI_ISL_406798.json
-    // https://www.epicov.org/acknowledgement/99/98/EPI_ISL_2839998.json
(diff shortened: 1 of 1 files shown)


corneliusroemer:
>This commit adds a check to verify the gisaid_epi_isl value matches
our expected format with the regex. The regex specifies that the
accession number must be at least 5 digits so that it can be sliced
to create the proper URL.

What happens when an invalid epi isl is passed? Is there a reason we need to validate so tightly?

joverlee521:
> What happens when an invalid epi isl is passed? Is there a reason we need to validate so tightly?

Ah, sorry my original commit message was not clear here. Values that do not match the regex are given the default GISAID URL ("https://gisaid.org/"). This strict check only affects the generated link for the accession number when a user clicks on a tip in the tree. (I've also add the explanation to the commit message in 2c35f5a)

PR #1533, opened 2022-07-13
Docs fixes

jameshadfield:
Changes originally motivated by [nextstrain/auspice.us#34](https://github.com/nextstrain/auspice.us/pull/34) - see commit messages for more.


docs/server/overview.rst
```diff
@@ -62,24 +62,17 @@ But if you _really_ want to implement your own server, then you only need to imp
...
-`We use option (b) <https://github.com/nextstrain/auspice/blob/master/package.json>`_ by specifing a npm script called `heroku-postbuild`.
-
 
-.. toctree::
-   :maxdepth: 0
-   :titlesonly:
-   :caption: Pages available
+1. You must set the config var `HOST=0.0.0.0` for the app.
+2. You will need to either create a `Procfile` or a `npm run start` script which calls `auspice view` (or `npx auspice view` depending on how you implement auspice).
+3. Make sure the datasets to be served are either (a) included in your git repo or (b) downloaded by the heroku build pipeline. `We use option (b) <https://github.com/nextstrain/auspice/blob/master/package.json>`_ by specifing a npm script called `heroku-postbuild`.
```
victorlin, line 77: Fixup of 61b87b2a0a2fd52d9cda8041229ff5eebb45fb11:
```suggestion
1. You must set the config var ``HOST=0.0.0.0`` for the app.
2. You will need to either create a ``Procfile`` or a ``npm run start`` script which calls ``auspice view`` (or ``npx auspice view`` depending on how you implement auspice).
3. Make sure the datasets to be served are either (a) included in your git repo or (b) downloaded by the heroku build pipeline. `We use option (b) <https://github.com/nextstrain/auspice/blob/master/package.json>`_ by specifing a npm script called ``heroku-postbuild``.
```

victorlin (changes requested):
One more formatting fix, but otherwise looks good!

docs/server/overview.rst
jameshadfield: Thanks @victorlin - fixed up (as well as other examples in that file). It's frustrating how hard rst is after using markdown for a long time

tsibley: You can change the default role (what `\`…\`` introduces in rST) for a file using the `default-role` directive, e.g. to make single backticks mean the `literal` role instead of the `title` role (the default), do:

```rst
.. default-role:: literal

and now `foo` will be treated as :literal:`foo` or ``foo``
instead of :title:`foo`.
```

PR #1535, opened 2022-07-18
docs: Install nextstrain.sphinx.theme extension

victorlin:

The new extension has sphinx_copybutton pre-installed and configured. This change enables it here.
This also makes it easier to configure other extensions across multiple docs projects.

## Related issue(s)

https://github.com/nextstrain/sphinx-theme/issues/30

## Testing

- [ ] copy button works on RTD preview build


tsibley (changes requested):
+1 but collapse the commits together since they're co-dependent?

PR #1550, opened 2022-09-20
controls/color-by: make nucleotide the first genotype option

### LINKED ISSUE #1548: ENH: Show nucleotides as first item in list for genotype coloriung
### Context
I often colour by genotype. Particularly in monkeypox with ~200 genes, it's annoying that nucleotides come last. The chance that I care about the first gene in the list is low as there are 200 of them. It would be nice as a result if nucleotides came first.

As nucleotide is probably the most relevant unless there's a gene of particular interest.

It could be best if one could specify which genes/nt should come first. In SC2, for example, it'd be great to start with S, then have nucleotides.

Maybe it's possible to customize the sequence already and I'm just not aware of it? In that case, please close @jameshadfield

joverlee521:
Users have requested that the "nucleotide" option should be the first option in the genotype color-by dropdown. This is especially useful for monkeypox, which has ~200 genes.

Resolves #1548.

### DIFF
--- a/src/components/controls/color-by.js
+++ b/src/components/controls/color-by.js
@@ -136,10 +136,12 @@ class ColorBy extends React.Component {
   }, 400);
 
   getGtGeneOptions() {
-    let options = [];
+    const options = [];
     if (this.props.geneMap) {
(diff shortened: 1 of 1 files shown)


jameshadfield (approved):
Thanks @joverlee521 🙏

PR #1551, opened 2022-09-21
docs: Update Node.js references

victorlin:

The current docs are unclear about which versions of Node.js are supported, and recommend an outdated version for installation.

### Related issue(s)

_N/A_

### Testing

- [x] Preview of changes looks good

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ We recommend using a conda environment, but this is not the only way.
 Create and activate a [conda](https://docs.conda.io) environment:
 
 ```bash
-conda create --name auspice nodejs=12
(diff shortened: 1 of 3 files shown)


jameshadfield (approved):
Looks good. In anther PR we should update `engines.node`. I'll make an issue...

[v12 has reached end-of-life and should be dropped](https://github.com/nodejs/release#release-schedule)

We should support v18. I'm running it day-to-day without problems, apart from needing the the `NODE_OPTIONS=--openssl-legacy-provider` flag (I can't remember the specifics of _why_ this is needed).

PR #1552, opened 2022-09-21
Bugfix when tree zooms back too far

### LINKED ISSUE #1529: BUG: `-` lens sometimes zooms out 2 steps rather than 1
### Current Behavior
When clicking the `-` lens, in some cases, Auspice zooms out more than one step

### Expected behavior
Auspice zooms out one internal node at a time

### How to reproduce
1. Open https://nextstrain.org/monkeypox/hmpxv1?c=recency&label=clade:B.1
2. Click on the top branch
3. Click the `-` lens
4. Observe result

The error is reproducible given this tree, I've attached the json here for reproducibility in case we replace the build

Remove the `.txt` ending, it's just present so that I can upload into Github:
[zoomerror.json.txt](https://github.com/nextstrain/auspice/files/8986943/zoomerror.json.txt)

### Your environment: if running Nextstrain locally
 - Operating system:
 - Browser:
 - Version (e.g. `auspice 2.7.0`): 2.37.3

jameshadfield:
Current behavior (from https://github.com/nextstrain/auspice/issues/1529#issuecomment-1253071116):

We want to zoom back to the parent branch, unless that branch is part of some polytomy-like structure, in which case we want to keep walking up the tree until we're beyond the polytomy. Determining whether a branch length is so small it's in a polytomy is the root of the problem here. We use a heuristic which takes the length of the branch from the root node to the rightmost node in the current view (in the example here when zoomed into B.1.1 this is 563, but the actual numerical value can differ by many orders of magnitude depending on the dataset and what zoom you're using); if a branch is less than 1/50th of this then we keep walking up the tree. In the example here, the branch for B.1.1 has length 1 and 1<563/50 so we try the next branch, B.1, which is 12 and 12>563/50 so we stop there.

New behavior:

For trees viewed with the divergence metric and with mutations defined, it is sensible (and easy) for the zoom back button ("-" button) to walk up the tree until a branch with mutations is found. This will behave nicely with the common scenario of polytomies encoded by a bifurcating tree (which often happens during coalescent reconstruction in augur refine).

The original heuristic remains for those trees where mutations are not defined on the branches. There are cleverer ways to infer the correct threshold here if we want to, but I don't think it's worth it at the moment.

Closes #1529

### DIFF
--- a/src/components/tree/phyloTree/helpers.js
+++ b/src/components/tree/phyloTree/helpers.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-param-reassign */
 import { max } from "d3-array";
-import {getTraitFromNode, getDivFromNode} from "../../../util/treeMiscHelpers";
+import {getTraitFromNode, getDivFromNode, getBranchMutations} from "../../../util/treeMiscHelpers";
 
 /** get a string to be used as the DOM element ID
  * Note that this cannot have any "special" characters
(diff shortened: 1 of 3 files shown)


src/components/tree/phyloTree/helpers.js
```diff
@@ -128,9 +128,18 @@ function isWithinBranchTolerance(node, otherNode, distanceMeasure) {
     const tolerance = (node.shell.that.dateRange[1]-node.shell.that.dateRange[0])/50;
     return (getTraitFromNode(node, "num_date") - tolerance < getTraitFromNode(otherNode, "num_date"));
   }
-  /* Compute the divergence tolerance similarly to above. This uses the approach used to compute the
-  x-axis grid within phyotree, and could be refactored into a helper function. Note that we don't store
-  the maximum divergence on the tree so we use the in-view max instead */
+
+  /* divergence tree metric -- if there are any actual mutations
+  (i.e. not deletions / undeletions / gaps) then we are outside the tolerance */
+  const categorisedMutations = getBranchMutations(node, {}); // 2nd param of `{}` means we skip homoplasy detection
```
joverlee521, line 134: > walk up the tree until a branch with mutations is found

Based on this comment, I was expecting to get mutations from `otherNode` here.

jameshadfield: Hmm, could you say more? The mutations on the branch between `otherNode` and `node` are encoded as a branch_attr on `node`

joverlee521: Hmm, maybe I'm not understanding the underlying branch/node structure?

Within `getParentBeyondPolytomy`, the `potentialNode` is the only thing changing in the while loop, so I thought we had to check the branch mutations of the `potentialNode`. 

https://github.com/nextstrain/auspice/blob/07b66e80c61aea344b73f8c9aee12213a54c1a60/src/components/tree/phyloTree/helpers.js#L148-L154

jameshadfield: Huh. Now i'm confused as to how it worked when I tested it. What we want to do is check `node` first, then `node.parent`, then `node.parent.parent`, but it seems we're only ever checking `node`. I'll dive into this now. Thanks for the good 👀 !

jameshadfield: Ahh yeah this is broken in two places :( 

1. I assumed the both `node` and `otherNode` were updated each time through the loop, but they're not!
2. On branches without mutations, my code interpreted these as a tree without mutations and so it fell back to the old heuristic. 

I'll fix things up and re-push

jameshadfield: Fixed! @joverlee521 would you take a look at [2b3086b](https://github.com/nextstrain/auspice/pull/1552/commits/2b3086b8d0057ee897a4724d19efff3904b83cf5)? 🙏

src/components/tree/phyloTree/helpers.js
```diff
@@ -128,32 +128,63 @@ function isWithinBranchTolerance(node, otherNode, distanceMeasure) {
...
   while (isWithinBranchTolerance(node, potentialNode, distanceMeasure)) {
     if (potentialNode === potentialNode.parent) break; // root node of tree
     potentialNode = potentialNode.parent;
   }
   return potentialNode;
 };
 
+function areNucleotideMutationsPresent(observedMutations) {
+  const mutList = Object.keys(observedMutations);
+  for (let idx=mutList.length-1; idx>=0; idx--) { // start at end, as nucs come last in the key-insertion order
```
joverlee521, line 180: Can you point to where the key-insertion order can be found? (I'm still trying to piece everything together 😅 )

From what I can find, it looks like the order is dependent on keys in `n.branch_attrs.mutations`

https://github.com/nextstrain/auspice/blob/07b66e80c61aea344b73f8c9aee12213a54c1a60/src/util/treeJsonProcessing.js#L145-L156

Which I think is based on the order of insertion [within `augur export`](https://github.com/nextstrain/augur/blob/e1e8bb155c25a2901bcd41423b0f67d14386b6b3/augur/export_v2.py#L700-L707)? 
(boo, cross-repo code snippets don't work)
```python
    def _transfer_mutations(node, raw_data):
        if "aa_muts" in raw_data or "muts" in raw_data:
            node["branch_attrs"]["mutations"] = {}
            if "muts" in raw_data and len(raw_data["muts"]):
                node["branch_attrs"]["mutations"]["nuc"] = raw_data["muts"]
            if "aa_muts" in raw_data:
                aa = {gene:data for gene, data in raw_data["aa_muts"].items() if len(data)}
                node["branch_attrs"]["mutations"].update(aa)
```

joverlee521 (approved):
Changes look good to me, thanks for fixing it up!
I just left a question for my own understanding 🙏

src/components/tree/phyloTree/helpers.js
jameshadfield: I don't actually know - but I would presume you're right here. This means it's related to https://github.com/nextstrain/augur/issues/1052, which if we move to OrderedDict will make "nuc" come first for GenBank files (created by `load_features`, since "source" comes before CDSs in genbank files) and last for VCF files (as the translate code adds them explicitly). (That's from reading the code, I haven't double checked everything.)

In testing for this PR the "nuc" block consistently came last. In terms of code performance it's not overly important, but it seemed sensible to iterate this way. I'll make a note on the augur issue to revisit this auspice code in the future.

jameshadfield: A further note: it would be more efficient to calculate and store a "areNucleotideMutationsPresent" flag in a redux store rather than re-calculating it each time one presses the "goBack" button. I decided that the slight performance hit (not noticeable in my testing) was preferable to the overhead of this state & keeping it in sync as trees change etc.

corneliusroemer:
How can I test this on a dataset where I've encountered this bug? None of them are included as test datasets. Can I drag and drop? Would be nice if test builds were like auspice.us :)

Or is it easy to push so that any test dataset appears there? What's the bucket?

jameshadfield:
> How can I test this on a dataset where I've encountered this bug? 

This prompted #1558 - apologies that testing wasn't easier here. I'm going to merge & release now as I've just re-tested things.

PR #1554, opened 2022-09-21
toggle to show all branch labels

jameshadfield:
We implement a heuristic to only show certain branch labels, which changes as you zoom the tree. This works well in most situations (indeed, showing all labels is the wrong choice in most situations!) however it's really useful to have the option to show them all. I commonly want this when using nextclade's nCoV reference tree where I want to see every clade.


emmahodcroft:
This would also be useful on things like the MPX build where one wants to just look at all the lineages when zoomed in, even if small. 

I didn't read through all the source code - is there a chance this can be toggled by auspice config? For some builds it may be nice to have this always-on (thinking of D68 here, as always).

jameshadfield:
> I didn't read through all the source code - is there a chance this can be toggled by auspice config? For some builds it may be nice to have this always-on (thinking of D68 here, as always).

Good point. It's not possible via this PR, but it would be nice to have in the future (as well as URL-configurable).

In general I think a good development path is to do things in 2 steps:
(1) Add the feature (e.g. this PR)
(2) If it's working well then make it URL-settable + dataset JSON settable.

Issue #1463, opened 2022-02-14
Measurements panel improvements


joverlee521:
**Context**  
Improvements for the Measurements panel added in https://github.com/nextstrain/auspice/pull/1452.

**Description**   
- [ ] Add URL parameters for the measurements panel options
- [ ] Order subplots by custom order from JSON
- [ ] Change ordering of subplots by field or calculation across fields (see [Slack discussion](https://bedfordlab.slack.com/archives/C2G8C5LUD/p1643305652009300?thread_ts=1643246971.008600&cid=C2G8C5LUD))
- [ ] 2 x group-bys (see [Slack discussion](https://bedfordlab.slack.com/archives/C2G8C5LUD/p1642068645003300))

huddlej:
Playing with the measurements panel more today, I had a few comments/thoughts to add to the list above.

1. +1 for handling the long card titles. In [the example community build](https://nextstrain.org/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17?p=grid), I can only get the tree and measurements to appear side-by-side when I close the control panel.

Control panel open:
![FE1B2B16-696F-4431-9ABF-C1A041135BA9](https://user-images.githubusercontent.com/85372/157347983-f83660db-e74b-4823-a0cf-c75d910694b8.png)

Control panel closed:
![5EEC8CB8-01D6-4EC1-A79D-FCF6C7A07893](https://user-images.githubusercontent.com/85372/157347999-abd3f4b7-bf92-4080-ba94-e8a7e50cbf94.png)

2. In grid mode, there is no legend for measurement panel colors. It would be nice if the tooltip for grouped mean +/- SD mode also displayed the value associated with the color-by. 

<img width="470" alt="name of color-by value in tooltip would help understand plots" src="https://user-images.githubusercontent.com/85372/157348143-f08460a0-867a-47e9-8844-f08f2ab2722f.png">

3. What do colors mean in measurements panel when coloring the tree by a continuous attribute like number of epitope mutations? Related to the point above about tooltips, it would be helpful to include the group info in the tooltips for the mean +/- SD mode.

![D2C4F6B1-4726-432F-AC44-474D5E308E9F](https://user-images.githubusercontent.com/85372/157348268-e259038e-8b70-4b7a-a075-dd6ea32ac02e.png)

[...]

PR #1557, opened 2022-10-03
Measurements improvements

joverlee521:
Small improvements to the measurements panel that includes:
- truncate long panel titles to prevent panel from being pushed to new line in grid mode
- allow "0" values for display in hover panel
- add color-by attribute as title to hover panel
- scale y-axis label to fit 
- add color-by attribute to the y-axis labels if available (e.g. reference strains)

### Related issue(s)

Partially addresses #1463

### DIFF
--- a/src/components/measurements/measurementsD3.js
+++ b/src/components/measurements/measurementsD3.js
@@ -24,12 +24,12 @@ export const layout = {
   thresholdStroke: "#DDD",
   subplotFill: "#adb1b3",
   subplotFillOpacity: "0.15",
-  diamondSize: 25,
-  standardDeviationStroke: 2,
+  diamondSize: 50,
+  standardDeviationStroke: 3,
   overallMeanColor: "#000",
   yAxisTickSize: 6,
-  yAxisColorByLineHeight: 7,
-  yAxisColorByLineStrokeWidth: 2
+  yAxisColorByLineHeight: 9,
+  yAxisColorByLineStrokeWidth: 4
 };
 // Display overall mean at the center of each subplot
 layout['overallMeanYValue'] = layout.subplotHeight / 2;
--- a/src/components/measurements/measurementsD3.js
+++ b/src/components/measurements/measurementsD3.js
@@ -105,7 +105,7 @@ export const clearMeasurementsSVG = (ref) => {
     .selectAll("*").remove();
 };
 
-const drawMeanAndStandardDeviation = (values, d3ParentNode, containerClass, color, xScale, yValue, handleHover) => {
+const drawMeanAndStandardDeviation = (values, d3ParentNode, containerClass, color, xScale, yValue) => {
   const meanAndStandardDeviation = {
     mean: mean(values),
     standardDeviation: deviation(values)
@@ -122,13 +122,7 @@ const drawMeanAndStandardDeviation = (values, d3ParentNode, containerClass, colo
     .attr("class", classes.mean)
     .attr("transform", (d) => `translate(${xScale(d.mean)}, ${yValue})`)
     .attr("d", symbol().type(symbolDiamond).size(layout.diamondSize))
-    .attr("fill", color)
-    .on("mouseover", (d) => {
-      // Get mouse position for HoverPanel
-      const { clientX, clientY } = d3event;
-      handleHover(d, "mean", clientX, clientY);
-    })
-    .on("mouseout", () => handleHover(null));
+    .attr("fill", color);
 
   if (meanAndStandardDeviation.standardDeviation !== undefined) {
     container.append("line")
@@ -138,17 +132,11 @@ const drawMeanAndStandardDeviation = (values, d3ParentNode, containerClass, colo
       .attr("y1", yValue)
       .attr("y2", yValue)
       .attr("stroke-width", layout.standardDeviationStroke)
-      .attr("stroke", color)
-      .on("mouseover", (d) => {
-        // Get mouse position for HoverPanel
-        const { clientX, clientY } = d3event;
-        handleHover(d, "mean", clientX, clientY);
-      })
-      .on("mouseout", () => handleHover(null));
+      .attr("stroke", color);
   }
 };
 
-export const drawMeasurementsSVG = (ref, svgData, handleHover) => {
+export const drawMeasurementsSVG = (ref, svgData) => {
   const {xScale, yScale, x_axis_label, threshold, groupingOrderedValues, groupedMeasurements} = svgData;
 
   // Do not draw SVG if there are no measurements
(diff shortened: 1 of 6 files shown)


src/components/measurements/measurementsD3.js
```diff
@@ -138,17 +132,11 @@ const drawMeanAndStandardDeviation = (values, d3ParentNode, containerClass, colo
...
       .attr("y2", yValue)
       .attr("stroke-width", layout.standardDeviationStroke)
-      .attr("stroke", color)
-      .on("mouseover", (d) => {
-        // Get mouse position for HoverPanel
-        const { clientX, clientY } = d3event;
-        handleHover(d, "mean", clientX, clientY);
-      })
-      .on("mouseout", () => handleHover(null));
+      .attr("stroke", color);
```
jameshadfield: > With the color-by attribute titles, the handleHover function will be dependent on color-by changes, which we do _not_ want to force SVG redrawing.

Could you elaborate on this a bit more? Changes in color-by will necessitate some degree of SVG redrawing -- at least the lines (mean+sd) / circles (raw). Does this change stop the `drawMeasurementsSVG` re-rendering, as this does not actually set the colouring [1], whereas (e.g.) `drawMeansForColorBy` will still rerun as required (e.g. when color-by changes)?

[1] If I understand the code, it does run `drawMeanAndStandardDeviation` which sets the color to `layout.overallMeanColor`, but this just adds the elements to the DOM in a non-visible fashion. We need a second call to `drawMeanAndStandardDeviation` to show/color them. Similar for the raw dots.

src/components/measurements/measurementsD3.js
```diff
@@ -27,14 +27,17 @@ export const layout = {
   diamondSize: 25,
   standardDeviationStroke: 2,
   overallMeanColor: "#000",
-  yAxisTickSize: 6
+  yAxisTickSize: 6,
+  yAxisColorByLineHeight: 7,
```
jameshadfield, line 31: I found the current line widths too subtle, but I understand you've picked them to mirror the SD lines. I personally preferred things a bit bigger  👇 but I haven't tested things on all resolutions. Not blocking, just a comment.

```js
  diamondSize: 50,
  standardDeviationStroke: 3,
  overallMeanColor: "#000",
  yAxisTickSize: 6,
  yAxisColorByLineHeight: 9,
  yAxisColorByLineStrokeWidth: 4
```

src/components/measurements/measurementsD3.js
```diff
@@ -365,3 +368,36 @@ export const addHoverPanelToMeasurementsAndMeans = (ref, handleHover, treeStrain
...
+export const addColorByAttrToGroupingLabel = (ref, treeStrainColors) => {
+  const svg = select(ref);
+  // Remove all previous color-by labels for the y-axis
+  svg.selectAll(`.${classes.yAxisColorByLabel}`).remove();
+  // Loop through the y-axis labels to check if they have a corresponding color-by
+  svg.selectAll(`.${classes.yAxis}`).select(".tick")
+    .each((_, i, elements) => {
+      const groupingLabel = select(elements[i]);
+      const groupingValue = groupingLabel.text();
+      const groupingValueColorBy = treeStrainColors[groupingValue];
```
jameshadfield, line 381: These colours are only set if a reference strain is visible - for instance change the date range slider and notice the reference strain colours start to disappear. I haven't walked up the code here to find out why this is, but ideally the colours should be here regardless of the visibility of the underlying reference strain. If this proves hard to implement we could add it as a to-do to #1463

src/components/measurements/measurementsD3.js
```diff
@@ -211,8 +212,26 @@ export const drawMeasurementsSVG = (ref, svgData) => {
...
-          .tickFormat(() => groupingValue))
-      .call((g) => g.attr("font-family", null));
+          .tickSize(layout.yAxisTickSize)
+          .tickFormat(groupingValue))
+      .call((g) => {
+        g.attr("font-family", null);
+        // If necessary, scale down the text to fit in the available space for the y-Axis labels
+        // This does mean that if the text is extremely long, it can be unreadable.
+        // We can improve on this by manually splitting the text into parts that can fit on multiple lines,
+        // but there's always limits of the available space so punting that for now.
```
jameshadfield, line 225: I think it's worth spending 30 minutes to see if an acceptable solution can be found here. There's _heaps_ of vertical pixel space [1]. Splitting on spaces, hypens, slashes etc should allow a reasonably good result.

[1] at least, there is in the measurements paper's h3n2 dataset I'm using to test. Even when colouring by a unchanging genotype, so there can only ever be one category. Perhaps in some configurations this is different?

jameshadfield (changes requested):
Nice work @joverlee521! I don't actually use the panel day-to-day, so it'd be good to get :eyes: from someone who does as well.

jameshadfield:
More of a future to-do, I think, but here are some suggestions to enhance hovering by calling the `updateTipRadii` thunk:
* hovering over (raw) circles should highlight the corresponding tip in the tree. 
* hovering over mean+sd lines/diamonds should highlight the corresponding tips in the tree.
* hovering over the category (the y-axis label) should  highlight the corresponding tip in the tree if it's a reference sera. 

Since I'm here, we currently only colour the y-axis if it's a reference sera (and happens to be present in the tree). But for other groupings we can map them to their corresponding reference sera, and then find the set of colours they represent. So there is a valid colour / set of colours for the other groupings as well. I imagine this'd require a bit of prototyping to get right, but could be an interesting direction.

src/components/measurements/measurementsD3.js
joverlee521: > Does this change stop the drawMeasurementsSVG re-rendering, as this does not actually set the colouring [1], whereas (e.g.) drawMeansForColorBy will still rerun as required (e.g. when color-by changes)?

Yes, this is exactly why I made this change. Previously, `handleHover` was directly tied to the redrawing of the entire SVG. Now, it its only tied to the coloring changes so we don't have to worry about unnecessary SVG redrawings.

src/components/measurements/measurementsD3.js
joverlee521: Ah thanks for pointing that out! This should be easy to tweak. I filtered colors to only visible strains since we only display visible strains for the measurements, but I can remove the filter.

src/components/measurements/measurementsD3.js
joverlee521: Yeah, I didn't want to go too far down the SVG text rabbit hole here...From what I understand, we have to draw the text to get the actual rendered width. So we split the text by spaces/hyphens/slashes, then render each substring to figure out the best configuration to fit the substrings into the allotted width. I can play around with this for a little bit.

Would we want to split reference strain names into multiple lines or would we have to special case them to preserve them in one line?

src/components/measurements/measurementsD3.js
joverlee521: Updated in c120508.

Things can look a little crammed when there are too many color groupings, but I do think it's an overall improvement.

jameshadfield (approved):

huddlej (commented):
This is awesome, @joverlee521! I really like the tooltips on the mean/std view and the color-by annotation and color bar below each reference strain. Those changes alone improve the usability dramatically!

I understand the challenge (and rabbit hole) of the SVG text scaling. In the original implementation, the text got truncated arbitrarily by length. Now, the issue is that longer strings become unreadably small as in the example below when I group by haplotype:

image

I like that I can copy and paste the text somewhere else to read it, at least, which feels like an improvement over the previous state. In addition to the small font size, the variation in label sizes makes the y-axis appear buggy or inconsistent in a way you'd want to avoid for a publication figure, for example (assuming you use the Auspice SVG for a paper). Interestingly, the coloring annotation text below the reference strain does not get scaled down, so it can extend into the main plot. The combination of these factors can lead to a slightly unreadable reference strain name and a normal-sized coloring annotation that extends into the plot like below:

image

I notice that the mean/std horizontal lines appear a bit thicker, which is nice when there are not many groups in the current coloring. When there are more groups, though, the lines get quite tightly packed together. I wish the height of each row could scale with the number of coloring groups, so one could have more space between items at the cost of showing fewer rows at a time on the screen.

I also like the 0 values in the raw data tooltips and the ability to consistently display the measurements panel in grid view with truncated panel titles!

My only major issue from the list above is the readability of the scaled y-axis labels. I don't have a great solution idea other than the same that @jameshadfield mentioned of somehow using the abundant vertical space by wrapping the text. Whatever solution we use for the main y-axis label, though, we should probably apply the same approach to the coloring text annotation that can appear below it.

PR #1560, opened 2022-10-04
Remove support for Node.js version 12, add 18

### LINKED ISSUE #1553: Update supported node versions
Currently our supported versions are "^12 || ^14 || ^16"

[v12 has reached end-of-life and should be dropped](https://github.com/nodejs/release#release-schedule)

We should support v18. I'm running it day-to-day without problems, apart from needing the `NODE_OPTIONS=--openssl-legacy-provider` flag. My understanding is that this should go away with webpack 5, introduced in #1520.

victorlin:

See commit messages.

### Related issue(s)

- Closes #1553

### Tasks

- [x] First merge https://github.com/nextstrain/auspice/pull/1520 since [webpack 4 does not support Node.js v18](https://github.com/webpack/webpack/issues/14532) without the [`--openssl-legacy-provider` option](https://nodejs.org/api/cli.html#--openssl-legacy-provider).
- [ ] Post-release: Update nodejs requirement on the [bioconda recipe](https://github.com/bioconda/bioconda-recipes/blob/-/recipes/auspice/meta.yaml)

### Testing

- [x] Ran with Node.js v18 locally and the example zika dataset seems to work fine
- [ ] Checks pass

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [12, 14, 16, 18]
+        node: [14, 16, 18]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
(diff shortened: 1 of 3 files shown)


tsibley (approved):
Seems good to me by inspection.

jameshadfield (approved):
> Seems good to me by inspection.

Same for me!

PR #1565, opened 2022-10-04
Restore nextstrain.org PRs

### LINKED ISSUE #1564: reenable nextstrain.org PR action
We should re-enable the action disabled due to Heroku's review apps temporarily not working. Disabled via https://github.com/nextstrain/auspice/commit/db858051b69db63ce725a673f8dd8e7edd3b922a.

Previous behavior was to create a nextstrain.org PR for every auspice PR, but this proved to be a bit excessive. I think it'd be better to make this a manual action, as it's only useful for auspice PRs which may affect consumers such as nextstrain.org. 

@victorlin you're our local GitHub actions expert, would you mind taking a look at this?

victorlin:

Follow-up to db858051b69db63ce725a673f8dd8e7edd3b922a.

Heroku is building review apps for nextstrain.org now.

### Related issue(s)

- Closes #1564 

### Testing

- [x] Manual trigger using `gh workflow run "Make PRs for Nextstrain projects which depend on Auspice" --repo nextstrain/auspice --ref victorlin/restore-nextstrain.org-prs` succeeds ([link](https://github.com/nextstrain/auspice/actions/runs/3199546724/jobs/5225418043))
- [x] The [resulting PR](https://github.com/nextstrain/nextstrain.org/pull/603) looks good, and a new PR doesn't get created upon updates
- [ ] ~Figure out why runs are taking so long (>15m)~ [not worth it](https://github.com/nextstrain/auspice/pull/1565#issuecomment-1270768343)

### DIFF
--- a/.github/workflows/make_prs_for_other_repos.yaml
+++ b/.github/workflows/make_prs_for_other_repos.yaml
@@ -1,45 +1,68 @@
 name: "Make PRs for Nextstrain projects which depend on Auspice"
 on:
-  pull_request:
+  workflow_dispatch:
 jobs:
   make-pr-on-nextstrain-dot-org: # <job_id>
+    # I don't see this being used for tags, so ensure it's only run on branches
+    # to make subsequent logic and wording easier.
+    if: github.ref_type == 'branch'
+
     runs-on: ubuntu-latest
+
+    env:
+      DESTINATION_REPO_DIR: nextstrain.org
+
     steps:
+      # Outputs:
+      # - pr-number: The PR number from the branch name (exits if no PR exists).
+      - name: Detect PR from branch
+        id: detect-pr
+        run: |
+          PR_NUMBER=$(gh pr view $GITHUB_REF_NAME --repo nextstrain/auspice --json 'number' --jq '.number') || true
+          if [[ -z $PR_NUMBER ]]; then
+            echo "ERROR: This branch is not associated with a PR in Auspice." >&2
+            exit 1
+          fi
+          echo "::set-output name=pr-number::$PR_NUMBER"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - uses: actions/setup-node@v2
         with:
           node-version: '14'
+
       - name: Checkout nextstrain.org repo
         uses: actions/checkout@v2
         with:
           repository: nextstrain/nextstrain.org
+          path: ${{ env.DESTINATION_REPO_DIR }}
+
       - name: Install Auspice from PRs HEAD commit
-        if: ${{ github.event_name == 'pull_request' }}
-        # Note: $GITHUB_SHA is _not_ the same commit as the HEAD commit on the PR branch
-        # see https://github.community/t/github-sha-not-the-same-as-the-triggering-commit/18286/2
         shell: bash
+        working-directory: ${{ env.DESTINATION_REPO_DIR }}
         run: |
-          AUSPICE_COMMIT=$(cat $GITHUB_EVENT_PATH | jq -r .pull_request.head.sha)
-          echo "auspice_commit=$AUSPICE_COMMIT" >> $GITHUB_ENV
           npm ci
-          npm install nextstrain/auspice#${AUSPICE_COMMIT}
+          npm install nextstrain/auspice#${GITHUB_SHA}
           git add package.json package-lock.json
+
       - name: Create Pull Request for testing on nextstrain.org repo
-        if: ${{ github.event_name == 'pull_request' }}
         id: cpr
         uses: peter-evans/create-pull-request@v3
         with:
+          path: ${{ env.DESTINATION_REPO_DIR }}
           token: ${{ secrets.NEXTSTRAIN_BOT_PAT }}
-          branch: "nextstrain-bot/test-auspice-pr/${{ github.event.pull_request.number }}"
-          commit-message: "[testing only] Upgrade Auspice to ${{ env.auspice_commit }}"
+          branch: "nextstrain-bot/test-auspice-pr/${{ steps.detect-pr.outputs.pr-number }}"
+          commit-message: "[testing only] Upgrade Auspice to ${{ github.sha }}"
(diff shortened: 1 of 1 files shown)


tsibley:
> Heroku is still not building review apps

Uh, yes they are? Maybe we forgot to re-enable them for Auspice, but they work for nextstrain.org.

tsibley (approved):
+1 for the re-enabling here.

Not the fault of this PR, but I'm still not thrilled that we're passing decently powerful credentials to some random third-party code:

https://github.com/nextstrain/auspice/blob/5ba0b26b7575312e0919eca5f38ab3bf3d0660e5/.github/workflows/make_prs_for_other_repos.yaml#L29-L31

jameshadfield:
Good timing! As per https://github.com/nextstrain/auspice/issues/1564 I'd recommend making this a workflow dispatch rather than automatically making a (nextstrain.org) PR for each auspice PR -- in practice this resulted in too many unnecessary PRs which became confusing.

victorlin:
@tsibley oh you're right, they are working on nextstrain.org PRs. I updated the commit/PR descriptions to reflect that.

tsibley:
@victorlin

From the message of af56b03:

```
Two notes:

1. The title and description of the PR have been updated to use the
   source ref (e.g. branch, tag) rather than PR number, since that is
   not guaranteed or retrievable with the manual trigger.
2. The commit to install Auspice from is changing from the
   GitHub-managed PR merge HEAD to the source ref HEAD. This can have
   implications when run on a PR's source branch if there are notable
   changes in the target branch unincorporated in the source branch. To
   mitigate this, make sure to update the source branch (e.g. with a git
   rebase) before triggering this workflow.
```

To avoid both of these issues, can we instead make the workflow accept a PR number? e.g.:

```
gh workflow run "Make PRs for Nextstrain projects which depend on Auspice" --repo nextstrain/auspice -f pr=1565
```

The workflow would then have the PR number to use and be able to turn `refs/pull/${{inputs.pr}}/merge` into a commit id (SHA) and use that as it previously did.

victorlin:
@tsibley I've opted for automatic retrieval of PR number using `gh pr view`: 9c9bfd32da8c3a1e5059a4ca7854f26c6609b2eb Still WIP but the PR number retrieval should work. Good point about using `refs/pull/${{inputs.pr}}/merge`, I'll add that back.

jameshadfield:
@victorlin is your intention to run these on the head commit of the branch (which is what I did when I set up the action) or the merge commit (`GITHUB_SHA`)? I don't know which is the best for this purpose, but I thought it'd be difficult to try to debug a failing nextstrain.org build if the problem was only introduced by the merge and thus didn't exist on the auspice PR branch.

victorlin:
@jameshadfield oops, I got your original intention backwards in my commit message. You're right, `GITHUB_SHA` for `pull_request` events reflect the GitHub-managed PR merge and your implementation.

In the currently pushed implementation, I intend to run these on the head commit of the branch (not merge), which is simply `GITHUB_SHA` for the `workflow_dispatch` event. So this would be the same as before.

However, @tsibley noted from my (incorrect) commit description that it's still possible to get the merge ref from the PR number. This would change behavior as you mention:

> I thought it'd be difficult to try to debug a failing nextstrain.org build if the problem was only introduced by the merge and thus didn't exist on the auspice PR branch.

I'd actually be in favor of surfacing these failures, since they are bound to come up once actually merged, and better to catch them sooner than later.

jameshadfield:
> I'd actually be in favor of surfacing these failures, since they are bound to come up once actually merged, and better to catch them sooner than later.

Fair enough. Perhaps add a message to this effect in the PR body, as this will catch someone out!

> the head commit of the branch (not merge), which is simply GITHUB_SHA for the workflow_dispatch event

I guess this makes some sense, but I find it super confusing to have GITHUB_SHA refer to different things depending on the trigger. I didn't actually know a branch-merge-commit (which isn't found anywhere in your local git tree) even exists until I wrote the original action here.

victorlin:
Updated to only allow branches with PRs, for simplicity and parity with previous behavior.

victorlin:
This is ready for review again. I went down a deep rabbit hole on the last task in the PR tasklist - figuring out why runs are taking so long [[1]] [[2]], and [trying alternatives unsuccessfully](https://github.com/victorlin/github-test/actions/workflows/test-nextstrain-auspice-install.yml). 15 minute `npm install`s aren't ideal, but I think we can live with it.

[1]: https://github.com/npm/cli/issues/4896
[2]: https://github.com/npm/npm/issues/17246

.github/workflows/make_prs_for_other_repos.yaml
```diff
@@ -1,48 +1,74 @@
...
+      # - auspice-sha: The GitHub-managed merge ref. Used for npm install.
+      - name: Detect PR from branch
+        id: detect-pr
+        run: |
+          PR_NUMBER=$(gh pr view $GITHUB_REF_NAME --repo nextstrain/auspice --json 'number' --jq '.number') || true
+          if [[ -z $PR_NUMBER ]]; then
+            echo "ERROR: This branch is not associated with a PR in Auspice." >&2
+            exit 1
+          fi
+          MERGE_SHA=$(gh pr view $GITHUB_REF_NAME --repo nextstrain/auspice --json 'potentialMergeCommit' --jq '.potentialMergeCommit.oid')
```
jameshadfield, line 27: Just double checking I understand what's going on here (via https://github.com/nextstrain/auspice/pull/1565#issuecomment-1269040879) -- the salient thing is that it's a `workflow_dispatch` trigger and so `$GITHUB_SHA` here refers to the head of the branch (which in this case is a PR, but that's not relevant for how the action machinery operates). However, if the action _had_ been triggered by a push-to-branch then `$GITHUB_SHA` would refer to the merge commit SHA?

Perhaps add a one-line comment here explaining?

jameshadfield (approved):
[Looks good](https://github.com/nextstrain/nextstrain.org/pull/603) @victorlin -- thanks for persevering 😄

PR #1566, opened 2022-10-07
measurements: order means by legend values

joverlee521:
Keep the color-by means in a stable order so that order does not change when applying different filters to the data. This makes it easier to do comparisons across groups and across different filters.

This is the easier route for stable ordering, but in the future we can consider ordering by display order of groups in the tree.

### DIFF
--- a/src/components/measurements/measurementsD3.js
+++ b/src/components/measurements/measurementsD3.js
@@ -280,7 +280,7 @@ export const colorMeasurementsSVG = (ref, treeStrainColors) => {
     .style("fill", (d) => getBrighterColor(treeStrainColors[d.strain].color));
 };
 
-export const drawMeansForColorBy = (ref, svgData, treeStrainColors) => {
+export const drawMeansForColorBy = (ref, svgData, treeStrainColors, legendValues) => {
   const { xScale, groupingOrderedValues, groupedMeasurements } = svgData;
   const svg = select(ref);
   // Re move all current color by means
@@ -305,24 +305,28 @@ export const drawMeansForColorBy = (ref, svgData, treeStrainColors) => {
     // 2 x subplotPadding for padding around the overall mean display
     const ySpacing = (layout.subplotHeight - 4 * layout.subplotPadding) / (numberOfColorByAttributes - 1);
     let yValue = layout.subplotPadding;
-    Object.entries(colorByGroups).forEach(([attribute, {color, values}]) => {
-      drawMeanAndStandardDeviation(
-        values,
-        subplot,
-        classes.colorMean,
-        {attribute, color},
-        xScale,
-        yValue
-      );
-      // Increate yValue for next attribute mean
-      yValue += ySpacing;
-      // If the next yValue is near the overall mean display,
-      // shift to below the overall mean display + subplotPadding
-      if (yValue > (layout.overallMeanYValue - layout.subplotPadding) &&
-          yValue < (layout.overallMeanYValue + layout.subplotPadding)) {
-        yValue = layout.overallMeanYValue + layout.subplotPadding;
-      }
-    });
+    // Order the color groups by the legend value order so that we have a stable order for the means
+    legendValues
+      .filter((attribute) => String(attribute) in colorByGroups)
+      .forEach((attribute) => {
+        const {color, values} = colorByGroups[attribute];
+        drawMeanAndStandardDeviation(
+          values,
+          subplot,
+          classes.colorMean,
+          {attribute, color},
+          xScale,
+          yValue
+        );
+        // Increase yValue for next attribute mean
+        yValue += ySpacing;
+        // If the next yValue is near the overall mean display,
+        // shift to below the overall mean display + subplotPadding
+        if (yValue > (layout.overallMeanYValue - layout.subplotPadding) &&
+            yValue < (layout.overallMeanYValue + layout.subplotPadding)) {
+          yValue = layout.overallMeanYValue + layout.subplotPadding;
+        }
+      });
   });
 };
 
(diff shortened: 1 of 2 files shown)


huddlej (approved):
I can't comment much on the implementation details, but the ordering works as expected for me with our measurements panel build! 🚀

src/components/measurements/measurementsD3.js
```diff
@@ -305,7 +305,12 @@ export const drawMeansForColorBy = (ref, svgData, treeStrainColors) => {
     // 2 x subplotPadding for padding around the overall mean display
     const ySpacing = (layout.subplotHeight - 4 * layout.subplotPadding) / (numberOfColorByAttributes - 1);
     let yValue = layout.subplotPadding;
-    Object.entries(colorByGroups).forEach(([attribute, {color, values}]) => {
+    // Order the color groups by the legend value order so that we have a stable order for the means
+    orderBy(
+      Object.entries(colorByGroups),
+      ([attribute]) => legendValues.indexOf(attribute),
```
jameshadfield, line 311: This doesn't work when the `legendValues` are numerics (`legendValues.indexOf(attribute)=-1`) and thus the ordering isn't working for continuous colour scales (date, antigenic advance etc).

The `legendValues` are inherently ordered, so could we not simply replace the `orderBy` call with the following?

```js
    legendValues.filter((v) => String(v) in colorByGroups)
      .forEach((attribute) => {
        const {color, values} = colorByGroups[attribute];
        ...
```

This same incantation is used in one other place in auspice, so might be worth reviewing that as well (I haven't done so): https://github.com/nextstrain/auspice/blob/2eb4bb35f28375914f638eb894b9eb00c254fce3/src/components/measurements/index.js#L176

src/components/measurements/measurementsD3.js
```diff
@@ -305,7 +305,12 @@ export const drawMeansForColorBy = (ref, svgData, treeStrainColors) => {
     // 2 x subplotPadding for padding around the overall mean display
     const ySpacing = (layout.subplotHeight - 4 * layout.subplotPadding) / (numberOfColorByAttributes - 1);
     let yValue = layout.subplotPadding;
-    Object.entries(colorByGroups).forEach(([attribute, {color, values}]) => {
+    // Order the color groups by the legend value order so that we have a stable order for the means
+    orderBy(
```
jameshadfield, line 309: I'm not against lodash per-se (we use it a little bit in lots of places in auspice, and `orderBy` is used elsewhere in the measurements panel) but each time I feel there's a trade-off between learning another function vs writing out the slightly longer JS code. This trade-off varies across different lodash functions. Here I think a `sort` call would be simpler (at least, for others reading the code), but I'm interested in your general thoughts here? (It's not a request to change, especially as we're already importing it elsewhere).

src/components/measurements/measurementsD3.js
```diff
@@ -305,7 +305,12 @@ export const drawMeansForColorBy = (ref, svgData, treeStrainColors) => {
     // 2 x subplotPadding for padding around the overall mean display
     const ySpacing = (layout.subplotHeight - 4 * layout.subplotPadding) / (numberOfColorByAttributes - 1);
     let yValue = layout.subplotPadding;
-    Object.entries(colorByGroups).forEach(([attribute, {color, values}]) => {
+    // Order the color groups by the legend value order so that we have a stable order for the means
```
jameshadfield, line 308: Did we ever consider doing this for the raw dots?

jameshadfield (changes requested):
👍 for this feature -- it'll be a nice improvement 😄

src/components/measurements/measurementsD3.js
joverlee521: Yup, we've discussed ordering the raw dots by color before (it's part of #1463). 
I think that will be a bigger change because of the jitter. Currently, the jitter is calculated during loading of the sidecar JSON, so will need to think through how to rework that...

src/components/measurements/measurementsD3.js
joverlee521: Ah, thank you for catching that! I'll update to your suggestion. That's much easier to read and understand

src/components/measurements/measurementsD3.js
joverlee521: `lodash` is my personal preference because I trust it to do the right thing more than my own version of a `sort` call. I think that has definitely made me over-reliant on `lodash` and then miss the simpler ways of doing things like your suggestion above.

src/components/measurements/measurementsD3.js
joverlee521: In the other use of `orderBy` that you pointed out, the `fieldOrder` is not guaranteed to include every single display field. We want to order `displayFields` by the `fieldOrder` first then the remaining fields can fall into their existing order.

src/components/measurements/measurementsD3.js
jameshadfield: > I think that will be a bigger change because of the jitter. Currently, the jitter is calculated during loading of the sidecar JSON, so will need to think through how to rework that...

Thanks for the reference. I agree it's not for this PR, it just became apparent while reviewing it 😄

src/components/measurements/measurementsD3.js
jameshadfield: > In the other use of orderBy that you pointed out, the fieldOrder is not guaranteed to include every single display field. We want to order displayFields by the fieldOrder first then the remaining fields can fall into their existing order.

👍 Sounds good. And since `displayFields` is a list of keys, they won't be numeric (which is the underlying potential bug here).

jameshadfield (approved):

PR #1567, opened 2022-10-10
Enable preview PRs for auspice.us

victorlin:

Currently, preview PRs are only created on the nextstrain.org repo. Also create them for auspice.us and add inputs to select which PRs to make. This looks like:

workflow-dispatch-ui

### Related issue(s)

_N/A_

### Testing

- [x] Checks pass
- [x] Manually triggered run is successful
- [x] Generated PRs look good

### DIFF
--- a/.github/workflows/make_pr_for_downstream_repo.yaml
+++ b/.github/workflows/make_pr_for_downstream_repo.yaml
@@ -1,8 +1,12 @@
-name: "Make PRs for Nextstrain projects which depend on Auspice"
+name: "Make PR for a repository which depends on Auspice"
 on:
-  workflow_dispatch:
+  workflow_call:
+    inputs:
+      repository:
+        required: true
+        type: string
 jobs:
-  make-pr-on-nextstrain-dot-org:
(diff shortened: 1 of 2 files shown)


jameshadfield:
Awesome! This is a great idea & the code looks good. 

GitHub actions do constantly confuse me... I think I've tested this by dispatching [this action](https://github.com/nextstrain/auspice/actions/runs/3222850824/jobs/5272343954) from this PR's branch (which I'm sure used the updated action definition, as the repo-checkboxes came up when I selected this branch and they're implemented here for the first time). I'll approve this PR when the action finishes (it's running into the `npm install` slowness I think).

victorlin:
@jameshadfield yes, that looks right! It'll use the workflow file from the branch that's selected for run.

`npm install` slowness indeed... and I noticed that it's not just here in the "make PRs" workflow, but also in the downstream PR checks (note how long `npm ci` took to run on https://github.com/nextstrain/nextstrain.org/pull/608).

jameshadfield (approved):
Let's get this merged!

PR #1575, opened 2022-10-20
Add URL query for show all (branch) labels

### LINKED ISSUE #1573: ENH: Make more view settings a URL parameter for more reproducible views
I love the `show all` branch label switch

Would it be possible to add this as a URL parameter to make views more reproducible?

There are other view settings that would be great to add to URL parameters, can't think of an example right now but will add it to the issue once I come across one.

image

Also, I'd love to be able to set this switch to _on_ by default 🙃

jameshadfield:
Tested manually & via /narratives/test/simultaneous-tree-updates

Closes #1573

### DIFF
--- a/docs/advanced-functionality/view-settings.md
+++ b/docs/advanced-functionality/view-settings.md
@@ -77,6 +77,7 @@ All URL queries modify the view away from the default settings -- if you change 
 | `n`        | Narrative page number | `n=1` goes to the first page |
 | `s`        | Selected strain | `s=1_0199_PF` |
 | `branchLabel` | Branch labels to display | `branchLabel=aa` |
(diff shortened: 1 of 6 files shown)


jameshadfield:
The review-app appears to have been build ("-----> Launching... Released v5. https://auspice-feat-branch-lab-hdcfc5.herokuapp.com/ deployed to Heroku") however I see the following when loading that URL. I don't think this :bug: is the fault of this PR however
image

joverlee521:
Huh, the review app works for me now. There must have been something weird on Heroku's side.

joverlee521 (approved):
LGTM!

I pushed up commits to update the docs for URL params.

jameshadfield:
Thanks @joverlee521!

PR #1576, opened 2022-10-20
Fix three small bugs in the legend

jameshadfield:
See commit messages for details & which issues they close.

@joverlee521 could you take a look at this? 🙏


src/components/tree/legend/legend.js
```diff
@@ -159,7 +159,7 @@ class Legend extends React.Component {
             rectFill={getBrighterColor(this.props.colorScale.scale(d))}
             rectStroke={rgb(this.props.colorScale.scale(d)).toString()}
             transform={this.getTransformationForLegendItem(maxNumPerColumn, i)}
-            key={d}
+            key={this.props.colorScale.colorBy+d+i /* eslint-disable-line react/no-array-index-key */}
```
joverlee521, line 162: > The legend entries (array) do not change for a given coloring, and thus
I believe it's safe to use the index in the key.

Hmm, this may not be true for the legend items here. The legend entries do not change, but we are using the `visibleLegendValues` here which _does_ change index.

jameshadfield: Good catch. Hmm. [027e3b6](https://github.com/nextstrain/auspice/pull/1576/commits/027e3b6fb98f198cbe651e30d3fbd622fde2bdab) should fix the duplicate entries so maybe [7bfe09c](https://github.com/nextstrain/auspice/pull/1576/commits/7bfe09cf2abc778e38124671336d6219049524f6) is just not needed?

joverlee521: Yeah, if we can reliably say that the legend value will always be unique, I think it's fine to drop this commit.

joverlee521:
I triggered the [action](https://github.com/nextstrain/auspice/actions/runs/3293315962) to create a PR in nextstrain.org so we can test the [original community build](https://nextstrain.org/community/KelianP/RYMV-WA?c=rate) that had this bug.

src/components/tree/legend/legend.js
jameshadfield: I _think_ these will always be unique, but it's always possible they'll slip through. In cases such as these, would `this.props.colorScale.colorBy+d+i` be better than simply `d` since then (a) we cannot ever have duplicate (react) keys this way and (b) I actually think this would prevent reconciliation errors even when the array changes underneath (it's been a while since I did a deep-dive into keys, but by combining the value and the index then I think you'll catch any change in order)

jameshadfield: I'll certainly update the commit message with this concern, but I can't see how `d` would be _better_ than `this.props.colorScale.colorBy+d+i`

joverlee521: Ah true! `this.props.colorScale.colorBy+d+i` is definitely better than just `d`

jameshadfield: Updated the commit message of [b8f36f0](https://github.com/nextstrain/auspice/pull/1576/commits/b8f36f0b5413fef050e3c2ba237bfb40658c77f1) to reflect this thread. Thanks for testing!

PR #1577, opened 2022-10-20
Reenable tree confidence intervals

jameshadfield:
Previously we checked for the presence of temporal confidence intervals on the first node (nodes[0]). Prior to auspice handling subtrees (arrays of trees), nodes[0] was guaranteed to be the root node, but this is no longer the case. Checking a single node wasn't the best approach anyway as we should check for _any_ node with confidence set, which this commit now does.


emmahodcroft:
Thanks so much James!

PR #1578, opened 2022-10-20
Fix frequency colors for ordinal scales

### LINKED ISSUE #843: Ordinal values are not assigned colors from color ramp in the frequencies panel
On the live flu builds, [epitope mutations](https://nextstrain.org/flu/seasonal/h3n2/ha/2y?c=ep&d=tree,frequencies&p=full) and other mutation metrics are defined as “ordinal” types in the auspice config JSON. These metrics get colors in the tree from a color ramp, but the frequencies panel shows gray for all values. This issue does not affect categorical or continuous values.

For example, epitope mutations appear like so in the frequencies panel:

image

While we would expect them to appear like the continuous distance to the future values:

image

jameshadfield:
The underlying bug was a type-inconsistency whereby color-by values can be numeric however the matrix was stored as an Object (where keys are always stringified) and so `matrix[numeric_colorBy_value]=undefined`. Switching to a Map() allows typed keys and fixes this bug.

Closes #843

### DIFF
--- a/src/components/frequencies/functions.js
+++ b/src/components/frequencies/functions.js
@@ -175,10 +175,10 @@ const turnMatrixIntoSeries = (categories, nPivots, matrix) => {
     const x = [];
     for (let j = 0; j < nPivots; j++) {
       if (i === 0) {
-        x.push([0, matrix[categories[i]][j]]);
+        x.push([0, matrix.get(categories[i])[j]]);
       } else {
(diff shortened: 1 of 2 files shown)


jameshadfield:
Tested using the flu dataset from the original issue. Will merge now so that we can release in time for the next nCoV build which will include (ordinal) RBD Levels

PR #1584, opened 2022-10-23
Branch labels don't capture click events

### LINKED ISSUE #1580: BUG: Labels make branch covered by it unclickable
### Current Behavior
When there is label text that overlays a branch, the branch becomes unclickable

### Expected behavior
Labels do not act as a real layer that prevents clicks of the branch below 

### Your environment: if running Nextstrain onlinee
 - Operating system: M1 Mac
 - Browser: Brave
 - Version (e.g. `auspice 2.7.0`): 2.40.0

jameshadfield:
There is currently no on-hover behavior for branch labels and they have a tendency to obscure branches behind them and make them un-clickable.

Closes #1580

### DIFF
--- a/src/components/tree/phyloTree/labels.js
+++ b/src/components/tree/phyloTree/labels.js
@@ -154,6 +154,7 @@ export const drawBranchLabels = function drawBranchLabels(key) {
     .style("font-family", this.params.branchLabelFont)
     .style("font-weight", fontWeight)
     .style("font-size", labelSize)
+    .style("pointer-events", "none")
     .text((d) => d.n.branch_attrs.labels[key]);
 };
 


joverlee521 (approved):
Cool! Love it when a bug can be fixed with a single CSS change!
I didn't realize that `none` pointer-events worked like this!

> In addition to indicating that the element is not the target of pointer events, the value none instructs the pointer event to go "through" the element and target whatever is "underneath" that element instead.

corneliusroemer:
This works well indeed, just tested ;)

Is there a master deploy by `nextstrain-bot` @jameshadfield? I had quite some trouble building master locally so any preview would be very handy.

PR #1591, opened 2022-11-03
Fix download empty tree

### LINKED ISSUE #1267: Download Data producing empty trees
**Current Behavior**  
When I try to download a Newick tree after I've filtered to specific tips, the downloaded file is empty. The download "dashboard" says currently 523/4047 tips are displayed, but the downloaded file only contains a semicolon.

**Expected behavior**  
The tree file should contain the tips shown & selected in the tree display.

**How to reproduce**  
1. Open  https://nextstrain.org/groups/blab/ncov/wa/4m
2.   Filter to Recency --> 1-2 days ago.
3.  Click Download data
4. Download Newick tree
5. File is empty.

 - Operating system:  Windows 10 Pro
 - Browser:  Chrome 

**Additional context**  
This doesn't seem to happen with every type of filter every time. For example, when testing this bug, I initially filtered to clade 20G, and the file contained the expected number of tips. But then when I switched to filter by recency, the file was empty. If I then tried to filter by clade again, the file was also empty. So the bug may only be occurring after multiple filters have been toggled on and off.

This only seems to be a bug for .nwk files. With filters toggled on, all other files available to download (nexus tree, metadata, etc) were complete.

joverlee521:

Fixes the empty tree downloads when the tree is filtered. 
This does _not_ support the case of downloading multiple subtrees. I'm going to punt that for now and just added a note to the users that we do not support this feature. We can open another issue if this feature is ever requested.

I've made the `treeToNewick` function raise an error if it ever creates an empty tree since this seems like generally undesired behavior. I can drop this change if people think otherwise. 

### Related issue(s)
Resolves #1267

### Testing

- [x] Checks pass

### DIFF
--- a/src/components/download/downloadButtons.js
+++ b/src/components/download/downloadButtons.js
@@ -41,6 +41,9 @@ export const DownloadButtons = ({dispatch, t, tree, entropy, metadata, colorBy, 
         Downloaded data represents the currently displayed view.
         By zooming the tree, changing the branch-length metric, applying filters etc, the downloaded data will change accordingly.
         <p/>
+        NOTE: We do not support downloads of multiple subtrees, which are usually created with the date range filter or genotype filters.
+        Downloading multiple subtrees will result in an empty Newick tree!
+        <p/>
         {partialData ? `Currently ${selectedTipsCount}/${totalTipCount} tips are displayed and will be downloaded.` : `Currently the entire dataset (${totalTipCount} tips) will be downloaded.`}
       </div>
       <Button
--- a/src/components/download/helperFunctions.js
+++ b/src/components/download/helperFunctions.js
@@ -36,7 +36,12 @@ const treeToNewick = (tree, temporal, internalNodeNames=false, nodeAnnotation=()
     return leaf;
   }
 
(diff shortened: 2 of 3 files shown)


tsibley (approved):
I'm not well-versed in the interactions here, but code looks ok to me by inspection and clicking around the review app to exercise the filtering + tree download seems to work as expected.

jameshadfield (approved):
Thanks Jover! I agree that erroring in most situations like this is preferable -- newick trees can't represent the states that auspice can be in! We could probably work this out on download-modal rendering & not even allow downloading to be attempted, but the approach in this PR is just fine.

As an aside, the logic around zooming / inView nodes / visible nodes is getting pretty complex. In part this is because a lot of those features have been added over time and we haven't rethought the underlying architecture. Zooming is well overdue a rethink in auspice so hopefully we can address this then!

joverlee521:
> As an aside, the logic around zooming / inView nodes / visible nodes is getting pretty complex. In part this is because a lot of those features have been added over time and we haven't rethought the underlying architecture. Zooming is well overdue a rethink in auspice so hopefully we can address this then!

Definitely! Would love to revisit the [xy zoom idea](https://github.com/nextstrain/auspice/issues/1368).

PR #1593, opened 2022-11-09
measurements: fix disappearing mean/SD on entropy click

joverlee521:

Previously, when clicking on an entropy rect, the colored mean/SD in the measurements panel would flash and disappear. I tracked this down to the following:
1. Clicking on an entropy rect would trigger two NEW_COLOR dispatches.
2. The second NEW_COLOR action does _not_ update the colors, but the react-redux `useSelector` hook does a strict reference comparison by default, so it thinks the `legendValues` (an array) was updated.
3. This would trigger the side-effect `drawMeansForColorBy` but _not_ the `changeMeasurementsDisplay` so the colored mean/SD would get hidden.

This commit updates the `legendValues` to use `lodash.isEqual` equality function to ensure that we don't run any unnecessary updates. It also updates the dependencies for the `changeMeasurementsDisplay` side-effect to make sure they match the dependencies for the `drawMeansForColorBy` side-effect.

### Testing

- [x] Checks pass

### DIFF
--- a/src/components/measurements/index.js
+++ b/src/components/measurements/index.js
@@ -127,7 +127,7 @@ const filterMeasurements = (measurements, treeStrainVisibility, filters) => {
 const MeasurementsPlot = ({height, width, showLegend, setPanelTitle}) => {
   // Use `lodash.isEqual` to deep compare object states to prevent unnecessary re-renderings of the component
   const { treeStrainVisibility, treeStrainColors } = useSelector((state) => treeStrainPropertySelector(state), isEqual);
-  const legendValues = useSelector((state) => state.controls.colorScale.legendValues);
+  const legendValues = useSelector((state) => state.controls.colorScale.legendValues, isEqual);
   const colorings = useSelector((state) => state.metadata.colorings);
   const colorBy = useSelector((state) => state.controls.colorBy);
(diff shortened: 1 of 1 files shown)


joverlee521:
> Clicking on an entropy rect would trigger two NEW_COLOR dispatches.

I'd like to know why this is happening, but I haven't dug into it yet...

jameshadfield:
Thanks @joverlee521 - i'll take a look at this today/tomorrow.

> I'd like to know why this is happening, but I haven't dug into it yet...

In dev mode or production mode? React fires certain events twice in dev mode (e.g. `useEffect`, from memory) basically to test how resilient the component is to side-effects.

jameshadfield (approved):
The two `NEW_COLOR` actions are coming from the entropy onClick (as expected) and the sidebar color-by drop-down (unexpected). While this should be fixed, it's also important that components are resilient to extra dispatches such as this, which is fixed in this PR 👍

joverlee521:
> The two NEW_COLOR actions are coming from the entropy onClick (as expected) and the sidebar color-by drop-down (unexpected).

Ah thank you for looking into this! This is good to know and to fix in the future, I'll make an issue to track it.

Issue #1599, opened 2022-11-21
Parse narratives as a single document before breaking into slides


tsibley:
While implementing https://github.com/nextstrain/cli/pull/235, I realized that currently narrative source documents are first split into slides via string manipulation and then converted from Markdown → HTML.

https://github.com/nextstrain/auspice/blob/a416642765e15881b025c23967bf2625f444b3e1/src/util/parseNarrative.js#L90-L118

This causes issues with intra-document references like links and images that reference a definition found elsewhere in the file, if the definition is in a different "slide" than the link/image reference:

```markdown
[some link that's a reference][1]
![an image that's the same][cat]

[1]: https://example.com
[cat]: https://http.cat/599
```

Anyone previewing the document outside of Auspice's narrative rendering would see the link/image references work, but they'd be broken when viewed thru Auspice. Since narrative documents are intended to gracefully degrade with conventional Markdown rendering and still be a viewable document (e.g. [on GitHub](https://github.com/nextstrain/narratives/blob/master/how-to-write_basics.md)), it would be nice to address this issue for that reason.

One approach I'd like to pursue is inverting the order of parsing to render to HTML first and then split into slides. We can make use of the DOM to split the HTML into slides, and it might even be easier than line-wise string manipulation. :-)

jameshadfield:
Yup, this sounds like a good plan!

PR #1600, opened 2022-11-22
narratives: Convert Markdown → HTML before splitting into separate slides

tsibley:
Fixes handling of intra-document references like links and images that reference a definition found elsewhere in the file, e.g. before the first slide or on a different slide than the link/image reference.  Such features are soon to be used for embedding images (see related PR below).

Anyone previewing a document with such features outside of Auspice's narrative rendering would see the link/image references work, but they'd be broken when viewed thru Auspice.  Since narrative documents are intended to gracefully degrade with conventional Markdown rendering and still be a viewable document (e.g. on GitHub), it's nice to address this issue for that reason alone.

It's also nice to use the DOM as a structured way to split apart the document into slides, rather than relying on string manipulation and parsing.  An alternative approach would be to use a Markdown parser that emits an AST or stream of nodes as an intermediate we could split into slides while still preserving references/definitions, but that seemed like a bigger change (and I'd have to go survey the parsers available instead of using the browser's built-in DOM).

While the entire internals of parseNarrativeBody() change, the shape of the output "slide objects" is the same.  The one minor exception is that slides will always have a "mainDisplayMarkdown" property (though it will usually be the empty string) even if they don't have a corresponding code block in the source text.  All downstream usages I found condition not on property existence but on truthiness of the value, so this should result in no functional change.

Resolves: <https://github.com/nextstrain/auspice/issues/1599>
Related-to: <https://github.com/nextstrain/cli/pull/235>

### Testing

- [x] Manually tested with [this silly narrative](https://gist.githubusercontent.com/tsibley/0f128232ae9a27da3b0f6cadeda93eb6/raw/ba621eab7cd8383106bced054435731e599c11fb/test_trs.md)
- [x] Checks pass

### DIFF
--- a/cli/server/getNarrative.js
+++ b/cli/server/getNarrative.js
@@ -3,17 +3,6 @@ const path = require("path");
 const fs = require("fs");
 const utils = require("../utils");
 const marked = require('marked');
-const { parseMarkdownNarrativeFile } = require("../../src/util/parseNarrative");
-
-/**
- * A thin wrapper around the client-side `parseMarkdownNarrativeFile` function.
- * The main difference is that we pass in a different markdown parser
- * than the client uses.
- */
-const parseNarrative = (fileContents) => {
-  utils.verbose("Deprecation warning: Server-side parsing of narrative files is no longer needed!");
-  return parseMarkdownNarrativeFile(fileContents, marked);
-};
 
 const setUpGetNarrativeHandler = ({narrativesPath}) => {
   return async (req, res) => {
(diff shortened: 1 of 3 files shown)


jameshadfield (changes requested):
Cool to see a generator function work so seamlessly with the spread syntax we were already using. I've not got a lot of experience with `DOMParser()` but it works well and code is easy to read.

There is one regression introduced by this PR related to 👇 

https://github.com/nextstrain/auspice/blob/a416642765e15881b025c23967bf2625f444b3e1/src/util/parseNarrative.js#L2-L9

You can see this by making a request to http://localhost:4000/charon/getNarrative?prefix=/narratives/test/main-display-markdown (for instance) which now results in `Narratives couldn't be served -- DOMParser is not defined`. One solution would be to drop this backwards compatibility.

tsibley:
Ah! I missed that entirely. Thanks for flagging it.

When does this backwards compat come into play? i.e. what's making `/charon/getNarrative?type=json` requests to the server code in this repo? Is it only ever the client code in this repo? If so, then it seems to me we could drop the backwards compat without any ill effects since the current client never makes those requests?

tsibley:
Repushed with [0679589](https://github.com/nextstrain/auspice/pull/1600/commits/06795890ad957aa5522fd74dcdb97d4479d53886) to remove the legacy `type=json` (and missing `type`) support (plus [some typo fixes](https://github.com/nextstrain/auspice/pull/1600/commits/739b802c7c428d1b8145a586164a784b04ee909d)).

jameshadfield (approved):
I can't remember what situations we were guarding against with the client-server backwards compat, but it's been over 2 years so pretty confident we can remove this!

PR #1606, opened 2022-12-03
Remove prepare script

victorlin:

The prepare script is automatically run with every call to `npm ci` / `npm install`, among other scripts¹. As it is just an alias to the build script (which can take several minutes), it should be safe to remove this automation and speed up the dev experience.

Updated README accordingly.

¹ https://docs.npmjs.com/cli/v8/using-npm/scripts?v=true#life-cycle-operation-order

### Related issue(s)

_N/A_

### Testing

- [ ] Checks pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ If you wish to view the [nextstrain-maintained narratives](https://nextstrain.or
 ### Run auspice
 
 ```bash
+auspice build --verbose
 auspice view --datasetDir data
 ```
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,6 +45,7 @@ jobs:
           node-version: ${{ matrix.node }}
       - run: npm ci
       - run: npm run get-data
+      - run: npm run build
       - run: npm run smoke-test:ci
   lint:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -90,6 +90,7 @@ If you wish to view the [nextstrain-maintained narratives](https://nextstrain.or
 ### Run auspice
 
 ```bash
+auspice build --verbose
```
jameshadfield, line 93: FYI - the `prepare` script was so that installs of auspice via `npm install --global .` (line 58 of this README) would leave the user with a built instance of auspice they could run immediately with `auspice view ...`. I'll double check that this is still the case (I expect it is). 

I do agree with the intention of this PR tho -- would be nice to not have time-consuming build steps occurring when we don't want them.

victorlin: >the `prepare` script was so that installs of auspice via `npm install --global .` would leave the user with a built instance of auspice they could run immediately with `auspice view ...`.

Thanks for the context – you're right, this is still necessary. Also makes me think about some side effects of #1608, especially for anyone using `npm install --global auspice`. I'll probably create another PR or start a discussion about the effects of these two PRs.

jameshadfield: (I think you realised my typo Victor, but I was actually referring to line 50 `npm install --global auspice` -- the dev install separates out the build step for clarity.)

PR #1611, opened 2022-12-06
CLI improvements

victorlin:

See commit messages.

### Testing

- [ ] Checks pass


cli/view.js
```diff
@@ -75,28 +75,39 @@ const loadAndAddHandlers = ({app, handlersArg, datasetDir, narrativeDir}) => {
...
+    baseDir: baseDir,
+    distDir: path.join(baseDir, "dist")
   };
 };
 
+const hasAuspiceBuild = (directory) => {
+  return (
+    fs.existsSync(path.join(directory, "dist")) &&
+    fs.existsSync(path.join(directory, "dist/index.html")) &&
+    fs.readdirSync(path.join(directory, "dist")).filter((fn) => fn.match(/^auspice.bundle.[a-z0-9]+.js$/)).length === 1
```
jameshadfield, line 107: As noted by @tsibley  in #1520:

> those . should be escaped in the regex since they're meant to be literals.

cli/view.js
```diff
@@ -74,28 +75,39 @@ const loadAndAddHandlers = ({app, handlersArg, datasetDir, narrativeDir}) => {
...
+      utils.log(`Auspice build files not found under ${cwd}. Using build files under ${sourceDir}.`)
+      baseDir = sourceDir;
+    }
   }
   return {
-    message: `Serving auspice version ${version}`,
-    baseDir: sourceDir,
-    distDir: path.join(sourceDir, "dist")
+    message: `Serving the auspice build in ${baseDir}.`,
+    baseDir: baseDir,
```
jameshadfield, line 98: FWIW, in JS you can replace `baseDir: baseDir,` with simply `baseDir,`

cli/view.js
```diff
@@ -24,6 +24,7 @@ const addParser = (parser) => {
   subparser.addArgument('--handlers', {action: "store", metavar: "JS", help: "Overwrite the provided server handlers for client requests. See documentation for more details."});
   subparser.addArgument('--datasetDir', {metavar: "PATH", help: "Directory where datasets (JSONs) are sourced. This is ignored if you define custom handlers."});
   subparser.addArgument('--narrativeDir', {metavar: "PATH", help: "Directory where narratives (Markdown files) are sourced. This is ignored if you define custom handlers."});
+  subparser.addArgument('--customBuildOnly', {action: "storeTrue", help: "Error if a custom build is not found."});
```
jameshadfield, line 27: I like the intention here but the word "custom" is a bit overloaded here -- this is checking if an auspice build is present in the current working directory, not whether the build has been customised.

cli/view.js
```diff
@@ -74,28 +75,39 @@ const loadAndAddHandlers = ({app, handlersArg, datasetDir, narrativeDir}) => {
...
+  // Default to current working directory.
+  let baseDir = cwd;
+  if (!hasAuspiceBuild(cwd)) {
+    if (cwd === sourceDir || customBuildOnly) {
+      utils.error(`Auspice build files not found under ${cwd}. Did you run \`auspice build\` in this directory?`);
+      process.exit(1);
+    } else if (!hasAuspiceBuild(sourceDir)) {
+      utils.error(`Auspice build files not found under ${cwd} or ${sourceDir}. Did you run \`auspice build\` in either directory?`);
+      process.exit(1);
+    } else if (cwd !== sourceDir) {
```
jameshadfield, line 91: This conditional will always be true (if cwd is the sourceDir, then the if statement on line 85 will be true). I suggest this be changed to

```js
if (!hasAuspiceBuild(cwd)) {
    if (cwd === sourceDir || customBuildOnly) {
      utils.error(`Auspice build files not found under ${cwd}. Did you run \`auspice build\` in this directory?`);
      process.exit(1);
    } else if (!hasAuspiceBuild(sourceDir)) {
      utils.error(`Auspice build files not found under ${cwd} or ${sourceDir}. Did you run \`auspice build\` in either directory?`);
      process.exit(1);
    }
   utils.log(`Auspice build files not found under ${cwd}. Using build files under ${sourceDir}.`)
   baseDir = sourceDir;
}
```

I also wouldn't use an `else if` after a preceding return / exit call, but I think this is probably a stylistic thing.

cli/view.js
victorlin: Fixed in 4b9ca4dc416daca9b231162d70bead3b01259c42.

cli/view.js
victorlin: Fixed in 1d2a023e22004df7492e2cc9b7847c941ec8efeb.

cli/view.js
victorlin: Updated with your suggestions in 1d2a023e22004df7492e2cc9b7847c941ec8efeb.

cli/view.js
victorlin: I've split the addition of this argument into its own commit: 00902657df8e4391d039613712462c81fa6cc655

Ok I understand the difference now. I chose `--customBuildOnly` following the previous existence of `--customBuild`, which seemed to be doing something similar? I can't think of a better name without being verbose like `--onlyUseLocalAuspiceBuild`. Do you have suggestions?

jameshadfield:
Rebased onto master -- will re-review tomorrow

cli/view.js
jameshadfield: You're totally right!

jameshadfield (approved):

PR #1618, opened 2022-12-09
Add 'none' option to hide tip labels

joverlee521:

Adds a new 'none' option to turn off the tip labels. We should consider using a Symbol for the 'none' value so that it can't clash with potential coloring values, but then it's not as straightforward to specify this option via the URL query.

### Testing

- [x] Checks pass

### DIFF
--- a/src/components/controls/choose-tip-label.js
+++ b/src/components/controls/choose-tip-label.js
@@ -49,9 +49,15 @@ export default WithTranslation;
  */
 export function collectAvailableTipLabelOptions(colorings) {
   return [
+    /**
+     * We should consider using a Symbol for the 'none' value so that it
+     * can't clash with potential coloring values, but then it's not as
+     * straightforward to specify this option via the URL query
+     */
+    {value: 'none', label: "none"},
     {value: strainSymbol, label: "Sample Name"},
     ...Object.entries(colorings)
-      .filter((keyValue) => keyValue[0] !== 'gt')
+      .filter((keyValue) => keyValue[0] !== 'gt' && keyValue[0] !== 'none')
       .map(([key, value]) => {
         return {value: key, label: value.title};
       })
--- a/src/components/tree/phyloTree/change.js
+++ b/src/components/tree/phyloTree/change.js
@@ -392,7 +392,7 @@ export const change = function change({
 
   const extras = { removeConfidences, showConfidences, newBranchLabellingKey };
   extras.timeSliceHasPotentiallyChanged = changeVisibility || newDistance;
-  extras.hideTipLabels = animationInProgress;
+  extras.hideTipLabels = animationInProgress || newTipLabelKey === 'none';
   if (useModifySVGInStages) {
     this.modifySVGInStages(elemsToUpdate, svgPropsToUpdate, transitionTime, 1000, extras);
   } else {


src/components/controls/choose-tip-label.js
```diff
@@ -49,9 +49,15 @@ export default WithTranslation;
  */
 export function collectAvailableTipLabelOptions(colorings) {
   return [
+    /**
+     * We should consider using a Symbol for the 'none' value so that it
+     * can't clash with potential coloring values, but then it's not as
+     * straightforward to specify this option via the URL query
+     */
+    {value: 'none', label: "none"},
```
tsibley, line 57: We could use a Symbol and support selecting it via the URL query if we differentiate between `tl=` and `` in the query. Right now it seems they're treated the same, but they don't have to be.

joverlee521: If I'm understanding you correctly, that method is already used for the default tip label "Strain Name", which is also a Symbol.

tsibley: Hmm, I mean something like this:

```diff
diff --git a/src/actions/recomputeReduxState.js b/src/actions/recomputeReduxState.js
index ecefe85f..510227bb 100644
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -1,7 +1,7 @@
 import queryString from "query-string";
 import { cloneDeep } from 'lodash';
 import { numericToCalendar, calendarToNumeric } from "../util/dateHelpers";
-import { reallySmallNumber, twoColumnBreakpoint, defaultColorBy, defaultGeoResolution, defaultDateRange, nucleotide_gene, strainSymbol, genotypeSymbol } from "../util/globals";
+import { reallySmallNumber, twoColumnBreakpoint, defaultColorBy, defaultGeoResolution, defaultDateRange, nucleotide_gene, noneSymbol, strainSymbol, genotypeSymbol } from "../util/globals";
 import { calcBrowserDimensionsInitialState } from "../reducers/browserDimensions";
 import { getIdxMatchingLabel, calculateVisiblityAndBranchThickness } from "../util/treeVisibilityHelpers";
 import { constructVisibleTipLookupBetweenTrees } from "../util/treeTangleHelpers";
@@ -86,7 +86,9 @@ const modifyStateViaURLQuery = (state, query) => {
   if (query.p && state.canTogglePanelLayout && (query.p === "full" || query.p === "grid")) {
     state["panelLayout"] = query.p;
   }
-  if (query.tl) {
+  if (query.tl === "") {
+    state["tipLabelKey"] = noneSymbol;
+  } else if (query.tl) {
     state["tipLabelKey"] = query.tl;
   }
   if (query.dmin) {
diff --git a/src/components/controls/choose-tip-label.js b/src/components/controls/choose-tip-label.js
index 91bdc48d..4b12023c 100644
--- a/src/components/controls/choose-tip-label.js
+++ b/src/components/controls/choose-tip-label.js
@@ -3,7 +3,7 @@ import { connect } from "react-redux";
 import { withTranslation } from 'react-i18next';
 import { CHANGE_TIP_LABEL_KEY } from "../../actions/types";
 import { SidebarSubtitle } from "./styles";
-import { controlsWidth, strainSymbol } from "../../util/globals";
+import { controlsWidth, noneSymbol, strainSymbol } from "../../util/globals";
 import CustomSelect from "./customSelect";
 
 @connect((state) => ({
@@ -54,7 +54,7 @@ export function collectAvailableTipLabelOptions(colorings) {
      * can't clash with potential coloring values, but then it's not as
      * straightforward to specify this option via the URL query
      */
-    {value: 'none', label: "none"},
+    {value: noneSymbol, label: "none"},
     {value: strainSymbol, label: "Sample Name"},
     ...Object.entries(colorings)
       .filter((keyValue) => keyValue[0] !== 'gt' && keyValue[0] !== 'none')
diff --git a/src/util/globals.js b/src/util/globals.js
index 4b3c1575..8bbcf984 100644
--- a/src/util/globals.js
+++ b/src/util/globals.js
@@ -201,6 +201,7 @@ export const isValueValid = (value) => {
   // booleans, valid strings & numbers are valid.
   return true;
 };
+export const noneSymbol = Symbol('none');
 export const strainSymbol = Symbol('strain');
 export const genotypeSymbol = Symbol('genotype');
 export const measurementIdSymbol = Symbol('measurementId');
```

Untried, and there's probably a little more code to make sure that the URL query is updated to `tl=` when `noneSymbol` instead of omitting `tl` entirely from the query (like with the default `strainSymbol`).

tsibley: [Expanded patch](https://github.com/nextstrain/auspice/commit/3d5d1e7b7f7fa92f1759d076c509179afda51388) with that bit of extra bookkeeping (though still untried).

jameshadfield: P.S. this PR works well, and there's no necessity to use Symbols, so if this PR doesn't get revisited in the next few days I'm happy to merge as-is to get this functionality out.

tsibley: > None of this stops us using Symbols in the code.

True, but if we use `"none"` and forbid it as a valid field name, then there's no reason to use Symbols. :-)

joverlee521: Hmm, `tl=` looks weird to me too. I'm in favor of making the "none" key invalid in the schema.

jameshadfield (approved):
Thanks Jover! Merging now

PR #1631, opened 2023-02-03
Measurements paper improvements

joverlee521:

Minor improvements for the measurements panel that addresses review for the [measurements panel manuscript](https://github.com/blab/measurements-panel), which also align with improvements listed in #1463. 
- sticky x-axis
- multiple thresholds
- group raw measurements by color-by values

### Related issue(s)
- https://github.com/blab/measurements-panel/pull/4
- https://github.com/nextstrain/augur/pull/1148

### Testing

- [x] Checks pass

### DIFF
--- a/src/components/controls/measurementsOptions.js
+++ b/src/components/controls/measurementsOptions.js
@@ -112,9 +112,12 @@ const MeasurementsOptions = () => {
       />
       <Toggle
         // Only display threshold toggle if the collection has a valid threshold
-        display={typeof collection.threshold === "number"}
+        display={
+          Array.isArray(collection.thresholds) &&
+          collection.thresholds.some((threshold) => typeof threshold === "number")
+        }
         on={showThreshold}
-        label="Show measurement threshold"
+        label="Show measurement threshold(s)"
         callback={() => dispatch({type: TOGGLE_MEASUREMENTS_THRESHOLD, data: !showThreshold})}
       />
     </div>
--- a/src/components/measurements/measurementsD3.js
+++ b/src/components/measurements/measurementsD3.js
@@ -14,7 +14,7 @@ export const layout = {
   leftPadding: 180,
   rightPadding: 30,
   topPadding: 20,
-  bottomPadding: 50,
+  xAxisHeight: 50,
   subplotHeight: 100,
   subplotPadding: 10,
   circleRadius: 3,
@@ -103,9 +103,12 @@ export const groupMeasurements = (measurements, groupBy, groupByValueOrder) => {
     "asc");
 };
 
-export const clearMeasurementsSVG = (ref) => {
+export const clearMeasurementsSVG = (ref, xAxisRef) => {
   select(ref)
     .attr("height", null)
+    .selectAll("*").remove();
+
+  select(xAxisRef)
     .selectAll("*").remove();
 };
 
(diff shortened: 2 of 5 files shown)


src/components/controls/measurementsOptions.js
```diff
@@ -112,9 +112,12 @@ const MeasurementsOptions = () => {
...
       <Toggle
         // Only display threshold toggle if the collection has a valid threshold
-        display={typeof collection.threshold === "number"}
+        display={
+          Array.isArray(collection.thresholds) &&
+          collection.thresholds.some((threshold) => typeof threshold === "number")
+        }
         on={showThreshold}
-        label="Show measurement threshold"
+        label="Show measurement threshold(s)"
```
tsibley, line 120: _non-blocking_

This diff made me realize two things.

Labels like this should go thru our i18n framework (i18next) so they can be translated in the future.

```jsx
label={t("Show measurement threshold(s)")}
```

Once they do, we can avoid the ugly "(s)" by using the [standard pluralization/number agreement features](https://www.i18next.com/translation-function/plurals), e.g. something like:

```jsx
label={t("Show measurement threshold", {count: collection.thresholds.length})}
```

with corresponding translations, e.g. for `en` something like:

```json
{
  "Show measurement threshold_one": "Show measurement threshold",
  "Show measurement threshold_others": "Show measurement thresholds",
}
```

(I realize these are out of scope for this PR.)

src/components/measurements/measurementsD3.js
```diff
@@ -194,19 +194,21 @@ export const drawMeasurementsSVG = (ref, xAxisRef, svgData) => {
...
-      .attr("x2", thresholdXValue)
-      .attr("y1", layout.topPadding)
-      .attr("y2", svgHeight)
-      .attr("stroke-width", layout.thresholdStrokeWidth)
-      .attr("stroke", layout.thresholdStroke)
-      // Hide threshold by default since another function will toggle display
-      .attr("display", "none");
+  // Add threshold(s) if provided
+  if (thresholds !== null) {
+    thresholds.forEach((threshold, index) => {
```
tsibley, line 199: _non-blocking_

FWIW, you can do

```suggestion
    for (let threshold of thresholds) {
```

and not have to create a closure.

tsibley (commented):
A few things I noticed when laying 👀 on this.

joverlee521:
Saw a weird behavior with the sticky x-axis in the test app. Non-blocking, but would be good to investigate:

1. Scroll to the bottom of the measurements panel. 
2. Change filtering/options that shortens the main SVG, e.g. change "Group By" from "reference_strain" to "source".
3. See the x-axis is stuck at the bottom with a bunch of white space above.
4. Scrolling back up to the plots will fix the issue.

Screen Shot 2023-02-03 at 11 25 19 AM
Screen Shot 2023-02-03 at 11 25 30 AM
Screen Shot 2023-02-03 at 11 26 19 AM

src/components/measurements/measurementsD3.js
joverlee521: Ah, thanks for pointing this out! It also made me realize that I no longer need the `index` value here.

src/components/controls/measurementsOptions.js
joverlee521: Oh yeah, I'm totally missing the translation feature for all measurements options...

Very cool that i18next supports plurals formatting like this! 
I decided that I'm going to punt this for now as I was reading the i18next docs since we are still using i18next v19.9.2 and I haven't dug into which JSON format we are using:
> Starting with i18next>=21.3.0 you can use the built-in formatting functions based on the Intl API

> You may need to polyfill the Intl.PluralRules API, in case it is not available it will fallback to the i18next JSON format v3 plural handling.

src/components/measurements/measurementsD3.js
joverlee521: Updated in force-push.

PR #1588, opened 2022-10-25
docs: Convert to rST, add CI, and fix warnings

victorlin:
### Description of proposed changes

[docs: Fix typo](https://github.com/nextstrain/auspice/pull/1588/commits/62f684d0b221dc2d99f88996799891bb9c699a8b) 
 
[Update CI workflow to catch warnings in docs](https://github.com/nextstrain/auspice/pull/1588/commits/8ee1dbfc634d461e3619eeb8f5411e48a1a7aa2b) 

[Remove references to unused _static path](https://github.com/nextstrain/auspice/pull/1588/commits/2dfc68dd1a33d4987f049fca7acde9c3f983cee2) 
 
[Update link to nextstrain.org handlers example](https://github.com/nextstrain/auspice/pull/1588/commits/11dca19e41e534b5be3623f6fd0655208fbb498c) 
 
[Use markdown heading for changelog](https://github.com/nextstrain/auspice/pull/1588/commits/03423b54cb0a61221d0b88ab36c3891193b4e52f) 

[Convert markdown pages to reStructuredText](https://github.com/nextstrain/auspice/pull/1588/commits/7b8e45e204d68a6de1bd65d63e829d13e72ea9c6) 

[Fix links and overall rST formatting](https://github.com/nextstrain/auspice/pull/1588/commits/420a74e2ed8f3f0d5f9f57f86f04cba08133c980) 

[Remove DEV_DOCS.md symlinking](https://github.com/nextstrain/auspice/pull/1588/commits/65a82cab81ee089dbb004adc83f009d6b692a6db) 

[Use '//' to denote comments in JSON code block](https://github.com/nextstrain/auspice/pull/1588/commits/78b943999287065e85f5072252cb549d9b3abb52) 

[Document Plausible on client requests page](https://github.com/nextstrain/auspice/pull/1588/commits/653f4403211b2e038b2084f1e7244fe1bc143ac9)

### Testing

- [x] Checks pass
- [x] Manually check pages in [docs preview](https://nextstrain--1588.org.readthedocs.build/projects/auspice/en/1588/).

PR #1638, opened 2023-03-10
docs: Improve TSV examples

tsibley:
See commit messages.

### Related issue(s)

#1588, https://github.com/nextstrain/auspice/pull/1588#discussion_r1131738370

### Testing

- [x] [Preview](https://nextstrain--1638.org.readthedocs.build/projects/auspice/en/1638/advanced-functionality/drag-drop-csv-tsv.html#example) looks good
- [x] Checks pass


victorlin (approved):
Slick! I learn something new about Sphinx every time you make a PR. Looks like I should be referencing [Sphinx docs](https://www.sphinx-doc.org/) more, or just browsing it in general...

tsibley:
@victorlin

> I learn something new about Sphinx every time you make a PR. Looks like I should be referencing [Sphinx docs](https://www.sphinx-doc.org/) more, or just browsing it in general...

:blush: \o/

Yeah, I think _browsing_ official docs, in addition to using them for reference and search, is really underrated. The main reason is that browsing lets you build a mental index of what's available (and what's not), which is very useful before the reference/search phase of doc usage. ([I'm not alone either](https://twitter.com/hillelogram/status/1529509327840722946).)

<p align="center">⁂ ⁂ ⁂ </p>

I'm going to wait a bit for @jameshadfield to chime in here since he had [previous thoughts on these examples](https://github.com/nextstrain/auspice/pull/1588#discussion_r1131738370).

jameshadfield (approved):
Aesthetics are much improved here. I advised against tables as I wanted to reinforce that these are just plain-old text files, but the download button probably achieves this.

tsibley:
@jameshadfield Thanks for the :eyes:. Agreed good to reinforce these are plain text files! I think the download links maybe do that a _bit_ less than inlined verbatim blocks, but also think that the improved reading experience of table + download is probably worth it.

PR #1641, opened 2023-03-13
Update CI

victorlin:

Small improvements here and there to improve build times and give the lockfile a little more attention. See commit messages.

### Related issue(s)

_N/A_

### Testing

- [x] New check-lockfile job works ([proof](https://github.com/nextstrain/auspice/actions/runs/4471191373/jobs/7855742618?pr=1641))
- [x] Manually test that steps 2-3 in `releaseNewVersion.sh` make the right changes to package.json and package-lock.json
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ I don't think this is necessary as there are no changes to the Auspice source code.

### DIFF
--- a/releaseNewVersion.sh
+++ b/releaseNewVersion.sh
@@ -70,8 +70,8 @@ select yn in "major-new-release" "feature-release" "minor-fix"; do
     esac
 done
 echo -e "\n"
-# now replace the version in packages.json and version.js (inplace!)
-perl -pi -e "s/\"version\": \"${packagesVersion}\"/\"version\": \"${newVersion}\"/" package.json
+# now replace the version in NPM files and version.js (inplace!)
+npm version "${newVersion}" --no-git-tag-version
 perl -pi -e "s/version = \"${srcVersion}\";/version = \"${newVersion}\";/" src/version.js
 unset packagesVersion srcVersion parts bumps yn
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,7 +81,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version: 18
           registry-url: https://registry.npmjs.org/
       - run: node --version
       - run: npm ci
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,8 +19,8 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - run: node --version
+      # Build is implicit with the "prepare" life cycle script
       - run: npm ci
-      - run: npm run build
 
   unit-test:
     runs-on: ubuntu-latest
@@ -33,7 +33,7 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - run: node --version
-      - run: npm ci
+      - run: npm ci --ignore-scripts
       - run: npm test
 
   smoke-test:
@@ -60,7 +60,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node }}
-      - run: npm ci
+      - run: npm ci --ignore-scripts
       - run: npm run lint
 
   bundlesize:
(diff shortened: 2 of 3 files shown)


releaseNewVersion.sh
```diff
@@ -74,6 +74,8 @@ echo -e "\n"
 perl -pi -e "s/\"version\": \"${packagesVersion}\"/\"version\": \"${newVersion}\"/" package.json
 perl -pi -e "s/version = \"${srcVersion}\";/version = \"${newVersion}\";/" src/version.js
 unset packagesVersion srcVersion parts bumps yn
+# apply version changes to the NPM lockfile
+npm install --ignore-scripts
```
jameshadfield, line 78: Hmm. This will automatically bump dependencies (within the allowed range) which may result in the released version differing from the version we tested and "signed off on". I do like the idea of updating dependencies, but not sure this is the way to go about it...

jameshadfield (commented):
Changes looks good by inspection, save for the commented line. And CI passed, so that's something!

releaseNewVersion.sh
victorlin: Changed – see cf66d0332653ff2ec133dbe8c7b951d9fac7abd6

jameshadfield: Ahh - I interpreted the intention here as version changes (updates) to dependencies, not version changes of auspice!

releaseNewVersion.sh
```diff
@@ -70,8 +70,10 @@ select yn in "major-new-release" "feature-release" "minor-fix"; do
     esac
 done
 echo -e "\n"
-# now replace the version in packages.json and version.js (inplace!)
-perl -pi -e "s/\"version\": \"${packagesVersion}\"/\"version\": \"${newVersion}\"/" package.json
+# now replace the version in NPM files and version.js (inplace!)
+npm version "${newVersion}"
```
jameshadfield, line 74: >    If run in a git repo, it will also create  a  version  commit  and  tag.
       This  behavior  is controlled by git-tag-version (see below), and can be
       disabled on the command line by running  npm  --no-git-tag-version  ver-
       sion.  It will fail if the working directory is not clean, unless the -f
       or --force flag is set.
       
So I think we want to run with `--no-git-tag-version`, as we use our own tagging from the release branch. (There's really no need for a release branch, we should just tag the master branch and publish from that... but that's beyond this PR!)

victorlin: I overlooked this option! Updated to use `--no-git-tag-version` in force push.

jameshadfield (approved):
Reading through the code looks good but I haven't tested it out. Happy for you to merge whenever you wish and then we could do a release together?

releaseNewVersion.sh
corneliusroemer: We should update dependencies straight _after_ the release so we have it tested for the next one

tsibley:
From the message of e29c01cdd2d9d8ba35dee04bc190af1b3b23349c:

> Note that `npm ci --ignore-scripts && npm run build` is NOT the same as `npm ci` alone, since --ignore-scripts also ignores life cycle scripts defined by dependencies. This is why `npm ci` is still necessary for smoke-test even though it doesn't explicitly require the Auspice build.

This seems precarious… If we don't currently have deps with lifecycle scripts that are required to function, it seems like we could easily find ourselves with one in the future?

victorlin:
@tsibley That commit only skips lifecycle scripts in the `unit-test` and `lint` jobs. It's possible that those might require lifecycle scripts in the future, but when it comes to that it should be obvious and the change can be reverted.

tsibley:
Yeah, I saw that. Those jobs still both invoke a lot of dependency code though, right?

> but when it comes to that it should be obvious and the change can be reverted.

I would put money on the reason for such a failure being non-obvious. :-) It'll be obvious, probably, that "module X suddenly broke", but not _why_ (e.g. "X released a new version, which relies on a lifecycle script for proper functioning, and so it no longer works").

But maybe lifecycle scripts are uncommon in published modules…

In any case, we can wait and see!

victorlin:
@tsibley I see. How about #1652?

tsibley:
Ha, sure. :-)

PR #1642, opened 2023-03-13
Replace bundlesize with bundlesize2

victorlin:
### Description of proposed changes

bundlesize was last published 2 years ago and has indirect dependencies that are deprecated, such as [iltorb](https://www.npmjs.com/package/iltorb/v/2.4.5).

The author has published a new version under the name bundlesize2 with a [migration guide](https://github.com/siddharthkp/bundlesize2/blob/5d5b61387d59e7aaf63e1b662c6b0b6a6592203c/README.md#migration-from-bundlesize0180). I can't tell what benefits it has over the original bundlesize, but it seems backwards compatible (at least with usage in this repo) and comes with more up-to-date dependencies.

### Related issue(s)

N/A

### Testing

- [x] Checks pass
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ I don't think this is necessary as there are no changes to the Auspice source code.

[1]: https://github.com/nextstrain/auspice/blob/-/DEV_DOCS.md#test-on-downstream-repositories

PR #1645, opened 2023-03-14
Remove bundlesize from list of devDependencies

victorlin:

bundlesize has some indirect dependencies which require extra build time. Currently, it is only used for CI, so the extra time required to build/install is unnecessary for most use cases.

Instead of installing bundlesize alongside Auspice, install it through npx where it is used. Installing through npx installs to a folder in the npm cache.

### Related issue(s)

Alternative to #1642 (closes #1642)

### Testing

- [x] Checks pass
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ I don't think this is necessary as there are no changes to the Auspice source code.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,19 +60,12 @@ jobs:
       - run: npm run lint
   bundlesize:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node: [14, 16, 18]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node }}
+          node-version: 18
       - run: npm ci
-      # Node.js 14 comes with NPM 6 which auto-installs packages by default.
-      - if: ${{ matrix.node == 14 }}
-        run: npx bundlesize
-      - if: ${{ matrix.node != 14 }}
         run: npx --yes bundlesize
   publish:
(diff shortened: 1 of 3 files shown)


tsibley (approved):

.github/workflows/ci.yaml
```diff
@@ -69,7 +69,11 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - run: npm ci
-      - run: npx bundlesize
+      # Node.js 14 comes with NPM 6 which auto-installs packages by default.
+      - if: ${{ matrix.node == 14 }}
+        run: npx bundlesize
+      - if: ${{ matrix.node != 14 }}
```
jameshadfield, line 75: I think it's enough to compute bundlesizes on a single node version, but equally happy to do them across all of them (not much compute).

jameshadfield (approved):

.github/workflows/ci.yaml
victorlin: Good point. It looks like the bundles have identical filenames and sizes on Node.js [14](https://github.com/nextstrain/auspice/actions/runs/4418719566/jobs/7746203399#step:5:51), [16](https://github.com/nextstrain/auspice/actions/runs/4418719566/jobs/7746203686#step:6:6), [18](https://github.com/nextstrain/auspice/actions/runs/4418719566/jobs/7746204105#step:6:6). I'll push a commit to use 18 only, then merge this.

PR #1649, opened 2023-03-18
Create download TSVs with Papa Parse

### LINKED ISSUE #1648: Unescaped quotes in downloaded TSV file
### Current Behavior
Some acknowledgement files cannot be read correctly into Excel (and also sometimes R, depending on quote settings), with rows going 'missing'. 
An acknowlegement file downloaded from [23A.Omicron focal build](https://nextstrain.org/groups/neherlab/ncov/23A.Omicron) on 14 Mar (the build has since been updated, haven't tried the newer file) contains 6346 sequences & examining the resulting acknowledgement file in Visual Studio Code shows it also has 6347 lines (including header).

However, if the file is opened in Excel it seems to have 5308 lines (including header) and if read into R with `read.csv(...quote="\'")` (allowing `"` to quote leads to an EOF error) results in 3224 rows (excluding header).

This is confusing to users.

### Expected behavior
Downloaded acknowledgement files should have as many rows/lines as sequences in the downloaded set.

### How to reproduce
1. At least on the 15 Mar 23, download acknowledgements files from [23A.Omicron build](https://nextstrain.org/groups/neherlab/ncov/23A.Omicron) -- should have 6567 sequences
2. Drag-and-drop into blank Excel 
3. Scroll to see number of lines - is 5623

### Possible reason
For unclear reasons, sometimes originating & submitting labs and authors can be surrounded by double-quotes in GISAID:
image
This (unclosed `"`) happens at least 17 times in a dataset of 6300 seqs, so it's not so uncommon. 
I haven't seen any signs it also happens with single quotes, but I can't rule it out entirely

When this happens with originating/submitting labs, this doesn't seem to be an issue - they come out with the quotes. However, for authors, if we then want to shorten to `et al` ending (as we do [here](https://github.com/nextstrain/ncov-ingest/blob/59b2a36171d8379a284abb21f450f8c640e38c28/lib/utils/transformpipeline/transforms.py#L370)), this replaces the end, leaving a hanging `"`. 
From the above example, it goes from:
`"Sylvanus Okogbenin, Cyril Erameh, Sophie Duraffour, Stephan Guenther"`
to
`"Sylvanus Okogbenin et al`

### Possible solution
We should either decide to:
1. Strip all quotes entirely (could displease places that would prefer to have internal quotes, ex `Mühlenkreiskliniken Standort Minden: Institut für Laboratoriumsmedizin Mikrobiologie Hygiene und Transfusionsmedizin; wird auch unter dem Namen "LaborZentrum Weser" geführt`)
2. Strip all leading/ending quotes before any further transformation (could lead to more issues without careful matching- example `RSE "National Center of Expertise" and RSE "National center for Biotechnology"` (ending quote, but no leading quote)), 
3. Remote all quotes from Authors only
4. Add a trailing quote back in when converting to `et al.` for Authors.

joverlee521:

Creates download TSVs with Papa Parse instead of naively joining with tabs to ensure that we have properly formatted TSVs. 

### Related issue(s)

Fixes #1648 

### Testing

- [x] Checks pass

### DIFF
--- a/src/components/download/helperFunctions.js
+++ b/src/components/download/helperFunctions.js
@@ -1,4 +1,5 @@
 /* eslint no-restricted-syntax: 0 */
+import { unparse } from "papaparse";
 import { infoNotification, warningNotification } from "../../actions/notifications";
 import { spaceBetweenTrees } from "../tree/tree";
 import { getTraitFromNode, getDivFromNode, getFullAuthorInfoFromNode, getVaccineFromNode, getAccessionFromNode } from "../../util/treeMiscHelpers";
@@ -107,6 +108,43 @@ const treeToNexus = (tree, colorings, colorBy, temporal) => {
   ].join("\n");
 };
 
+/**
+ * Create a properly formatted TSV string for given data using Papa.unparse().
+ *
+ * Each object within the data array should represent a single row in the
+ * TSV string. All values of the object will be converted to their string
+ * representation via `toString` within unparse
+ * (see https://github.com/mholt/PapaParse/blame/824bbd9daf17168bddfc5485066771453cab423e/papaparse.js#L464).
+ *
+ * The optional columns parameter allows you to specify the specific keys to
+ * use as columns in the TSV string. Note, order of column names will
+ * determine order of output columns in the TSV string.
+ *
+ * If columns are not specified, then parser will use the keys of the first
+ * Object in the data array as the columns for the TSV string.
+ *
+ * See Papa Parse docs for more details about config options: https://www.papaparse.com/docs#json-to-csv
+ *
+ * @param {Array<Object>} data
+ * @param {Array<string>|null} columns
+ * @returns {string}
+ */
+const createTsvString = (data, columns=null) => {
+  return unparse(
+    data,
+    {
+      quotes: false,
+      quoteChar: '"',
+      escapeChar: '"',
+      delimiter: "\t",
+      header: true,
+      newline: "\n",
+      skipEmptyLines: true,
+      columns
+    }
+  );
+};
+
 const write = (filename, type, content) => {
   /* https://stackoverflow.com/questions/18848860/javascript-array-to-csv/18849208#comment59677504_18849208 */
   const blob = new Blob([content], { type: type });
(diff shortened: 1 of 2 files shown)


src/components/download/helperFunctions.js
```diff
@@ -107,6 +108,37 @@ const treeToNexus = (tree, colorings, colorBy, temporal) => {
...
+ * The optional columns parameter allows you to specify the specific keys to
+ * use as columns in the TSV string. Note, order of column names will
+ * determine order of output columns in the TSV string.
+ *
+ * If columns are not specified, then parser will use the keys of the first
+ * Object in the data array as the columns for the TSV string.
+ *
+ * See Papa Parse docs for more details about config options: https://www.papaparse.com/docs#json-to-csv
+ *
+ * @param {Array<Object>} data
```
jameshadfield, line 122: [not blocking] if a data element is `{key: <Array>}`, as it is for the "strains" column in `authorTSV`, then the array is stringified via `array.join(",")` (at least that what seems to be happening), but I couldn't see this in the linked docs. Would be nice to add this to the docstring here.

jameshadfield (approved):
Much improved code as well as the bugfix - thanks @joverlee521!

---

For posterity, using the [23A.Omicron](https://nextstrain.org/groups/neherlab/ncov/23A.Omicron) build updated 2023-03-16 with 6533 tips, here's a summary of what was happening:

The Auspice JSON has entries such as

```json
"author": {
  "author": "\"Flores et al",
  "value": "\"Flores et al"
},
```

A diff between the previous (broken) TSV and the TSV from this PR:

```diff
314c314
< USA/SC-0708/2021      EPI_ISL_17137584                Microbiology Division, SC DHEC  Microbiology Division, SC DHEC  "Flores et al
---
> USA/SC-0708/2021      EPI_ISL_17137584                Microbiology Division, SC DHEC  Microbiology Division, SC DHEC  """Flores et al"
```

Which uses the [seemingly correct](https://webapps.stackexchange.com/a/26867) escaped quotes syntax and is read correctly by MacOS Numbers. The broken TSV, when parsed in Numbers, joined multiple lines into a single "author" field due to the un-escaped `"` characters which resulted in a 3600-row spreadsheet in this case.

src/components/download/helperFunctions.js
joverlee521: Ah good point! I did a little digging in the papaparse source code and this happens because they call [`str.toString()` for all values](https://github.com/mholt/PapaParse/blame/824bbd9daf17168bddfc5485066771453cab423e/papaparse.js#L464) in the data object.

I'll add a note a here that all values will be converted to the string representation of the object.

PR #1655, opened 2023-03-22
Type cast continuous values in tree

### LINKED ISSUE #1626: Scatterplot not displaying negative values for ACE2 binding
The scatterplot view does not display the negative values for [ACE2 binding](https://nextstrain.org/ncov/gisaid/global/6m?c=ace2_binding&l=scatter&scatterX=ace2_binding):

Screen Shot 2023-01-19 at 2 16 14 PM

The negative values _**are**_ displayed properly for other continuous color-bys such as [Logistic Growth](https://nextstrain.org/ncov/gisaid/global/6m?branches=hide&c=logistic_growth&l=scatter&scatterX=logistic_growth) and [Mutational Fitness](https://nextstrain.org/ncov/gisaid/global/6m?branches=hide&c=mutational_fitness&l=scatter&scatterX=mutational_fitness):

 Screen Shot 2023-01-19 at 2 31 57 PM

This was a bug reported/discovered during Nextstrain office hours on 2023-01-19.

jameshadfield:
Conceptually these values only make sense if they are numeric. Previously some implicit casting was happening, but this resulted in bugs such as negative values not being displayed in the scatterplot.

Note one slight side-effect which may be considered a regression: previously string values which were not numbers (e.g. "abc" but not "1.0") would show up in the hover/click info-boxes. Now such values will be changed to `undefined` and therefore not shown in those info-boxes.

Timing of the changes introduced here (AppleM1, Firefox 111):
* nextclade/sars-cov-2/21L   7-9ms
* ncov/gisaid/global/6m      4-10ms
* zika                       2ms

Closes #1626

### DIFF
--- a/src/util/castJsonTypes.js
+++ b/src/util/castJsonTypes.js
@@ -1,0 +1,56 @@
+/**
+ * Values in the JSON should be appropriately typed however this may not always be the case.
+ * For instance, certain values of a continuous trait may be strings, and we should cast
+ * these to numbers. See https://github.com/nextstrain/auspice/issues/1626 for more discussion
+ * of this particular case. Feel free to add more checks / casts to this function!
+ * @param {Object} metadata
+ * @param {Object} tree
+ * @returns {undefined} Any type casting is in-place
+ */
+export const castIncorrectTypes = (metadata, tree) => {
+  try {
+    const continuousKeys = new Set();
+    Object.entries(metadata.colorings || {}).forEach(([key, details]) => {
+      if (details.type==="continuous") {
+        continuousKeys.add(key);
+      }
+    });
+    tree.nodes.forEach((node) => {
+      for (const key of continuousKeys) {
+        const attr = node?.node_attrs?.[key];
+        if (!attr) continue;
+        continuousAttrValueToNumber(attr);
+      }
+    });
+  } catch (err) {
+    // type casting shouldn't be required (the JSON should be correctly typed)
+    // so any errors shouldn't prevent Auspice loading
+    console.error(err);
+  }
+};
+
+/**
+ * Cast `attr.value` to either a Number or undefined.
+ * Note that "Infinity" (string) is changed to undefined, but `Infinity` (Number)
+ * is left alone as the latter is not possible to encode in JSON.
+ * @param {Record<string, any>} attr
+ */
+export function continuousAttrValueToNumber(attr) {
+  switch (typeof attr.value) {
+    case "number":
+      break;
+    case "string":
+      const value = attr.value.trim();
+      if (value === "" || isNaN(value) || value ==="Infinity" || value ==="-Infinity") {
+        // Note: Number("")=0
+        // undefined values are handled appropriately (e.g. scatterplots, tooltips etc)
+        attr.value = undefined;
+      } else {
+        attr.value = Number(value);
+      }
+      break;
+    default:
+      // any other types (Boolean, Null ('object')) are not valid for a continuous scale
+      attr.value = undefined;
+  }
+}
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -11,6 +11,7 @@ import { getDefaultMeasurementsState } from "../reducers/measurements";
 import { countTraitsAcrossTree, calcTotalTipsInTree } from "../util/treeCountingHelpers";
 import { calcEntropyInView } from "../util/entropy";
 import { treeJsonToState } from "../util/treeJsonProcessing";
+import { castIncorrectTypes } from "../util/castJsonTypes";
 import { entropyCreateState } from "../util/entropyCreateStateFromJsons";
 import { determineColorByGenotypeMutType, calcNodeColor } from "../util/colorHelpers";
 import { calcColorScale, createVisibleLegendValues } from "../util/colorScale";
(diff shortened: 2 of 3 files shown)


corneliusroemer:
Thanks for working on this!

It doesn't fix the issue in my test (maybe I'm testing wrong?)

Repro:

```
mkdir tmp
curl https://gist.githubusercontent.com/corneliusroemer/ecf99b3ba5bce5aef57a5538ac50da20/raw/98099bf349c50b7c23d73d58fc6e1a4d46744287/auspice.json -o tmp/nextclade_21L.json
npm run view -- --datasetDir tmp       
```

Then open in browser: http://localhost:4000/nextclade/21L?branches=hide&c=ace2_binding&l=scatter&p=full&scatterX=ace2_binding&scatterY=immune_escape

See results (still not showing negative):
image

Maybe I'm testing wrong, see #1656

tsibley:
@corneliusroemer You'll need to `npm run build` when switching branches. `npm run view` doesn't automatically rebuild (it's not like `cargo run`, [as I just noted elsewhere](https://github.com/nextstrain/auspice/issues/1656#issuecomment-1481562945)). Alternatively, you can `npm run dev` instead of `npm run view` to use a auto-rebuilding, hot-reloading version.

corneliusroemer:
Thanks @tsibley! I'm still struggling unfortunately.

Can you adapt this script to show me how to use what you suggest?

This:

```
npm run build
npm run view -- --datasetDir tmp 
```

doesn't seem to work.

Neither does

```
npm run dev -- --datasetDir tmp  
```

src/util/castJsonTypes.js
```diff
@@ -0,0 +1,37 @@
...
+    Object.entries(metadata.colorings || {}).forEach(([key, details]) => {
+      if (details.type==="continuous") {
+        continuousKeys.add(key);
+      }
+    });
+    tree.nodes.forEach((node) => {
+      Object.entries(node.node_attrs || {}).forEach(([key, details]) => {
+        if (continuousKeys.has(key)) {
+          if ((typeof details.value) !== "number") {
+            if (details.value==="" || isNaN(details.value)) { // Note: Number("")=0
```
joverlee521, line 22: Any string with only whitespaces also get converted to 0 because Number() ignores leading and trailing whitespace/line terminators. 

`details.value.trim() === ""` should catch those cases here.

joverlee521: Should we also check for `null` and boolean values here? 
`null` gets converted to 0 and booleans get converted to 1/0.

src/util/castJsonTypes.js
```diff
@@ -0,0 +1,37 @@
...
+    const continuousKeys = new Set();
+    Object.entries(metadata.colorings || {}).forEach(([key, details]) => {
+      if (details.type==="continuous") {
+        continuousKeys.add(key);
+      }
+    });
+    tree.nodes.forEach((node) => {
+      Object.entries(node.node_attrs || {}).forEach(([key, details]) => {
+        if (continuousKeys.has(key)) {
+          if ((typeof details.value) !== "number") {
```
joverlee521, line 21: (_nitpick_) Can these be a single if statement?

joverlee521 (changes requested):
Thanks @jameshadfield! Reviewing this forced me to read more about `isNaN` and `Number`. Never realized they worked in such unexpected ways.

src/util/castJsonTypes.js
```diff
@@ -0,0 +1,37 @@
...
+            if (details.value==="" || isNaN(details.value)) { // Note: Number("")=0
+              // undefined values are handled appropriately (e.g. scatterplots, tooltips etc)
+              details.value = undefined;
+            } else {
+              details.value = Number(details.value);
```
tsibley, line 26: Not sure if intended or not, but this will allow `Infinity` and `-Infinity` strings to be parsed into their respective global float values. There's not normally a way to represent those as a number in JSON.

src/util/castJsonTypes.js
jameshadfield: Good catches! I'll refactor next week to make any values other than strings & numbers `undefined`, and also trim the strings & deal with infinities.

src/util/castJsonTypes.js
jameshadfield: (New design means this is not applicable)

src/util/castJsonTypes.js
jameshadfield: As of [c7ac36f](https://github.com/nextstrain/auspice/pull/1655/commits/c7ac36f12ff9a282964831ec32df77b44ce29bef) these will now be cast to undefined.

src/util/castJsonTypes.js
jameshadfield: As of [c7ac36f](https://github.com/nextstrain/auspice/pull/1655/commits/c7ac36f12ff9a282964831ec32df77b44ce29bef) those examples will now all be `undefined`.

src/util/castJsonTypes.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+    });
+  } catch (err) {
+    // type casting shouldn't be required (the JSON should be correctly typed)
+    // so any errors shouldn't prevent Auspice loading
+    console.error(err);
+  }
+};
+
+export function toNumber(attr) {
```
joverlee521, line 32: _nitpick_

I would rename this function to be more specific, e.g. `continuousAttrValueToNumber`, since it designed to only be used in this specific case.

jameshadfield: Done in [eca540e](https://github.com/nextstrain/auspice/pull/1655/commits/eca540e209ffcc56121b260c9ec5152b08e7a0a2)

PR #1659, opened 2023-03-30
wip! Upgrade to React 18

tsibley:
Testing an upgrade to React 18 in order to use some new features in downstream projects.

The version spec should potentially be something like `^16 | ^18` so that Auspice can be flexible and let downstream integrations take their pick. This would avoid forcing all downstream integrations to upgrade.

### Testing

- [ ] Checks pass


jameshadfield:
Nice! Auspice uses a bunch of class methods that are now removed (at least, [that's my understanding from these release notes](https://legacy.reactjs.org/blog/2018/03/29/react-v-16-3.html) which is why I've been holding off doing this but it would be great to upgrade here.

tsibley:
@victorlin I just pushed what I had in progress. It resolves all peer deps like `react ^16` and so npm is now happy, finally. However, I wouldn't rule out packages with peer deps like `react >=16` actually not working on 18. Compilation with Webpack still fails, as only packages have been updated, not our usage of them.

I'm going to put this down now, so please feel free to pick it up if you'd like!

tsibley:
Also, to clarify, I don't need this upgrade in order to prototype the thing I need React 18 for in nextstrain.org.

jameshadfield:
Rebased onto master & moved this forward a little bit. 

`auspice build` and `auspice develop` work well for me locally. CI only failing due to bundlesize & linting.

> Auspice uses a bunch of class methods that are removed in React 18 ... which is why I've been holding off doing this but it would be great to upgrade here.

We had already renamed these lifecycle methods to `UNSAFE_` and there's seemingly no timeline for when these will be removed, so this is not a blocker.

Remaining tasks:

- [x] Upgrade redux packages (`react-redux` 7.2.9 → 9.1.0, `@reduxjs/toolkit` 1.9.7 → 2.1.0)
- [x] Reinstate hot reloading (using the new-but-still-experimental fast-refresh approach)
- [ ] `CSSTransitionGroup`. Used for notification styling. Could be changed completely, no need to restore prior behaviour exactly.
- [ ] `react-virtualized` Without this the react-select menu is (was) slow, and we render (hundreds of) thousands entries here. 
- [ ] A bunch more testing!

PR #1660, opened 2023-03-31
Move dev installation to DEV_DOCS

### LINKED ISSUE #1656: Add "how to build/run local version of Auspice CLI" to dev docs
I'm trying to test @jameshadfield's PR #1655. The new feature applies only to very specific datasets, none of which are in the sample datasets that come with the heroku test build.

So I need to run auspice locally and pass in my own dataset through the CLI.

However, being a JS noob, I can't figure out how. And unfortunately the dev_docs.md don't seem to help me neither.

It would be nice if there was a simple section explaining how to run the auspice CLI based off the current code base, the equivalent of `pip install -e .` (which would also make the CLI entry point available in path), or something more like `pipx` not putting it in path but allowing me to pass execute it once, something like that must exist!

I want something like `cargo run` (Rust) where it runs the CLI off the current code base.

victorlin:

This keeps the Sphinx docs user-focused and makes DEV_DOCS.md more complete.

### Related issue(s)

Resolves #1656

### Testing

- [x] Checks pass, though this change shouldn't affect that.
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ N/A, no functional change

### DIFF
--- a/docs/introduction/install.rst
+++ b/docs/introduction/install.rst
@@ -36,18 +36,7 @@ If you look at the :doc:`release notes <../releases/changelog>` you can see the 
 Install Auspice as a developer
 ==============================
 
-This is useful for debugging, modifying the source code, or using an unpublished feature branch.
-
-.. code:: bash
-
-   # grab the GitHub auspice repo
-   git clone https://github.com/nextstrain/auspice.git
-   cd auspice
-
-   # install dependencies and make `auspice` available globally
-   npm install --global .
-
-Updating Auspice should only require pulling the new version from GitHub - it shouldn't require any ``npm`` commands. You will, however, have to re-build Auspice whenever the client-related code has changed, via ``auspice build``.
+See `DEV_DOCS.md <https://github.com/nextstrain/auspice/blob/-/DEV_DOCS.md#developer-installation>__`.
 
 Testing if it worked
 ====================
--- a/DEV_DOCS.md
+++ b/DEV_DOCS.md
@@ -19,9 +19,24 @@ This project strictly adheres to the [Contributor Covenant Code of Conduct](http
 
 Please see the [project boards](https://github.com/orgs/nextstrain/projects) for currently available issues.
 
-## Contributing code
+## Developer Installation
 
-Code contributions are welcomed! [Please see the main auspice docs](https://docs.nextstrain.org/projects/auspice/en/latest/introduction/install.html) for details on how to install and run auspice from source.
+This is useful for debugging, modifying the source code, or using an unpublished feature branch.
+
+You can install using your system `npm` or [within a Conda environment](https://docs.nextstrain.org/projects/auspice/en/stable/introduction/install.html#install-dependencies) to keep things isolated.
+
+```sh
+# grab the GitHub auspice repo
+git clone https://github.com/nextstrain/auspice.git
+cd auspice
+
+# install dependencies and make `auspice` available globally
+npm install --global .
+```
+
+Updating Auspice should only require pulling the new version from GitHub - it shouldn't require any `npm` commands. You will, however, have to re-build Auspice whenever the client-related code has changed, via `auspice build`.
+
+## Contributing code
 
 Please comment on an open issue if you are working on it.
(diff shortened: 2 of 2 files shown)


docs/introduction/install.rst
```diff
@@ -36,18 +36,7 @@ If you look at the :doc:`release notes <../releases/changelog>` you can see the
...
-
-   # grab the GitHub auspice repo
-   git clone https://github.com/nextstrain/auspice.git
-   cd auspice
-
-   # install dependencies and make `auspice` available globally
-   npm install --global .
-
-Updating Auspice should only require pulling the new version from GitHub - it shouldn't require any ``npm`` commands. You will, however, have to re-build Auspice whenever the client-related code has changed, via ``auspice build``.
+See `DEV_DOCS.md <https://github.com/nextstrain/auspice/blob/-/DEV_DOCS.md#developer-installation>__`.
```
jameshadfield, line 39: Tangential question - `blob/-` redirects to `blob/master` - does `-` mean "default branch"? Is this a git thing or a GitHub thing?

jameshadfield (approved):

docs/introduction/install.rst
victorlin: Yeah, this is a GitHub thing. Note that all of these are similar:

- https://github.com/nextstrain/auspice/blob/master/DEV_DOCS.md
- https://github.com/nextstrain/auspice/blob/-/DEV_DOCS.md
- https://github.com/nextstrain/auspice/blob/HEAD/DEV_DOCS.md
- https://github.com/nextstrain/auspice/blob/@/DEV_DOCS.md

`master` is the explicit default branch, `-` redirects to the default branch, and `HEAD`/`@` redirect to the last commit on the default branch. Also, `-` seems to only work on certain endpoints and not others, while `HEAD` is more widely supported:

- https://github.com/nextstrain/auspice/raw/-/DEV_DOCS.md
- https://github.com/nextstrain/auspice/raw/HEAD/DEV_DOCS.md

tsibley: In this case, GitHub aped Git. Git accepts `@` as shorthand for `HEAD` (which on any git remote (not just GitHub) is the default ref and in a git clone is the currently checked out ref) and `-` as shorthand for "the previous ref" (similar to `cd -` for example).

tsibley, line 39: This move separates the developer setup from the beginning of the install doc which recommends installing a separate Node.js version.

https://github.com/nextstrain/auspice/blob/471fab1424be784bccf3d4470a944adabaeb3442/docs/introduction/install.rst?plain=1#L11-L21

Do we want to keep that suggestion intact?

tsibley: Sure.

tsibley (approved):

docs/introduction/install.rst
jameshadfield: How does "the previous ref" refer to the default _branch_ here?

Interestingly, in GitHub URLs `@` doesn't redirect to `HEAD` (but `-` does redirect to the default branch), so I wonder if there is some subtlety there on GitHub's side, or whether they really are the same.

tsibley: Well, "previous ref" doesn't make much sense in the context of GitHub, so I assume they chose to use it as a "default branch" alias instead. I'm guessing that GitHub supports `-` mainly for forward-proofing against default branch name changes, and that it was intentional and implemented closer to the frontend as a redirect (e.g. the backend never sees `-`). Whereas with `@` and `HEAD`, I'm guessing those are supported more by happenstance because their Git backend (which is not official git for most things, but their own implementation) has to support it for other reasons (e.g. git interop), and the frontend merely passes it thru to the backend.

But that's all speculation!

tsibley: (But I think reasonable speculation that fits all the observed evidence, for example, why you can use other [revision syntax](https://git-scm.com/docs/git-rev-parse#_specifying_revisions) like `@{2.years.ago}` in those URLs too.)

PR #1663, opened 2023-04-19
Fix measurements narratives

### LINKED ISSUE #1603: Narratives with multiple datasets with measurements
### Current Behavior
Narratives error when navigating between slides that contain different datasets that both have measurements sidecar JSONs.

### How to reproduce
1. Drag this [test narrative](https://github.com/nextstrain/auspice/files/10118176/test_multiple_dataset_measurements.md) onto the [narratives debugger](https://next.nextstrain.org/edit/narratives).
2. Successfully navigate between slide 1 and 2. 
3. Navigating to slide 3 or 4 will result in error. 

### Additional context
First saw this error in a private narrative for seasonal flu that switched between builds of h3n2, h1n1pdm, and vic that all had measurements sidecars.

joverlee521:
See commits for details.

Fixes #1603 

### Testing

- [x] Checks pass

### DIFF
--- a/src/actions/measurements.js
+++ b/src/actions/measurements.js
@@ -95,13 +95,7 @@ const getCollectionDisplayControls = (controls, collection) => {
   return newControls;
 };
 
-export const loadMeasurements = (json) => (dispatch, getState) => {
-  // TODO: Load controls from state to get potential url query parameters
-  const { tree, controls } = getState();
-  if (!tree.loaded) {
-    throw new Error("tree not loaded");
-  }
-
+export const parseMeasurementsJSON = (json) => {
   const collections = json["collections"];
   if (!collections || collections.length === 0) {
     throw new Error("Measurements JSON does not have collections");
@@ -193,9 +187,19 @@ export const loadMeasurements = (json) => (dispatch, getState) => {
     );
   });
 
+  return {collections, defaultCollection: json["default_collection"]};
+};
+
+export const loadMeasurements = ({collections, defaultCollection}) => (dispatch, getState) => {
+  // TODO: Load controls from state to get potential url query parameters
+  const { tree, controls } = getState();
+  if (!tree.loaded) {
+    throw new Error("tree not loaded");
+  }
+
   // Get the collection to display to set up default controls
   // TODO: consider url query parameter?
-  const collectionToDisplay = getCollectionToDisplay(collections, json["default_collection"]);
+  const collectionToDisplay = getCollectionToDisplay(collections, defaultCollection);
 
   dispatch({
     type: LOAD_MEASUREMENTS,
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -4,7 +4,7 @@ import { getServerAddress } from "../util/globals";
 import { goTo404 } from "./navigation";
 import { createStateFromQueryOrJSONs, createTreeTooState, getNarrativePageFromQuery } from "./recomputeReduxState";
 import { loadFrequencies } from "./frequencies";
-import { loadMeasurements } from "./measurements";
+import { parseMeasurementsJSON, loadMeasurements } from "./measurements";
 import { fetchJSON, fetchWithErrorHandling } from "../util/serverInteraction";
 import { warningNotification, errorNotification } from "./notifications";
 import { parseMarkdownNarrativeFile } from "../util/parseNarrative";
@@ -298,8 +298,9 @@ Dataset.prototype.fetchSidecars = async function fetchSidecars() {
   }
   if (mainJson.meta.panels && mainJson.meta.panels.includes("measurements") && !this.measurements) {
     this.measurements = fetchJSON(this.apiCalls.measurements)
+      .then((json) => parseMeasurementsJSON(json))
       .catch((err) => {
-        console.error("Failed to fetch measurements collections", err.message);
+        console.error("Failed to fetch and parse measurements collections", err.message);
       });
   }
 };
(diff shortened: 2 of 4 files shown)


src/actions/measurements.js
```diff
@@ -102,7 +102,15 @@ export const loadMeasurements = (json) => (dispatch, getState) => {
     throw new Error("tree not loaded");
   }
 
-  const collections = json["collections"];
+  /**
```
jameshadfield, line 105: An alternative approach here would be to use a flag to indicate whether the (non-idempotent) transform has been done. This should save memory & be faster as moves back to a previously loaded measurements panel won't need to clone + retransform the data. Something like:

```diff
diff --git a/src/actions/measurements.js b/src/actions/measurements.js
index 6909c560..4740e2d7 100644
--- a/src/actions/measurements.js
+++ b/src/actions/measurements.js
@@ -107,6 +107,8 @@ export const loadMeasurements = (json) => (dispatch, getState) => {
     throw new Error("Measurements JSON does not have collections");
   }
 
+  if (json?.__collections_transformed !== true) {
+  // following forEach block would need to be indented. Could also be pulled out into a function - i.e. collections.forEach(transform).
   collections.forEach((collection) => {
     /**
      * Keep backwards compatibility with single value threshold.
@@ -192,6 +194,8 @@ export const loadMeasurements = (json) => (dispatch, getState) => {
       })
     );
   });
+  json.__collections_transformed = true;
+  }
 
   // Get the collection to display to set up default controls
   // TODO: consider url query parameter?
```

jameshadfield (changes requested):
Thanks Jover - nice debugging! I have one suggested change (see thread). I tested exclusively using your example from https://github.com/nextstrain/auspice/issues/1603#issue-1468851609 (plus some variation in URL queries of that narrative). Looking at John's video in that issue I think this PR would fix that too. 

For my own notes, if I need to revisit this (not a request for changes):
Slides which change datasets need to reset measurements state to the defaults (`getDefaultMeasurementsState()` - introduced in this PR) and then subsequently `LOAD_MEASUREMENTS` will fire and update state. Other sidecars behave similarly. Neither of these steps need to happen when changing between slides of the same dataset. 


P.S. Are the commits in the right order? Or maybe I wasn't testing the right thing -- I couldn't see any improvement in behavior after 6e7598accb5ce50fccac7949da718585c9a16822 but cherry-picking de4201365b93a4f29b0276729cbf96a6db2cb75e allowed me to move between datasets (but not back again!). No need for changes here, I'm just interested!

src/actions/measurements.js
jameshadfield: (Maybe use a [symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) instead of a string key)

joverlee521: Thank you for the thoughts on performance here!

I went down a slightly different route to combine the fetch and transformation so that the transformation only happens on the initial fetch of the JSON, see 338db0aa28f61252808aeda25d45d2ef42ecb1cc.

joverlee521:
> P.S. Are the commits in the right order? Or maybe I wasn't testing the right thing -- I couldn't see any improvement in behavior after https://github.com/nextstrain/auspice/commit/6e7598accb5ce50fccac7949da718585c9a16822 but cherry-picking https://github.com/nextstrain/auspice/commit/de4201365b93a4f29b0276729cbf96a6db2cb75e allowed me to move between datasets (but not back again!). No need for changes here, I'm just interested!

Huh, for me, the first commit fixes the complete crash that happens when I transition back to previous different datasets. The second commit fixes some seemingly random crashes that happen when transitioning between slides.

jameshadfield (approved):

src/actions/measurements.js
joverlee521: Hmm, just realized this method breaks the loading of the measurements sidecar JSON in auspice.us since it handles the dropped files instead of fetching from server.

joverlee521: Refactored in 69e1d9c9466897441139a6d23e512b34955907a3 to keep the parsing in a separate function so that it can be called in auspice.us `handleDroppedFiles` to parse the dropped measurements sidecar JSON.

joverlee521: See updates for auspice.us in https://github.com/nextstrain/auspice.us/pull/57

joverlee521:
Tested again after latest changes. 
- [x] downloaded the latest nextflu-private narrative and worked locally
- [x] drag-and-drop same private narrative onto auspice.us [test app](https://auspice-us-parse-measur-b6us7j.herokuapp.com/)
- [x] viewed [test narrative](https://nextstrain-s-nextstrain-uklbsq.herokuapp.com/community/narratives/joverlee521/nextstrain-testing/test-multiple-dataset-measurements) in nextstrain.org test app

src/actions/measurements.js
```diff
@@ -95,45 +95,38 @@ const getCollectionDisplayControls = (controls, collection) => {
...
     /**
      * Keep backwards compatibility with single value threshold.
-     * Make sure thresholds are an array of values so that we don't have to
-     * check the data type in the D3 drawing process
-     * `collection.thresholds` takes precedence over the deprecated `collection.threshold`
-     */
+      * Make sure thresholds are an array of values so that we don't have to
+      * check the data type in the D3 drawing process
+      * `collection.thresholds` takes precedence over the deprecated `collection.threshold`
+      */
```
tsibley, line 109: Unexpected indent change here (and in following comments as well).

tsibley (commented):
Put my 👀 on this while plowing thru PRs, so just mentioning I did so and that it makes sense to me. :+1: One minor note on an unexpected change.

src/actions/measurements.js
joverlee521: Ah, thanks for catching this!

PR #1664, opened 2023-04-20
Add helper script for downloading dataset+sidecars

jameshadfield:
Adds a small dev-use-only script to download the dataset + sidecar files from a given nextstrain.org URL and give them the correct filename so that you can test the same URL pathname from localhost auspice.

I've wanted this for ages but #1603 finally motivated me to do it.

### DIFF
--- a/scripts/get-dataset.sh
+++ b/scripts/get-dataset.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+# A simple, non-battle-hardened, script to help downloading a dataset
+# and corresponding sidecars for testing locally with auspice.
+#
+# Examples:
+# ./scripts/get-dataset.sh nextstrain.org/zika data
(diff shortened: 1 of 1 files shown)


joverlee521:
FWIW, you can do the same thing with the nextstrain CLI: 
```
nextstrain remote download nextstrain.org/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17 datasets
```
produces the same results as 
```
./scripts/get-dataset.sh nextstrain.org/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17 datasets
```

genehack:
Any reason not to merge this? Or is Jover's commentary sufficient evidence that what's actually needed is some documentation around how to use `nextstrain remote` to solve this problem?

PR #1665, opened 2023-04-22
Update eslint rules

victorlin:

See commit messages.

### Related issue(s)

- Motivated by a [Slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1682096157511009).
- https://github.com/nextstrain/auspice/pull/1450/commits/9b0c47f33159816c19358d7b54c735491ace2f15 would have to be updated for these changes.
- If this goes well, similar changes should be made in the nextstrain.org codebase.

### Testing

- [x] Checks pass
- [x] Added `console.log` statements within src/components/map.js's `Map.UNSAFE_*` methods, ran dev server locally, clicked around successfully, verified presence of console outputs to indicate that the methods were invoked.
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ I don't think this is necessary but feel free to create one if you'd like for reviewing.


.eslintrc.yaml
```diff
@@ -18,7 +18,6 @@ globals:
 rules:
   no-unused-vars: ["error", { "argsIgnorePattern": "^_", "destructuredArrayIgnorePattern": "^_" }]
   react/jsx-filename-extension: [1, { "extensions": [".js", ".jsx"] }]
-  react/no-direct-mutation-state: off
```
victorlin: Could someone help me with a90938b7d10672c59e7b548e050c3d38dff5c42a? Currently, this is the only thing that is causing the CI job to [fail](https://github.com/nextstrain/auspice/actions/runs/4770552082/jobs/8481885730?pr=1665):

```
/home/runner/work/auspice/auspice/src/components/map/map.js
Error:   584:5  error  Do not mutate state directly. Use setState()  react/no-direct-mutation-state
```

The [`react/no-direct-mutation-state` doc page](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-direct-mutation-state.md) says that changes to `this.state` should only happen in the constructor, however I'm unsure how to refactor the code to do that.

.eslintrc
```diff
@@ -103,4 +103,3 @@ env:
   es6: true
   browser: true
   jest: true
-# https://gist.github.com/odedw/cf14a55aa154adbc3bf7#file-eslintrc-yaml-L4
```
tsibley, line 106: :tada:

src/components/controls/color-by.js
```diff
@@ -69,7 +69,7 @@ class ColorBy extends React.Component {
   }
 
   // State from the outside world enters via props.
-  componentWillReceiveProps(nextProps) {
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
tsibley, line 72: Does this `UNSAFE_…` naming work on React 16?

.eslintrc.yaml
tsibley: That's complaining about:

https://github.com/nextstrain/auspice/blob/1ff15addb24d7672c2d3802ecfef5a588b521714/src/components/map/map.js#L584

But is `currentBounds` actually _used_ anywhere? I don't see where it is.

It was introduced in 8a213ac8976c34df1182a441336bc18cc48401e2 and the last usage aside from the assignment above was seemingly removed in 43ee2ed6220944f57b5cc2f1eecaaa639f0e208b.

In general though, I think you could almost mechanically replace `this.state.X = Y` with approximately `this.setState({X: Y})`, as long as it was ok for subsequent code in the same tick to not notice the state change.

tsibley (commented):
Put 👀 on this while plowing thru PRs. In general I'm very much +1 for this direction (even if we do add back some code-style linting as well if others desire). The resulting config and linewise ignores are so much simpler.

src/components/controls/color-by.js
jameshadfield: Should work [from 16.3 onwards](https://legacy.reactjs.org/blog/2018/03/27/update-on-async-rendering.html#gradual-migration-path)

src/util/colorScale.js
```diff
@@ -203,7 +203,8 @@ function createOrdinalScale(colorBy, t1nodes, t2nodes) {
       const spread = minMax[1] - minMax[0];
       const dp = spread > 5 ? 2 : 3;
       legendValues = genericDomain.map((d) => parseFloat((minMax[0] + d*spread).toFixed(dp)));
-      if (legendValues[0] === -0) legendValues[0] = 0; /* hack to avoid bugs */
+      // Hack to avoid a bug: https://github.com/nextstrain/auspice/issues/540
+      if (Object.is(legendValues[0], -0)) legendValues[0] = 0;
```
jameshadfield, line 207: Haven't actually tried this out, but interestingly the docs say

> The only difference between Object.is() and === is in their treatment of signed zeros and NaN values. The === operator ...  treats the number values -0 and +0 as equal, but treats [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) as not equal to each other.

So I don't see how the original hack worked in practice. Anyway ,as long as the 🐛  doesn't return this change is great.

tsibley: AIUC, the original `===` comparison meant that both +0 and -0 values got reset to +0, where with the `Object.is` comparison, only -0 values will get reset to +0. The outcome is the same, though (assuming no side effects in the property setter).

.eslintrc.yaml
victorlin: Nice digging, that's helpful context! I've replaced a90938b7d10672c59e7b548e050c3d38dff5c42a with 172917602a382bc05ed62c3b247ccba38211af99...c7d3362439ee833661e2ec99af5bcb8c1105f451.

src/components/controls/color-by.js
victorlin: Also, I assume that CI should be able to catch if it doesn't work, since Auspice pins React 16.8.6.

src/util/colorScale.js
victorlin: I assume the reason for the bug is that something else using `legendValues[0]` treated `-0` as different from `+0`, when they should be treated the same. So, as @tsibley points out, either `===` or `Object.is()` can be used to correct that behavior, but the former is discouraged by [no-compare-neg-zero](https://eslint.org/docs/latest/rules/no-compare-neg-zero).

victorlin:
@jameshadfield let me know if you have specific code-style rules that you've found beneficial. We can consider (re-)adding them in this PR, or they could come in a separate PR.

src/store/index.js
```diff
@@ -2,15 +2,13 @@ import { createStore, applyMiddleware, compose } from "redux";
...
 import rootReducer from "../reducers";
-import { loggingMiddleware } from "../middleware/logActions";
 import { keepScatterplotStateInSync } from "../middleware/scatterplot";
 
 const configureStore = (initialState) => {
   const middleware = [
     thunk,
     keepScatterplotStateInSync,
-    changeURLMiddleware,
-    // loggingMiddleware
```
joverlee521, line 13: `loggingMiddleware` is very helpful during development and debugging. I think we've kept it commented out here for "easy access". Should we make it just part of the dev environment with something like 
```js
(process.env.NODE_ENV !== 'production' && loggingMiddleware)
```

joverlee521 (approved):
I really like this direction! I only left a single suggestion, but it may be out of scope for this PR.

src/components/controls/color-by.js
tsibley: Thanks both!

I tried briefly to find the changelog/docs for the `UNSAFE_…` prefixing and came up empty, but didn't then expand to a general search that would have found the blog post.

@victorlin Ah, it wasn't clear to me that CI here would catch if these method names would actually get used, i.e. I wasn't sure if CI runs Auspice in a live browser environment and if tests would break if these were not called.

victorlin: I take back my comment – I just tried renaming the method to something random and none of the npm scripts run in CI produced any errors.

It looks like `eslint-plugin-react` [automatically detects](https://github.com/jsx-eslint/eslint-plugin-react/blob/4a92667e4b99221013e1d2aa61e97296895cecc1/README.md?plain=1#L43-L45) the installed version of React, however specifying `settings.react.version: "16.0"` doesn't produce any "unused method" messages.

This requires some manual testing.

victorlin:
The above force-push is the result of changing 82e266f7dfccfb1445e059304b619440d64e8d16 to 933d4b93ad8f874d313ad52646ffdba3cfdac200.

victorlin:
The above force-push is the result of:

1. Replacing 933d4b93ad8f874d313ad52646ffdba3cfdac200 with b5d2ec06 + 7878ac41 + cf09c3d4
2. Replacing 1d45bd326efad15ab859e7bbe21de13d55fa35d8 with b71817a7...95cc4659

src/store/index.js
jameshadfield: It's a bit too noisy for "normal" development work, but I'd be fine with it being under a different environment variable (e.g. `LOG_ALL_ACTIONS=1 auspice develop`). Not necessary for this PR tho.

victorlin: > please keep it there!

Dropped 1e1d441094951003f252f741cba5b876258b988f.

tsibley:
Latest rework LGTM by inspection.

src/components/controls/color-by.js
tsibley: I do note this warning from eslint-plugin-react when running linting:

    Warning: React version not specified in eslint-plugin-react settings. See https://github.com/yannickcr/eslint-plugin-react#configuration .

victorlin: Oh, I misread the snippet linked above. Pushed 4adfc9b275a38d19fe5f144477c4720a881ce9f0 to address the warning.

victorlin: Added manual testing notes for `UNSAFE_*` methods in the PR description.

tsibley (approved):

.eslintrc
```diff
@@ -1,106 +0,0 @@
...
-  no-useless-constructor: off
-  no-nested-ternary: off
-  object-curly-spacing: off # allow { a: 1 } and {a: 1}
-  block-spacing: off # same as above for function blocks on a single line
-  dot-notation: off # allows someObject["a"] as well as someObject.a
-  comma-dangle: ["error", "never"]
-  padded-blocks: off
-  no-plusplus: off # we are old enought to use i++
-  arrow-body-style: off
-  arrow-parens: ["error", "always"] # use brackets always, (a) => body // () => body // (a, b) => body
```
jameshadfield, line 28: I'd like this rule re-enabled please 🙏

.eslintrc
```diff
@@ -1,106 +0,0 @@
...
-  no-case-declarations: off # we don't surround case blocks with { } in our switch statements as we use fallthroughs deliberately
-  one-var: off
-  one-var-declaration-per-line: off
-  no-console: [1, { "allow": ["warn", "error"] }]
```
jameshadfield, line 32: Can we bring this rule back please -- it's too easy for a debugging console.log to slip through.

.eslintrc
```diff
@@ -1,106 +0,0 @@
...
-  react/sort-comp: off # possibly reinstate
-  jsx-a11y/no-static-element-interactions: off
-  jsx-a11y/anchor-has-content: off
-  jsx-a11y/heading-has-content: off
-  import/prefer-default-export: off
-  no-multi-spaces: ["error", { ignoreEOLComments: true }]
-  no-labels: off
-  no-unused-labels: off
-  no-continue: off
-  no-unneeded-ternary: ["error", { "defaultAssignment": true }]
```
jameshadfield, line 55: This rule is nice to have - it produces easier to read code.

.eslintrc
```diff
@@ -1,106 +0,0 @@
...
-  indent: ["error", 2, {"MemberExpression": "off", "SwitchCase": 1}]
-  react-hooks/rules-of-hooks: "error"
-  react-hooks/exhaustive-deps: "warn"
-  object-curly-newline: off
-  operator-linebreak: ["error", "after", { "overrides": { "?": "ignore", ":": "ignore" } }]
-  react/destructuring-assignment: off
-  prefer-destructuring: off
-  implicit-arrow-linebreak: off
-  jsx-quotes: off
-  no-use-before-define: ["error", { "functions": false }]
```
jameshadfield, line 67: This rule should be brought back. See [this comment](https://github.com/nextstrain/auspice/pull/1450#discussion_r1152650310) in the TS PR for context.

.eslintrc
```diff
@@ -1,106 +0,0 @@
...
-  no-unused-labels: off
-  no-continue: off
-  no-unneeded-ternary: ["error", { "defaultAssignment": true }]
-  quote-props: ["error", "as-needed"]
-  prefer-const: ["error", {"destructuring": "all"}]
-  indent: ["error", 2, {"MemberExpression": "off", "SwitchCase": 1}]
-  react-hooks/rules-of-hooks: "error"
-  react-hooks/exhaustive-deps: "warn"
-  object-curly-newline: off
-  operator-linebreak: ["error", "after", { "overrides": { "?": "ignore", ":": "ignore" } }]
```
jameshadfield, line 62: This is stylistic, but I really like it! I know others don't, which is why the overrides have been added in.

.eslintrc
```diff
@@ -1,106 +0,0 @@
...
-  jsx-a11y/anchor-has-content: off
-  jsx-a11y/heading-has-content: off
-  import/prefer-default-export: off
-  no-multi-spaces: ["error", { ignoreEOLComments: true }]
-  no-labels: off
-  no-unused-labels: off
-  no-continue: off
-  no-unneeded-ternary: ["error", { "defaultAssignment": true }]
-  quote-props: ["error", "as-needed"]
-  prefer-const: ["error", {"destructuring": "all"}]
```
jameshadfield, line 57: This should remain, I think. I know JavaScript's definition of _constant_ is slightly opaque, but I do find this rule useful.

.eslintrc
```diff
@@ -1,106 +0,0 @@
...
-  jsx-a11y/no-static-element-interactions: off
-  jsx-a11y/anchor-has-content: off
-  jsx-a11y/heading-has-content: off
-  import/prefer-default-export: off
-  no-multi-spaces: ["error", { ignoreEOLComments: true }]
-  no-labels: off
-  no-unused-labels: off
-  no-continue: off
-  no-unneeded-ternary: ["error", { "defaultAssignment": true }]
-  quote-props: ["error", "as-needed"]
```
jameshadfield, line 56: Purely stylistic, but nice to have (IMO).

jameshadfield (changes requested):
Thanks @victorlin for this! It's hard to track exactly what's changed, but given the ad-hoc nature of our previous config I'm not too worried about this (and one could argue it was hard to know what was set in the first place!). All the rules we had turned "off' (e.g. [these react rules](https://github.com/nextstrain/auspice/blob/master/.eslintrc#L84-L95), must have been enabled via the air-bnb preset (have verified a couple of these, but the air-bnb config is 1000s of lines of JS 😬). I spent most of my review looking at the rules we had previously enabled; most of them are pretty marginal so I don't mind them being removed by this PR, but there's a bunch I'd like to keep (and which aren't in the recommended presets). At one time or another each has been explicitly added to our config, so there is a rational for them (even if we perhaps didn't document it as well as we could have).

.eslintrc
jameshadfield: (Only really applicable for `./src/` files, so I realise that scripts will need to manually override this rule, which is unfortunate.)

src/components/framework/title.js
```diff
@@ -39,7 +39,7 @@ class Title extends React.Component {
     const title = "nextstrain";
     return title.split("").map((letter, i) => (
       <span
-        key={i} // eslint-disable-line react/no-array-index-key
```
jameshadfield, line 42: I'm surprised this rule isn't in [the recommended set](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/configs/recommended.js) of react rules. _Missing_ keys have caused a bunch of bugs in auspice before (recently a bug in the legend), and I thought key collisions were just as bad?

.eslintrc
victorlin: We can take advantage of [ESLint config cascading](https://eslint.org/docs/latest/use/configure/configuration-files#cascading-and-hierarchy) to enable this in `src/` only.

.eslintrc.yaml
```diff
@@ -0,0 +1,39 @@
+---
+extends:
+  - eslint:recommended
+  - plugin:react/recommended
```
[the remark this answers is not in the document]
victorlin: Added to b3cedb71 in [force-push](https://github.com/nextstrain/auspice/compare/4adfc9b275a38d19fe5f144477c4720a881ce9f0..902895a79e0cd6d7b390461c518b170034b99a3f).

.eslintrc.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+plugins:
```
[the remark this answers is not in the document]
victorlin: The behavior is addressed by this line in the docs (emphasis mine):

> A configuration file, once extended, can inherit all the traits of another configuration file (including rules, **plugins**, and language options)

Yes, I agree it's better to keep it if we are using rules from the `react` plugin directly.

.eslintrc
victorlin: Added in 71c07097.

.eslintrc
victorlin: Done in a9aab1d4 + cf2b5416.

.eslintrc
victorlin: Added in 71c07097 + eb0c5928.

.eslintrc
victorlin: Added in 71c07097 + 9fc42f01.

.eslintrc
victorlin: Added in 71c07097.

.eslintrc
victorlin: Added in 71c07097 + 93ee800c.

.eslintrc
victorlin: Added in 71c07097.

src/components/framework/title.js
victorlin: Pushed 09b5f3f4b1020deb104c2fe8257483187ab3ac3d to enable it.

src/util/googleAnalytics.js
victorlin, line 1: I added 5c9c8caf as a follow-up to cf2b5416.

victorlin (commented):
@jameshadfield thanks for the going through the removed custom rules and picking out what's useful – I've added everything that you requested. There is still a gap in checking the rules previously enabled by Airbnb config for anything that's now disabled but might be useful to add back (such as `react/no-array-index-key`), but as you mentioned, it's a lot to go through and not something I want to do at the moment.

jameshadfield (approved):
LGTM!

victorlin:
For future reference, here is where you can find the Airbnb base config that was extended prior to b3cedb712c1fdb38b4e6e759e2652376075ea0a6:

https://github.com/airbnb/javascript/tree/eslint-config-airbnb-v18.2.0/packages/eslint-config-airbnb-base/rules

PR #1672, opened 2023-06-13
Remove integration tests

jameshadfield:
See commit messages for details, especially 2047696d859602db920b938f9637174bdac813eb.

If you had git-lfs installed, you may uninstall it via `git lfs uninstall` if needed.

Beyond this PR, we should try to stop using GitHub's LFS, but [this seems difficult](https://docs.github.com/en/repositories/working-with-files/managing-large-files/removing-files-from-git-large-file-storage#git-lfs-objects-in-your-repository). At the least, this PR will mean we are within the free storage & bandwidth tier.


jameshadfield:
CI failing on node14 because the `??=` syntax is unsupported. Node 14 [reached "end-of-life" on 2023-04-30](https://github.com/nodejs/release#release-schedule). I'll tackle this in a separate PR and then circle back.

jameshadfield:
Moved commit 59b6ac76 ('Update puppeteer and jest-puppeteer') to #1674 as that requires us dropping support for node.js v14, which [this comment](https://github.com/nextstrain/auspice/pull/1673#discussion_r1228688263) indicates might not be possible for a while. This also decouples this PR from #1673.

victorlin (approved):

PR #1673, opened 2023-06-13
Remove node14, add node20

jameshadfield:
See commit messages for details


.github/workflows/ci.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [14, 16, 18]
```
victorlin, line 15: The Docker image still uses Node.js 14 for installing/using Auspice. There is a [PR to use Node.js 16](https://github.com/nextstrain/docker-base/pull/122) but it hasn't been pushed through since it's based on a bigger change.

victorlin: Yes, let's keep Node.js 14 support until this is done. It's mostly dependent on https://github.com/nextstrain/docker-base/pull/98, so I'll push on that a bit harder now and update here when it's done.

victorlin (approved):

.github/workflows/ci.yaml
victorlin, line 1: Can [bundlesize](https://github.com/nextstrain/auspice/blob/e9b19e68768a2060651c46e743df1c9c4074f6cc/.github/workflows/ci.yaml#LL85C1-L93C34) be updated to use 20 as well?

jameshadfield: Probably, but f we had to pick one node version I'd pick the one we use to publish auspice (v18)

victorlin: Oh, I didn't notice there are other jobs that also use v18. Yes, those should be updated together later.

PR #1684, opened 2023-08-04
Entropy panel mk2

### LINKED ISSUE #299: entropy gene annotation

### LINKED ISSUE #1145: consolidate Diversity panel genome zoom state
**Current Behavior**  
In the redux state, we track the current zoom state of the diversity panel, which for example looks like this in the query `?gmax=4757&gmin=2379` and like this in the UI: 
Screen Shot 2020-05-28 at 5 23 39 PM

To track this, we have both 
* `state.controls.zoomMin, state.controls.zoomMax`
* `state.entropy.zoomMin, state.entropy.zoomMax`. 

@jameshadfield and I agree that you should only need one of these two, since they do seem to be redundant (only the `state.controls` version is used, though perhaps it should be the other way around). James drew up this nice diagram to understand the current behavoir:

Untitled-2020-05-29-1109

**Expected behavior**  
Just use one place in the redux state to track the zoom status of the Diversity panel.

**How to reproduce**  
Technically this works right now, hence the low priority and lack of bug tag. If we wanted to read this part of the state in some new code, however, we might easily write something that doesn't work by using the wrong part of the state.

**Possible solution**  
Replace use of `state.controls.zoomMin` and `state.controls.zoomMax` with the same fields from `state.entropy`, essentially collapsing the two boxes on the right of the diagram into one.

**Additional context**  
`absoluteZoomMin` and `absoluteZoomMax` are also [set here on `state.controls`](https://github.com/nextstrain/auspice/blob/master/src/actions/recomputeReduxState.js#L688-L689) and used in middleware to unset the url query when the zoom goes back to the ends.

Also important to also note from our conversation:
> It’s also complicated how you actually perform zooming in the d3 component. For instance, when you drag the brushes to zoom do you:
(a) dispatch an action, change redux state, then on a react lifecycle see that it’s changed and perform the imperative d3  command to zoom into the (new) value
(b) as you zoom change the d3 object as well as dispatching redux actions (or debounced dispatches, or only a single dispatch at the end). Then, when the react lifecycle updates (as redux state has changed) do you have a check to see that you don’t actually need to zoom in as it’d already been done in this case?

We do (b)
> which is why when dragging the brushes you only see the URL change when you release the brush

This should be able to stay that way as it seems like it should give better performance.

### LINKED ISSUE #1516: ENH: Show gene names and maybe more info in tooltip on hover in entropy panel
The mpox genemap is very dense. It's hard to read gene names in the genemap part of the entropy panel.

The white gene names also seem to overflow into neighbouring names, causing some undesirable long-range interactions.

One solution could be to have a tooltip that shows gene information. This would be a great place to include extra gene information, like protein function etc, if we wanted to include it - I think this could be useful to help understand what's going on in a virus with a lot of genes.

image

image

### LINKED ISSUE #1596: Entropy click triggers two `NEW_COLOR` actions
### Current Behavior
Clicking on an entropy bar will trigger two `NEW_COLOR` Redux actions. 
Although this is an internal issue, it can lead to unexpected side effects such as the bug fixed in https://github.com/nextstrain/auspice/pull/1593. 

According to @jameshadfield:
> The two NEW_COLOR actions are coming from the entropy onClick (as expected) and the sidebar color-by drop-down (unexpected).

Digging into this a little more:
1. Entropy onClick triggers a `NEW_COLOR` action that updates the Redux state `controls.colorBy`.
2. This Redux state update triggers `componentWillReceiveProps` of the `ColorBy` component, which sets the local React state to the new `colorBy` state. This ensures the displayed color-by in the dropdown reflects the current color by. 
3. The new state triggers `componentDidUpdate` of the `ColorBy` component, which triggers a `NEW_COLOR` action. This is necessary because manually changing the color-by via the dropdown menu updates local React state to trigger the Redux action to update other components.

I think what needs to happen is for the `ColorBy` component to only trigger the `NEW_COLOR` Redux action only when the local React state is changed via the dropdown menu, but _not_ when the state is updated by the props of the Redux state.

### LINKED ISSUE #1651: Diversity panel extends beyond genome end on 3' end
### Current Behavior
The diversity panel x-axis extends beyond the 3' end of the genome.

e.g. for SARS-CoV-2 it goes roughly 150 bases beyond the 3' end at 29903 (Wuhan-Hu-1)
image

This is asymmetric with the 5' end, where it starts at 0 and doesn't go negative:
image

### Expected behavior
The position axis does not extend beyond the genome end (at least not more than a few bp)

The fact that the axis extends creates the user impression that the real genome is longer than it actually is, and that there is a lack of mutations - when in fact there can't be any mutations in this non-existent region.

See in our ncov builds: https://next.nextstrain.org/ncov/gisaid/global/1m?gmin=29677&l=unrooted&m=div

jameshadfield:
For the full story please read through the commit messages. I'll try to highlight the main UI changes here, but perhaps it's better to just jump in and [use the review app](https://auspice-entropy-updates-rgiu2h.herokuapp.com/) with fresh eyes.

### Overlapping CDSs are better shown

<details>
  <summary>Click for details</summary>

By using opacity and jittering them based on the frame the CDS is in. The aim is not to have a perfect genome viewer (e.g. see [how much space a "proper" genome browser](https://www.ncbi.nlm.nih.gov/gene/43740578) needs!) but to expose what's there and allow you to hover over them / click them (clicking a CDS selects it, see next section).

image

_nCoV, current Auspice on top, [this PR on bottom](https://auspice-entropy-updates-rgiu2h.herokuapp.com/entropy-test-data/ncov?gmin=26141); much better communication of the CDSs present (they were always there in the JSON, just visually obscured)._

Hovering over the CDSs is now possible too:

image

image

</details>

### Selecting a CDS focuses the entropy panel on the translated CDS

<details>
  <summary>Click for details</summary>

When viewing a CDS we now have the main axis represent this and only this CDS, always shown going L→R, with tick labels representing amino acids. This better represents biology as we are able to ignore any artifacts upstream of a translation. The UI to view a CDS is to either click on a CDS or select a genotype color-by for that CDS. 

This screenshot ([top](https://next.nextstrain.org/staging/entropy-test-data/hepB?gmax=2327&gmin=1810), [bottom](https://auspice-entropy-updates-rgiu2h.herokuapp.com/entropy-test-data/hepB)) shows the pre-capsid CDS for HepB. In the top view other CDSs are overlapping so most of the bars you see are from a different CDS (pre-capsid CDS is the orange coloured CDS, and then the green CDS is overlapping), and _it was not possible_ to see the actual diversity of pre-capsid¹. 

image

¹ _I guess in this example the CDSs are in frame, so the diversity is identical, but barcharts from different CDSs obscuring each other has always been a problem in auspice_

</details>

### Segmented CDSs (nCoV example)

<details>
  <summary>Click for details</summary>

Top view is our current genome annotation for nCoV, [bottom](https://auspice-entropy-updates-rgiu2h.herokuapp.com/entropy-test-data/ncov) is an example of what it could be (incomplete, as I've only added in 2 of the 16 proteins cleaved from the polyproteins)

image

Specifically, we can now view ORF1ab (PP1ab) in its entirety:

image

(The main axis now has two segments, which is what we have been referring to as ORF1a and ORF1b, separated by a -1 ribosomal frameshift.)

And since overlapping CDSs are better shown we can (finally!) show all the proteins which are produced, e.g. RdRp (a complex one, since it spans the -1 frameshift). The entropy panel conveys the expected lack of diversity here well I think. 

image

</details>

### CDSs which wrap around the origin (HepB example)

<details>
  <summary>Click for details</summary>

Most of HepB's CDSs wrap the origin! Displaying the translation (i.e. the top half of the panel) is no different from any segmented gene, it's the zooming that's tricky here. My solution was to ~invert the brush; perhaps easier to [explore](https://auspice-entropy-updates-rgiu2h.herokuapp.com/entropy-test-data/hepB?c=gt-pol_457) than to explain!

image

Segmented CDSs may have codons which span multiple segments (referred to as the phase of the segment), and we handle this well now:

image

</details>

### Entropy panel state is decoupled from the color-by

<details>
  <summary>Click for details</summary>

Previously the entropy panel was tightly coupled to the color-by. It is still linked to it, but it can also operate independently if (and only if) you start interacting with it. Any change to the (genotype) color-by will bring it back in sync. For instance, I can now select Spike 452 color-by (which will always jump to that position in the entropy panel + select Spike), and then use the UI to select another CDS in the entropy panel without changing the colorBy. If (via the sidebar) I change to a different Spike position then the entropy panel will jump back to show Spike. 

</details>

### Proposed JSON schema

The current proposal is mostly backwards compatible, except segmented CDSs are ignored by previous Auspice versions. For instance here's how ORF1ab looks:

```json
  "ORF1ab": {
    "gene": "Multiple CDSs with the same gene will be grouped together",
    "strand": "+",
    "segments":[
      {"start": 266, "end": 13468},
      {"start": 13468, "end": 21555}
    ],
    "display_name": "Text here is shown when hovering on the CDSs"
  },
```

And you can specify a CDS which wraps around the origin by using an end value beyond the end value of the genome (or create the segments yourself, either way works):

```json
"nuc": {
  "start": 1, "end": 3182, "strand": "+"
},
"envL": {
  "start": 2850, "end": 4019, "strand": "+", "gene": "S"
},
```

### Test datasets

Available via the [review app](https://auspice-entropy-updates-rgiu2h.herokuapp.com/):
* Moneypox (no changes from staging site version) - useful for a genome with lots of CDSs
* Zika (no changes from staging site version) - a very simple genome
* HepB - a short genome but complex: overlapping CDSs, many CDSs which wrap the origin
* ncov - A modified nCoV dataset which removes our ORF1a / 1b artifact and shows the ORF1ab polyprotein (2 segments, one with a -1 frameshift), the PP1a polyprotein (aka ORF1a) and 2 of the 16 proteins cleaved out of those polyproteins. Created using a small script to modify a nextclade nCoV tree so it shouldn't be taken as 100% perfect!

## Things still to do (either here or in subsequent work):
**styling**

- [x] The y-jitter (currently using the CDS's frame) still results in hard-to-click / hard-to-see CDSs. I'm not sure of the ideal solution here, but we could probably allow a little more height in overall entropy panel (perhaps calculated at load time depending on how complicated the genome map is). Please come up with good suggestions!!!
- [x] I must better communicate which CDS is selected - currently I change the text/stroke colour, but it's not clear enough.
- [x] CDS text overlapping
- [x] "Simple" genomes (e.g. zika) have too much whitespace because all CDSs are in a single frame
- [x] The zoom elements are a little bit too small - originally the grey brush element extended above the annotations (but behind them), but this obscured the fact that you could pan the zoom by dragging the grey part. Arguably it's so small now it's still obscured. 

**Negative strand CDSs**
- [x] All that's required is writing the code to handle the projection from genome space to protein space. And a whole lot of testing.

**misc**

- [x] I've done a lot here to remove duplicate / conflicting redux state but there is one which still exists: `state.metadata.genomeAnnotations` (used by the tree info boxes to order the genes). 
- [x] Sometimes the entropy zoom URL params aren't quite right (e.g. missing when they should be there or vice versa)
- [x] Button to expand zoom (where possible)
- [x] Remove the wireframes / console.log statements
- [x] Help popup (similar to the ones we have in the sidebar)
- [x] Failing tests
- [ ] Changelog
- [ ] Add documentation / tutorial? (I won't let this hold up the PR, but it is important)

**future work?**

- [x] For nucleotide positions we used to check if it fell inside a gene, which allowed on-hover info boxes to show information such as "Nucleotide x (Codon y in protein z)". This never considered the case where a nucleotide fell in multiple genes (CDSs), so it was somewhat misleading. It was a generally useful idea, so we should bring it back in an improved form. It's plausible that we could find the AA positions for all CDSs which overlapped that nucleotide.
- [ ] Main axis uses ticks which go above and below the axis, with an additional label to communicate the corresponding nt position in the genome (information which exists in the on-hover box)
- [ ] Mousewheel zooming has been disabled. I should bring it back, but it needs a proper debounce. 

### Issues closed

Closes #1596
Closes #1651
Closes #1145
Closes #299
Closes #1516


jameshadfield:
For the full story please read through the commit messages. I'll try to highlight the main UI changes here.

## Main UI changes

### Selecting a CDS focuses the entropy panel on the translated CDS

<details>
  <summary>Click for details</summary>


When viewing a CDS we now have the main axis represent this and only this CDS, always shown going L→R, with tick labels representing amino acids. This better represents biology as we are able to ignore any artifacts upstream of a translation. The UI to view a CDS is to either click on a CDS or select a genotype color-by for that CDS. 

This screenshot ([top](https://next.nextstrain.org/staging/entropy-test-data/hepB?gmax=2327&gmin=1810), [bottom](https://auspice-entropy-updates-rgiu2h.herokuapp.com/entropy-test-data/hepB)) shows the pre-capsid CDS for HepB. In the top view other CDSs are overlapping so most of the bars you see are from a different CDS (pre-capsid CDS is the orange coloured CDS, and then the green CDS is overlapping). We now focus on one CDS at a time, and display things in aa-coordinates.

image

</details>


### Segmented CDSs (nCoV example)

<details>
  <summary>Click for details</summary>

Top view is our current genome annotation for nCoV, [bottom](https://auspice-entropy-updates-rgiu2h.herokuapp.com/entropy-test-data/ncov) is an example of what it could be (incomplete, as I've only added in 2 of the 16 proteins cleaved from the polyproteins)

image

Specifically, we can now view ORF1ab (PP1ab) in its entirety:

image

(The main axis now has two segments, which is what we have been referring to as ORF1a and ORF1b, separated by a -1 ribosomal frameshift ~some of the time, and sometimes only ORF1a is translated which is shown here as well as PP1a.)

And since overlapping CDSs are better shown we can (finally!) show all the proteins which are produced, e.g. RdRp (a complex one, since it spans the -1 frameshift). The entropy panel conveys the expected lack of diversity here well I think. 

image

</details>


### CDSs which wrap around the origin (HepB example)

<details>
  <summary>Click for details</summary>

Most of HepB's CDSs wrap the origin! The zooming here is ~inverted, perhaps easier to [explore](https://auspice-entropy-updates-rgiu2h.herokuapp.com/entropy-test-data/hepB?c=gt-pol_457) than to explain!

image

Segmented CDSs may have codons which span multiple segments (referred to as the phase of the segment), and we handle this well now:

image

</details>

### Better tooltips

<details>
  <summary>Click for details</summary>

image


image


</details>

### Entropy panel state is decoupled from the color-by

_TODO_


### Proposed JSON schema

The current proposal is mostly backwards compatible, except segmented CDSs are ignored by previous Auspice versions. For instance here's how ORF1ab looks:

```json
  "ORF1ab": {
    "gene": "Multiple CDSs with the same gene will be grouped together",
    "strand": "+",
    "segments":[
      {"start": 266, "end": 13468},
      {"start": 13468, "end": 21555}
    ],
    "display_name": "Text here is shown when hovering on the CDSs"
  },
```

And you can specify a CDS which wraps around the origin by using an end value beyond the end value of the genome (or create the segments yourself, either way works):

```json
"nuc": {
  "start": 1, "end": 3182, "strand": "+"
},
"envL": {
  "start": 2850, "end": 4019, "strand": "+", "gene": "S"
},
```

### Test datasets

Available via the [review app](https://auspice-entropy-updates-rgiu2h.herokuapp.com/):
* Moneypox (no changes from staging site version) - useful for a genome with lots of CDSs
* Zika (no changes from staging site version) - a very simple genome
* HepB - a short genome but complex: overlapping CDSs, many CDSs which wrap the origin
* ncov - A modified nCoV dataset to show the ORF1ab polyprotein (2 segments with a -1 frameshift), the PP1a polyprotein, some of the 16 proteins cleaved out of those polyproteins etc. Created using a small script to modify a nextclade nCoV tree so it shouldn't be taken as 100% perfect!

## Things still to do (either here or in subsequent work):
**styling**

- [ ] The y-jitter (currently using the CDS's frame) still results in hard-to-click / hard-to-see CDSs. I'm not sure of the ideal solution here, but we could probably allow a little more height in overall entropy panel (perhaps calculated at load time depending on how complicated the genome map is). For an example of how "proper" genome browsers do it, see https://www.ncbi.nlm.nih.gov/gene/43740578. Please come up with good suggestions!!!

- [ ] "Simple" genomes (e.g. zika) have too much whitespace because all CDSs are in a single frame

- [ ] The zoom elements are a little bit too small

**Negative strand CDSs**

- [ ] !

**misc**

- [ ] I've done a lot here to remove duplicate / conflicting redux state but there is one which still exists: `state.metadata.genomeAnnotations` (used by the tree info boxes to order the genes). 

- [ ] Sometimes the entropy zoom URL params aren't quite right (e.g. missing when they should be there or vice versa)

- [ ] Button to expand zoom (where possible)

- [ ] Remove the wireframes / console.log statements

- [ ] Help popup (similar to the ones we have in the sidebar)

- [ ] Failing tests

- [ ] Changelog

**future work?**

- [ ] For nucleotide positions we used to check if it fell inside a gene, which allowed on-hover info boxes to show information such as "Nucleotide x (Codon y in protein z)". This never considered the case where a nucleotide fell in multiple genes (CDSs), so it was somewhat misleading. It was a generally useful idea, so we should bring it back in an improved form. It's plausible that we could find the AA positions for all CDSs which overlapped that nucleotide.

- [ ] Main axis uses ticks which go above and below the axis, with an additional label to communicate the corresponding nt position in the genome (information which exists in the on-hover box)



### Issues closed

Closes #1596
Closes #1651
Closes #1145
Closes #299

rneher:
this is very cool! Looking forward to finally having this in auspice!

A few points, some of which you touch upon above. 

 - When using the MPXV build, the zoom elements for the CDS get very small. We might need some way to zoom into the genome first. Or maybe put the zoom elements for the CDS on the CDS and keep the genome zoom below? 
 - as you say, negative strand CDS are not properly accounted for. The majority of MPXV genes in the first third of the genome are on the negative strand. Auspice used to know about that. 
 - we could add `localAnnotation` to the CDS along side the ranges to specify things that should only be shown in the CDS view, like RBD.

jameshadfield:
Thanks @rneher 

-ve strand CDSs are now working, except that zooming within a CDS is backwards (will fix tomorrow). I've abandoned the use of frame for the y-offsets and instead just stack them as efficiently as possible to avoid any CDSs overlapping (in the bottom view, anyways, the top view used too much space that way so they overlap a little bit). 

image

> we could add localAnnotation to the CDS along side the ranges to specify things that should only be shown in the CDS view, like RBD.

Yeah! [I played around with this yesterday](https://github.com/nextstrain/auspice/commit/f102a119b3d59787ea359f136a77060796a50e17), and I think it'll be very helpful. Here's how it looked for Spike (clicking on the local annotation toggled the highlighting). I'm going to leave it out of this PR, or at least get this PR finished before adding it in.

image

trvrb:
This is such cool work. Thanks for putting this together @jameshadfield 🌟. Some of this we talked through, but here are my UI impressions / suggestions from working with the review app:

1. If I've zoomed to a particular CDS. I still get the left/right arrow ↔ mouse cursor when hovering over the edges in the bottom track. However, I'm not able to drag this bottom track past the gene edges until the "↵ nuc" table is clicked on.

This behavior of only being able to zoom into the selected CDS makes total sense. However, it would be nice to have a bit more indication that this is what's going on. Could it be a unidirectional mouse cursor when at the end of the rope? 

2. For consistency and as well as for immediate understanding, I'd think replacing "↵ nuc" with "RESET ZOOM" (or "RESET LAYOUT") would be more appropriate. It would be grayed out like "RESET LAYOUT" is in the Phylogeny panel when the zoom is the entire genome. This would reset to full genome in nucleotide view if handles have been dragged and would also reset to full nucleotide genome if currently focused to a CDS.

I do realize this doesn't catch the case where someone has clicked on a CDS to focus it and then zoomed within the CDS. I think having a single universal reset will be friendlier in general. And having too separate buttons will create confusion.

3. The two labels for ORF1ab at https://auspice-entropy-updates-rgiu2h.herokuapp.com/entropy-test-data/ncov are a bit confusing to me. I think these should just be "ORF1a" and "ORF1b". I'm trying to align people's expectations and what we've done in the past with proper biology. I'm used to diagrams like this https://media.springernature.com/full/springer-static/image/art%3A10.1038%2Fs41392-021-00653-w/MediaObjects/41392_2021_653_Fig1_HTML.png or this https://media.springernature.com/lw685/springer-static/image/art%3A10.1007%2Fs00284-021-02396-x/MediaObjects/284_2021_2396_Fig3_HTML.png.

4. Giving handle triangles a white center and gray outline could help to indicate that they are draggable. Just feels a bit more tactile, this is what we do for the date slider.

5. How difficult would it be to draw faint lines from the gray bounding box on the bottom track to the left and right edges of the top track? This could help orient people to the relationship between the two. Figure 2 from https://bedford.io/papers/kinganda-lusamaki-ebola-drc/ did this nicely.

6. Is there a reason for why CDSs are strictly rectangles in the lower track, while they have the helpful arrows on the top track?

Fine to consider things like 4-6 out of scope for the PR.

cassiawag:
This is so cool, James! Would it be possible to add an `amino acids` or `nucleotides` label for the plotted units.

jameshadfield:
Thanks for all the reviews. I think this PR is at a good point to wrap up and release which I plan to do ~early next week (and we can keep improving this after merge if we notice things). 

> Would it be possible to add an AA or NT label for the plotted units?

Good idea! I've added a `AA pos` label when a CDS is selected, but having "NT" felt a bit too cramped. I've also made the panel title indicate whether we're viewing the genome or a CDS, and the little help icon also explains the coordinates. 

>  Could it be a unidirectional mouse cursor when at the end of the rope? ... Giving handle triangles a white center and gray outline could help to indicate that they are draggable. 

Yup! Both done.

> The two labels for ORF1ab ... are a bit confusing to me. I think these should just be "ORF1a" and "ORF1b"

(Discussed elsewhere) the best way forward is to continue to annotate these as separate CDSs, rather than a single segmented CDS which I demoed here. This allows people to continue to label mutations in an ORF1b coordinate system which has become quite common.

> How difficult would it be to draw faint lines from the gray bounding box on the bottom track to the left and right edges of the top track?

> Is there a reason for why CDSs are strictly rectangles in the lower track, while they have the helpful arrows on the top track?

Both possible -- I tried the arrows approach on the lower track and it looked quite messy (lots of aliasing), but I'm sure if we played around with it enough it'd work. I haven't tried the faint lines but think it's worth pursuing. The zooming approach will need further work to support large (>200kb) genomes well. 

> I'd think replacing "↵ nuc" with "RESET ZOOM" (or "RESET LAYOUT") would be more appropriate.

Done. It's a bridge too far to make this greyed out at the present moment, but we can tackle that aspect later on.

PR #1688, opened 2023-08-29
Allow inlined root sequence

### LINKED ISSUE #1682: Inline root sequence
_This is written as a note to myself, but if anyone would like to implement this please do!_

Allow the root-sequence JSON contents to be inlined under a top level key `"root_sequence"` (key name can change in review). This should be trivial to parse within `createMetadataStateFromJSON`. Rather than using this information to skip fetching the sidecar, @tsibley recommends "firing off the two fetches simultaneously but then only using the root-sequence fetch result if it's not inlined in the main dataset". Remember to test the implementation using narrative files with multiple datasets.

jameshadfield:
See commits for further details (Let me know if anyone would like datasets for testing).

[Corresponding Augur PR](https://github.com/nextstrain/augur/pull/1295)

Closes #1682

### DIFF
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -299,7 +299,7 @@ Dataset.prototype.fetchSidecars = async function fetchSidecars() {
       .catch((reason) => Promise.resolve(reason))
   }
 
-  if (!this.rootSequence) {
+  if (!mainJson.root_sequence && !this.rootSequence) {
     // Note that the browser may log a GET error if the above 404s
     this.rootSequence = fetchJSON(this.apiCalls.rootSequence)
       .catch((reason) => Promise.resolve(reason))
--- a/src/actions/loadData.js
+++ b/src/actions/loadData.js
@@ -8,7 +8,7 @@ import { parseMeasurementsJSON, loadMeasurements } from "./measurements";
 import { fetchJSON, fetchWithErrorHandling } from "../util/serverInteraction";
 import { warningNotification, errorNotification } from "./notifications";
 import { parseMarkdownNarrativeFile } from "../util/parseNarrative";
-import { NoContentError } from "../util/exceptions";
+import { NoContentError, FetchError} from "../util/exceptions";
 import { parseMarkdown } from "../util/parseMarkdown";
 import { updateColorByWithRootSequenceData } from "../actions/colors";
 import { explodeTree } from "./tree";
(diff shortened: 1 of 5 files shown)


src/actions/loadData.js
```diff
@@ -283,47 +283,78 @@ Dataset.prototype.fetchMain = function fetchMain() {
     .then((res) => res.json());
 };
 Dataset.prototype.fetchSidecars = async function fetchSidecars() {
+  /**
+   * If deemed appropriate, fetch sidecars and store the resulting promise as `this.<sidecarName>`.
+   * The returned promise is used by `loadSidecars` which will append `.then` clauses to the promise;
+   * often the load function is called immediately (and so the promises are unresolved) but for
+   * narratives it may be called some time later. Because of this the promises should not reject -
+   * if there is a fetch error then we resolve to the error and it is the responsibility of
+   * `loadSidecars` to react appropriately.
```
tsibley, line 292: _non-blocking_

It's not clear to me from anything in this commit why the rejected promises have to be converted to promises that resolve to an `Error`. Why can't the rejection be chained thru? It's possible to chain a `.catch()` on a rejected promise well after it's already been fulfilled.

Hmm… Is the sole purpose here to avoid the browser's default behaviour of emitting a console error message for an "unhandled rejection" (incorrectly/prematurely sometimes, though it doesn't know that). That default [rejection event behaviour](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#promise_rejection_events) is changeable.

tsibley (commented):
Looked at the code, looks fine, didn't try the functionality. One question about promise handling just so I understand better, but its non-blocking. I also noticed that the commits omit any reason _why_ this inlining is now being supported/allowed. That would be nice for posterity.

src/actions/loadData.js
jameshadfield: My understanding is that you don't have control over unhandled rejections on a per-promise level. In general, I _do_ want unhandled rejections to result in error logs as most of the time they are coding errors, but not in this case. I'll update the message to indicate this is why the promises should not reject.

tsibley: Yeah, not easily on a per-promise level; you do always have to interact in some way with the global events/handlers.

Definitely want to surface unhandled rejections somehow for the reason that they're a programming error, yes, though I think the default global events/handlers are a very minimal, not very good example of that surfacing. As a example of an alternative, I'd consider the approach I used here as an improvement:

https://github.com/nextstrain/nextstrain.org/blob/01724a5d3669a936c1322b79b04289c26901952b/src/utils/scripts.js#L40-L76

This example is from Node.js and uses process exit as the trigger for reporting as-of-yet-unhandled rejections. In a browser context, you could use other triggers (likely more frequent), such as "finished loading" events specific to the app (e.g. a periodic barrier point by which all promises "should" be handled) or even something as simple as a regular interval (e.g. of a ~10s or something) so that promises have some space to get dealt.

jameshadfield: In this case these promises should be handled when Auspice renders the dataset they represent, which may never happen (e.g. if a narrative never renders the slide backed by this dataset). I think in every other case we attach `catch` clauses (or use try/catch + async) when the promise is set up, so the default behavior seems fine. At least, I can't remember seeing an "unhandled promise rejection" log which was not indicative of a bug (in the code modified here we were implicitly resolving to `undefined`, which had its own problems).

tsibley (approved):
Tested this out a little and it seems to work (as expected). New commit message and comments are appreciated!

corneliusroemer:
I'm copying this from the slack convo as it gives good context as to why the inlined root sequence feature is important:
> Even many of our builds lack the root sequence, resulting in uninformative colo(u)rings (we should fix this):
https://nextstrain.org/flu/seasonal/h3n2/ha/2y?c=gt-HA1_11
The two step procedure of having to explicitly ask for the file in export and then putting it in the right place (coupled to the fact that you only need the root sequence for sites without variation) seems to result in a lot of missing root-sequences.
Conceptually, the mutations in the tree.json are in reference to the root sequence and the root sequence is in some sense tightly linked to this info in the tree.
Also [auspice.us](http://auspice.us/) would be easier to use (just drag one file).

In all my augur usage, the root seuqence is the only side car I come across and it's always a bit of a headache to handle two files rather than one. Inlining will be the default for all of my usage moving forward. Thanks!

Issue #1690, opened 2023-08-30
v2.47.0 crashes when changing color for dataset that does not have entropy panel


kvargha:
### Current Behavior
A clear and concise description of what is happening / what the bug is.
When I change the color by selection on the UI it causes the page to turn white and logs these errors:
![image](https://github.com/nextstrain/auspice/assets/35252220/ee585524-70c7-4dee-86dc-9a2ef553e69b)

### Expected behavior
It should highlight the samples by the attribute selected.

### How to reproduce
Steps to reproduce the current behavior:
1. Visit https://nextstrain.org/fetch/storage.googleapis.com/ucsc-gi-cdph-bigtree/CDPH_08_28_2021/CDPH-subtree-0.json
2. Select any field in the color by selector
3. The screen should turn white

### Your environment: if browsing Nextstrain online
 - Operating system: macOS Ventura 13.4.1
 - Browser: Chrome Version 116.0.5845.110

### Additional context
This wasn't happening last week.

joverlee521:
Hi @kvargha,

Thank you for reporting this issue! This is most likely due to the latest release of Auspice ([2.47.0](https://github.com/nextstrain/auspice/releases/tag/v2.47.0)) which we just promoted to use on the production nextstrain.org site this Monday. 

I've rolled back the promotion so the issue is fixed on the production site, but the behavior is still visible at https://next.nextstrain.org/fetch/storage.googleapis.com/ucsc-gi-cdph-bigtree/CDPH_08_28_2021/CDPH-subtree-0.json

kvargha:
The issue is gone. Thank you @joverlee521!

joverlee521:
Keeping this issue open since we still need to fix the underlying bug within Auspice.

PR #1691, opened 2023-08-30
Replace `equalArrays` with `lodash.isEqual`

joverlee521:

The custom `equalArrays` function can raise an error when one of the provided arguments does not have a `length` property. Instead of updating the function, just replace with `lodash.isEqual` which DTRT.
    
### Related issue(s)

Fixes https://github.com/nextstrain/auspice/issues/1690

### Testing

- [x] Checks pass
- [x] Fixes the crashing issue for the fetch dataset in the [nextstrain.org test app](https://nextstrain-s-nextstrain-i1pajb.herokuapp.com/fetch/storage.googleapis.com/ucsc-gi-cdph-bigtree/CDPH_08_28_2021/CDPH-subtree-0.json)


jameshadfield (approved):
Changing to lodash's `isEqual` command does fix this, however the underlying bug is that `changeEntropyCdsSelection` should be a no-op if the entropy data isn't present, i.e.

jameshadfield (changes requested):
Changing to lodash's `isEqual` command does fix this, however it also allows non-array arguments which is not the intention of the code. The underlying bug is that `changeEntropyCdsSelection` should be a no-op if the entropy data isn't present, i.e.


```diff
--- a/src/actions/entropy.js
+++ b/src/actions/entropy.js
@@ -55,6 +55,7 @@ export const updateEntropyVisibility = debounce((dispatch, getState) => {
 export const changeEntropyCdsSelection = (arg) => (dispatch, getState) => {
   const action = {type: types.CHANGE_ENTROPY_CDS_SELECTION}
   const entropy = getState().entropy;
+  if (!entropy.loaded) return;
```

This is a good example of where TypeScript would have helped ensure arguments passed to `equalArrays` were indeed arrays, which was the intention of the code. We don't want cases where we pass two non-array-but-equal arguments to `isEqual`.

We should also improve `equalArrays` to catch cases where arguments are not arrays:

```
export const equalArrays = (a,b) => Array.isArray(a) && Array.isArray(b) && a.length===b.length && a.every((el, idx) => b[idx]===el);
```

I'll do a little more digging to see if there are other situations where we assume the presence of entropy data.

joverlee521:
> While the intention of the code was to never call equalArrays with non-array arguments, we should also improve equalArrays to catch cases where we didn't do the right thing

```
export const equalArrays = (a,b) => Array.isArray(a) && Array.isArray(b) && a.length===b.length && a.every((el, idx) => b[idx]===el);
```

Hmm, this means it will return False if either arg is not an array and when the arrays are not equal which can make falsy checks problematic 
https://github.com/nextstrain/auspice/blob/cdf9f7c10508186d34a21d0c2341d9ad27d2f95a/src/components/entropy/index.js#L231-L232

I think it would be safer to do type checks separately than equality checks.

jameshadfield (approved):
> Hmm, this means it will return False if either arg is not an array and when the arrays are not equal which can make falsy checks problematic

Yes, that's fair, although `isEqual` will do the same. I think typescript is the way to go here, but that's well out of scope for this fix.

PR #1668, opened 2023-06-08
Extend usefulness of tip labels

jameshadfield:
The motivating use case is the proliferation of datasets using an accession for `node.name` but who wish to see node names of another piece of metadata. See commit messages for more details.

Using our current monkeypox/mpxv dataset with the following differences:
* `node.name` is the accession (which is the uniq ID used throughout the pipeline). This removes the need for the pipelines to use JSON-modification scripts such as `set_final_strain_name.py`.
* "strain" is added to `node_attrs` reflecting the original strain name. (Any name can be used here.)
* `"tip_label": "strain"` added as a `display_defaults`
* "strain" added to filters and colorings

On hover boxes show both the chosen tip-label and the underlying node name (accession). (On-click modals are similar)
image

Filtering to duplicate strain names works intuitively
image

PR #1692, opened 2023-08-30
Tip label can be defined in display_defaults

jameshadfield:
The ability to encode the tip-label in the URL (?tl=...) was implemented at the same time as the ability to change tip-label keys was done - see https://github.com/nextstrain/auspice/commit/0e55b3022db5ac8c5ffa52deddec61b7182f8103

The ability to encode this in the JSON's display_defaults wasn't done, presumably due to time constraints.

This commit was originally part of the larger https://github.com/nextstrain/auspice/pull/1668, however that's had no comments on it and so I don't know if/when it'll be merged. 

Augur PR https://github.com/nextstrain/augur/pull/1298 will update the schema to allow this field

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 # Changelog
 
+
+* Allow the tip label key to be defined in the JSON via `display_defaults.tip_label`, in addition to being settable via a URL query.
+See [the docs](https://docs.nextstrain.org/projects/auspice/en/stable/advanced-functionality/view-settings.html) or [PR #1668](https://github.com/nextstrain/auspice/pull/1668) for more.
+
 ## version 2.48.0 - 2023/08/31
-
-
 
 * Root sequence data may be inlined in the main dataset JSON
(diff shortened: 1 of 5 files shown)


corneliusroemer:
Let's merge this soon as the schema already advertises tip labels as settable because https://github.com/nextstrain/augur/pull/1299/files was merged. I didn't check if there was already a PR and redid the work in #1703

corneliusroemer (approved):
Tested an it works - though I don't know all the details of why the middleware changes are necessary, in my PR I seemed to be able to get away without.

This one has much better validation and documentation.

Tree to test on: [nextclade_mpox_b1.json.txt](https://github.com/nextstrain/auspice/files/12740815/nextclade_mpox_b1.json.txt)

jameshadfield:
This should have been merged & released _before_ Augur PR https://github.com/nextstrain/augur/pull/1298 but that's not the way it played out. Thanks for the reminder @corneliusroemer!

>  though I don't know all the details of why the middleware changes are necessary, in my PR I seemed to be able to get away without.

The middleware (in this context) keeps the URL query `tl=...` up-to-date - if the selected tip label matches the JSON-set default then we don't want any URL query, as the URL query is intended to change state from the defaults.

jameshadfield:
[Rebased onto master]

Issue #1384, opened 2021-07-28
Extra (broken) grid lines in radial layout


jameshadfield:
Something strange is happening in radial views where we have extra major & minor grids in (some) radial views. [For example:](https://nextstrain.org/flu/seasonal/h3n2/ha/3y?c=cTiter&l=radial)

![image](https://user-images.githubusercontent.com/8350992/127276145-7affb55f-468c-496b-808f-cd21d987088f.png)

PR #1693, opened 2023-09-01
Align first major grid circle in radial view

victorlin:
Compare radial tree view of Zika dataset [before](https://nextstrain.org/zika?l=radial) and [after](https://auspice-victorlin-fix-r-st5x6f.herokuapp.com/zika?l=radial).

When plotting the first major grid, its position value can be below the domain of the x-axis. I'm not sure why. To reduce the effects of this, impose a lower limit on grid positions.

Previously, these situations would render a first major grid circle offset from the center. With this change, it draws a point at the actual center.

The text labels are similarly adjusted.

## Related issue(s)

Prompted by [a discussion post](https://discussion.nextstrain.org/t/1429/5).

## Checklist

- [x] Checks pass
- [x] PR preview looks good - no more weird circle, and the text is aligned.
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ I don't think this is necessary.

### DIFF
--- a/src/components/tree/phyloTree/grid.js
+++ b/src/components/tree/phyloTree/grid.js
@@ -273,7 +273,8 @@ export const addGrid = function addGrid() {
           " " +
           yScale.range()[0].toString();
       } else if (layoutShadow==="radial") {
-        const xPos = xScale(gridPoint.position-xmin);
+        const relativePosition = gridPoint.position - xmin;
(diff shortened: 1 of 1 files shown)


jameshadfield:
Closes #1384

jameshadfield (approved):

PR #1698, opened 2023-09-12
Hide tree options when tree is off

victorlin:

This conditional rendering is in place for all other panels. Make the tree panel behave the same way.

### Testing

- [x] Checks pass
- [x] [Preview](https://auspice-victorlin-hide--vafoqb.herokuapp.com/zika?d=entropy) works to conditionally hide tree options
~- [ ] (to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ Probably not necessary


jameshadfield (approved):
Thanks!

Victor & I also discussed lifting the panel on-off toggles from their own section at the bottom of the sidebar up to be inline with the respective panel options ("Tree options" etc). That work can build on this PR (either here or most probably in a separate PR).

victorlin:
I started on that and realized it'll take a bit more work. I'll merge this PR now and open another PR later.

Issue #1633, opened 2023-02-23
Redesign long side control panel to improve usability


joverlee521:
### Context
With each new feature that we add to Auspice, we usually add corresponding controls for it in the side control panel. This has resulted in the side control panel becoming way too long to the point that it impedes usability. We should redesign the long control panel to improve usability.

### Possible solution
The team has discussed this in conversations in the past, but I don't think we've ever officially decided on a new design. Some ideas we've discussed in the past:

- Move some core controls to the top of the main panel
- Make collapsible sections in the control panel
- Add "normal" + "expert" modes to the panel, where "normal" is a simplified version and "expert" includes all controls

victorlin:
I think a very broad goal would be to reduce the amount of clicks/scrolling/eye-scanning necessary across various use cases.

A couple more ideas (both these can fit with the "collapsible sections" idea):

#### Related items should be close together

This is mostly implemented with the per-panel sections, but a few other ideas:

1. Move the animation options to be under the date slider.
2. Move the toggles for showing/hiding panels to be under each panel section.

#### Conditional rendering

This is implemented in some places already. For example, turning off a panel will hide panel-specific options (except for tree, might be a bug).

It could be expanded to more than just panels, for example the animation buttons + options could live directly under a new "animation on/off toggle", and only shown when that is toggled on.

emmahodcroft:
It might be cool (if not too laborious) to store settings for either all builds or specific builds. If one uses a specific panel quite a lot you could leave it expanded knowing it stays that way, whereas others could remain collapsed. (I feel like I toggle full/panel view a lot, but only rarely use the build selectors at the top, usually using direct URLs, for example).

joverlee521:
> If one uses a specific panel quite a lot you could leave it expanded knowing it stays that way when you come back later, whereas others could remain collapsed. 

This would definitely be useful! This aligns with the idea that we implement "user settings" as suggested in https://github.com/nextstrain/auspice/issues/1574

rneher:
Another option would be to include some some controls in the individual panels. 
We have this for the entropy panel. Similarly, the tree layout could be little icons in the tree. We already have the zoom/layout buttons there. `Normalize frequencies` could also be in the frequency panel. The side bar could be reserved for things that span multiple panels.

PR #1704, opened 2023-09-29
Redesign panel sections in sidebar

victorlin:

Hopefully a good improvement to user experience.

## Related issue(s)

- #1633

## Checklist

- [x] Add back behavior described in dd37b3f1822b16f7608ceb6b92f87420840d90ad
- [x] Add entropy toggle
- [x] Finalize design
    - [x] Overall look ([thread](https://github.com/nextstrain/auspice/pull/1704#discussion_r1346511625))
    - [x] Border placement ([thread](https://github.com/nextstrain/auspice/pull/1704#discussion_r1357498130))
    - [x] Border color ([thread](https://github.com/nextstrain/auspice/pull/1704#discussion_r1366273502))
    - [x] Click behavior ([thread](https://github.com/nextstrain/auspice/pull/1704#discussion_r1366246718))
- [x] Update info tooltips ([thread](https://github.com/nextstrain/auspice/pull/1704#discussion_r1369115003))
- [x] Make issue for modifying `panelsAvailable` when `showTreeToo` is changed ([thread](https://github.com/nextstrain/auspice/pull/1704#discussion_r1366231693))
- [x] Make issue for how to define types for prop drilling ([thread](https://github.com/nextstrain/auspice/pull/1704#discussion_r1369073836))
- [x] Fix alignment of toggle with panel title ([thread](https://github.com/nextstrain/auspice/pull/1704#discussion_r1366262888))
- [x] Update changelog
- [x] Checks pass
    - [x] `unselectedBackground` still works


victorlin:
@jameshadfield this PR is unfinished but would like to get your opinion on general implementation direction.

All the remaining tasks that I can think of are in the PR description - please add anything that I might be missing!

src/components/controls/controls.tsx
```diff
@@ -25,6 +25,17 @@ import {TreeOptionsInfo, MapOptionsInfo, AnimationOptionsInfo, PanelOptionsInfo,
   ExplodeTreeInfo, FrequencyInfo, MeasurementsOptionsInfo} from "./miscInfoText";
 import { AnnotatedHeader } from "./annotatedHeader";
 import MeasurementsOptions from "./measurementsOptions";
+import { useSelector } from "react-redux";
+
+// TODO: Move this to src/store.
+// https://react-redux.js.org/using-react-redux/usage-with-typescript#define-root-state-and-dispatch-types
+interface RootState {
```
jameshadfield, line 32: I haven't kept up to date with Redux's TS stuff (and our usage predates `@reduxjs/toolkit`) but this looks good. We might want to add `[propName: string]: any;` to indicate that there is more state than just controls, and then add it within controls to indicate that the controls state is much bigger than just these 3 properties.

src/components/controls/controls.tsx
```diff
@@ -94,7 +135,6 @@ function Controls({ treeOn, mapOn, frequenciesOn, measurementsOn, mobileDisplay
       <span style={{ paddingTop: "10px" }} />
       <AnnotatedHeader title={t("sidebar:Panel Options")} tooltip={PanelOptionsInfo} mobile={mobileDisplay}/>
```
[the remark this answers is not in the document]
victorlin: Good catch, I've addressed it here: 9714d4234bac0ea17c81ee1e00957b1583ab9c4a

> An alternate approach would be to combine the Animation Options + Panel Options into a "Other options" section.

My inclination is to avoid binning of irrelevant options together in an "other" category. Since the options would be harder to find, it seems like this idea only makes sense under the assumption that these options are used by few users - something we don't have the user data to confirm.

I have an idea to toggle the Animation Options, I'll try that out in another PR.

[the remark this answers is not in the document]
victorlin: Also addressed in 9714d4234bac0ea17c81ee1e00957b1583ab9c4a.

src/components/controls/controls.tsx
victorlin: Good idea! Done in [force-push](https://github.com/nextstrain/auspice/compare/9714d4234bac0ea17c81ee1e00957b1583ab9c4a..c7dd987b9d0ceaca288839205689e4cb4fc61db5).

victorlin: Added some more context in [force-push](https://github.com/nextstrain/auspice/compare/82015f3afd43c409a602b9b113d78a999be90a1d..16e680b03e4f8373d9964fe695bf1012a57074eb).

src/components/controls/controls.tsx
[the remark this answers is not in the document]
victorlin: > I have an idea to toggle the Animation Options, I'll try that out in another PR.

#1705

src/components/controls/panel-toggle.tsx
```diff
@@ -0,0 +1,32 @@
...
+  // visual label.
+  // FIXME: Add a hidden label?
+
+  return (
+    <Toggle
+      display={true}
+      on={on}
+      callback={() => dispatch(togglePanelDisplay(panel))}
+      label={""}
+      style={{display: "inline"}}
```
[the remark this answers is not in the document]
victorlin: I tried making prototypes but found myself going back to code changes for the interactive feel, which is an important factor in this change, and something that a static picture can't provide.

Starting with two options: [1](https://auspice-victorlin-move--vboshc.herokuapp.com/ncov/gisaid/global), [2](https://auspice-victorlin-move--vok2df.herokuapp.com/ncov/gisaid/global). The only difference between them is the toggle being visual or textual.

My thoughts that influenced the initial options:

1. The horizontal divider line is generally useful. There shouldn't be a need to indent if the horizontal divider line is there.
2. I like having the header text be the first thing on the left. It's visually more of a "header" to the section contents than if a toggle was on the far left.
3. Having the visual toggle to the right of the text label can be confusing with other visual toggles appearing on the other side. Is this what prompted you to try a textual toggle in your mockup above?
4. It can be good to keep the toggles, whether visual or textual, vertically aligned so that a minimal amount of visual searching and mouse movement is required to interact with multiple toggles.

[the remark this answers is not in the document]
victorlin: [Cross-posted to Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1696895285086759) with nextstrain.org preview links, including one for #1705.

src/components/controls/controls.tsx
victorlin: I looked into Redux Toolkit a bit more. Well, this is actually [bad practice](https://redux.js.org/usage/migrating-to-modern-redux#migrating-typescript-for-redux-logic) since Redux Toolkit is capable of automatically inferring types. I'm going to try some [migration](https://redux.js.org/usage/migrating-to-modern-redux) in a separate PR.

victorlin: Redux Toolkit PR: #1710

victorlin: Rebased this PR onto the above PR so there are less changes here.

src/components/controls/panel-toggles.tsx
```diff
@@ -1,34 +0,0 @@
...
-
-export default function PanelToggles() {
-  const dispatch = useAppDispatch();
-  const { t } = useTranslation();
-
-  const panelsAvailable = useSelector((state: RootState) => state.controls.panelsAvailable);
-  const panelsToDisplay = useSelector((state: RootState) => state.controls.panelsToDisplay);
-  const showTreeToo = useSelector((state: RootState) => state.controls.showTreeToo);
-
-  const panels = panelsAvailable.slice();
```
jameshadfield, line 18: > Add entropy toggle, keeping [conditions](https://github.com/nextstrain/auspice/blob/8377c4c47003468dab5b34cc02ab2d06eb952ca1/src/actions/recomputeReduxState.js#L289-L296) in mind

The intention of the code is that if the entropy panel is available for display it'll be in  `state.controls.panelsAvailable`. The conditions in that linked comment should are used to determine if "entropy" is an entry in `panelsAvailable` and shouldn't need to be re-examined.

src/components/controls/styles.js
```diff
@@ -24,19 +24,23 @@ export const ControlsContainer = styled.div`
 export const HeaderContainer = styled.div`
   display: flex;
   justify-content: space-between;
-  font-family: ${(props) => props.theme["font-family"]};
-  font-size: 16px;
   line-height: 28px;
   min-height: 28px; /* needed for safari, else the div height is 0 */
   margin-top: 15px;
   margin-bottom: 5px;
+  border-top: 1px solid #d3d3d3;
```
jameshadfield, line 31: I'd suggest the border is only used to demarcate panels, not everything which uses `<AnnotatedHeader>`. (The sidebar element below the last panel may also want to use this to indicate it's not part of the panel options.)

victorlin: Good suggestion. Updated in 66f82bdb901b37e7875278b4c4a00e251cbdf92b.

src/components/controls/panelToggle.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import Toggle from "./toggle";
+import { togglePanelDisplay } from "../../actions/panelDisplay";
+
+type Props = {
+  panel: string
+  on: boolean
+}
+
+const PanelToggle = ({ panel, on }: Props) => {
```
jameshadfield, line 14: The positioning of the toggle is currently so that the bottom of the toggle is in-line with the text baseline (perhaps by design, perhaps not). I think it's aesthetically nicer to align the (vertical) centers, similar to how other toggles are positioned in the sidebar. 
image

src/components/controls/panelHeader.tsx
```diff
@@ -4,22 +4,39 @@ import React from "react";
...
+  }
+
   return (
     <HeaderContainer>
-      <AnnotatedTitle
-        title={title}
-        tooltip={tooltip}
-        mobile={mobile} />
+      <span {...titleContainerProps}>
+        {hasOptions && <PanelChevron show={optionsAreVisible} />}
```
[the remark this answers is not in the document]
jameshadfield: If we do switch the behavior here, I would remove the chevron entirely for mobile display.

src/components/controls/styles.js
```diff
@@ -30,6 +30,23 @@ export const HeaderContainer = styled.div`
...
 `;
 
+export const PanelSectionContainer = styled.div`
+  // Less padding is necessary on the top because there is already some space
+  // from HeaderContainer's top margin.
+  padding-top: 2px;
+  padding-bottom: 8px;
+
+  // Add borders to delineate panel sections from other sections.
+  border-top: 1px solid ${(props) => props.theme.color};
```
jameshadfield, line 40: This is personal preference, but I'd prefer a more subtle colour here (especially when using the nextstrain.org theme). Something like `props.theme.unselectedBackground` works, but it's not exactly great naming. What do you think? It's not major, and it's not something that should hold up this PR.

src/components/controls/controls.tsx
```diff
@@ -16,85 +17,122 @@ import GeoResolution from "./geo-resolution";
...
-        <AnnotatedHeader title={t("sidebar:Animation Options")} tooltip={AnimationOptionsInfo} mobile={mobileDisplay}/>
+        <ControlHeader title={t("sidebar:Animation Options")} tooltip={AnimationOptionsInfo} mobile={mobileDisplay}/>
         <AnimationOptions />
       </span>
 
-      <span style={{ paddingTop: "10px" }} />
-      <AnnotatedHeader title={t("sidebar:Panel Options")} tooltip={PanelOptionsInfo} mobile={mobileDisplay}/>
-      <PanelLayout />
-      <PanelToggles />
       <Language />
```
jameshadfield, line 136: For consistency, we should change the rendering of "Language" to be a `<ControlHeader>`. Minor.

src/components/controls/controlHeader.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { HeaderContainer } from "./styles";
+import { AnnotatedTitle } from "./annotatedTitle";
+
+type Props = {
+  title: string
+  tooltip: JSX.Element
+  mobile: boolean
+}
+
```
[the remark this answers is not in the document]
victorlin: [Force-pushed](https://github.com/nextstrain/auspice/compare/c9e44807721923586f25c182b50d3fef83e3032a..b3bceb5a7d2a672bea05e7e4f8320122280ca01c) all docstring updates.

I agree to not detail arguments/types in component docstrings, but I think it's still valuable to have a description in the prop type definition (which I've added in the force-push). This involved a lot of copy/pasting due to the prop drilling that's going on in the current implementation, and is prone to getting out of sync over time. Maybe using [contexts](https://react.dev/learn/passing-data-deeply-with-context) is a better approach here.

src/components/controls/controls.tsx
```diff
@@ -54,52 +52,87 @@ function Controls({ mobileDisplay }: Props) {
...
-          <ChooseMetric />
-          <ChooseBranchLabelling />
-          <ChooseTipLabel />
-          <ChooseSecondTree />
-          <ChooseExplodeAttr tooltip={ExplodeTreeInfo} mobile={mobileDisplay} />
-          <ToggleTangle />
-        </span>
+      <FilterData measurementsOn={panelsToDisplay.includes("measurements")} />
+
+      {canTogglePanelLayout &&
```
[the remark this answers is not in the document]
victorlin: Done in [force-push](https://github.com/nextstrain/auspice/compare/b3bceb5a7d2a672bea05e7e4f8320122280ca01c..bf037f1bfc8a318efd68454465843ea910fbf1e6). I moved most changes for this control to a separate commit: 13ba87c83aeb33884e9bfab2ff2333f0860cd821

src/components/controls/miscInfoText.js
```diff
@@ -27,11 +27,15 @@ export const AnimationOptionsInfo = (
...
-    Control which panels are being displayed and whether to show the tree and the map side-by-side (<em>grid</em>) or expanded (<em>full</em>).
-    <br/>
-    Note that what options are available here are dataset specific!
+    Control whether to show the tree and the map side-by-side (<em>grid</em>) or expanded (<em>full</em>).
+  </>
+);
+
+export const EntropyInfo = (
+  <>
+    FIXME: Add some info text that describes the entropy panel.
```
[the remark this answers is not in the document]
victorlin: Done in [force-push](https://github.com/nextstrain/auspice/compare/b3bceb5a7d2a672bea05e7e4f8320122280ca01c..bf037f1bfc8a318efd68454465843ea910fbf1e6). I copied the text from the [in-panel info hover](https://github.com/nextstrain/auspice/blob/bf037f1bfc8a318efd68454465843ea910fbf1e6/src/components/entropy/index.js#L280).

I'm also realizing now that the other info tooltips were written to describe panel options, not panels themselves. I think, now that each section clearly represents a panel as a whole, it would be good to at least add an introductory line to describe each panel. What do you think?

src/components/controls/panelToggle.tsx
victorlin: This is something I noticed but after spending a long time inspecting the elements, I couldn't figure out how to fix it.

It's not aligned to the bottom, though it may look like that coincidentally with the font size of "Tree". It, specifically `ToggleBackground`, is aligned to the top of `HeaderContainer`.

image

I would like to get this fixed before merging. Let me know if you have any ideas.

src/components/controls/controls.tsx
victorlin: Done: 7608d33a3fb278c258ba84c9adf67bf3291fa585

src/components/controls/styles.js
victorlin: Good suggestion. I don't think the border color should be coupled with either the font color (which I currently have) or the unselected background color. Some options:

1. Add an additional entry `borderColor`. This would be the easiest change, but thinking long-term it's one more specific thing that needs to be documented and defined.
2. Rename `unselectedBackground` to encompass both the border and toggle background. I can't think of a good name.
3. Use a scale `neutral1`, `neutral2`, … to go from lightest neutral to darkest neutral (or inverse for dark mode). I think the hard part would be coming up with the number of shades in the scale ([this article says 5 shades is good](https://uxdesign.cc/defining-colors-in-your-design-system-828148e6210a)), but should be more future-proof and can be extended to other elements where neutrals are currently hardcoded ([1](https://github.com/nextstrain/auspice/blob/ba7e1c20d463f69d570cd46cc00e9bd9e51d49fa/src/components/controls/styles.js#L41), [2](https://github.com/nextstrain/auspice/blob/ba7e1c20d463f69d570cd46cc00e9bd9e51d49fa/src/components/controls/slider.js#L45), [3](https://github.com/nextstrain/auspice/blob/ba7e1c20d463f69d570cd46cc00e9bd9e51d49fa/src/components/framework/fine-print.js#L29)).

victorlin:
@jameshadfield it's probably hard to track in the sea of notifications/threads. Here's links to places where I'd like your input, as of now:

- [Border color](https://github.com/nextstrain/auspice/pull/1704#discussion_r1369409269)
- [Alignment issue](https://github.com/nextstrain/auspice/pull/1704#discussion_r1369162881)
- [Update panel info tooltips?](https://github.com/nextstrain/auspice/pull/1704#discussion_r1369115003)
- [Use React context instead of prop drilling?](https://github.com/nextstrain/auspice/pull/1704#discussion_r1369073836)

src/components/controls/miscInfoText.js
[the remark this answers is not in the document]
jameshadfield: > Done in [force-push](https://github.com/nextstrain/auspice/compare/b3bceb5a7d2a672bea05e7e4f8320122280ca01c..bf037f1bfc8a318efd68454465843ea910fbf1e6). I copied the text from the [in-panel info hover](https://github.com/nextstrain/auspice/blob/bf037f1bfc8a318efd68454465843ea910fbf1e6/src/components/entropy/index.js#L280).

👍 

> now that each section clearly represents a panel as a whole, it would be good to at least add an introductory line to describe each panel.

I agree

src/components/controls/annotatedTitle.tsx
```diff
@@ -0,0 +1,36 @@
...
+export const AnnotatedTitle = ({title, tooltip=undefined, mobile}: Props) => {
+  return (
+    <span>
+      <HeaderTitle>{title}</HeaderTitle>
+      {tooltip && !mobile && (
+        <>
+          <HeaderIconContainer data-tip data-for={title}>
+            <FaInfoCircle/>
+          </HeaderIconContainer>
+          <StyledTooltip place="bottom" type="dark" effect="solid" id={title}>
```
jameshadfield, line 29: The size of the info-icon has become much smaller in this PR - I assume this wasn't on purpose? I think the larger size is preferable

src/components/controls/styles.js
jameshadfield: Long term 3 is the best option if we really want to help whitelabelling efforts, but it's a large amount of work and I think needs a handful of customised versions to develop across. 

For this task I'd suggest renaming `unselectedBackground` to `alternateBackground`, and allowing background compatibilty by renaming the former to the latter on load (e.g. in `styles.js:49`)

src/components/controls/panelToggle.tsx
jameshadfield: The following works, but you'll have to play with exactly how many pixels to use!

```diff
diff --git a/src/components/controls/toggle.js b/src/components/controls/toggle.js
index e0ba253a..c52fb2c6 100644
--- a/src/components/controls/toggle.js
+++ b/src/components/controls/toggle.js
@@ -7,6 +7,7 @@ const ToggleBackground = styled.label`
   display: inline-block;
   width: 35px;
   height: 21px;
+  margin-top: 3px;
   input {
     display: none;
   }
```

src/components/controls/panel-toggle.tsx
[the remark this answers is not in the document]
jameshadfield: The previews have gone dark, but FWIW I'm happy with the current styles in this PR as it relates to this conversation

src/components/controls/controlHeader.tsx
```diff
@@ -0,0 +1,28 @@
...
+  tooltip?: JSX.Element
+
+  /** Indicates mobile display. */
+  mobile: boolean
+}
+
+/**
+ * A header used by all non-panel controls, containing an informative title.
+ */
+export const ControlHeader = ({title, tooltip=undefined, mobile }: Props) => {
```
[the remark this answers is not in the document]
victorlin: This is because [exactOptionalPropertyTypes](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) in tsconfig is enabled. I don't think it's necessary, so I've pushed a commit to disable it: 67bffee2b0dd80066b32eba028d5b0b6b701d7ec

src/components/controls/annotatedTitle.tsx
victorlin: Oops, this wasn't on purpose. Thanks for catching!

src/components/controls/panelToggle.tsx
victorlin: Adding `margin-top` is more of a patch that doesn't address the alignment issue. I'll use it if we can't figure out how to align properly. I think I might have figured it out though: dc76d27311a6868cba97a83db93203c32c81765a...2198a244a41ba398ef7b2de05981253c168d39cb

src/components/controls/styles.js
victorlin: Sounds good. Done in 67bffee2b0dd80066b32eba028d5b0b6b701d7ec...aa6e3df788853bb12cde4c8968847c5cc05dfbc3

src/components/controls/annotatedTitle.tsx
victorlin: Fixing with da91b15b7c25b705b86fca80917df4ef2e7cb8ae

src/components/controls/controlHeader.tsx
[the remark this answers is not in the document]
victorlin: That's a good suggestion - applied in [force-push](https://github.com/nextstrain/auspice/compare/1864c9cf1a83f655b87cbd101403c98e2ff10c25..ea97343643bc052a90f667025c4fa76b74d7cbd3). It's definitely much more maintainable.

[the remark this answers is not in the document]
victorlin: The way props are typed really shouldn't block this PR, but I'd like to take this opportunity to think about the options more since this is the first extended usage of TypeScript in this codebase.

---

[Here's](https://github.com/nextstrain/auspice/compare/ea97343643bc052a90f667025c4fa76b74d7cbd3..369425a3a5f895c2f3cd2ab4d14896d528db7d65) another option which exports prop types in its entirety, which is useful for cases where a group of props is often passed around together. This PR seems like an ideal place to consider it given many scenarios are covered:

1. [[ex](https://github.com/nextstrain/auspice/blob/369425a3a5f895c2f3cd2ab4d14896d528db7d65/src/components/controls/controlHeader.tsx#L5)] All props of a children are props of the component, and the names are the same.
2. [[ex](https://github.com/nextstrain/auspice/blob/369425a3a5f895c2f3cd2ab4d14896d528db7d65/src/components/controls/panelHeader.tsx#L8-L9)] All props of a children are props of the component, but some have different names. This requires using [indexed access types](https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html).
3. [[ex](https://github.com/nextstrain/auspice/blob/369425a3a5f895c2f3cd2ab4d14896d528db7d65/src/components/controls/panelSection.tsx#L10)] Some props of a children are props of the component. This requires indexed access types even if name is the same.
4. [[ex](https://github.com/nextstrain/auspice/blob/369425a3a5f895c2f3cd2ab4d14896d528db7d65/src/components/controls/panelSection.tsx#L12-L13)] The component has its own types.
5. [[ex](https://github.com/nextstrain/auspice/blob/369425a3a5f895c2f3cd2ab4d14896d528db7d65/src/components/controls/panelHeader.tsx#L7-L19)] A mix of everything.

This has even less duplication and (to me) is more readable when aliases such as `AnnotatedTitleProps` are used, but I can see how it adds complexity.

@jameshadfield what do you think?

src/components/controls/panelToggle.tsx
jameshadfield: 👍

src/components/controls/controlHeader.tsx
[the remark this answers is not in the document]
jameshadfield: I've not written enough TS + React code to make a decision, but I'd like to spend some time (here & there) looking at what others have done / reading things on this subject. 

For the purposes of this PR can you pick either of your implementations, and then break this out into a separate issue that can bubble along in the background for a while?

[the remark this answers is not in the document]
victorlin: Yes, for sure. I'll stick your suggestion (what's currently included in this PR).

src/components/controls/miscInfoText.js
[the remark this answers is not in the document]
victorlin: Done: 830b73de2d4d03f42b7272e8fd6f83e6616cfed1

src/components/controls/miscInfoText.js
```diff
@@ -27,23 +28,31 @@ export const AnimationOptionsInfo = (
...
     <em>Normalize frequencies</em> controls whether the vertical axis represents the entire dataset or only the samples currently visible (e.g. due to filtering).
     This option is not available when data is limited to prevent numerical issues.
   </>
 );
 
-export const MeasurementsOptionsInfo = (
+export const MeasurementsInfo = (
   <>
+    This panel displays measurements.
+    <br/>
```
victorlin, line 55: I've repurposed each panel options section to represent the panel as a whole. Naturally, this wants a short description of each panel. I've added most in 830b73de2d4d03f42b7272e8fd6f83e6616cfed1, but don't know what to write for the measurements panel.

@joverlee521 are you able to suggest a short description for this panel?

joverlee521: Ah sure! How about:

```
This panel displays multidimensional data such as serological experimental results 
that are linked to samples in the tree.
```

victorlin: That's perfect, thank you!

victorlin: Actually, since one can toggle the tree off and keep measurements on, does it make sense to word it without assuming the tree is visible? Or should the behavior be changed such that measurements can only be shown when the tree is also shown?

src/components/controls/panelHeader.tsx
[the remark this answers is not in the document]
victorlin: In this [force-push](https://github.com/nextstrain/auspice/compare/830b73de2d4d03f42b7272e8fd6f83e6616cfed1..6753065b16fafa781fa44424e66852cd56a2f33d), among other smaller changes, I made it so that clicking most of the header changes the toggle.

Notes:

- Clicking the info icon does not change the toggle, since I imagine that can only be unintentional.
- Clicking the chevron is the only thing that changes options visibility, and is not rendered on mobile display.

src/components/controls/miscInfoText.js
joverlee521: Hmm, the measurements can definitely be shown even if the tree is off. 

I guess what I'm trying to get across is that the panel cannot contain arbitrary data not related to at least one sample in the tree. But maybe that's obvious since they are all part of the same dataset?

jameshadfield: I think the text looks good as is. A (the?) central design of Auspice is that all data is derived from / linked to nodes in the tree.

victorlin: I see. I think your snippet is good! Could be read implicitly as
> … results that are linked to samples in the tree (which may be hidden).

victorlin: Added: 40ff36dfddf1771fbfdd3c48acc32938f43925de

src/components/controls/panelHeader.tsx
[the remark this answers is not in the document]
jameshadfield: Thanks. Behaviour works well. One minor bug - the "filter data", "color by" headers shouldn't have a "pointer" cursor.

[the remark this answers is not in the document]
victorlin: Oh right, good catch.

victorlin:
Force-pushing a rebase onto latest `master`.

src/components/controls/panelHeader.tsx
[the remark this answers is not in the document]
victorlin: Fixed in [force-push](https://github.com/nextstrain/auspice/compare/b25e783edcc63d11798541f2ee207bd8aacb3f06..a8f8b0dad73419da5b42d5edbd979cb6752888d7).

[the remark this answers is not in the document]
victorlin: After collecting feedback [on Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1699577373284099?thread_ts=1696895285.086759&cid=C01LCTT7JNN), I've [force-pushed](https://github.com/nextstrain/auspice/compare/a8f8b0dad73419da5b42d5edbd979cb6752888d7..e3eabae6ddec32dbcd68e7159d24efcee5d0b4ec) another change in behavior.

Instead of making most of the click area toggle panel visibility, make it toggle the options visibility.

victorlin:
[Force-pushed](https://github.com/nextstrain/auspice/compare/e3eabae6ddec32dbcd68e7159d24efcee5d0b4ec..8a69ca2c2890839bc2c8fc8f0148d4f788a4e0dc) two more design touch-ups:

1. Move full/grid display control back to the bottom, since its conditional rendering when placed at the top worsens usability (noted in 404e2d4fa5a99356a11df2e172b6c04e5a4b93f6).
2. Prevent the option-less Entropy section from showing a pointer on hover.

src/components/controls/panelHeader.tsx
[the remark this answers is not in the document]
jameshadfield: 👍

jameshadfield (approved):
I haven't re-reviewed the code changes, but UI-3 is working well and, judging by the commentary on slack, looks good to merge!

src/components/controls/controls.tsx
```diff
@@ -54,52 +52,87 @@ function Controls({ mobileDisplay }: Props) {
...
 
-      {measurementsOn &&
-        <span style={{ marginTop: "10px" }}>
-          <ControlHeader title={t("sidebar:Measurements Options")} tooltip={MeasurementsOptionsInfo} mobile={mobileDisplay}/>
-          <MeasurementsOptions />
-        </span>
+      {/* Prevent the map from being toggled on when a second tree is visible.
+          It is hidden by logic elsewhere.
+      */}
+      {panelsAvailable.includes("map") && !showTreeToo &&
```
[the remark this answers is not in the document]
victorlin: #1722

src/components/controls/controlHeader.tsx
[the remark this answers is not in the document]
victorlin: #1723

PR #1710, opened 2023-10-11
Start using modern Redux

victorlin:

Mostly following the [migration doc](https://redux.js.org/usage/migrating-to-modern-redux). See commit messages.

### Related issue(s)

#1711

### Checklist

- [x] Checks pass
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ Not necessary - no functional changes.
- [x] Manually clicking around preview app, everything seems to work fine


src/store/index.js
```diff
@@ -12,7 +12,14 @@ const middleware = [
 
 const store = configureStore({
   reducer: rootReducer,
-  middleware: (getDefaultMiddleware) => getDefaultMiddleware().concat(middleware),
+  middleware: (getDefaultMiddleware) =>
+    getDefaultMiddleware({
+      // TODO: Go through reducers and see why the state is not immutable nor serializable.
+      // These were not checked prior to the adoption of Redux Toolkit, and were not
+      // investigated to minimize conversion efforts.
+      immutableCheck: false,
```
jameshadfield: Re: immutability, there are objects "controlled" by our d3-tree-rendering code, and these are referenced by our redux tree state. It's not following the redux paradigm but there were reasons we chose to do it this way! Off the top of my head there shouldn't be any _other_ redux state that's mutated outside of the canonical dispatch loop, but I'd be interested to find out!

Re: serializable, we use lots of circular references in our state, so I presume this is what's preventing serializable.

victorlin: That's good to know! If you want, you can check by commenting these lines out and investigating the errors when loading Auspice in the browser. Otherwise, I'll take a look tomorrow – I'm about to sign off for the day.

jameshadfield: The errors are (on the surface) somewhat unhelpful maximum recursion reached errors, which I suspect are due to the circular references we use. It [looks like](https://github.com/reduxjs/redux-toolkit/blob/e657098ad8e8d13e3757b5c2cfe17c594320f6b9/packages/toolkit/src/immutableStateInvariantMiddleware.ts#L205C29-L205C29) we could use `ignoredPaths` to narrow things down. It would be interesting to check that only the parts of state we are knowingly and deliberately modifying fail this check; however looking at the middleware code it looks like we'd have to remove circular references before this check is possible.

jameshadfield (approved):

src/store/index.js
victorlin: Continued in #1712.

PR #1712, opened 2023-10-12
Consider enabling default middleware checks on Redux store

victorlin:
Follow-up to dcd83c0dcb10cabee1fab827f314d97176d4ae94.

### Checklist

- [x] Checks pass
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ N/A, no functional changes
- [x] (local test) Setting `ignoredPaths` to an empty array and running `export SKIP_REDUX_CHECKS=true && npm run dev` shows no console warnings/errors related to the immutability middleware.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,18 @@ If you are editing source code, running the following command will allow hot-rel
 auspice develop --datasetDir data
 ```
 
+#### Environment variables
+
+The client looks for some environment variables. All are optional.
+
+> [!NOTE]
+> This is an incomplete list. For other variables, search for `process.env.` in the codebase.
+
+- `SKIP_REDUX_CHECKS`: Set this to a truthy value to improve dev server responsiveness. Useful when you see a console warning like this:
+
+    > ImmutableStateInvariantMiddleware took 200ms, which is more than the warning threshold of 32ms.
+    > If your state or actions are very large, you may want to disable the middleware as it might cause too much of a slowdown in development mode.
+
 ### CLI (Command Line Interface)
 
 Run `auspice --help` or `auspice view --help` to see all the available command line options.
--- a/src/store.ts
+++ b/src/store.ts
@@ -15,15 +15,19 @@ const store = configureStore({
   middleware: (getDefaultMiddleware) =>
     // This adds the thunk middleware, and for development builds, other useful checks.
     getDefaultMiddleware({
-      immutableCheck: {
-        // Immutability can't be checked in the following due to circular references.
-        ignoredPaths: [
-          'tree.nodes',
-          'tree.vaccines',
-          'treeToo.nodes',
-          'treeToo.vaccines',
-        ],
-      },
+      // Immutability can't be checked in some parts of the state due to circular references.
+      // Allow the option to disable this check through an environment variable.
+      // Note that it is never enabled when NODE_ENV=production.
+      immutableCheck: process.env.SKIP_REDUX_CHECKS
+        ? false
+        : {
+          ignoredPaths: [
+            'tree.nodes',
+            'tree.vaccines',
+            'treeToo.nodes',
+            'treeToo.vaccines',
+          ],
+        },
 
       // By design, the state contains many values that are non-serializable.
       // Instead of adding several ignoredPaths, disable this check entirely.
--- a/src/store.ts
+++ b/src/store.ts
@@ -19,7 +19,12 @@ const store = configureStore({
       // These were not checked prior to the adoption of Redux Toolkit, and were not
       // investigated to minimize conversion efforts.
       immutableCheck: false,
-      serializableCheck: false
+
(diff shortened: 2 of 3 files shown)


src/store.ts
victorlin, line 1: @jameshadfield so `ignoredPaths` does work, but not with just `tree.nodes`/`treeToo.nodes` as you expected. I'm sure we can make it more specific than `tree`/`treeToo`.

It also seems like `serializableCheck` has something other than `tree`/`treeToo` that is causing the "Maximum call stack size exceeded" error.

jameshadfield: Ok. I guess we use circular references in other places then? (Not altogether unsurprising.) Perhaps we can try ignoring almost everything then adding parts back in to see where the recursion errors are coming from?

jameshadfield: Re: serialization. From a cursory look at redux-toolkit this is necessary for time-travel debugging to work. With the size of state we have in Auspice I'm sceptical this would be performant enough to be usable. We should evaluate what the benefit of it will be for us before spending time implementing it. 

(It's been a while since I used redux & react advanced dev tooling a lot, but in the days before redux-toolkit very few features worked in Auspice because of the size of our state.)

victorlin: > Perhaps we can try ignoring almost everything then adding parts back in to see where the recursion errors are coming from?

Yeah this would be my approach, but I imagine it'll be a bit tedious.

> From a cursory look at redux-toolkit this is necessary for time-travel debugging to work.

Is time-travel debugging something that you've done previously? Curious if the update to Redux Toolkit went backwards on that.

> We should evaluate what the benefit of it will be for us before spending time implementing it.

💯 I'd think to keep `serializableCheck` disabled and update the note with something along the lines of "we store many values that are non-serializable by design".

jameshadfield: The following works (at least for a single tree, and not testing the measurements panel). From limited testing there's no out-of-redux state mutation happening, although it notes `ImmutableStateInvariantMiddleware took 62ms, which is more than the warning threshold of 32ms. 
If your state or actions are very large, you may want to disable the middleware as it might cause too much of a slowdown in development mode.`

```js
      immutableCheck: {
        ignoredPaths: [
          // Immutability can't be checked here because of circular references.
          'tree.nodes',
          'tree.vaccines',
        ],
      },
      serializableCheck: false
```

> Is time-travel debugging something that you've done previously? 

Yes, at least that's my memory of early-days Auspice (although we had circular references from very early on, so perhaps I'm misremembering). But it was unusably slow, so it never formed part of my day-to-day development process.

victorlin: @jameshadfield thanks for narrowing it down to those paths. I tested the measurements panel using [files from blab/measurements-panel](https://github.com/blab/measurements-panel/tree/c6d7e2cbbfbdfebe95e0da60d7b4cfaa8b2612df/auspice) and it seems to work fine.

I've updated this PR with new commits reflecting our discussion above. I think merge-able now, let me know what you think.

jameshadfield: Nice one. My concern is around the speed penalty, but I don't want to discard this as it seems a sensible sanity check. I'd suggest we add an opt-out env variable, e.g. `SKIP_REDUX_CHECKS=1 auspice develop`

victorlin: Ok, I've added 7d143341455e68ec81988bec98d1e264977bedb2.

Note that the console warning message is not super helpful since it doesn't mention the middleware's field name for customization (`immutableCheck`) nor `SKIP_REDUX_CHECKS`. Ideally we'd catch that warning and give a more helpful one (e.g. "re-run dev server with `SKIP_REDUX_CHECKS=true`"). I haven't looked into how to do that.

The next best thing that I can think of is to document this value. I think it'd be better to move all docs about running Auspice to `DEV_DOCS.md` and document all env vars there, but those are beyond the scope of this PR. Added this bit of documentation to a new sub-section in `README.md`.

README.md
```diff
@@ -75,6 +75,18 @@ If you are editing source code, running the following command will allow hot-rel
 auspice develop --datasetDir data
 ```
 
+#### Environment variables
+
+The server looks for some environment variables. All are optional.
```
jameshadfield, line 80: Nitpick, but `/server/client/`. (It's been a perpetual source of confusion to people over the years as Auspice can mean both the server and the client.)

The server does also look for some variables, the most common being `PORT` and `HOST`, so over time we can include them.

And then there's variables which are essentially compiler arguments, e.g. `BABEL_*`, which I would categorise separately to the server.

jameshadfield (approved):
Looks good, pending the one minor syntax change.

README.md
victorlin: Thanks for the clarification. This is something I haven't looked into much and doesn't seem to be spelled out in docs. I'll look further and update docs in a future PR.

PR #1717, opened 2023-10-23
Move mobile display detection to Redux

victorlin:

Addressing a task defined in a comment:

https://github.com/nextstrain/auspice/blob/ba7e1c20d463f69d570cd46cc00e9bd9e51d49fa/src/components/main/index.js#L53-L54

and getting rid of the prop drilling before more layers are added in #1704.

### Testing

- [x] Checks pass
- [x] There should be no functional changes, and it'd be good to check that all modified components still behave properly.
    - [x] Confirm existing behavior: Interface on a window with narrow width defaults to hidden sidebar, and when the sidebar is opened, there is no info icon for hover. Tested locally on Chromium and Safari.


jameshadfield (approved):
Thanks!

PR #1726, opened 2023-11-16
Properly use useAppDispatch hook

### LINKED ISSUE #1725: Narratives editor broken
_reported by @huddlej [on Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1700093470010439)_

I tested locally and can reproduce. Determined that the narratives editor works on v2.49.0 but not v2.50.0. Relevant parts of error message:

```
Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

The above error occurred in the <DebugNarrative> component
```

Steps to reproduce:

1. Download [example narrative](https://github.com/nextstrain/narratives/blob/master/intro-to-narratives.md).
2. Go to [nextstrain.org/edit/narratives](https://nextstrain.org/edit/narratives).
3. Drag the narratives file onto the page
4. Observe error

victorlin:

Function call parentheses were accidentally removed in 3e77a0a0f572db0e4621d9dfc3614a7353311be2.

The code is only invoked during runtime when accessing the narrative editor, which was not tested before merging that change.

### Related issue(s)

Fixes #1725

### Testing

- [x] Checks pass
- [x] [Preview narrative editor](https://auspice-victorlin-fix-n-c3t4ly.herokuapp.com/edit/narratives) works

### DIFF
--- a/src/components/narrativeEditor/useDatasetFetch.js
+++ b/src/components/narrativeEditor/useDatasetFetch.js
@@ -14,7 +14,7 @@ import { FetchError } from "../../util/exceptions";
  */
 
 export function useDatasetFetch(datasets) {
-  const dispatchRedux = useAppDispatch;
+  const dispatchRedux = useAppDispatch();
   const [datasetResponses, dispatchDatasetResponses] = useReducer(
     (state, action) => {
       if (action.reset) return {};


victorlin:
Test passed - merging and releasing.

PR #1731, opened 2023-12-12
Improvements to updated date

victorlin:

See commit messages.

## Related issue(s)

N/A

## Checklist

- [ ] Checks pass


package.json
victorlin, line 1: CI is [failing](https://github.com/nextstrain/auspice/actions/runs/7175501775/job/19538899725#step:5:30) on bundlesize. Looks like adding `moment` might have caused the bundle to tip over its suggested size limits. I'll probably just bump the limit.

src/components/info/byline.js
```diff
@@ -106,11 +109,13 @@ function renderMaintainers(t, metadata) {
...
  */
-function renderDataUpdated(t, metadata) {
-  if (metadata.updated) {
+function renderDataUpdated(t, metadata, language) {
+  if (moment(metadata.updated).isValid()) {
     return (
       <span>
-        {`${t("Data updated")} ${metadata.updated}. `}
+        {`${t("Data updated")} `}
+        <Moment locale={language} date={metadata.updated} fromNow />
```
victorlin, line 117: This is splitting translation on a English boundary, so it might not be proper grammar for all languages. Hopefully it still gets point across.

package.json
jameshadfield: This is my biggest concern with this direction - is the size of the imported library worth it?

Our bundlesize has exceeded the limit by 73kB (gzipped) so we're adding at least this much, and [bundlephobia](https://bundlephobia.com/package/moment@2.29.4) indicates moment's around 72kB (gzipped). If we do want to implement this functionality I'd suggest switching to luxon ([23kB gzipped](https://bundlephobia.com/package/luxon@3.4.4)) or some other even-smaller library.

jameshadfield: And react-moment is another [8kB gzipped](https://bundlephobia.com/package/react-moment@1.1.3)! I would expect the functionality we need here would be a single function which calls moment.

src/components/info/byline.js
jameshadfield, line 117: (I haven't run the code, just reading it)

How does this work for very recent dates? In `augur export` the updated string uses [`time.strftime('%Y-%m-%d')`](https://github.com/nextstrain/augur/blob/master/augur/export_v2.py#L1086) which'll use the [users local time](https://docs.python.org/3/library/time.html#time.strftime). Our schema [says nothing](https://github.com/nextstrain/augur/blob/master/augur/data/schema-export-v2.json#L24C1-L28C19) about time zones. I presume (?) moment's also interpreting the date string in the (Auspice) users' local time zone (or is it UTC?). 

So what happens if, e.g., I update a dataset right now, thus getting "2023-12-13" and someone in the US views it (the US is currently Dec 12th)... is the dataset updated "tomorrow"? Do we ignore it because it's a bad value? What about the inverse situation, would the dataset appear to me to be updated "yesterday" even if it'd been updated 5 min ago? Time zones make this hard!

jameshadfield: It'd be useful to create a table or similar testing different values here. From the datasets on the Auspice review app, long-ago updated datasets do look better than before, e.g. "Data updated 2 years ago" is better than a YYYY-MM-DD string!

victorlin: Good point, and very important to consider since we have daily ncov builds.

There's really no good way to handle this with the current schema. The same thing could be said for the current implementation of displaying `meta.updated` as-is without any time zone info.

I think the difference is that relative dates must assume a time zone, and that its effects are much worse when recent. Continuing your example, "tomorrow" sounds much worse than "2023-12-13", but "3 years ago" zooms the perspective out to a point where time zone doesn't matter.

How about we only show relative dates if it's at least a week old?

victorlin: Going further, we could extend the schema to support an exact ISO datetime (e.g. `2023-12-12T21:47:18Z`) which then allows proper conversion to the user's timezone.

victorlin, line 117: One scenario where this change would be unfavorable is in screenshots, where relative dates don't age well. (I just noticed this while watching someone present their Nextstrain build).

victorlin: Displaying both would make things more interpretable without compromising screenshots, but maybe that's getting too wordy?

> Data updated 3 years ago (2020-12-11)

package.json
victorlin: Thanks for the suggestion. [Updated to use luxon](https://github.com/nextstrain/auspice/compare/f48f0711019811b2b0f377bd5f21e6bf131de0dc..c826b5a490ffa96ece8fc4b20a37c095ee3370f4), which required bumping the bundlesize limit by ~30kb

You're right about react-moment - I had added it because I was specifying `language` improperly but now realize that it's not necessary.

src/components/info/byline.js
joverlee521: Personally, I get irked by relative dates without the ability to see the exact date. I think displaying both would be better.

victorlin: Updated to display both:

https://github.com/nextstrain/auspice/blob/a89abcfc0ecfb338e037bf50e1dfc2658358ae10/src/components/info/byline.js#L117

victorlin: Updated so that the most granular relative date is "in the last week":

https://github.com/nextstrain/auspice/blob/a89abcfc0ecfb338e037bf50e1dfc2658358ae10/src/components/info/byline.js#L114-L119

victorlin: Moved this to a comment:

https://github.com/nextstrain/auspice/blob/dff1f6ef8a91ed08c6202d6c2548f3a6d92faf5c/src/components/info/byline.js#L114-L122

PR #1733, opened 2023-12-28
Replace CleanWebpackPlugin with output.clean

victorlin:
The built-in option is available starting with webpack 5.20.0.¹ It's also recommended in the output management guide.²
    
¹ <https://webpack.js.org/configuration/output/#outputclean>
² <https://webpack.js.org/guides/output-management#cleaning-up-the-dist-folder>


jameshadfield (approved):
Looks good. Tested by adding some files to `./dist` and verifying that they were removed when auspice was built.

PR #1738, opened 2024-01-30
Revert to using CleanWebpackPlugin

### LINKED ISSUE #1737: [dev] hot-reloading broken
Running `auspice develop ...` works fine initially, but as soon as the code rebuilds it is no longer viewable in the browser. Firefox displays "The page isn’t redirecting properly. Firefox has detected that the server is redirecting the request for this address in a way that will never complete". It looks like we're getting into a infinite loop of 302 redirects to the same page ("/"). 

### How to reproduce
1. `auspice develop`
2. load localhost:4000 - note that all works as expected
3. Change some code, you should see "webpack building..." in the terminal 
4. Refresh localhost:4000

### Possible solution
The bug is present at current master (40befc6ea4ac02226db2f5658faa12beda52a3e1).
Bisecting indicates the bug was introduced in #1733

victorlin:

This reverts "Replace CleanWebpackPlugin with output.clean" (798f3ee42) due to issues with Hot Module Replacement and output.clean.

## Related issue(s)

- Closes #1737

## Checklist

- [x] hot reloading works (can no longer repro #1737)
- [x] Checks pass
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ not necessary

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "binomial": "^0.2.0",
         "buffer": "^6.0.3",
         "chalk": "^2.4.1",
+        "clean-webpack-plugin": "^4.0.0",
         "compression": "^1.7.3",
         "compression-webpack-plugin": "^3.1.0",
         "core-js": "^3.6.5",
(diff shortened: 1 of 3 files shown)


jameshadfield (approved):
Thanks @victorlin!

One of the reasons I've been hesitant to dive back into webpack is due to bugs such as this. Hot reloading, extendability, aliases, transpilation, env-variable handling, bundling and differential cacheing all rely on webpack and the interaction between webpack and a dozen other packages. From experience it's all too easy to break some of this functionality inadvertantly. This part of the JS ecosystem keeps evolving, and so we fall out of date over time. (Notice all the deprecation warnings we are getting both at build time and run time.)

I wonder if the best way to update this part of Auspice is a concerted push to update all of React, hot reloading, redux (?), babel, webpack, webpack plugins, etc. As part of this we'd define a number of tests beforehand so we can be more confident nothing's broken.

victorlin:
A concentrated push sounds like the right approach here (instead of incremental pushes such as #1737). That would require much more effort and intention. I've added it to the "Project ideas" section in the [priorities doc](https://docs.google.com/document/d/1hVJBa3d4SsW-WY5Ag4SYVtQXmwMgcoCwQDUyK_5rRbY/edit#heading=h.klsxtvmi840h):

> - Modernize front-end codebases
>     - TypeScript adoption in Auspice
>     - **Replace outdated usage of React, Webpack, Babel, Redux, etc.**

Issue #1251, opened 2020-12-20
Sidebar filtering options should include all node/branch attrs


jameshadfield:
The original intent of the sidebar filtering was to allow filtering by all traits which are defined via `metadata.filters` as well as those defined via `metadata.colorings`. The latter has not yet been implemented.

Note that the filters which appear in the footer of Auspice should still only reflect those defined in `metadata.filters`.

joverlee521:
With the work to export metadata fields without defining them as coloring options (https://github.com/nextstrain/augur/pull/1384), we should extend filtering options to include _all_ available metadata fields.

jameshadfield:
I think this issue should be extended a bit (but I'm including here as it'll almost certainly be simpler to do it in a single push). 

I think _all_ node and branch attrs defined across a tree should be searchable via the dropdown sidebar. That is, the `meta.filters` array should no longer be used. For the filters listed in the footer we don't want this behaviour, and we could repurpose the `meta.filters` array to suit this purpose.

Issue #1275, opened 2021-01-25
Filtering does not work as expected if internal traits are defined


jameshadfield:
**Context**

Currently, when filtering by a trait (e.g. [SARS-CoV-2 filtered to clade 20I/501Y.v1](https://nextstrain.org/ncov/global?f_clade_membership=20I/501Y.V1)) we determine node visibility via the following steps
1. Find the set of tips `X` which match the filter (in this case, `X` is tips annotated with `clade_membership = 20I/501Y.V1`)
2. Postorder traversal to add the ancestral nodes of these tips to the selection, `X`. This will always end up visiting and including the root, i.e. `rootNode \in X`.
3. Preorder traversal to find the MRCA of `X` and remove nodes from `X` which are above this. The tree can then be zoomed to this (MRCA) node via the "zoom to selected button". 

Importantly, this approach does not take into account trats annotated on internal nodes, should they exist. Most of the time this works well, but it there are situations where it does not.

**Example dataset**
The following 

**Bug 1: 

Given a trait whose values are known and annotated across the tree, because filtering does not examine internal node states the filtering may be wrong. As en 

**Posible solution**

PR #1743, opened 2024-01-30
Enable filtering for all available metadata

jameshadfield:
See commit messages for more details, but here's the main change:

The sidebar filtering now surfaces all valid node-attrs defined across
the (terminal) nodes. URL queries (`?f_${attrName}=value1,value2,...`)
also work for all known attrs. Attributes which are known to be
continuous (via a colorings definition) are excluded, as the filtering
UI is not (yet) able to handle these; if a non-coloring continuous
attribute is set on the nodes then this will end up as a multitude of
numerical options in the sidebar.

As part of this implementation we have removed `stateCountAttrs` from
redux state and improved the validation of (filtering) URL queries.

The behaviour of filtering, and the restriction to collecting attributes
from terminal nodes only, is unchanged. See https://github.com/nextstrain/auspice/issues/1275 for more context.

Closes https://github.com/nextstrain/auspice/issues/1251

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -669,8 +669,7 @@ const modifyTreeStateVisAndBranchThickness = (oldState, zoomSelected, controlsSt
   const newState = Object.assign({}, oldState, visAndThicknessData);
   newState.stateCountAttrs = Object.keys(controlsState.filters);
   newState.idxOfInViewRootNode = newIdxRoot;
-  newState.visibleStateCounts = countTraitsAcrossTree(newState.nodes, newState.stateCountAttrs, newState.visibility, true);
-  newState.totalStateCounts   = countTraitsAcrossTree(newState.nodes, newState.stateCountAttrs, false,               true);
+  newState.totalStateCounts = countTraitsAcrossTree(newState.nodes, newState.stateCountAttrs, false, true);
 
   return newState;
 };
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -8,7 +8,7 @@ import { constructVisibleTipLookupBetweenTrees } from "../util/treeTangleHelpers
 import { getDefaultControlsState, shouldDisplayTemporalConfidence } from "../reducers/controls";
 import { getDefaultFrequenciesState } from "../reducers/frequencies";
 import { getDefaultMeasurementsState } from "../reducers/measurements";
-import { countTraitsAcrossTree, calcTotalTipsInTree } from "../util/treeCountingHelpers";
+import { countTraitsAcrossTree, calcTotalTipsInTree, gatherTraitNames } from "../util/treeCountingHelpers";
 import { calcEntropyInView } from "../util/entropy";
 import { treeJsonToState } from "../util/treeJsonProcessing";
 import { castIncorrectTypes } from "../util/castJsonTypes";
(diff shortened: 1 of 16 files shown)


src/components/controls/filter.js
```diff
@@ -54,22 +55,38 @@ class FilterData extends React.Component {
...
+    /**
+     * First set of options is from the totalStateCounts -- i.e. every node attr
+     * which we know about (minus any currently selected filters). Note that we
+     * can't filter the filters to those set on visible nodes, as selecting a
+     * filter from outside this is perfectly fine in many situations.
+     *
+     * Those which are colorings appear first (and in the order defined in
+     * colorings). Within each trait, the values are alphabetical
+     */
+    const coloringKeys = Object.values(this.props.colorings||{}).map((c) => c.key);
```
joverlee521, line 68: From my testing of a couple datasets in the test app, the filter options do not follow the colorings order. 

I think this needs to be updated to
```suggestion
    const coloringKeys = Object.keys(this.props.colorings||{});
```

src/components/controls/filter.js
```diff
@@ -54,22 +55,38 @@ class FilterData extends React.Component {
...
+    const traitNames = Object.keys(this.props.totalStateCounts).sort((a,b) => {
+      return coloringKeys.indexOf(a) - coloringKeys.indexOf(b)
+    })
```
joverlee521, line 71: If a trait is not included in `coloringKeys` then this will sort them to the beginning of the `traitNames` array. 

I think this needs to be updated to 
```suggestion
    const traitNames = Object.keys(this.props.totalStateCounts).sort((a,b) => {
      if(coloringKeys.indexOf(b) === -1) return -1;
      if(coloringKeys.indexOf(a) === -1) return 1;
      return coloringKeys.indexOf(a) - coloringKeys.indexOf(b)
    })
```

joverlee521 (approved):
The new filters work as expected in the test app! I tested filtering by accessions in the ncov and seasonal-flu builds.

+1 for removing a feature that has been broken for ~4 years 😅 
I left some comments to change the order of the filter options, but overall looks good to me.

src/actions/recomputeReduxState.js
```diff
@@ -225,19 +230,16 @@ const modifyStateViaMetadata = (state, metadata, genomeMap) => {
...
   }
   if (metadata.filters) {
-    /* the `meta -> filters` JSON spec should define which filters are displayed in the footer.
-    Note that this UI may change, and if so then we can change this state name. */
+    /**
+     * this spec previously defined both the footer-filters and the
+     * sidebar-filters, however it now only defines the former as the sidebar
+     * surfaces all available attributes.
+     */
     state.filtersInFooter = [...metadata.filters];
```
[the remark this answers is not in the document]
jameshadfield: This actually caused Auspice to crash with the changes in this PR, so thanks for bringing it up.

Fixed in force push.

src/components/controls/filter.js
jameshadfield: Ha! I had gotten all mixed up between the JSON structure and the auspice structure. Fixed in force push.

src/components/controls/filter.js
jameshadfield: Yeah, I've switched to a simpler to understand construction:

```js
    const coloringKeys = Object.keys(this.props.colorings||{});
    const unorderedTraitNames = Object.keys(this.props.totalStateCounts);
    const traitNames = [
      ...coloringKeys.filter((name) => unorderedTraitNames.includes(name)),
      ...unorderedTraitNames.filter((name) => !coloringKeys.includes(name))
    ]
```

PR #1749, opened 2024-02-05
Improve interaction between selected node modal and strain filters

### LINKED ISSUE #1243: Deselecting a single strain doesn't remove the info panel overlay

### LINKED ISSUE #1701: Filtering being deactivated after clicking away from metadata popup
### Current Behavior
When you click on a filtered sample and click away from the popup, the filter becomes deactivated.

### Expected behavior
The filtered shouldn't be deactivated.

### How to reproduce
1. Visit https://nextstrain.org/fetch/storage.googleapis.com/ucsc-gi-cdph-bigtree/CDPH_09_20_2023/CDPH-subtree-178.json?s=USA/CA-CDPH-2000016698/2021%7CEPI_ISL_2782613%7C2021-01-04,USA/CA-CDPH-2000016731/2021%7CEPI_ISL_2782386%7C2021-01-03
2. Click "Zoom To Selected"
3. Click on a sample
4. Click away the popup.
5. Samples filter should be deactivated.

### Your environment: if browsing Nextstrain online
 - Operating system: macOS 13.5.2 (22G91)
 - Browser: Chrome Version 117.0.5938.88 

### Additional context
Auspice Recording

jameshadfield:
I think the first four commits are reasonably straight forward in terms of code + UI, but the final commit represents my best attempt to fix the situation described in #1701 while still maintaining the expected behaviour when we click on tips without having  any strain-filters enabled. 

Closes #1701
Closes #1243


jameshadfield:
_Merging after discussion with @trvrb_

PR #1764, opened 2024-04-09
Add download JSON button

jameshadfield:
The main (i.e. not sidecar) dataset is downloaded (or datasets in the case of tangletrees). Downloading sidecar files is also possible if we want to expose them.

Given that we don't store the (un-modified) dataset JSON in Auspice to download it we can either (a) convert the modified data format back to the original JSON or (b) re-download it. (a) is by far the simpler implementation and with appropriate caching should have good performance.

- [ ] Checks pass

### DIFF
--- a/src/components/download/downloadButtons.js
+++ b/src/components/download/downloadButtons.js
@@ -10,6 +10,7 @@ import { getNumSelectedTips } from "../../util/treeVisibilityHelpers";
 const RectangularTreeIcon = withTheme(icons.RectangularTree);
 const PanelsGridIcon = withTheme(icons.PanelsGrid);
 const MetaIcon = withTheme(icons.Meta);
+const DatasetIcon = withTheme(icons.Dataset);
 const iconWidth = 25;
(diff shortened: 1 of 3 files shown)


joverlee521 (commented):
The user-experience when download multiple datasets, i.e. tangletrees, is not ideal with multiple download popups. Consider using something like [JSZip](https://stuk.github.io/jszip/) to zip the multiple files? This should also make it easier to include sidecar files.

jameshadfield:
> The user-experience when download multiple datasets, i.e. tangletrees, is not ideal with multiple download popups. Consider using something like [JSZip](https://stuk.github.io/jszip/) to zip the multiple files? This should also make it easier to include sidecar files.

The multiple download popups aren't ideal. But it's not clear that zipped bundles are more user friendly really. I'm going to leave this for now - downloading JSONs for tangletrees will be a rare occurrence I think - but let's revisit it when/if we allow sidecar downloading.

PR #1776, opened 2024-06-05
Nextclade linkout

jameshadfield:
Most discussion about this functionality has been happening within the nextclade repo, see <https://github.com/nextstrain/nextclade/pull/1455> and <https://github.com/nextstrain/nextclade/pull/1460> for a good summary.

**Testing**

via nextstrain.org review app, e.g. https://nextstrain-s-nextstrain-ytpxv6.herokuapp.com/avian-flu/h5n1-cattle-outbreak/genome then click the button in the footer "view in other platforms"

### DIFF
--- a/src/components/modal/LinkOutModalContents.jsx
+++ b/src/components/modal/LinkOutModalContents.jsx
@@ -189,6 +189,58 @@ function MicrobeTraceLinkOut() {
 }
 
 
+function NextcladeLinkOut() {
+  const displayName = 'nextclade';
+  const {pathname, origin} = clientDetails();
+  const {showTreeToo} = useSelector((state) => state.controls)
+  const {mainTreeNumTips, rootSequence} = useSelector((state) => state.metadata);
+
+  // All datasets which have a root-sequence (either in-line or sidecar) can theoretically work as Nextclade
+  // datasets. See <https://github.com/nextstrain/nextclade/pull/1455> for more thorough discussion here.
+  // Excessively big trees may be problematic (as Nextclade was designed around smaller reference trees), but
+  // exactly what the threshold is isn't known. Here I use a rather ad-hoc tip-count threshold:
+  const largeTreeWarning = mainTreeNumTips > 4000;
+
+  if (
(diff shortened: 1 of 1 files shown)


src/components/modal/LinkOutModalContents.jsx
```diff
@@ -73,103 +73,169 @@ const ButtonContainer = styled.div`
...
+      </ButtonContainer>
+    )
+  }
+
+  const url = `https://clades.nextstrain.org?dataset-json-url=${encodeURIComponent(`${origin}${pathname}`)}`
+  return (
+    <ButtonContainer key={displayName}>
+      <ButtonText href={url} target="_blank" rel="noreferrer noopener">{displayName}</ButtonText>
+      <ButtonDescription>
+        View this data in Nextclade (<a href='https://clades.nextstrain.org' target="_blank" rel="noreferrer noopener">learn more</a>).
```
jameshadfield, line 227: This isn't the right language. It should be something more descriptive like

"Use this tree as a nextclade reference dataset which allows you to add new sequences (via drag-and-drop) and see them placed on the tree. Alignment and QC will use default parameters and it may be better to use a curated Nextclade reference dataset for this pathogen [1]. Visit [clades.nextstrain.org](https://clades.nextstrain.org) to learn more about Nextclade or read [these docs](???) for more details about using Auspice datasets in Nextclade"

Copy edits welcome!

[1] We could examine the (currently unparsed) `extensions` field in the Auspice JSON for any nextclade data

jameshadfield: cc @ivan-aksamentov

ivan-aksamentov: LGTM.

(Perhaps instead of the longer explanation, we could link to Nextclade docs? The nuances should be explained there.)

jameshadfield: Thanks - settled on a terser version of the above description with links out to docs

jameshadfield:
Merging after testing in the nextstrain.org review app

PR #1779, opened 2024-06-05
Add support for Node.js version 22

corneliusroemer:
v22 is the current version, with scheduled end-of-life 2027-04-30¹.

The associated npm version (as of node v22.2.0) is npm v10 (v.10.7.0 for v22.2.0)².

¹ https://github.com/nodejs/release#release-schedule
² https://nodejs.org/dist/index.json

Commit adapted from #1673

## Checklist

- [x] Checks pass


corneliusroemer:
We can use the .nvmrc as single source of truth for CI versions of node as
well! Then we only need to change one string rather than one each step that
installs node.

It is not something we need to keep religiously up to date. It can be
bumped whenever we update versions, say once a year.

I tested with 22 and it worked locally with `npm run dev`

Also, the tests passed apparently.

.nvmrc
victorlin, line 1: > We can use the .nvmrc as single source of truth for CI versions of node as well

How would this work if CI needs to test a matrix but only one version is set in this file?

The single source of truth should be package.json, since that's what `npm` checks when running scripts.

https://github.com/nextstrain/auspice/blob/d6fd0aa7ff428f46a3844c7bb3d88901cf4d4321/package.json#L10

.nvmrc
```diff
@@ -0,0 +1 @@
+20
```
victorlin, line 1: The addition of `.nvmrc` made sense for nextstrain.org since we only support a single version of Node.js there, and `.nvmrc` serves to define that single version for devs who use nvm. However, Auspice supports a range of Node.js versions so there's no reason to constrain to a single version.

victorlin (changes requested):

.github/workflows/ci.yaml
```diff
@@ -132,9 +132,9 @@ jobs:
     needs: [publish]
     runs-on: ubuntu-latest
     steps:
-    - run: gh workflow run ci.yml --repo nextstrain/docker-base
-      env:
-        GITHUB_TOKEN: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH }}
+      - run: gh workflow run ci.yml --repo nextstrain/docker-base
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH }}
```
tsibley, line 137: These unrelated formatting changes should not be part of this commit.

jameshadfield:
I fixed up some unrelated changes introduced here and removed the `.nvmrc` (you're welcome to submit another PR with this change; while I don't use it I'm happy to have it in the codebase if it improves the developer experience for others).

Our CI is passing, including the build-docs step. Read the docs' [automated build is failing](https://readthedocs.org/projects/nextstrain-auspice/builds/24712371/), which is also seen on other PRs, but I don't think it's related to these changes. Merging.

victorlin (commented):
✅

PR #1784, opened 2024-06-10
Increase time limit for immutability checks

jameshadfield:
The default resulted in very noisy console logging.

Docs: <https://redux-toolkit.js.org/api/serializabilityMiddleware>

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -21,6 +21,7 @@ const store = configureStore({
       immutableCheck: process.env.SKIP_REDUX_CHECKS
         ? false
         : {
+          warnAfter: 100, // ms. Default is 32.
           ignoredPaths: [
             'tree.nodes',
             'tree.vaccines',


jameshadfield:
Merging now, but thoughts welcome post-merge (and I'm happy to back this out if need be)

src/store.ts
```diff
@@ -21,6 +21,7 @@ const store = configureStore({
       immutableCheck: process.env.SKIP_REDUX_CHECKS
         ? false
         : {
+          warnAfter: 100, // ms. Default is 32.
```
victorlin, line 24: Had to look back at a003a322c7d09a9fd4f1bdb169733af3019f4aa1 for context here. Change is good, just noting that while it cleans up the console for smaller trees such as `/flu/seasonal/h3n2/na/6y`, the middleware still takes ~400ms on `/tb/global` on my computer.

Actually, `/tb/global` in dev mode is noticeably slow on my computer. I tried setting `immutableCheck: false` but not sure if it does anything (would need to play around more with profiling in console dev tools which I'm not very familiar with)

victorlin (commented):
✅

src/store.ts
jameshadfield: Yeah I wanted to keep the warning so that we can remember to turn off the check for big trees (when it really really slows things down).

PR #1785, opened 2024-06-10
Tangletree genotype colouring

### LINKED ISSUE #1773: [tanglegrams] genotype is only computed for LHS tree
When looking at genotype colouring (dynamically generated by Auspice) the RHS tree is not computing / displaying genotype. This information is available in the RHS tree (i.e. look at the tree on its own and everything works as expected).

Perhaps I never enabled this due to the potential for mismatched coordinate systems? If the annotations block is identical in each tree we should allow it as it's super useful.

image

jameshadfield:
See commit messages for details.

Closes #1773 

- [ ] Checks pass

### DIFF
--- a/src/util/setGenotype.js
+++ b/src/util/setGenotype.js
@@ -67,15 +67,20 @@ export const setGenotype = (nodes, prot, positions, rootSequence) => {
   // console.log(`set ${ancNodes.length} nodes to the ancestral state: ${ancState}`)
 };
 
-export const orderOfGenotypeAppearance = (nodes, aaGenotype) => {
+export const orderOfGenotypeAppearance = (nodes, nodesSecondTree, aaGenotype) => {
   const seen = {};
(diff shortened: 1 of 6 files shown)


src/util/setGenotype.js
```diff
@@ -95,5 +100,5 @@ export const orderOfGenotypeAppearance = (nodes, aaGenotype) => {
   if (ordered.indexOf("X") !== -1) {
     orderedBases.push("X");
   }
-  return orderedBases;
-};
+  return orderedBases
+}
```
genehack, line 104: probably want to keep the `;` on the `return` expression?

genehack (approved):

genehack (approved):

src/util/setGenotype.js
jameshadfield: Thanks - fixed up

PR #1787, opened 2024-06-12
Hide all parts of footer with `?onlyPanels` query

jameshadfield:
The "data updated...", "download data", "powered by" content is actually rendered by the `<FinePrint>` component not the `<Footer>`, but it is conceptually part of the footer and should be hidden when using the `onlyPanels` query.

- [ ] Checks pass


jameshadfield:
Merging sans review due to the limited size of these changes

PR #1790, opened 2024-06-12
Streamline display with onlyPanels

trvrb:
Following up on #1787.

When `controls.showOnlyPanels` is active via the URL param `&onlyPanels` switch each primary panel to be an "infoPanel". This removes the title and the line above the title in the display card giving the panel a more streamlined appearance. This results in a more pleasant embedding into slide decks.

Here's example display in the context of a Reveal.js slidedeck.

#### Current
Screenshot 2024-06-11 at 8 55 26 PM

#### With PR
Screenshot 2024-06-11 at 8 53 32 PM

#### Current
Screenshot 2024-06-11 at 8 55 06 PM

#### With PR
Screenshot 2024-06-11 at 8 54 33 PM

Note that I couldn't figure out how to do the equivalent for measurements panel due to it not using the same `@connect((state)` pattern that's used by tree, map, entropy and frequencies.

### DIFF
--- a/src/components/entropy/index.js
+++ b/src/components/entropy/index.js
@@ -79,7 +79,8 @@ const getStyles = (width) => {
     zoomMax: state.controls.zoomMax,
     defaultColorBy: state.controls.defaults.colorBy,
     panelLayout: state.controls.panelLayout,
-    narrativeMode: state.narrative.display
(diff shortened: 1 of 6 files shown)


jameshadfield (approved):
Very nice. I added a commit to mirror this for measurements as well:
image

PR #1794, opened 2024-06-25
Fix filter badge display for single trees

jameshadfield:
Due to a truthy default value in redux state we incorrectly displayed the 2-tree display when viewing a single tree.

It's a one-line bugfix, so I plan to merge and review as soon as CI passes


jameshadfield:
(Our CI passed, red-cross [is read-the-docs failure](https://github.com/nextstrain/docs.nextstrain.org/issues/209))

PR #1796, opened 2024-06-28
Convey uncertainty via tip colors

jameshadfield:
The previous code conveyed uncertainty in node attrs for _branches_ by making them appear grey-er, but we never implemented this for _tips_; most likely because we never had a dataset with such data when this was built.

Here we use the same approach for tips as for branches, but with a slightly different parameterisation of the interpolation. The mapping of the entropy value into `[0,1]` (`tipOpacityFunction`) was chosen so that tips with no (or very little) uncertainty look unchanged from previous Auspice versions, and uncertainty makes them appear more similar to the branch colour (for an equivalent uncertainty).

There should be no visible changes for views without any uncertainty (genotype is a good one to use to test this), as well as traits where there is uncertainty in the dataset but not in the tips (e.g. ebola country / division reconstruction). Here's a side-by-side with the h5n1-cattle-flu dataset from https://github.com/nextstrain/avian-flu/pull/66, which identified this issue in Auspice (this PR LHS, current Auspice RHS):

Frame 12(1)

- [x] Checks pass _(CI passes, RTD is currently broken however that's unrelated to this PR)_

### DIFF
--- a/src/util/colorHelpers.js
+++ b/src/util/colorHelpers.js
@@ -68,11 +68,14 @@ export const calcNodeColor = (tree, colorScale) => {
 // scale entropy such that higher entropy maps to a grayer less-certain branch
 const branchInterpolateColour = "#BBB";
 const branchOpacityConstant = 0.6;
-export const branchOpacityFunction = scalePow()
+const branchOpacityFunction = scalePow()
   .exponent([0.6])
-  .domain([0, 2.0])
-  .range([0.4, 1])
+  .domain([0, 2.0]) // entropy values close to 0 -> ~100% confidence, close to 2 -> very little confidence
+  .range([0.4, 1])  // 0 -> return original node colour, 1 -> return branchInterpolateColour
   .clamp(true);
+const tipOpacityFunction = branchOpacityFunction
+  .copy()
+  .range([0, 0.9]); // if entropy close to 0 return the original node color
 
 
 // entropy calculation precomputed in augur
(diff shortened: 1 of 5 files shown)


src/util/colorHelpers.js
```diff
@@ -68,35 +68,54 @@ export const calcNodeColor = (tree, colorScale) => {
...
+const branchOpacityFunction = scalePow()
   .exponent([0.6])
-  .domain([0, 2.0])
-  .range([0.4, 1])
+  .domain([0, 2.0]) // entropy values close to 0 -> ~100% confidence, close to 2 -> very little confidence
+  .range([0.4, 1])  // 0 -> return original node colour, 1 -> return branchInterpolateColour
   .clamp(true);
+const tipOpacityFunction = branchOpacityFunction
+  .copy()
+  .range([0, 0.9]); // if entropy close to 0 return the original node color
```
jameshadfield, line 78: The range (and the domain) of this scale are the magic values which control how entropy values affect the stroke colour (via an interpolation between the original colour & grey). The tip fill colour is a brighter version of the stroke colour. Very happy for adjustments here, although I think it's important that `tipOpacityFn(close-to-zero) -> 0` so that we don't change how the majority of datasets appear.

jameshadfield:
Some URLs to compare this PR on nextstrain.org vs released Auspice on nextstrain.org:

Cattle-flu [new](https://nextstrain-s-nextstrain-lgk81r.herokuapp.com/avian-flu/h5n1-cattle-outbreak/genome?c=division) & [old](https://nextstrain.org/avian-flu/h5n1-cattle-outbreak/genome?c=division)

Zika (country) [new](https://nextstrain-s-nextstrain-lgk81r.herokuapp.com/zika) & [old](https://nextstrain.org/zika). Note that this does have uncertainty for tips, which is kind of strange, but that's how `augur traits` currently works.

H3N2 (genotype view) [new](https://nextstrain-s-nextstrain-lgk81r.herokuapp.com/seasonal-flu/h3n2/ha/2y?c=gt-HA1_53&gmax=1052&gmin=66) & [old](https://nextstrain.org/seasonal-flu/h3n2/ha/2y?c=gt-HA1_53&gmax=1052&gmin=66) - no uncertainty here.

src/util/colorHelpers.js
```diff
@@ -68,35 +68,54 @@ export const calcNodeColor = (tree, colorScale) => {
...
+ * depending on the confidence we have in the trait (the entropy), with more confidence resulting
+ * in more saturated colours. For branches we always make them slightly greyer (even in the absence
+ * of uncertainty) for purely aesthetic reasons.
  * @param {obj} tree phyloTree object
+ * @param {bool} branch will this color be used for the branch or the tip?
  * @param {bool} confidence enabled?
  * @return {array} array of hex's. 1-1 with nodes.
  */
-export const calcBranchStrokeCols = (tree, confidence, colorBy) => {
+export const calculateStrokeColors = (tree, branch, confidence, colorBy) => {
```
joverlee521, line 97: _not a request for change, just curious_

Why combine branch/tip colors into one function with the `branch` flag when they are essentially completely different paths within the function?

joverlee521 (commented):
How would one be able to differentiate the grey scale for uncertainty vs grey scale for unprovided colorings? For example, imagine if [zika's region](https://next.nextstrain.org/zika@2024-06-17?c=region) had uncertainty, it would be mixed in with the "Asia" grey colorings. 

(The default "Asia" coloring issue will be fixed when augur is released with https://github.com/nextstrain/augur/pull/1490, but the question still stands for any other unprovided colorings)

trvrb:
Awesome! This behavior looks spot on to me. Here's the current H5N1 cattle outbreak

Screenshot 2024-06-28 at 11 42 31 AM

Note that recent SRA tips are not completely gray. For example, the top clade descends from viruses sampled from South Dakota. These are appropriately colored a gray/green indicating potential South Dakota, but with little certainty.

Screenshot 2024-06-28 at 11 45 19 AM

The interpolation between SRA tips close to known Ohio viruses in blue to the the Michigan human case in lime also seems very appropriate.

Screenshot 2024-06-28 at 11 47 02 AM

I think Auspice is now doing exactly what it should be doing. However, we still should have a way to have a more data-informed decision about how to set `--sampling-bias-correction`. We could be doing leave 10% out cross validation as Gytis did in the [2019 BMC Evol Biol paper](https://bedford.io/papers/dudas-genomic-horizon/).

trvrb:
>How would one be able to differentiate the grey scale for uncertainty vs grey scale for unprovided colorings? For example, imagine if [zika's region](https://next.nextstrain.org/zika@2024-06-17?c=region) had uncertainty, it would be mixed in with the "Asia" grey colorings.

This is a really good point @joverlee521. The issue is that currently we use gray to mean either:
1. Unknown or uncertain
2. Uninteresting

This uninteresting take can be seen here https://nextstrain.org/ncov/gisaid/north-america/6m@2020-05-01 for example. This felt semantically appropriate to distinguish focal samples from background samples.

I _think_ this is okay however... This example does DTA on samples with a focal vs contextual color ramp so that uncertain nodes and contextual nodes are both gray. This feels okay and appropriate (perhaps not ideal, but not broken). It highlights clades that are more certain to be in a focal region.

That said, we should be fixing colorings like the Zika example so that random location is not gray. In the Zika example, "Asia" should be blue, like it is for country.

jameshadfield:
> How would one be able to differentiate the grey scale for uncertainty vs grey scale for unprovided colorings? For example, imagine if [zika's region](https://next.nextstrain.org/zika@2024-06-17?c=region) had uncertainty, it would be mixed in with the "Asia" grey colorings.

Very difficult at the moment! Let's continue discussion in [[maybe] differentiate between nodes with uncertainty vs nodes missing from colour scale](https://github.com/nextstrain/auspice/issues/1797)

src/util/colorHelpers.js
jameshadfield: While they are ~separate code paths, they both do the same thing: take a colour and modify it according to the node's uncertainty. Co-locating them feels natural to me and should help them to stay in-sync. 

I actually thought about taking this further and refactoring it into a function `calculateNodeColors -> {branchColors, tipStrokeColors, tipFillColors}` so we calculate everything at once, but I'll leave that for another day (and I need to check that we never have a situation where we only need to recompute one of those sets).

Issue #1758, opened 2024-03-05
measurements panel does not honor the value of `raw` in the `measurements_display` field of the measurements JSON


huddlej:
## Current behavior

The `augur measurements export` command provides a `--measurements-display` argument that accepts value of "mean" or "raw". Setting the value to `raw` produces a measurements JSON with a display defaults entry of `"measurements_display": "raw"`. When loading Auspice JSONs with a measurements panel with that value, Auspice always displays the mean +/- std dev instead of the raw values.

## Expected behavior

Auspice should honor the value of the `measurements_display` entry in the default display of the measurements JSON.

## How to reproduce

Using the same dataset from #1757, I manually generated a measurements panel with the "raw" display default like so:

```
augur measurements export \
  --collection results/h3n2/antigenic_distances_between_strains.tsv \
  --collection-config results/h3n2/measurements_collection_config.json \
  --strain-column test_strain \
  --value-column log2_titer \
  --key h3n2_ha_cell_hi \
  --title 'H3N2 cell-passaged HI measurements' \
  --x-axis-label 'normalized log2 titer distance from reference' \
  --threshold 0.0 2.0 \
  --filters reference_strain clade_reference source serum \
  --show-threshold \
  --hide-overall-mean \
  --minify-json \
  --measurements-display raw \
  --output-json auspice/measurements-panel_flu_seasonal_h3n2_ha_measurements.json
```

The corresponding Auspice JSON and measurements JSON files for this build are in the attached zip file here: 
[measurements_auspice.zip](https://github.com/nextstrain/auspice/files/14489010/measurements_auspice.zip)

Unzip these files and drag them on to https://auspice.us to confirm that the display is the mean value.

Issue #1757, opened 2024-03-04
 measurements panel crashes on load in Auspice when no threshold is provided during `augur measurements export`


huddlej:
## Current Behavior

When loading a measurements panel JSON that did not have any thresholds provided through the `--thresholds` argument to `augur measurements export`, Auspice starts to load the panel and then crashes. This behavior occurs even when the user explicitly disables the threshold display by default with the `--hide-threshold` argument. The measurements JSON validates with augur despite this behavior in Auspice.

## Expected behavior

Auspice should load the measurements JSON even when a threshold isn't defined.

## How to reproduce

Reproducing this error requires building a minimal tree and measurements dataset. I ended up using [the public data from the measurements panel paper](https://github.com/blab/measurements-panel/tree/main/data) as a starting point. The sequences and metadata are attached here.

Clone the measurements panel repo and change into the directory:

```
git clone https://github.com/blab/measurements-panel.git
cd measurements-panel
```

Download sequences and metadata attached here into the `data/` subdirectory ([h3n2_data.zip](https://github.com/nextstrain/auspice/files/14488052/h3n2_data.zip)).

Run the workflow to build the Auspice JSONs with the latest Nextstrain CLI and image:

```
nextstrain build . \
  auspice/measurements-panel_flu_seasonal_h3n2_ha.json \
  auspice/measurements-panel_flu_seasonal_h3n2_ha_measurements.json \
  --forceall
```

Confirm that the Auspice JSONs work as expected by dragging them on to https://auspice.us/.

Run the `augur measurements export` command without the `--thresholds` argument and with a `--hide-threshold` instead of `--show-threshold`:

```
nextstrain shell .

augur measurements export \
  --collection results/h3n2/antigenic_distances_between_strains.tsv \
  --collection-config results/h3n2/measurements_collection_config.json \
  --strain-column test_strain \
  --value-column log2_titer \
  --key h3n2_ha_cell_hi \
  --title 'H3N2 cell-passaged HI measurements' \
  --x-axis-label 'normalized log2 titer distance from reference' \
  --filters reference_strain clade_reference source serum \
  --hide-threshold \
  --hide-overall-mean \
  --minify-json \
  --output-json auspice/measurements-panel_flu_seasonal_h3n2_ha_measurements.json
```

Drag the Auspice JSONs on to auspice.us again and confirm that it crashes with the console error of `TypeError: t is undefined`.

PR #1802, opened 2024-07-29
Fix measurements bugs

joverlee521:

Small fixes for bugs in the measurements panel.

Resolves https://github.com/nextstrain/auspice/issues/1757
Resolves https://github.com/nextstrain/auspice/issues/1758

## Checklist

- [x] Checks pass

### DIFF
--- a/src/components/measurements/measurementsD3.js
+++ b/src/components/measurements/measurementsD3.js
@@ -197,8 +197,9 @@ export const drawMeasurementsSVG = (ref, xAxisRef, svgData) => {
   drawStickyXAxis(xAxisRef, containerHeight, svgHeight, xScale, x_axis_label);
 
   // Add threshold(s) if provided
-  if (thresholds !== null && thresholds !== undefined) {
+  if (Array.isArray(thresholds)) {
     for (const threshold of thresholds) {
+      if (typeof threshold !== "number") continue;
       const thresholdXValue = xScale(threshold);
       svg.append("line")
         .attr("class", classes.threshold)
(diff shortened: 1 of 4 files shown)


src/components/measurements/measurementsD3.js
```diff
@@ -197,7 +197,7 @@ export const drawMeasurementsSVG = (ref, xAxisRef, svgData) => {
   drawStickyXAxis(xAxisRef, containerHeight, svgHeight, xScale, x_axis_label);
 
   // Add threshold(s) if provided
-  if (thresholds !== null) {
+  if (thresholds !== null && thresholds !== undefined) {
     for (const threshold of thresholds) {
```
genehack, line 201: _Overthinking nit; non-blocking_

Since you're making this change, I wonder if it's worth adding an explicit check for `thresholds` being iterable? If, somehow, `thresholds === 3`, L201 is going to throw `Uncaught TypeError: threasholds is not iterable`.

genehack (approved):

src/components/measurements/measurementsD3.js
joverlee521: Fair! Better to just check that `thresholds` is an array.

joverlee521: Updated in https://github.com/nextstrain/auspice/pull/1802/commits/172c55622a12e2838c0f106f2280f17be399791a

joverlee521:
Thank you @huddlej for the detailed steps to reproduce bug in https://github.com/nextstrain/auspice/issues/1758 and https://github.com/nextstrain/auspice/issues/1757. 

I was able to verify both issues are fixed in the downstream [auspice.us test app](https://auspice-us-nextstrain-b-xyf7zq.herokuapp.com/).

huddlej:
Awesome! Thank you, @joverlee521!

Issue #1803, opened 2024-07-30
Notify when no datasets are available for download


joverlee521:
Prompted by https://github.com/nextstrain/auspice.us/issues/93

The download Auspice JSON feature depends on fetching the original JSON by parsing the pathname. 
It silently skips invalid datasets so users are not notified if the dataset is not available:

https://github.com/nextstrain/auspice/blob/835fbb23d61549da81fde3725b031ad8a24c74cf/src/components/download/helperFunctions.js#L634-L635

### Possible solutions

1. Dispatch error notification that the dataset is not available for download if all dataset names are invalid. 
2. Hide the download Auspice JSON button if we cannot parse the dataset name from the URL pathname.

PR #1804, opened 2024-07-31
Only support downloading Auspice JSONs if dataset name can be parsed

joverlee521:

Only supports downloading Auspice JSON if the dataset name can be parsed from the URL pathname. 
This mainly hides the download Auspice JSON option in auspice.us. 

## Related issue(s)

Resolves https://github.com/nextstrain/auspice/issues/1803

## Checklist

- [x] Checks pass

### DIFF
--- a/src/components/download/downloadButtons.js
+++ b/src/components/download/downloadButtons.js
@@ -6,6 +6,7 @@ import { materialButton } from "../../globalStyles";
 import * as helpers from "./helperFunctions";
 import { getFullAuthorInfoFromNode } from "../../util/treeMiscHelpers";
 import { getNumSelectedTips } from "../../util/treeVisibilityHelpers";
+import { getDatasetNamesFromUrl } from "../../actions/loadData";
 
 const RectangularTreeIcon = withTheme(icons.RectangularTree);
 const PanelsGridIcon = withTheme(icons.PanelsGrid);
@@ -35,6 +36,10 @@ export const DownloadButtons = ({dispatch, t, tree, entropy, metadata, colorBy, 
       }
     }
   }
+  /* Verify that we can parse dataset name from URL to support Auspice JSON download */
+  const datasetNames = getDatasetNamesFromUrl(window.location.pathname);
+  const supportAuspiceJsonDownload = !gisaidProvenance && datasetNames.some(Boolean);
+
   const entropyBar = entropy?.selectedCds===nucleotide_gene ? "nucleotide" : "codon";
 
   return (
@@ -48,12 +53,12 @@ export const DownloadButtons = ({dispatch, t, tree, entropy, metadata, colorBy, 
         <p/>
         {partialData ? `Currently ${selectedTipsCount}/${totalTipCount} tips are displayed and will be downloaded.` : `Currently the entire dataset (${totalTipCount} tips) will be downloaded.`}
       </div>
-      {!gisaidProvenance && (
+      {supportAuspiceJsonDownload && (
         <Button
           name="Auspice (Nextstrain) JSON"
           description={`The main Auspice dataset JSON(s) for the current view`}
           icon={<DatasetIcon width={iconWidth} selected />}
-          onClick={() => helpers.auspiceJSON(dispatch)}
+          onClick={() => helpers.auspiceJSON(dispatch, datasetNames)}
         />
       )}
       <Button
--- a/src/components/download/helperFunctions.js
+++ b/src/components/download/helperFunctions.js
@@ -8,7 +8,7 @@ import { NODE_VISIBLE, nucleotide_gene } from "../../util/globals";
 import { datasetSummary } from "../info/datasetSummary";
 import { isColorByGenotype } from "../../util/getGenotype";
 import { EmptyNewickTreeCreated } from "../../util/exceptions";
-import { getDatasetNamesFromUrl, Dataset } from "../../actions/loadData";
+import { Dataset } from "../../actions/loadData";
 
 export const isPaperURLValid = (d) => {
   return (
@@ -623,15 +623,15 @@ export const entropyTSV = (dispatch, filePrefix, entropy) => {
 /**
  * Write out Auspice JSON(s) for the current view. We do this by re-fetching the original
  * JSON because we don't keep a copy of the unprocessed data around.
- * 
+ *
  * Sidecar files are not fetched, but we can also download them if desired.
- * 
+ *
  * Note that we are not viewing a narrative, as the download button functionality is disabled
  * for narratives.
  */
-export const auspiceJSON = (dispatch) => {
+export const auspiceJSON = (dispatch, datasetNames) => {
   const filenames = [];
-  for (const datasetName of getDatasetNamesFromUrl(window.location.pathname)) {
+  for (const datasetName of datasetNames) {
     if (!datasetName) continue; // e.g. no 2nd tree
     const filename = datasetName.replace('/', '_') + '.json';
     filenames.push(filename);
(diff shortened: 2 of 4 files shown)


src/components/download/helperFunctions.js
```diff
@@ -623,15 +623,19 @@ export const entropyTSV = (dispatch, filePrefix, entropy) => {
...
+ *
  * Note that we are not viewing a narrative, as the download button functionality is disabled
  * for narratives.
  */
-export const auspiceJSON = (dispatch) => {
+export const auspiceJSON = (dispatch, datasetNames) => {
   const filenames = [];
-  for (const datasetName of getDatasetNamesFromUrl(window.location.pathname)) {
+  if (datasetNames.every((x) => x === '')) {
+    console.error("Unable to fetch empty dataset names");
```
genehack, line 635: I'm totally unfamiliar with this code, but I wonder if logging the value of `dispatch` here would be helpful in figuring out what's going on, if this error is ever triggered? (I realize it's a "this should never happen" category thing.)
 
```suggestion
    console.error(`Unable to fetch empty dataset names for ${dispatch}`);
```

genehack (approved):

joverlee521:
Hmm, the downstream [auspice.us test app](https://auspice-us-nextstrain-b-8sgcjz.herokuapp.com/) still shows the download Auspice JSON option...I'll dig into this.

src/components/download/helperFunctions.js
joverlee521: Ah, `dispatch` is the [Redux dispatch](https://redux.js.org/api/store#dispatchaction) for dispatching the error notification below so it would not be meaningful here.

joverlee521: I ended up needing to see what the `datasetNames` value was in my debugging, so I'm logging out `${JSON.stringify(datasetNames)}`  in https://github.com/nextstrain/auspice/pull/1804/commits/45bd85d4cca351ab5447c02d507b8dba69cdce05

src/components/download/helperFunctions.js
```diff
@@ -623,15 +623,19 @@ export const entropyTSV = (dispatch, filePrefix, entropy) => {
...
  * Note that we are not viewing a narrative, as the download button functionality is disabled
  * for narratives.
  */
-export const auspiceJSON = (dispatch) => {
+export const auspiceJSON = (dispatch, datasetNames) => {
   const filenames = [];
-  for (const datasetName of getDatasetNamesFromUrl(window.location.pathname)) {
+  if (!datasetNames.every(Boolean)) {
+    console.error(`Unable to fetch empty dataset names: ${JSON.stringify(datasetNames)}`);
+    return dispatch(errorNotification({message: "Unable to download Auspice JSON (see console for more info)"}))
```
jameshadfield, line 636: `datasetNames` is always an array of length 2. So for `/zika` it's `[ "zika", undefined ]`, and thus we end up showing a error notification and not downloading the JSON. 

Perhaps the intention here was `.all(Boolean)`? But that's unnecessary because of the conditions under which `auspiceJSON()` gets called.

src/components/download/downloadButtons.js
```diff
@@ -35,6 +36,10 @@ export const DownloadButtons = ({dispatch, t, tree, entropy, metadata, colorBy,
       }
     }
   }
+  /* Verify that we can parse dataset name from URL to support Auspice JSON download */
+  const datasetNames = getDatasetNamesFromUrl(window.location.pathname);
+  const supportAuspiceJsonDownload = !gisaidProvenance && datasetNames.some(Boolean);
```
jameshadfield, line 41: There's a bug in auspice.us (?) where dragging on two trees results in them being added to the URL. As such, `datasetNames` contains 2 truthy values, and we allow the download, which then fails.

Using `dev_multi_h3n2.json` and `ebola.json` (any two JSONs will do) we get a URL of `https://auspice-us-nextstrain-b-q2kkxc.herokuapp.com/dev/multi/h3n2:ebola`, and the following error when trying to download:

image

jameshadfield (changes requested):

src/components/download/helperFunctions.js
joverlee521: Ah, thanks for catching that. I meant `.some(Boolean)` to verify at least one of them is truthy. 

> But that's unnecessary because of the conditions under which auspiceJSON() gets called.

Yeah, I added this guard to get error notifications on unexpected use of `auspiceJSON()` instead of just letting it hang on the "Preparing Auspice JSON..." notification.

src/components/download/downloadButtons.js
joverlee521: Oh, I did not realize that the URL gets updated in auspice.us when there's a two trees. 
Looks like it's coming from the `changeURLMiddleware` in auspice:

https://github.com/nextstrain/auspice/blob/69557289fe7b4a51c4d529c97ee02fea1cafa6a9/src/middleware/changeURL.js#L236-L243

I assume the URL should never be updated in auspice.us if you are calling this a bug? How is that configured?

joverlee521: Hmm, maybe we just shouldn't rely on the URL parsing here to determine if download is supported. We can add a `disableAuspiceJsonDownload` config param to the extensions and have auspice.us explicitly disable the downloads.

src/components/download/helperFunctions.js
joverlee521: Fixed up in [force-push](https://github.com/nextstrain/auspice/compare/45bd85d4cca351ab5447c02d507b8dba69cdce05..dbfce230392927f83a921c29e8447287d4133187)

src/components/download/downloadButtons.js
jameshadfield: > I assume the URL should never be updated in auspice.us if you are calling this a bug? How is that configured?

auspice.us, once it's parsed the dropped files, [initialises the Auspice viz via:](https://github.com/nextstrain/auspice.us/blob/96c8d5486f2852f2fa66aa995ee30a637efc5230/auspice_client_customisation/handleDroppedFiles.js#L200-L205)

```js
  /* load (into redux state) the main dataset(s) */
  try {
    dispatch({
      type: "CLEAN_START",
      pathnameShouldBe: "",
      ...createStateFromQueryOrJSONs({
```

and the intention is that Auspice updates the pathname (URL) to the empty string via:

https://github.com/nextstrain/auspice/blob/69557289fe7b4a51c4d529c97ee02fea1cafa6a9/src/middleware/changeURL.js#L231-L234

and then the switch statement ends. But `""` is falsey, so we keep going in the switch statement and then, when there are two trees, we hit the case you linked above. I think checking if `pathnameShouldBe` is a property on the action rather than just the truthyness of its value is the correct fix.

> Hmm, maybe we just shouldn't rely on the URL parsing here to determine if download is supported. We can add a `disableAuspiceJsonDownload` config param to the extensions and have auspice.us explicitly disable the downloads.

I think once the bugs identified here are fixed the URL parsing will still be appropriate. (The particular bug in this thread is, strictly speaking, separate to this PR, but it'd be good to just fix it here.)

joverlee521: Updated in https://github.com/nextstrain/auspice/pull/1804/commits/a9c461057f9258b88aa9e8882f1664cb2238b7e6.

joverlee521: Confirmed in [auspice.us test app](https://auspice-us-nextstrain-b-q2kkxc.herokuapp.com/) that the URL does not update with multiple trees.

src/components/download/helperFunctions.js
jameshadfield: > Ah, thanks for catching that. I meant `.some(Boolean)` to verify at least one of them is truthy.

Nice one. I don't know what I meant by suggesting `.all`, which isn't even an array prototype method!

jameshadfield (approved):
Thanks @joverlee521!

PR #1827, opened 2024-08-16
Limit x-axis by filtered measurements

### LINKED ISSUE #1814: Limit x-axis by visible measurements
@rneher originally noted:

> In the measurement panel, some limits on the axis might be useful as well. We currently have -6 to 14 in some views (which is a million fold titer difference)

The following example from [the measurements panel paper dataset](https://nextstrain.org/community/blab/measurements-panel/flu/seasonal/h3n2/ha?p=grid) shows how the x-axis retains its original domain even after filtering to a single reference with a smaller range of measurement values. In this example, the x-axis domain should reflect the minimum and maximum std dev values at the lower and upper bounds.

image

Similarly, the raw measurements display for the same data should use the minimum and maximum measurement value plus/minus some small padding on the lower and upper bounds.

image

joverlee521:
_[preview of measurements panel paper dataset](https://nextstrain-s-nextstrain-bvu2mb.herokuapp.com/community/blab/measurements-panel/flu/seasonal/h3n2/ha?p=grid)_

Resolves #1814 

### Raw measurements

Before:

Screenshot 2024-08-16 at 1 34 53 PM

After:

Screenshot 2024-08-16 at 1 34 58 PM

### Mean ± SD 

It's not perfect for the mean ± SD view because the xScale is based on the underlying raw data.

Before:

Screenshot 2024-08-16 at 1 32 17 PM

After:

Screenshot 2024-08-16 at 1 32 24 PM

## Checklist

- [x] Checks pass

### DIFF
--- a/src/components/measurements/measurementsD3.js
+++ b/src/components/measurements/measurementsD3.js
@@ -386,8 +386,13 @@ export const drawMeansForColorBy = (ref, svgData, treeStrainColors, legendValues
     const ySpacing = (layout.subplotHeight - 4 * layout.subplotPadding) / (numberOfColorByAttributes - 1);
     let yValue = layout.subplotPadding;
     // Order the color groups by the legend value order so that we have a stable order for the means
-    legendValues
-      .filter((attribute) => String(attribute) in colorByGroups)
+    const orderedColorGroups = orderBy(
+      Object.keys(colorByGroups),
+      (key) => legendValues.indexOf(key),
+      "asc"
+    );
+
+    orderedColorGroups
       .forEach((attribute) => {
         const {color, values} = colorByGroups[attribute];
         drawMeanAndStandardDeviation(
--- a/src/components/measurements/measurementsD3.js
+++ b/src/components/measurements/measurementsD3.js
@@ -58,12 +58,14 @@ const getSubplotDOMId = (groupingValueIndex) => `measurement_subplot_${groupingV
 /**
  * Creates the D3 linear scale for the x-axis with the provided measurements'
  * values as the domain and the panelWidth with hard-coded padding values as
- * the range. Expected to be shared across all subplots.
+ * the range. The optional paddingProportion can be provided to include additional
+ * padding for the domain. Expected to be shared across all subplots.
  * @param {number} panelWidth
  * @param {Array<Object>} measurements
+ * @param {number} [paddingProportion=0.1]
  * @returns {function}
  */
-export const createXScale = (panelWidth, measurements) => {
+export const createXScale = (panelWidth, measurements, paddingProportion = 0.1) => {
   // Padding the xScale based on proportion
   // Copied from https://github.com/d3/d3-scale/issues/150#issuecomment-561304239
   function padLinear([x0, x1], k) {
@@ -73,7 +75,7 @@ export const createXScale = (panelWidth, measurements) => {
 
   return (
     scaleLinear()
-      .domain(padLinear(extent(measurements, (m) => m.value), 0.1))
+      .domain(padLinear(extent(measurements, (m) => m.value), paddingProportion))
       .range([layout.leftPadding, panelWidth - layout.rightPadding])
       .nice()
   );
--- a/src/components/measurements/measurementsD3.js
+++ b/src/components/measurements/measurementsD3.js
@@ -64,9 +64,16 @@ const getSubplotDOMId = (groupingValueIndex) => `measurement_subplot_${groupingV
  * @returns {function}
  */
 export const createXScale = (panelWidth, measurements) => {
+  // Padding the xScale based on proportion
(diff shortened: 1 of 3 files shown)


src/components/measurements/measurementsD3.js
```diff
@@ -64,9 +64,16 @@ const getSubplotDOMId = (groupingValueIndex) => `measurement_subplot_${groupingV
...
+  // Copied from https://github.com/d3/d3-scale/issues/150#issuecomment-561304239
+  function padLinear([x0, x1], k) {
+    const dx = (x1 - x0) * k / 2;
+    return [x0 - dx, x1 + dx];
+  }
+
   return (
     scaleLinear()
-      .domain(extent(measurements, (m) => m.value))
+      .domain(padLinear(extent(measurements, (m) => m.value), 0.1))
```
huddlej, line 76: This proportion of 0.1 is the kind of parameter I could imagine exposing to users some day through the measurements config JSON, since there could be dataset-specific reasons to prefer a lower or higher value. Maybe for the sake of remembering what this value means here and eventually exposing it later, could we define 0.1 as a variable in this code?

huddlej (commented):
This is awesome, @joverlee521! I actually like how the axis scale doesn't change between the raw and mean/std displays, since it allows me to toggle between those displays without the panel context jumping around.

Playing around with the paper dataset, this view shows the expected domain on the x-axis for the two references:

image

When I dropped the first reference, the x-axis domain didn't update which seemed like a bug here.

image

Then I switched to the raw display and saw that there actually are measurements at the far left of the domain that are plotted in gray as an "undefined" clade and that don't appear in the mean/std view:

image

This seems like a bug with the dataset (that it doesn't define a clade for every reference) and a separate bug with the measurements panel (that it doesn't show a mean/std anyway even when the clade is "undefined"). That's not a bug for this PR, though, (it happens in the production app) and the way you've implemented a fixed x-axis domain for both raw and mean/std views actually reveals the underlying data issue which is great.

I only noticed a couple of edge case issues where the std dev could exceed the lower and upper bounds of the x scale. For example, in [the preview app](https://nextstrain-s-nextstrain-bvu2mb.herokuapp.com/community/blab/measurements-panel/flu/seasonal/h3n2/ha?d=measurements&p=full) if I filtered the source to "nimr-sep-2011-2011-06-24" and the clade reference to "158N/189K", I get the following mean/std view:

image

And this is the raw view:

image

That's a pretty extreme case, though, so maybe it's ok for now?

src/components/measurements/measurementsD3.js
joverlee521: Yeah, that's totally fair! I'll add it as an optional param with a default value for now and we can feed in a user defined value in the future if desired.

joverlee521:
> This seems like a bug with the dataset (that it doesn't define a clade for every reference) and a separate bug with the measurements panel (that it doesn't show a mean/std anyway even when the clade is "undefined"). 

Ah, this is a case where the test strain does not have a defined clade. Looks like they don't get included in the mean display because the panel was adding means for the legend values only. This should be fixed with https://github.com/nextstrain/auspice/pull/1827/commits/b201e5ce28e6f31ce23c26db3a514234f011939a. 

> I only noticed a couple of edge case issues where the std dev could exceed the lower and upper bounds of the x scale. For example, in [the preview app](https://nextstrain-s-nextstrain-bvu2mb.herokuapp.com/community/blab/measurements-panel/flu/seasonal/h3n2/ha?d=measurements&p=full) if I filtered the source to "nimr-sep-2011-2011-06-24" and the clade reference to "158N/189K"

Huh, I would think this is an issue outside of this PR if those were the only two data points the measurements collection. I'll leave it as-is for now since it's such an extreme case.

src/components/measurements/measurementsD3.js
joverlee521: Updated in https://github.com/nextstrain/auspice/pull/1827/commits/c3029397e21784498f9c2b94f8c33d635d98187f

joverlee521:
> > I only noticed a couple of edge case issues where the std dev could exceed the lower and upper bounds of the x scale. For example, in [the preview app](https://nextstrain-s-nextstrain-bvu2mb.herokuapp.com/community/blab/measurements-panel/flu/seasonal/h3n2/ha?d=measurements&p=full) if I filtered the source to "nimr-sep-2011-2011-06-24" and the clade reference to "158N/189K"
> 
> Huh, I would think this is an issue outside of this PR if those were the only two data points the measurements collection. I'll leave it as-is for now since it's such an extreme case. 

Tracking in https://github.com/nextstrain/auspice/issues/1828 as a future TODO. 

If there's no other feedback, I'll merge this tomorrow.

huddlej:
Awesome! Thank you again, @joverlee521!

Issue #1813, opened 2024-08-14
Surface console-based errors in the web UI


genehack:
Extracted from [Q3 planning meeting](https://docs.google.com/document/d/150k8VU2q5IXYU3a2_g4mLLJGCh-A5AJNguppu9dCU-A/edit) notes.

Currently, auspice will log certain warnings and errors to the console. Discoverability of this is obviously pretty poor. To make it these more clear, update the UI to have an indication that errors exist, and have a way to surface them in a popover UI element or similar.

victorlin:
I've updated the issue description with scope and potential solutions. Initial commentary:

> 1. Use the existing [error notification framework](https://github.com/nextstrain/auspice/blob/36ed728d1b75794a760a1a6f8ec199f78ac13419/src/actions/notifications.js#L24-L26) where appropriate.

I haven't closely inspected any of the errors in question, but I wouldn't be surprised if there are some that could benefit from switching to the more visible error notification popup that's already [used elsewhere](https://github.com/search?q=repo%3Anextstrain%2Fauspice+path%3Asrc%2F**+"errorNotification"&type=code).

> 2. Surface presence of console errors through a warning notification.
> 3. Surface presence of console errors through a new UI element.

These are similar, but (2) leverages an already existing framework. (3) would have to go through additional design decisions.

Issue #805, opened 2019-10-04
Remove hardcoded range for LBI


huddlej:
The range for LBI values is [currently hardcoded in auspice](https://github.com/nextstrain/auspice/blob/162824b/src/util/colorScale.js#L197-L198) from [0, 0.7]. Since default LBI values from augur range up to 1.0 and augur will soon support unnormalized LBI values that can exceed one, we should set the `vmin` and `vmax` for any LBI color-by in the auspice config JSON (when that is supported).

PR #1832, opened 2024-08-27
Allow YYYY-MM-DD values in a temporal scales

### LINKED ISSUE #1427: Add temporal colouring scale to use YYYY-MM-DD dates
**Context**  
Datasets often have date information additional to sampling date (e.g. symptom onset date, sequencing date, submission date). We have no good way of displaying / interrogating such data. An additional use case is datasets with sampling date but for which internal nodes have no temporal info. 

**Description**   
Add a "temporal" colour scale type to the [existing 4 types](https://github.com/nextstrain/augur/blob/master/augur/data/schema-export-v2.json#L175-L180). The values must be either floats (e.g. `2020.123`) or YYYY-MM-DD (e.g. `"2021-10-31"`). We may extend this to allow ambiguity, but this can be left for subsequent work. Auspice will interpret such a value the same way as we currently interpret sampling dates, and generate a colour scale appropriately. 

**Related issues**
#1402 will discuss how filtering UI should behave here. Which branches we show as selected is up for debate, but I propose we treat this the same as filtering to traits where we trace back to the MRCA of selected tips, ignoring internal node annotations.

jameshadfield:
Includes a number of improvements to date parsing and validation as well (see commits).

Temporal scales previously could only use numeric dates (2024.123 etc) but can now also use YYYY-MM-DD values including ambiguous values (YYYY-XX-XX, YYYY-MM-XX). For ambiguous values the midpoint of the ambiguous range is used to pick the colour / match the legend swatches to tips.

Example datasets (rabies):

* [YYYY-MM-DD sampling date, most of which are ambiguous, but only for tips (i.e. no timetree)](https://nextstrain-s-nextstrain-waprls.herokuapp.com/staging/rabies/YYYY-MM-DD-coloring?c=date)
* [same dataset with a custom scale provided](https://nextstrain-s-nextstrain-waprls.herokuapp.com/staging/rabies/YYYY-MM-DD-custom-coloring?c=date)

image

Closes #1427

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -249,23 +249,6 @@ const restoreQueryableStateToDefaults = (state) => {
 };
 
 const modifyStateViaMetadata = (state, metadata, genomeMap) => {
-  if (metadata.date_range) {
-    /* this may be useful if, e.g., one were to want to display an outbreak
-    from 2000-2005 (the default is the present day) */
-    if (metadata.date_range.date_min) {
-      state["dateMin"] = metadata.date_range.date_min;
-      state["dateMinNumeric"] = calendarToNumeric(state["dateMin"]);
-      state["absoluteDateMin"] = metadata.date_range.date_min;
-      state["absoluteDateMinNumeric"] = calendarToNumeric(state["absoluteDateMin"]);
-      state["mapAnimationStartDate"] = metadata.date_range.date_min;
-    }
-    if (metadata.date_range.date_max) {
-      state["dateMax"] = metadata.date_range.date_max;
-      state["dateMaxNumeric"] = calendarToNumeric(state["dateMax"]);
-      state["absoluteDateMax"] = metadata.date_range.date_max;
-      state["absoluteDateMaxNumeric"] = calendarToNumeric(state["absoluteDateMax"]);
-    }
-  }
   if (metadata.analysisSlider) {
     state["analysisSlider"] = {key: metadata.analysisSlider, valid: false};
   }
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -92,13 +92,31 @@ const modifyStateViaURLQuery = (state, query) => {
   if (query.tl) {
     state["tipLabelKey"] = query.tl===strainSymbolUrlString ? strainSymbol : query.tl;
   }
+
   if (query.dmin) {
-    state["dateMin"] = query.dmin;
-    state["dateMinNumeric"] = calendarToNumeric(query.dmin);
+    const dateNum = calendarToNumeric(query.dmin);
+    if (_validDate(dateNum, state.absoluteDateMinNumeric, state.absoluteDateMaxNumeric)) {
+      state["dateMin"] = query.dmin;
+      state["dateMinNumeric"] = dateNum;
+    } else {
+      console.error(`URL query "dmin=${query.dmin}" is invalid ${state.branchLengthsToDisplay==='divOnly'?'(the tree is not a timetree)':''}`);
+      delete query.dmin;
+    }
   }
   if (query.dmax) {
-    state["dateMax"] = query.dmax;
-    state["dateMaxNumeric"] = calendarToNumeric(query.dmax);
+    const dateNum = calendarToNumeric(query.dmax);
+    if (_validDate(dateNum, state.absoluteDateMinNumeric, state.absoluteDateMaxNumeric)) {
+      if ((query.dmin && dateNum <= state.dateMinNumeric)) {
+        console.error(`Requested "dmax=${query.dmax}" is earlier than "dmin=${query.dmin}", ignoring dmax.`);
+        delete query.dmax;
+      } else {
+        state["dateMax"] = query.dmax;
+        state["dateMaxNumeric"] = dateNum;
+      }
+    } else {
+      console.error(`URL query "dmax=${query.dmax}" is invalid ${state.branchLengthsToDisplay==='divOnly'?'(the tree is not a timetree)':''}`);
+      delete query.dmax;
+    }
   }
 
   /** Queries 's', 'gt', and 'f_<name>' correspond to active filters */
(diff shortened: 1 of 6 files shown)


src/actions/recomputeReduxState.js
```diff
@@ -92,13 +92,29 @@ const modifyStateViaURLQuery = (state, query) => {
   if (query.tl) {
     state["tipLabelKey"] = query.tl===strainSymbolUrlString ? strainSymbol : query.tl;
   }
+  
   if (query.dmin) {
-    state["dateMin"] = query.dmin;
-    state["dateMinNumeric"] = calendarToNumeric(query.dmin);
+    const dateNum = calendarToNumeric(query.dmin);
+    if (dateNum===undefined || dateNum < state.absoluteDateMinNumeric || dateNum > state.absoluteDateMaxNumeric) {
+      console.error(`Requested dmin (min date) "${query.dmin}" is invalid`);
```
genehack, line 99: _non-blocking_

Align this error message with the max date one. 

```suggestion
      console.error(`Requested min date "${query.dmin}" is invalid`);
```

src/util/colorScale.js
```diff
@@ -218,63 +221,31 @@ function createOrdinalScale(colorBy, t1nodes, t2nodes) {
...
   }
   const scale = scaleLinear().domain(domain).range(range);
 
   let legendValues;
-  if (isTemporal) {
-    legendValues = domain.slice(1);
-  } else if (colorBy==="lbi") {
+  if (colorBy==="lbi") {
     /* TODO: this is for historical reasons, and we should switch to a provided scale */
     legendValues = [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7];
```
huddlej, line 250: Not related at all to this PR, but whoa...I had no idea we hardcoded the LBI scale in Auspice. Can we (I?) remove this in a separate PR?

huddlej:
Super cool, @jameshadfield! How do you see the new `temporal` type interacting with the scatterplot view? From your example links above, I stumbled into [this view of divergence by date](https://nextstrain-s-nextstrain-waprls.herokuapp.com/staging/rabies/YYYY-MM-DD-coloring?c=date&l=scatter&scatterX=date&scatterY=div) which is what made me wonder.

src/util/dateHelpers.js
```diff
@@ -39,26 +39,52 @@ export const numericToCalendar = (numDate) => {
...
+      return numYear + 0.5;
+    }
+    if (isNaN(numYear) || isNaN(numMonth)) return undefined
+    const range = [
+      _yearMonthDayToNumeric(numYear, numMonth, 1),
+      _yearMonthDayToNumeric(numMonth===12?numYear+1:numYear, numMonth===12?1:numMonth+1, 1)
+    ]
+    return range[0] + (range[1]-range[0])/2
+  }
+  return _yearMonthDayToNumeric(numYear, numMonth, numDay)
```
genehack, line 77: _non-blocking_

the logic would be a little more apparent here with some "else"s 

```suggestion
  if (calDate.includes("X")) {
    if (!ambiguity) return undefined
    else if (year.includes("X")) return undefined;
    else if (month.includes("X")) {
      if (isNaN(numYear) || day!=="XX") return undefined
      else return numYear + 0.5;
    }
    else if (isNaN(numYear) || isNaN(numMonth)) return undefined
    const range = [
      _yearMonthDayToNumeric(numYear, numMonth, 1),
      _yearMonthDayToNumeric(numMonth===12?numYear+1:numYear, numMonth===12?1:numMonth+1, 1)
    ]
    else return range[0] + (range[1]-range[0])/2
  }
  else return _yearMonthDayToNumeric(numYear, numMonth, numDay)
```

src/util/dateHelpers.js
```diff
@@ -39,26 +39,52 @@ export const numericToCalendar = (numDate) => {
...
-  const [year, month, day] = calDate.split("-").map((n) => parseInt(n, 10));
+  const fields = calDate.split("-");
+  if (fields.length !== 3) return undefined;
+  const [year, month, day] = fields;
+  const [numYear, numMonth, numDay] = fields.map((d) => parseInt(d, 10));
+  
+  if (calDate.includes("X")) {
+    if (!ambiguity) return undefined
+    if (year.includes("X")) return undefined;
+    if (month.includes("X")) {
```
genehack, line 66: As written, this allows `1970-X-XX` or `1970-XXX-XX` to work, while `1970-01-X` or `1970-01-XXX` would fail; check for exactly `XX` to match the spec.

```suggestion
    if (month==="XX")) {
```

test/dates.test.js
```diff
@@ -93,3 +93,36 @@ test("dates are prettified as expected", () => {
...
+    expect(calendarToNumeric(calendarDate, true)).toBeCloseTo(numericDate);
+  }
+});
+
+const invalidCalendarDates = [
+  "string",
+  "-string",
+  "2024-XX-01",
+  2024.123,
+]
```
genehack, line 122:
```suggestion
  "2024-X-XX",
  "2024-XXX-XX",
]
```

genehack (approved):

src/actions/recomputeReduxState.js
joverlee521, line 99: I was confused why I kept seeing this error in the testing dataset when trying to view https://nextstrain-s-nextstrain-waprls.herokuapp.com/staging/rabies/YYYY-MM-DD-coloring?dmin=2020-01-01

It took me too long to realized this is a div only tree, so dmin/dmax/animate would not apply 🤦‍♀️ 
It would have been helpful to see a message that these queries do not work on a `divOnly` tree.

joverlee521 (approved):
Nice work! I'm slightly cautious that this might cause confusion for users based on my own experience. Not sure why, but the coloring of sample date in the rabies example _really_ confused me with the divOnly tree. Maybe fixing https://github.com/nextstrain/auspice/issues/1833 will lower the confusion, but I think there needs to be some way to highlight the distinction between the `num_date` of temporal trees and temporal colorings.

src/util/colorScale.js
jameshadfield: > I had no idea ...

Once upon a time 😉 https://github.com/nextstrain/auspice/issues/805

test/dates.test.js
jameshadfield: Thanks - this comment made me realise I wasn't testing dates which were ambiguous + erroneous correctly. We now have two separate tests:
* calendarToNumeric returns undefined for erroneous dates without ambiguity
* calendarToNumeric returns undefined for erroneous dates with ambiguity

src/util/dateHelpers.js
jameshadfield: Fixed up + tests added

genehack (approved):

src/actions/recomputeReduxState.js
jameshadfield: I've made the error messages more consistent + expanded them as per Jover's comment

src/actions/recomputeReduxState.js
```diff
@@ -92,13 +92,29 @@ const modifyStateViaURLQuery = (state, query) => {
...
+    } else {
+      state["dateMin"] = query.dmin;
+      state["dateMinNumeric"] = dateNum;
+    }
   }
   if (query.dmax) {
-    state["dateMax"] = query.dmax;
-    state["dateMaxNumeric"] = calendarToNumeric(query.dmax);
+    const dateNum = calendarToNumeric(query.dmax);
+    if (dateNum===undefined || dateNum > state.absoluteDateMaxNumeric || dateNum < state.absoluteDateMinNumeric) {
```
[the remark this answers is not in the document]
jameshadfield: Yeah, good call. Done.

src/actions/recomputeReduxState.js
jameshadfield: I think there are two cases of invalid URL queries - when the UI has put us in an invalid state (bad!) and when we've added an invalid query ourselves. This falls in the latter group and in general I don't think we need to do too much here, but in this case it's simple enough that I added it:
```
console.error(`URL query "dmin=${query.dmin}" is invalid ${state.branchLengthsToDisplay==='divOnly'?'(the tree is not a timetree)':''}`);

```

(There are plenty of weird situations similar to this -- e.g. ask for the frequencies panel in a dataset without one:  [zika?d=frequencies](https://nextstrain.org/zika?d=frequencies).)

src/util/dateHelpers.js
jameshadfield: ha! I actually don't like this style too much and have recommended the opposite to others 😄

jameshadfield:
This PR didn't introduce temporal scales for arbitrary metadata, it simply allowed values to be YYYY-MM-DD strings rather than numbers. Review comments have raised some valid points about the temporal scales themselves:

> How do you see the new temporal type interacting with the scatterplot view? From your example links above, I stumbled into [this view of divergence by date](https://nextstrain-s-nextstrain-waprls.herokuapp.com/staging/rabies/YYYY-MM-DD-coloring?c=date&l=scatter&scatterX=date&scatterY=div) which is what made me wonder.

Good find! I'll make an issue for this. This should work (but I'm not surprised it doesn't) and show a temporal x-axis, but of course there'll be no branches displayed.

> I'm slightly cautious that this might cause confusion for users based on my own experience. Not sure why, but the coloring of sample date in the rabies example really confused me with the divOnly tree. Maybe fixing https://github.com/nextstrain/auspice/issues/1833 will lower the confusion, but I think there needs to be some way to highlight the distinction between the num_date of temporal trees and temporal colorings.

Sure - if you can think of visual aids (beyond #1833) we can use them. The tree's x-axis units and title indicate that it's a divergence tree. I guess I don't see it as different to [colouring by date but looking at the divergence tree](https://nextstrain.org/zika?c=num_date&m=div) which has always been possible.

src/util/dateHelpers.js
genehack: As we discussed earlier on our call, this is peak "take it or leave it" feedback. I like the explicitness that the `else`s introduce; it means I don't have to reason as much about what's a fall-through condition and what's an alternative branch.

src/actions/recomputeReduxState.js
joverlee521: Yup, just the additional hint that the tree is not a timetree is enough for me! These types of hints will be especially useful when we [surface console errors in the UI](https://github.com/nextstrain/auspice/issues/1813)

joverlee521:
> I guess I don't see it as different to [colouring by date but looking at the divergence tree](https://nextstrain.org/zika?c=num_date&m=div) which has always been possible.

That's what I was expecting, but the available controls are slightly different when there's _only_ a div tree. I was mainly confused why the date range filter was not available. [Hiding the control sections](https://github.com/nextstrain/auspice/pull/1834) is enough for me for now, maybe we can revisit if questions come up from external users.

jameshadfield:
Oh yeah, that's fair. I actually [prototyped](https://github.com/nextstrain/auspice/pull/1519) a date range filter for this kind of metadata but I wasn't happy with the overall behaviour of the feature.

src/util/colorScale.js
huddlej: Oh no...

huddlej: @jameshadfield https://github.com/nextstrain/auspice/issues/541

I should stop pulling on this thread :D

jameshadfield: ha! too good

PR #1836, opened 2024-08-28
Scatterplots work with temporal YYYY-MM-DD scales

### LINKED ISSUE #1835: Temporal scales don't work in scatterplots
### Current Behavior

Colourings with a temporal scale (except for `num_date`) do not work in scatterplots, with NaNs appearing in the SVG coordinates, e.g.:
```
Error: <path> attribute d: Expected number, "MNaN 472.09999999…".
```

They should [behave similarly to how `num_date` currently does](https://nextstrain.org/zika?c=num_date&l=scatter&m=div), however all nodes without metadata values (usually all internal nodes and perhaps some tips) should not appear. Scatterplots already have the ability to hide nodes which don't have appropriate metadata ([e.g.](https://nextstrain.org/zika?c=author&l=scatter&m=div&scatterX=region)).

This was first identified in #1832 but it predates that PR.

### How to reproduce

1. Create a dataset with a temporal scale (values can be numerical or, if you're using a version of Auspice more recent than c5857225f2911091fac20395e091031c79683b11 they can be YYYY-MM-DD)

2. View a scatterplot with the temporal scale as the x or y axis.

jameshadfield:
Closes #1835

This code would be greatly enhanced by using TypeScript

- [ ] Checks pass

### DIFF
--- a/src/util/scatterplotHelpers.js
+++ b/src/util/scatterplotHelpers.js
@@ -98,9 +98,12 @@ export function getFirstMatchingScatterVariable(options, tryTheseFirst, notThisV
 export function addScatterAxisInfo(scatterVariables, axis, controls, tree, metadata) {
   const axisVar = scatterVariables[axis];
   const knownContinuousKeys = ["div", "num_date", "displayOrder"];
-  if (knownContinuousKeys.includes(axisVar) || metadata.colorings[axisVar].type==="continuous") {
+  if (knownContinuousKeys.includes(axisVar) || metadata.colorings[axisVar].type==="continuous" ||
+    metadata.colorings[axisVar].type==="temporal")
+  {
     scatterVariables[`${axis}Continuous`] = true;
     scatterVariables[`${axis}Domain`] = undefined;
+    scatterVariables[`${axis}Temporal`] = metadata.colorings[axisVar]?.type==="temporal";
     return scatterVariables;
   }
   const trait = scatterVariables[axis]==="gt" ? controls.colorBy : scatterVariables[axis];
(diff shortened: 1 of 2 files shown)


jameshadfield:
@huddlej you can see your (formerly broken) example at https://nextstrain-s-nextstrain-smwbzp.herokuapp.com/staging/rabies/YYYY-MM-DD-coloring?l=scatter&scatterX=date&scatterY=div

src/util/scatterplotHelpers.js
```diff
@@ -97,10 +97,14 @@ export function getFirstMatchingScatterVariable(options, tryTheseFirst, notThisV
  */
 export function addScatterAxisInfo(scatterVariables, axis, controls, tree, metadata) {
   const axisVar = scatterVariables[axis];
+  console.log("addScatterAxisInfo::axisVar", axis, axisVar)
```
genehack, line 100: intentional? (asking because the linting is complaining.)

genehack (approved):

src/util/scatterplotHelpers.js
jameshadfield: Ahh. Intentionally added, unintentionally neglected to remove.

jameshadfield: That's not why CI failed here (that's due to a 6-hour `npm ci` call timing out the action), but I agree with #1839 in that it should cause CI to fail.

jameshadfield: [dropped in force push](https://github.com/nextstrain/auspice/compare/681951aa9e7793db220cd3c9866a203a2ffcdac5..f5e388ef35cb700fafb5377be4f758d86a188f67)

huddlej:
Nice! Thanks, @jameshadfield!

PR #1863, opened 2024-10-08
Tree component refactoring

victorlin:

Changes made while prepping for #1373. These can be considered independently from that PR. See individual commit messages.

## Checklist

- [x] Checks pass
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ no functional changes

### DIFF
--- a/src/components/tree/tree.js
+++ b/src/components/tree/tree.js
@@ -35,19 +35,22 @@ class Tree extends React.Component {
       tree: null,
       treeToo: null
     };
+
     /* bind callbacks */
     this.clearSelectedNode = callbacks.clearSelectedNode.bind(this);
-    this.redrawTree = () => {
-      this.props.dispatch(updateVisibleTipsAndBranchThicknesses({
-        root: [0, 0]
-      }));
-    };
-    /* pressing the escape key should dismiss an info modal (if one exists) */
-    this.handlekeydownEvent = (event) => {
-      if (event.key==="Escape" && this.props.selectedNode) {
-        this.clearSelectedNode(this.props.selectedNode);
-      }
-    };
+  }
+
+  redrawTree = () => {
+    this.props.dispatch(updateVisibleTipsAndBranchThicknesses({
+      root: [0, 0]
+    }));
+  }
+
+  /* pressing the escape key should dismiss an info modal (if one exists) */
+  handlekeydownEvent = (event) => {
+    if (event.key==="Escape" && this.props.selectedNode) {
+      this.clearSelectedNode(this.props.selectedNode);
+    }
   }
 
   setUpAndRenderTreeToo(props, newState) {
--- a/src/components/tree/tree.js
+++ b/src/components/tree/tree.js
@@ -115,16 +115,13 @@ class Tree extends React.Component {
   }
 
   getStyles = () => {
-    const activeResetTreeButton = this.props.tree.idxOfInViewRootNode !== 0 ||
-      this.props.treeToo.idxOfInViewRootNode !== 0;
-
     const filteredTree = !!this.props.tree.idxOfFilteredRoot &&
       this.props.tree.idxOfInViewRootNode !== this.props.tree.idxOfFilteredRoot;
     const filteredTreeToo = !!this.props.treeToo.idxOfFilteredRoot &&
       this.props.treeToo.idxOfInViewRootNode !== this.props.treeToo.idxOfFilteredRoot;
     const activeZoomButton = filteredTree || filteredTreeToo;
 
-    const treeIsZoomed = this.props.tree.idxOfInViewRootNode !== 0 ||
+    const anyTreeZoomed = this.props.tree.idxOfInViewRootNode !== 0 ||
       this.props.treeToo.idxOfInViewRootNode !== 0;
 
     return {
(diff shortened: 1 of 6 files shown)


src/components/tree/tree.js
```diff
@@ -35,21 +35,25 @@ class Tree extends React.Component {
...
-    /* pressing the escape key should dismiss an info modal (if one exists) */
-    this.handlekeydownEvent = (event) => {
-      if (event.key==="Escape" && this.props.selectedNode) {
-        this.clearSelectedNode(this.props.selectedNode);
-      }
-    };
   }
+
+  clearSelectedNode = () => {
+    callbacks.clearSelectedNode.bind(this);
```
jameshadfield, line 41: [reminder to me] - check this change, I remember binding callbacks in the constructor was "the right way" when I was writing this code.

jameshadfield:
@genehack could you take a look at the types added / used here if you have a chance? 🙏

src/components/tree/tree.js
victorlin: I see:

https://github.com/nextstrain/auspice/blob/3cd0c1e60cd728e7c9bfc05f5d80b2607ba5af52/src/components/map/map.js#L84-L85

I've replaced 1eedee40 with a1aec383.

src/components/tree/tree.js
```diff
@@ -110,18 +115,17 @@ class Tree extends React.Component {
...
       this.props.tree.idxOfInViewRootNode !== this.props.tree.idxOfFilteredRoot;
     const filteredTreeToo = !!this.props.treeToo.idxOfFilteredRoot &&
       this.props.treeToo.idxOfInViewRootNode !== this.props.treeToo.idxOfFilteredRoot;
     const activeZoomButton = filteredTree || filteredTreeToo;
 
-    const treeIsZoomed = this.props.tree.idxOfInViewRootNode !== 0 ||
+    const anyTreeZoomed = this.props.tree.idxOfInViewRootNode !== 0 ||
       this.props.treeToo.idxOfInViewRootNode !== 0;
 
+    const activeResetTreeButton = anyTreeZoomed;
```
genehack, line 127: if you're gonna make these two variables the same thing why not just go the one extra step and rename `treeIsZoomed` directly to `activeResetTreeButton`, and then update the ternaries below to also use `activeResetTreeButton`? Seems less confusing in the longer run.

src/reducers/controls.ts
```diff
@@ -36,11 +40,24 @@ export interface MeasurementsControlState {
...
+  layout: Layout
+  geoResolution: string
+  filters: Record<string, any>
+  filtersInFooter: string[]
+  colorBy: string
+  selectedBranchLabel: string
+  tipLabelKey: typeof strainSymbol
+  showTransmissionLines: boolean
+  sidebarOpen?: boolean
+}
```
genehack, line 54: I'd put this up around L15, so it's defined before being used in the (current) L18 -- was briefly confused where `Defaults` was even coming from.

src/reducers/controls.ts
```diff
@@ -36,11 +40,24 @@ export interface MeasurementsControlState {
...
 
+interface Defaults {
+  distanceMeasure: string
+  layout: Layout
+  geoResolution: string
+  filters: Record<string, any>
+  filtersInFooter: string[]
+  colorBy: string
+  selectedBranchLabel: string
+  tipLabelKey: typeof strainSymbol
```
genehack, line 51: what's the advantage of doing this versus adding a proper type in `../utils/globals` and exporting it from there?

src/reducers/narrative.ts
```diff
@@ -1,19 +1,47 @@
...
+  blockIdx: undefined,
+  pathname: undefined,
   display: false,
   title: undefined
-}, action) => {
+};
+
+const narrative = (
+  state: NarrativeState = defaultState,
+  action: AnyAction,
```
genehack, line 36: can you provide a default value to a positional param if there's a subsequent positional param without a default value? or do you need to switch the order here, or convert this to being passed an object?

src/components/tree/tree.js
victorlin: Good point, done: 4d8a2e04

I chose to go the other way around and drop `activeResetTreeButton` in favor of `anyTreeZoomed` to convey the condition rather than end state.

src/reducers/controls.ts
victorlin: I didn't want to dig a deeper hole and convert `src/utils/globals.js` to TypeScript. Even if I were to do that, I'm not sure that adding a proper type would be useful in this case. In `globals.ts` it would simply be:

```diff
- export const strainSymbol = Symbol('strain');
+ export const strainSymbol: unique symbol = Symbol('strain');
```

and the same `typeof strainSymbol` would still be used here.

src/reducers/narrative.ts
victorlin: Good point – ESLint and TypeScript clearly had no complaints about this so I searched to confirm – it is allowed in JavaScript. From [MDN docs on default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters):

> Parameters are still set left-to-right, overwriting default parameters even if there are later parameters without defaults.
>
> ```js
> function f(x = 1, y) {
>   return [x, y];
> }
> 
> f(); // [1, undefined]
> f(2); // [2, undefined]
> ```

victorlin: I don't think it's a good pattern, but it is the pattern used for other files in `src/reducers/` so I will keep it as-is.

src/reducers/controls.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/46588036488a0132423f7eb01815aa658c1686e3/src/reducers/controls.ts#L15-L32

src/reducers/narrative.ts
jameshadfield: Just a quick comment on this (and similar ones such as `const foo = function foo() {...}`) - if you see stuff that is clearly poorly implemented it's probably because it was just a poor implementation. Please add a FIXME comment in the code or just fix it. I'd prefer inconsistent code over consistently bad code.

PR #1864, opened 2024-10-08
Add tree-related types

victorlin:

Converted most but not all tree-related files in the commit with the same title as this PR. There were many type errors exposed by conversion. They have been addressed in the following commits.

I removed a few comments that are now self-explanatory with the added types. I also moved all JSDoc types and descriptions into the function signature.

Note that there are many optional properties in the newly defined interfaces. This is not ideal and doesn't capture the fact that many properties are conditional on things such as layout. But it's a starting point to be improved over time.

## Checklist

- [x] Check bundle size ([ref](https://github.com/nextstrain/auspice/pull/1864#discussion_r1825178808))
- [x] Check type of `ReduxNode.currentGt` ([ref](https://github.com/nextstrain/auspice/pull/1864#discussion_r1825235689))
- [x] Figure out what to do with `strictNullChecks` violations ([ref](https://github.com/nextstrain/auspice/pull/1864#discussion_r1821713176))
- [x] Figure out how to sort the new types ([ref](https://github.com/nextstrain/auspice/pull/1864#discussion_r1826253299))
- [x] Which files to put types in? ([ref1](https://github.com/nextstrain/auspice/pull/1864#discussion_r1826252513), [ref2](https://github.com/nextstrain/auspice/pull/1864#discussion_r1827078152))
- [x] Checks pass

## For later

- https://github.com/nextstrain/auspice/issues/1888
- https://github.com/nextstrain/auspice/issues/1889
- https://github.com/nextstrain/auspice/issues/1890
- https://github.com/nextstrain/auspice/issues/1891

### DIFF
--- a/src/actions/tree.ts
+++ b/src/actions/tree.ts
@@ -234,7 +234,6 @@ export const updateTipRadii = (
     tipSelectedIdx = false,
     selectedLegendItem = false,
     geoFilter = [],
-    searchNodes = false
   }: {
     /** the strain to highlight (always tree 1) */
     tipSelectedIdx?: number | false,
@@ -244,8 +243,6 @@ export const updateTipRadii = (
 
     /** a filter to apply to the strains. Empty array or array of len 2. [0]: geoResolution, [1]: value to filter to */
     geoFilter?: [string, string] | [],
-
-    searchNodes?: PhyloNode[] | false
   } = {}
 ): ThunkFunction => {
   return (dispatch, getState) => {
@@ -262,8 +259,8 @@ export const updateTipRadii = (
         d.dataToo = calcTipRadii({tipSelectedIdx: idx, colorScale, tree: treeToo});
       }
     } else {
-      d.data = calcTipRadii({selectedLegendItem, geoFilter, searchNodes, colorScale, tree});
-      if (tt) d.dataToo = calcTipRadii({selectedLegendItem, geoFilter, searchNodes, colorScale, tree: treeToo});
+      d.data = calcTipRadii({selectedLegendItem, geoFilter, colorScale, tree});
+      if (tt) d.dataToo = calcTipRadii({selectedLegendItem, geoFilter, colorScale, tree: treeToo});
     }
     dispatch(d);
   };
--- a/src/actions/tree.ts
+++ b/src/actions/tree.ts
@@ -398,7 +398,7 @@ const _traverseAndCreateSubtrees = (
         subtreeRootNode.parent = root;
       });
       node.unexplodedChildren = originalChildren;
-      node.children = node.children.filter((c, idx) => {
+      node.children = node.children.filter((_c, idx) => {
         return !childrenToPrune.includes(idx);
       });
       /* it may be the case that the node now has no children (they're all subtrees!) */
--- a/src/actions/tree.ts
+++ b/src/actions/tree.ts
@@ -259,7 +259,7 @@ export const updateTipRadii = (
       d.data = calcTipRadii({tipSelectedIdx, colorScale, tree});
       if (tt) {
         const idx = strainNameToIdx(treeToo.nodes, tree.nodes[tipSelectedIdx].name);
-        d.dataToo = calcTipRadii({idx, colorScale, tree: treeToo});
+        d.dataToo = calcTipRadii({tipSelectedIdx: idx, colorScale, tree: treeToo});
       }
     } else {
       d.data = calcTipRadii({selectedLegendItem, geoFilter, searchNodes, colorScale, tree});
--- a/src/actions/tree.ts
+++ b/src/actions/tree.ts
@@ -1,3 +1,4 @@
+import { AnyAction } from "@reduxjs/toolkit";
 import { calcTipRadii } from "../util/tipRadiusHelpers";
 import { strainNameToIdx, calculateVisiblityAndBranchThickness } from "../util/treeVisibilityHelpers";
 import * as types from "./types";
@@ -10,7 +11,18 @@ import { createVisibleLegendValues, getLegendOrder } from "../util/colorScale";
 import { getTraitFromNode } from "../util/treeMiscHelpers";
 import { warningNotification } from "./notifications";
 import { calcFullTipCounts, calcTipCounts } from "../util/treeCountingHelpers";
+import { PhyloNode } from "../components/tree/phyloTree/types";
+import { Metadata } from "../metadata";
+import { AppDispatch, RootState } from "../store";
+import { ReduxNode, TreeState } from "../reducers/tree/types";
 
+type RootIndex = number | undefined
+
+/** [root idx tree1, root idx tree2] */
+export type Root = [RootIndex, RootIndex]
+
+/** A function to be handled by redux (thunk) */
+type ThunkFunction = (dispatch: AppDispatch, getState: () => RootState) => void
 
 /**
  * Updates the `inView` property of nodes which depends on the currently selected
@@ -18,10 +30,13 @@ import { calcFullTipCounts, calcTipCounts } from "../util/treeCountingHelpers";
  * Note that this property is historically the remit of PhyloTree, however this function
  * may be called before those objects are created; in this case we store the property on
  * the tree node itself.
- * @param {Int} idx - index of displayed root node
- * @param {ReduxTreeState} tree
  */
-export const applyInViewNodesToTree = (idx, tree) => {
+export const applyInViewNodesToTree = (
+  /** index of displayed root node */
+  idx: RootIndex,
+
+  tree: TreeState,
+): number => {
   const validIdxRoot = idx !== undefined ? idx : tree.idxOfInViewRootNode;
   if (tree.nodes[0].shell) {
     tree.nodes.forEach((d) => {
@@ -29,12 +44,12 @@ export const applyInViewNodesToTree = (idx, tree) => {
       d.shell.update = true;
     });
     if (tree.nodes[validIdxRoot].hasChildren) {
-      applyToChildren(tree.nodes[validIdxRoot].shell, (d) => {d.inView = true;});
+      applyToChildren(tree.nodes[validIdxRoot].shell, (d: PhyloNode) => {d.inView = true;});
     } else if (tree.nodes[validIdxRoot].parent.arrayIdx===0) {
       // subtree with n=1 tips => don't make the parent in-view as this will cover the entire tree!
       tree.nodes[validIdxRoot].shell.inView = true;
     } else {
-      applyToChildren(tree.nodes[validIdxRoot].parent.shell, (d) => {d.inView = true;});
+      applyToChildren(tree.nodes[validIdxRoot].parent.shell, (d: PhyloNode) => {d.inView = true;});
     }
   } else {
     /* FYI applyInViewNodesToTree is now setting inView on the redux nodes */
@@ -42,7 +57,7 @@ export const applyInViewNodesToTree = (idx, tree) => {
       d.inView = false;
     });
     /* note that we cannot use `applyToChildren` as that operates on PhyloNodes */
-    const _markChildrenInView = (node) => {
+    const _markChildrenInView = (node: ReduxNode) => {
       node.inView = true;
       if (node.children) {
         for (const child of node.children) _markChildrenInView(child);
@@ -61,14 +76,21 @@ export const applyInViewNodesToTree = (idx, tree) => {
  * this fn relies on the "inView" attr of nodes
  * note that this function checks to see if the tree has been defined (different to if it's ready / loaded!)
  * for arg destructuring see https://simonsmith.io/destructuring-objects-as-function-parameters-in-es6/
- * @param  {array|undefined} root Change the in-view part of the tree. [root idx tree1, root idx tree2].
- *                                [0, 0]: reset. [undefined, undefined]: do nothing
- * @param  {string | undefined} cladeSelected
- * @return {function} a function to be handled by redux (thunk)
  */
-export const updateVisibleTipsAndBranchThicknesses = (
-  {root = [undefined, undefined], cladeSelected = undefined} = {}
-) => {
+export const updateVisibleTipsAndBranchThicknesses = ({
+  root = [undefined, undefined],
+  cladeSelected = undefined,
+}: {
+  /**
+   * Change the in-view part of the tree.
+   *
+   * [0, 0]: reset. [undefined, undefined]: do nothing
+   */
+  root?: Root
+
+  cladeSelected?: string
+} = {}
+): ThunkFunction => {
   return (dispatch, getState) => {
     const { tree, treeToo, controls, frequencies } = getState();
     if (root[0] === undefined && !cladeSelected && tree.selectedClade) {
@@ -86,7 +108,7 @@ export const updateVisibleTipsAndBranchThicknesses = (
       controls,
       {dateMinNumeric: controls.dateMinNumeric, dateMaxNumeric: controls.dateMaxNumeric}
     );
-    const dispatchObj = {
+    const dispatchObj: AnyAction = {
       type: types.UPDATE_VISIBILITY_AND_BRANCH_THICKNESS,
       visibility: data.visibility,
       visibilityVersion: data.visibilityVersion,
@@ -141,11 +163,17 @@ export const updateVisibleTipsAndBranchThicknesses = (
  * date changes need to update tip visibility & branch thicknesses
  * this can be done in a single action
  * NB calling this without specifying newMin OR newMax is a no-op
- * @param  {string|false} newMin optional
- * @param  {string|false} newMax optional
- * @return {null} side-effects: a single action
+ * side-effects: a single action
  */
-export const changeDateFilter = ({newMin = false, newMax = false, quickdraw = false}) => {
+export const changeDateFilter = ({
+  newMin = false,
+  newMax = false,
+  quickdraw = false,
+}: {
+  newMin?: string | false
+  newMax?: string | false
+  quickdraw?: boolean
+}): ThunkFunction => {
   return (dispatch, getState) => {
     const { tree, treeToo, controls, frequencies } = getState();
     if (!tree.nodes) {return;}
@@ -154,7 +182,7 @@ export const changeDateFilter = ({newMin = false, newMax = false, quickdraw = fa
       dateMaxNumeric: newMax ? calendarToNumeric(newMax) : controls.dateMaxNumeric
     };
     const data = calculateVisiblityAndBranchThickness(tree, controls, dates);
-    const dispatchObj = {
+    const dispatchObj: AnyAction = {
       type: types.CHANGE_DATES_VISIBILITY_THICKNESS,
       quickdraw,
       dateMin: newMin ? newMin : controls.dateMin,
@@ -199,18 +227,31 @@ export const changeDateFilter = ({newMin = false, newMax = false, quickdraw = fa
 
 /**
  * NB all params are optional - supplying none resets the tip radii to defaults
- * @param  {string|number} selectedLegendItem value of the attr. if scale is continuous a bound will be used.
- * @param  {int} tipSelectedIdx the strain to highlight (always tree 1)
- * @param  {array} geoFilter a filter to apply to the strains. Empty array or array of len 2. [0]: geoResolution, [1]: value to filter to
- * @return {null} side-effects: a single action
+ * side-effects: a single action
  */
 export const updateTipRadii = (
-  {tipSelectedIdx = false, selectedLegendItem = false, geoFilter = [], searchNodes = false} = {}
-) => {
+  {
+    tipSelectedIdx = false,
+    selectedLegendItem = false,
+    geoFilter = [],
+    searchNodes = false
+  }: {
+    /** the strain to highlight (always tree 1) */
+    tipSelectedIdx?: number | false,
+
+    /** value of the attr. if scale is continuous a bound will be used. */
+    selectedLegendItem?: string | number | false,
+
+    /** a filter to apply to the strains. Empty array or array of len 2. [0]: geoResolution, [1]: value to filter to */
+    geoFilter?: [string, string] | [],
+
+    searchNodes?: PhyloNode[] | false
+  } = {}
+): ThunkFunction => {
   return (dispatch, getState) => {
     const { controls, tree, treeToo } = getState();
     const colorScale = controls.colorScale;
-    const d = {
+    const d: AnyAction = {
       type: types.UPDATE_TIP_RADII, version: tree.tipRadiiVersion + 1
     };
     const tt = controls.showTreeToo;
@@ -230,16 +271,22 @@ export const updateTipRadii = (
 
 /**
  * Apply a filter to the current selection (i.e. filtered / "on" values associated with this trait)
- * Explanation of the modes:
- *    "add" -> add the values to the current selection (if any exists).
- *    "inactivate" -> inactivate values (i.e. change active prop to false). To activate just use "add".
- *    "remove" -> remove the values from the current selection
- *    "set"  -> set the values of the filter to be those provided. All disabled filters will be removed. XXX TODO.
- * @param {string} mode allowed values: "set", "add", "remove"
- * @param {string} trait the trait name of the filter ("authors", "country" etcetera)
- * @param {Array of strings} values the values (see above)
  */
-export const applyFilter = (mode, trait, values) => {
+export const applyFilter = (
+  /** Explanation of the modes:
+   *  - "add" -> add the values to the current selection (if any exists).
+   *  - "inactivate" -> inactivate values (i.e. change active prop to false). To activate just use "add".
+   *  - "remove" -> remove the values from the current selection
+   *  - "set"  -> set the values of the filter to be those provided. All disabled filters will be removed. XXX TODO.
+   */
+  mode: "add" | "inactivate" | "remove" | "set",
+  
+  /** the trait name of the filter ("authors", "country" etcetera) */
+  trait: string | symbol,
+
+  /** the values (see above) */
+  values: string[],
+): ThunkFunction => {
   return (dispatch, getState) => {
     const { controls } = getState();
     const currentlyFilteredTraits = Reflect.ownKeys(controls.filters);
@@ -296,16 +343,15 @@ export const applyFilter = (mode, trait, values) => {
   };
 };
 
-export const toggleTemporalConfidence = () => ({
+export const toggleTemporalConfidence = (): AnyAction => ({
   type: types.TOGGLE_TEMPORAL_CONF
 });
 
 
 /**
  * restore original state by iterating over all nodes and restoring children to unexplodedChildren (as necessary)
- * @param {Array<Node>} nodes
  */
-const _resetExpodedTree = (nodes) => {
+const _resetExpodedTree = (nodes: ReduxNode[]): void => {
   nodes.forEach((n) => {
     if (Object.prototype.hasOwnProperty.call(n, 'unexplodedChildren')) {
       n.children = n.unexplodedChildren;
@@ -323,11 +369,17 @@ const _resetExpodedTree = (nodes) => {
  * create subtrees where branches have different attrs.
  * Note: because the children of a node may change, we store the previous (unexploded) children
  * as `unexplodedChildren` so we can return to the original tree.
- * @param {Node} root - root node of entire tree
- * @param {Node} node - current node being traversed
- * @param {String} attr - trait name to determine if a child should become subtree
  */
-const _traverseAndCreateSubtrees = (root, node, attr) => {
+const _traverseAndCreateSubtrees = (
+  /** root node of entire tree */
+  root: ReduxNode,
+
+  /** current node being traversed */
+  node: ReduxNode,
+
+  /** trait name to determine if a child should become subtree */
+  attr: string,
+): void => {
   // store original children so we traverse the entire tree
   const originalChildren = node.hasChildren ? [...node.children] : [];
 
@@ -363,7 +415,11 @@ const _traverseAndCreateSubtrees = (root, node, attr) => {
 /**
  * sort the subtrees by the order the trait would appear in the legend
  */
-const _orderSubtrees = (metadata, nodes, attr) => {
+const _orderSubtrees = (
+  metadata: Metadata,
+  nodes: ReduxNode[],
+  attr: string,
+): void => {
   const attrValueOrder = getLegendOrder(attr, metadata.colorings[attr], nodes, undefined);
   nodes[0].children.sort((childA, childB) => {
     const [attrA, attrB] = [getTraitFromNode(childA, attr), getTraitFromNode(childB, attr)];
@@ -380,43 +436,47 @@ const _orderSubtrees = (metadata, nodes, attr) => {
   });
 };
 
-export const explodeTree = (attr) => (dispatch, getState) => {
-  const {tree, metadata, controls} = getState();
-  _resetExpodedTree(tree.nodes); // ensure we start with an unexploded tree
-  if (attr) {
-    const root = tree.nodes[0];
-    _traverseAndCreateSubtrees(root, root, attr);
-    if (root.unexplodedChildren.length === root.children.length) {
-      dispatch(warningNotification({message: "Cannot explode tree on this trait - is it defined on internal nodes?"}));
-      return;
+export const explodeTree = (
+  attr: string | undefined,
+): ThunkFunction => {
+  return (dispatch, getState) => {
+    const {tree, metadata, controls} = getState();
+    _resetExpodedTree(tree.nodes); // ensure we start with an unexploded tree
+    if (attr) {
+      const root = tree.nodes[0];
+      _traverseAndCreateSubtrees(root, root, attr);
+      if (root.unexplodedChildren.length === root.children.length) {
+        dispatch(warningNotification({message: "Cannot explode tree on this trait - is it defined on internal nodes?"}));
+        return;
+      }
+      _orderSubtrees(metadata, tree.nodes, attr);
     }
-    _orderSubtrees(metadata, tree.nodes, attr);
-  }
-  /* tree splitting necessitates recalculation of tip counts */
-  calcFullTipCounts(tree.nodes[0]);
-  calcTipCounts(tree.nodes[0], tree.visibility);
-  /* we default to zooming out completely whenever we explode the tree. There are nicer behaviours here,
-  such as re-calculating the MRCA of visible nodes, but this comes at the cost of increased complexity.
-  Note that the functions called here involve a lot of code duplication and are good targets for refactoring */
-  applyInViewNodesToTree(0, tree);
-  const visData = calculateVisiblityAndBranchThickness(
-    tree,
-    controls,
-    {dateMinNumeric: controls.dateMinNumeric, dateMaxNumeric: controls.dateMaxNumeric}
-  );
-  visData.idxOfInViewRootNode = 0;
-  /* Changes in visibility require a recomputation of which legend items we wish to display */
-  visData.visibleLegendValues = createVisibleLegendValues({
-    colorBy: controls.colorBy,
-    genotype: controls.colorScale.genotype,
-    scaleType: controls.colorScale.scaleType,
-    legendValues: controls.colorScale.legendValues,
-    treeNodes: tree.nodes,
-    visibility: visData.visibility
-  });
-  dispatch({
-    type: types.CHANGE_EXPLODE_ATTR,
-    explodeAttr: attr,
-    ...visData
-  });
+    /* tree splitting necessitates recalculation of tip counts */
+    calcFullTipCounts(tree.nodes[0]);
+    calcTipCounts(tree.nodes[0], tree.visibility);
+    /* we default to zooming out completely whenever we explode the tree. There are nicer behaviours here,
+    such as re-calculating the MRCA of visible nodes, but this comes at the cost of increased complexity.
+    Note that the functions called here involve a lot of code duplication and are good targets for refactoring */
+    applyInViewNodesToTree(0, tree);
+    const visData = calculateVisiblityAndBranchThickness(
+      tree,
+      controls,
+      {dateMinNumeric: controls.dateMinNumeric, dateMaxNumeric: controls.dateMaxNumeric}
+    );
+    visData.idxOfInViewRootNode = 0;
+    /* Changes in visibility require a recomputation of which legend items we wish to display */
+    visData.visibleLegendValues = createVisibleLegendValues({
+      colorBy: controls.colorBy,
+      genotype: controls.colorScale.genotype,
+      scaleType: controls.colorScale.scaleType,
+      legendValues: controls.colorScale.legendValues,
+      treeNodes: tree.nodes,
+      visibility: visData.visibility
+    });
+    dispatch({
+      type: types.CHANGE_EXPLODE_ATTR,
+      explodeAttr: attr,
+      ...visData
+    });
+  };
 };
--- a/src/components/tree/index.ts
+++ b/src/components/tree/index.ts
@@ -1,8 +1,11 @@
-import { connect } from "react-redux";
+import { connect, MapStateToProps } from "react-redux";
 import UnconnectedTree from "./tree";
 import { RootState } from "../../store";
+import { TreeComponentOwnProps, TreeComponentStateProps } from "./types";
 
-const Tree = connect((state: RootState) => ({
+const mapStateToProps: MapStateToProps<TreeComponentStateProps, TreeComponentOwnProps, RootState> = (
+  state: RootState,
+): TreeComponentStateProps => ({
   tree: state.tree,
   treeToo: state.treeToo,
   selectedNode: state.controls.selectedNode,
@@ -32,6 +35,8 @@ const Tree = connect((state: RootState) => ({
   animationPlayPauseButton: state.controls.animationPlayPauseButton,
   showOnlyPanels: state.controls.showOnlyPanels,
   performanceFlags: state.controls.performanceFlags,
-}))(UnconnectedTree);
+});
+
+const Tree = connect(mapStateToProps)(UnconnectedTree);
 
 export default Tree;
--- a/src/components/tree/phyloTree/change.ts
+++ b/src/components/tree/phyloTree/change.ts
@@ -311,7 +311,7 @@ export const change = function change(
     branchThickness = undefined,
     focus = undefined,
     scatterVariables = undefined,
-    performanceFlags = {},
+    performanceFlags = undefined,
   }: ChangeParams
 ): void {
   // console.log("\n** phylotree.change() (time since last run:", Date.now() - this.timeLastRenderRequested, "ms) **\n\n");
--- a/src/components/tree/phyloTree/change.ts
+++ b/src/components/tree/phyloTree/change.ts
@@ -418,7 +418,7 @@ export const change = function change(
   }
   /* mapToScreen */
   if (
-    svgPropsToUpdate.has(["stroke-width"]) ||
+    svgPropsToUpdate.has("stroke-width") ||
     newDistance ||
     newLayout ||
     changeNodeOrder ||
--- a/src/components/tree/phyloTree/change.ts
+++ b/src/components/tree/phyloTree/change.ts
@@ -1,3 +1,4 @@
+import { Selection, Transition } from "d3";
 import { timerFlush } from "d3-timer";
 import { calcConfidenceWidth } from "./confidence";
 import { applyToChildren, setDisplayOrder } from "./helpers";
@@ -6,11 +7,15 @@ import { NODE_VISIBLE } from "../../../util/globals";
 import { getBranchVisibility, strokeForBranch } from "./renderers";
 import { shouldDisplayTemporalConfidence } from "../../../reducers/controls";
 import { makeTipLabelFunc } from "./labels";
+import { ChangeParams, PhyloNode, PhyloTreeType, PropsForPhyloNodes, SVGProperty, TreeElement } from "./types";
 
 /* loop through the nodes and update each provided prop with the new value
  * additionally, set d.update -> whether or not the node props changed
  */
-const updateNodesWithNewData = (nodes, newNodeProps) => {
+const updateNodesWithNewData = (
+  nodes: PhyloNode[],
+  newNodeProps: PropsForPhyloNodes,
+): void => {
   // console.log("update nodes with data for these keys:", Object.keys(newNodeProps));
   // let tmp = 0;
   nodes.forEach((d, i) => {
@@ -35,38 +40,45 @@ const updateNodesWithNewData = (nodes, newNodeProps) => {
 const svgSetters = {
   attrs: {
     ".tip": {
-      r: (d) => d.r,
-      cx: (d) => d.xTip,
-      cy: (d) => d.yTip
+      r: (d: PhyloNode) => d.r,
+      cx: (d: PhyloNode) => d.xTip,
+      cy: (d: PhyloNode) => d.yTip
     },
     ".branch": {
     },
     ".vaccineCross": {
-      d: (d) => d.vaccineCross
+      d: (d: PhyloNode) => d.vaccineCross
     },
     ".conf": {
-      d: (d) => d.confLine
+      d: (d: PhyloNode) => d.confLine
     }
   },
   styles: {
     ".tip": {
-      fill: (d) => d.fill,
-      stroke: (d) => d.tipStroke,
-      visibility: (d) => d.visibility === NODE_VISIBLE ? "visible" : "hidden"
+      fill: (d: PhyloNode) => d.fill,
+      stroke: (d: PhyloNode) => d.tipStroke,
+      visibility: (d: PhyloNode) => d.visibility === NODE_VISIBLE ? "visible" : "hidden"
     },
     ".conf": {
-      stroke: (d) => d.branchStroke,
+      stroke: (d: PhyloNode) => d.branchStroke,
       "stroke-width": calcConfidenceWidth
     },
     // only allow stroke to be set on individual branches
     ".branch": {
-      "stroke-width": (d) => d["stroke-width"] + "px", // style - as per drawBranches()
-      stroke: (d) => strokeForBranch(d), // TODO: revisit if we bring back SVG gradients
-      cursor: (d) => d.visibility === NODE_VISIBLE ? "pointer" : "default",
+      "stroke-width": (d: PhyloNode) => d["stroke-width"] + "px", // style - as per drawBranches()
+      stroke: (d: PhyloNode) => strokeForBranch(d), // TODO: revisit if we bring back SVG gradients
+      cursor: (d: PhyloNode) => d.visibility === NODE_VISIBLE ? "pointer" : "default",
       visibility: getBranchVisibility
     }
   }
 };
+
+
+type SelectionOrTransition =
+   Selection<SVGGElement, PhyloNode, SVGSVGElement | null, unknown> |
+  Transition<SVGGElement, PhyloNode, SVGSVGElement | null, unknown>
+
+type UpdateCall = (selectionOrTransition: SelectionOrTransition) => void;
 
 
 /** createUpdateCall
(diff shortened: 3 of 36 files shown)


jameshadfield:
@genehack could you take a look at the types added / used here if you have a chance? 🙏

src/util/globals.js
```diff
@@ -201,9 +201,9 @@ export const isValueValid = (value) => {
   // booleans, valid strings & numbers are valid.
   return true;
 };
-export const strainSymbol = Symbol('strain');
-export const genotypeSymbol = Symbol('genotype');
-export const measurementIdSymbol = Symbol('measurementId');
+export const strainSymbol = 'strain';
```
jameshadfield, line 204: I presume this change is to make types easier?

I'd prefer to keep symbols - they are perfect for this use case. One downside to them is that they are skipped with things like `Object.keys()`, but I'd hope TS would be kind enough to flag this up for us?

victorlin: Yeah mainly for ease of typing. I can look into working with them as-is.

victorlin:
Progress update: I think I've renamed all the files that need to be renamed for sufficient types. I've temporarily turned off `strictNullChecks` to focus on the other typing issues and have addressed those. There are still several FIXMEs and `any` types that I'll come back to next week.

src/util/globals.js
victorlin: Addressed with with 3c3ef2dcabf6e5af70f9460eeefbeaf3dbc36363

tsconfig.json
```diff
@@ -31,6 +32,7 @@ Visit https://aka.ms/tsconfig.json for a detailed list of options.
 
     /* Type Checking */
     "strict": true,                                      /* Enable all strict type-checking options. */
+    "strictNullChecks": false,                           /* Allow unhandled false/null/undefined values to make incremental TypeScript adoption easier. */
```
jameshadfield, line 35: Eslint has the option of disabling a rule for a particular file - does tsc / typescript-eslint have the same? If so, that may be better as we can add a "todo" comment in those files and then incrementally move to adhering to this rule

genehack: I haven't read the whole Typescript issue thread, but the answer to this seems to be somewhere between "no" and "kinda via a PITA workaround". See [this SO question](https://stackoverflow.com/questions/63894511/can-you-turn-down-strict-mode-et-al-on-a-single-file-or-code-block) which leads to [a Typescript issue](https://github.com/Microsoft/TypeScript/issues/28306).

victorlin: Thanks for digging @genehack. The closest thing I could find in TypeScript docs is [`ts-nocheck`](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html#ts-check) which means "disable all rules for this file" (not great).

victorlin: Another bad alternative is the per-line `ts-ignore`/`ts-expect-error`:

1. This disables all rules for the line, which is similarly not great
2. Doing this for 443 violations of `strictNullChecks` seems overkill

victorlin: I can update the commit message of 8a0f4fa043297e7a8a5b1b22f77279cc613f8c0a to note that we've considered these alternatives.

jameshadfield: Thanks for the responses - pity about the answers tho! 

> I can update the commit message of https://github.com/nextstrain/auspice/commit/8a0f4fa043297e7a8a5b1b22f77279cc613f8c0a to note that we've considered these alternatives.

Sounds good. Ideally we would turn this rule on in the future as I think it'll catch a whole lot of bugs. 443 is daunting tho!

src/components/tree/phyloTree/types.ts
```diff
@@ -0,0 +1,260 @@
+import { NODE_NOT_VISIBLE, NODE_VISIBLE_TO_MAP_ONLY, NODE_VISIBLE } from "../../../util/globals";
+import { PhyloTree } from "./phyloTree";
+
+export type StringLayouts = "scatter"
+
+export type NumericLayouts = "rect" | "radial" | "unrooted" | "clock"
+
+export type Layout = NumericLayouts | StringLayouts
```
jameshadfield, line 8: Question - why split into StringLayouts and NumericLayouts? (and what's the difference?) I don't see either of those variables used elsewhere.

src/components/tree/phyloTree/types.ts
```diff
@@ -0,0 +1,260 @@
...
+  }
+}
+
+export interface Genotype {
+  gene: string
+  positions: number[]
+  aa: boolean
+}
+
+export type ScaleType = "ordinal" | "categorical" | "continuous" | "temporal"
```
jameshadfield, line 39: Color scales can also have `scaleType = "boolean"`

src/components/tree/phyloTree/types.ts
```diff
@@ -0,0 +1,260 @@
...
+  arrayIdx?: number
+
+  branch_attrs?: {
+    mutations?: {
+      [gene: string]: string[]
+    }
+    labels?: Record<string, any>
+  }
+  children?: ReduxNode[]
+  currentGt?: number
```
jameshadfield, line 86: [to check] `currentGt` is (or at least can be) an array (e.g. `node.currentGt = [...newState]`). But perhaps this is the same property name on a different object?

src/components/tree/phyloTree/types.ts
```diff
@@ -0,0 +1,260 @@
...
+  showRegression?: boolean
+  x?: string
+  xContinuous?: boolean
+  xDomain?: number[]
+  xTemporal?: boolean
+  y?: string
+  yContinuous?: boolean
+  yDomain?: number[]
+  yTemporal?: boolean
+}
```
jameshadfield, line 205: [question] this type can take a few different forms I think - an empty object or one with at least some of these properties (e.g. `showBranches`, `showRegression`). Is this expressable in TS?

jameshadfield (commented):
(Incomplete review - just posting now as I won't return to this until next week)

Looks good @victorlin - I'm only part way through the really big commit so far but will pick this up later. No action necessary at the moment I don't think.

I think I can review the correctness, but I haven't written enough TS to give advice about the big scale approach.

src/components/tree/phyloTree/types.ts
victorlin: All the properties are optional. Do you mean certain sets of properties always appear together? There might be a way with [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

src/components/tree/phyloTree/types.ts
victorlin: Oops, this is leftover from b37a77c9555b11907c6698e6a1d07ab94533d00f. I'll update to `Layout = "rect" | "radial" | "unrooted" | "clock" | "scatter"`

src/actions/tree.ts
```diff
@@ -21,28 +25,28 @@ import { calcFullTipCounts, calcTipCounts } from "../util/treeCountingHelpers";
  * @param {Int} idx - index of displayed root node
```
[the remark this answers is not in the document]
victorlin: Yes, I meant to remove JSdoc for converted functions. This one slipped through the cracks. Let me know if you see any others!

src/components/tree/phyloTree/types.ts
genehack: IMO, if some of the options have to co-occur with others, defining a type for each validation combination and then combining them in a union type would be the way to go.

src/components/tree/phyloTree/phyloTree.ts
```diff
@@ -8,9 +8,89 @@ import * as grid from "./grid";
 import * as confidence from "./confidence";
 import * as labels from "./labels";
 import * as regression from "./regression";
+import { Callbacks, Layout, Params, PhyloNode, ReduxNode, ScatterVariables, Visibility } from "./types";
+import { Selection } from "d3";
+
+export interface PhyloTree {
```
genehack, line 14: why not put this in `./types`?

src/components/tree/phyloTree/renderers.ts
```diff
@@ -338,11 +391,11 @@ export const updateColorBy = function updateColorBy() {};
 /** given a node `d` which is being hovered, update it's colour to emphasize
  * that it's being hovered. This updates the SVG element stroke style in-place
  * _or_ updates the SVG gradient def in place.
- * @param {PhyloNode} d node
- * @param {string} c1 colour of the parent (start of the branch)
- * @param {string} c2 colour of the node (end of the branch)
+ * @param d node
+ * @param c1 colour of the parent (start of the branch)
+ * @param c2 colour of the node (end of the branch)
```
genehack, line 396: remove

src/components/tree/phyloTree/phyloTree.ts
```diff
@@ -8,9 +8,89 @@ import * as grid from "./grid";
...
+  hideGrid: typeof grid.hideGrid
+  addGrid: typeof grid.addGrid
+  showTemporalSlice: typeof grid.showTemporalSlice
+  hideTemporalSlice: typeof grid.hideTemporalSlice
+
+  confidencesInSVG: boolean
+  regression?: regression.Regression
+  layout: Layout
+  svg: Selection<SVGSVGElement | null, unknown, null, any>
+  distance: "num_date" | "div"
```
genehack, line 71: you define a `Distance` type in `./types`; use it here.

src/components/tree/phyloTree/types.ts
```diff
@@ -0,0 +1,260 @@
...
+
+export interface Callbacks {
+  onTipHover: (d: PhyloNode) => void
+  onTipLeave: (d: PhyloNode) => void
+  onTipClick: (d: PhyloNode) => void
+  onBranchHover: (d: PhyloNode) => void
+  onBranchLeave: (d: PhyloNode) => void
+  onBranchClick: (d: PhyloNode) => void
+  tipLabel: (d: PhyloNode) => void
+}
```
genehack, line 260: i would be tempted to define the function sig once and then use that for the callback values:

```
interface CallbackMethodSig {(d:PhyloNode)=> void }
export interface Callbacks {
  onTipHover: CallbackMethodSig
  onTipLeave: CallbackMethodSig
...
```

src/components/tree/phyloTree/types.ts
```diff
@@ -0,0 +1,260 @@
+import { NODE_NOT_VISIBLE, NODE_VISIBLE_TO_MAP_ONLY, NODE_VISIBLE } from "../../../util/globals";
+import { PhyloTree } from "./phyloTree";
```
genehack, line 2: yeah, this to me really argues the PhyloTree interface should be defined and exported from this file.

src/components/tree/phyloTree/types.ts
genehack, line 1: it would be nice if there was some sort of overall organization to this file -- whether just a simple alphasort on the interface names, or some sort of logical structure, i'm not sure, but right now it feels very much like a junk drawer.

src/util/colorScale.ts
```diff
@@ -10,19 +9,23 @@ import { isColorByGenotype, decodeColorByGenotype } from "./getGenotype";
...
 
 /**
  * calculate the color scale.
- * @param {string} colorBy - provided trait to use as color
- * @param {object} controls
- * @param {object} tree
- * @param {object} treeToo
- * @param {object} metadata
- * @return {{scale: function, continuous: string, colorBy: string, version: int, legendValues: Array, legendBounds: Array|undefined, genotype: null|object, scaleType: null|string, visibleLegendValues: Array}}
+ * @param colorBy - provided trait to use as color
```
genehack, line 20: remove

src/util/treeCountingHelpers.ts
```diff
@@ -1,16 +1,24 @@
...
-* @param {Array} nodes - list of nodes
-* @param {Array} traits - list of traits to count across the tree
-* @param {Array | false} visibility - if Array provided then only consider visible nodes. If false, consider all nodes.
-* @param {bool} terminalOnly - only consider terminal / leaf nodes?
-* @return {obj} keys: the traits. Values: an object mapping trait values -> INT
+* @param nodes - list of nodes
+* @param traits - list of traits to count across the tree
+* @param visibility - if Array provided then only consider visible nodes. If false, consider all nodes.
+* @param terminalOnly - only consider terminal / leaf nodes?
+* @return keys: the traits. Values: an object mapping trait values -> INT
```
genehack, line 11: delete

src/util/treeCountingHelpers.ts
```diff
@@ -95,9 +100,9 @@ export const calcTotalTipsInTree = (nodes) => {
 /**
 * for each node, calculate the number of subtending tips (alive or dead)
 * side effects: n.fullTipCount for each node
-* @param {Node} node - deserialized JSON root to begin traversal
+* @param node - deserialized JSON root to begin traversal
```
genehack, line 103: delete

genehack (commented):
picked some nits.

general feedback:
* if you're going to have a `types.ts` file, centralize _all_ the types and interfaces on that level into that file (e.g., phylotree)
* there are *many* methods where the arguments got typed, but the return value of the method did not -- i would urge you to try to always provide a type for the return value too

tsconfig.json
victorlin: An alternative is to use a baseline (a concept I discovered [elsewhere](https://github.com/nextstrain/augur/issues/1472#issuecomment-2452297580)). There's no official TypeScript support, but a 3rd party tool [tsc-baseline](https://github.com/TimMikeladze/tsc-baseline) seems promising. Added: 8992782bf1ebf8cb1a104567433ac00625224201

A slight downside is that, since tsc-baseline has not gained much popularity, there is no support beyond the command line. This means code editors such as VS Code's TypeScript integration will use `tsc` and still report all the rule violations. I wouldn't consider this blocking though. Here's a workaround: 1960d5a3f9dc55a7946c9129ba67732db09081d9

src/components/tree/phyloTree/change.ts
```diff
@@ -74,33 +77,43 @@ const svgSetters = {
  * the SVG elements.
  * svgSetters (see above) are used to actually modify the property on the element,
  * so the given property must also be present there!
- * @param {string} treeElem (e.g. ".tip" or ".branch")
- * @param {list} properties (e.g. ["visibiliy", "stroke-width"])
- * @return {function} used in a d3 selection, i.e. d3.selection().methods().call(X)
+ * @return function used in a d3 selection, i.e. d3.selection().methods().call(X)
```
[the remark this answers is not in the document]
victorlin: I will go through and add return types on all functions that are touched in this PR.

src/components/tree/phyloTree/types.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/1960d5a3f9dc55a7946c9129ba67732db09081d9/src/components/tree/phyloTree/types.ts#L4

src/actions/tree.ts
```diff
@@ -364,7 +402,7 @@ const _traverseAndCreateSubtrees = (root, node, attr) => {
 /**
  * sort the subtrees by the order the trait would appear in the legend
  */
-const _orderSubtrees = (metadata, nodes, attr) => {
+const _orderSubtrees = (metadata: any, nodes: ReduxNode[], attr: string) => {
```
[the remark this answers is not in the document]
victorlin: Added types: https://github.com/nextstrain/auspice/blob/3cf6768919650f4294a8c7a213cfa11dda5fc462/src/components/tree/phyloTree/types.ts#L357-L371

src/components/tree/index.ts
```diff
@@ -1,8 +1,9 @@
-import { connect } from "react-redux";
+import { connect, MapStateToProps } from "react-redux";
 import UnconnectedTree from "./tree";
 import { RootState } from "../../store";
+import { TreeComponentPropsFromState } from "./tree";
 
-const Tree = connect((state: RootState) => ({
+const mapStateToProps: MapStateToProps<TreeComponentPropsFromState, Record<string, never>, RootState> = (state: RootState) => ({
```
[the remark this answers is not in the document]
victorlin: Updated to `TreeComponentOwnProps`:

https://github.com/nextstrain/auspice/blob/3cf6768919650f4294a8c7a213cfa11dda5fc462/src/components/tree/index.ts#L7

src/components/tree/phyloTree/types.ts
victorlin: Yes, I've been throwing things in as they come. I'll try to come up with a logical structure once contents are finalized.

I've done alphasort on the long lists of keys within interfaces, but maybe a logical structure would be better there too.

src/components/tree/phyloTree/renderers.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/1960d5a3f9dc55a7946c9129ba67732db09081d9/src/components/tree/phyloTree/renderers.ts#L390-L395

src/components/tree/phyloTree/phyloTree.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/3cf6768919650f4294a8c7a213cfa11dda5fc462/src/components/tree/phyloTree/types.ts#L211

src/components/tree/phyloTree/types.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/3cf6768919650f4294a8c7a213cfa11dda5fc462/src/components/tree/phyloTree/types.ts#L211

src/components/tree/phyloTree/types.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/1960d5a3f9dc55a7946c9129ba67732db09081d9/src/components/tree/phyloTree/types.ts#L47

src/util/colorScale.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/1960d5a3f9dc55a7946c9129ba67732db09081d9/src/util/colorScale.ts#L18-L23

src/util/treeCountingHelpers.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/1960d5a3f9dc55a7946c9129ba67732db09081d9/src/util/treeCountingHelpers.ts#L5-L20

src/util/treeCountingHelpers.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/1960d5a3f9dc55a7946c9129ba67732db09081d9/src/util/treeCountingHelpers.ts#L102-L109

src/components/tree/phyloTree/helpers.ts
```diff
@@ -16,10 +17,13 @@ export const getDomId = (type, strain) => {
...
  * this function takes a call back and applies it recursively
  * to all child nodes, including internal nodes
- * @param {PhyloNode} node
- * @param {Function} func - function to apply to each children. Is passed a single argument, the <PhyloNode> of the children.
  */
-export const applyToChildren = (phyloNode, func) => {
+export const applyToChildren = (
+  phyloNode: PhyloNode,
+
+  /** function to apply to each children. Is passed a single argument, the <PhyloNode> of the children. */
```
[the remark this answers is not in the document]
victorlin: Applied: https://github.com/nextstrain/auspice/blob/3cf6768919650f4294a8c7a213cfa11dda5fc462/src/components/tree/phyloTree/helpers.ts#L24

src/components/tree/phyloTree/phyloTree.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/3cf6768919650f4294a8c7a213cfa11dda5fc462/src/components/tree/phyloTree/types.ts#L268

src/components/tree/phyloTree/types.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/2af5974e4631a6768727b2a21bae1182bb354047/src/components/tree/phyloTree/types.ts#L347-L357

src/components/tree/phyloTree/types.ts
victorlin: I see: https://github.com/nextstrain/auspice/blob/2af5974e4631a6768727b2a21bae1182bb354047/src/util/setGenotype.js#L28

That file hasn't been typed yet. Changing the type to `currentGt?: number | number[]` doesn't work as-is because it is assumed to be a scalar for `PhyloNode.px`:

https://github.com/nextstrain/auspice/blob/2af5974e4631a6768727b2a21bae1182bb354047/src/components/tree/phyloTree/layouts.ts#L95

Do you think it's worth addressing in this PR? I could also just leave a comment that the type is incomplete.

src/components/tree/phyloTree/change.ts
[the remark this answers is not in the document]
victorlin: Done: eec456cde045785312a8182e2157da0184a0a056

src/components/tree/phyloTree/types.ts
jameshadfield: Two points:

> That file hasn't been typed yet. Changing the type to currentGt?: number | number[] doesn't work as-is because it is assumed to be a scalar for PhyloNode.px:

Within that function we create an array `node.currentGt = [...newState];` then convert to a string `nodes.forEach((n) => {n.currentGt = n.currentGt.join(' / ');});`. I _think_ everywhere else in the code expects a string. So I guess keeping the type as `currentGt?: str` will flag this up when we eventually type this file? (And at that point the code should be fixed!)

Secondly, the `currentGt` is (AFAICT) always a string, not a number.

src/util/colorScale.ts
```diff
@@ -309,7 +309,7 @@ function createTemporalScale(
     vals = vals.sort();
     domain = [rootDate];
     const n = 10;
-    const spaceBetween = parseInt(vals.length / (n - 1), 10);
+    const spaceBetween = vals.length / (n - 1);
```
jameshadfield, line 345: There's a bug with this change which results in temporal colour scales failing (everything's black) when `vals.length` is not an integer multiple of `(n-1)` - `spaceBetween` must be a integer so that we can use it as an array index below (`vals[spaceBetween*i`).

src/util/colorScale.ts
```diff
@@ -183,13 +183,13 @@ function createScaleForGenotype(
   const range = [unknownColor, ...genotypeColors.slice(0, trueValues.length)];
   // Bases are returned by orderOfGenotypeAppearance in order, unknowns at end
   if (legendValues.indexOf("-") !== -1) {
-    range.push(rgb(217, 217, 217));
+    range.push(rgb(217, 217, 217).formatHex());
```
jameshadfield, line 198: FYI, [here's an example URL](https://next.nextstrain.org/mpox/clade-IIb?c=gt-OPG001_1&gt=OPG001.1-) and the results are unchanged with this PR:

image

jameshadfield (changes requested):
This is a big undertaking - thanks for persevering Victor. It's already apparent how many bugs this has identified and will really help going forward. A few of my comments are blocking, but most are questions only.

tsconfig.json
jameshadfield: I'll defer to @genehack but the added complexity of `tsc-baseline` seems too much, although I've not used it enough to make a call here.

src/components/tree/phyloTree/types.ts
```diff
@@ -0,0 +1,260 @@
...
+
+  crossDepth?: number
+  depth?: number
+  displayOrder?: number
+  displayOrderRange?: number[]
+  fill?: string
+  inView: boolean
+  n: ReduxNode
+  pDepth?: number
+  px?: number
```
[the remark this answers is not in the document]
victorlin: This is what I was trying to do with b37a77c9555b11907c6698e6a1d07ab94533d00f. I think it has to start at the `PhyloTree` level since that is where layout is set and checked. I've made a second attempt using [mapped types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html): 9da17b783f1e1d4d4c0fc05b33a639dc883c09b8

src/components/tree/phyloTree/types.ts
victorlin: Trying [mapped types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) in https://github.com/nextstrain/auspice/pull/1864#discussion_r1828775085

src/util/treeJsonProcessing.ts
```diff
@@ -1,27 +1,29 @@
...
- * @param  {array} nodes redux tree nodes
- * @return {Object} ret
- * @return {Set} ret.nodeAttrKeys collection of all `node_attr` keys whose values are Objects
- * @return {Array} ret.nodes input array (kinda unnecessary)
- *
- * side-effects: node.hasChildren (bool) and node.arrayIdx (INT) for each node in nodes
+ * Adds the following properties to each node:
+ * - fullTipCount
+ * - hasChildren
+ * - arrayIdx
```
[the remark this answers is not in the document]
victorlin: Yeah, there's an abundance of optional properties right now. It would be good to narrow those down by encoding the conditions somehow. I initially thought about putting those off in this PR by turning off `strictNullChecks`, but I think it's worth addressing now since it'll impact overall organization of types.

> Define `fullTipCount` as a non-optional property on `ReduxNode`

This is not trivial. The non-optional property will need to be used everywhere ReduxNode is initialized, such as: https://github.com/nextstrain/auspice/blob/5fccbc7d0aefd510e6aa68e162a699100a228859/src/util/treeJsonProcessing.ts#L74-L79

which will cause further conflicts such as the last line there where `ReduxNode.parent` is typed as `ReduxNode` which would require `fullTipCount` and error.

> `processNodes` would take an argument `nodes: Omit<ReduxNode, "fullTipCount">[]` and return `nodes: ReduxNode[]`

This wouldn't make any meaningful difference to the TypeScript compiler nor us. The docstring of this function makes it clear that `fullTipCount` is added here, and nothing is after calling this function to assert that `fullTipCount` is indeed set.

I think the only options are (note: I'm still kinda new to TypeScript)

1. non-null assertion operators (not great, [not allowed](https://typescript-eslint.io/rules/no-non-null-assertion) by `@typescript-eslint/recommended`)
2. type narrowing by handling alternate code paths where the value is unset

src/components/tree/tree.tsx
```diff
@@ -1,10 +1,11 @@
 import React from "react";
-import { withTranslation } from "react-i18next";
+import { withTranslation, WithTranslation } from "react-i18next";
 import { FaSearchMinus } from "react-icons/fa";
 import { updateVisibleTipsAndBranchThicknesses } from "../../actions/tree";
 import Card from "../framework/card";
 import Legend from "./legend/legend";
-import PhyloTree from "./phyloTree/phyloTree";
+import PhyloTreeConstructor, { PhyloTree } from "./phyloTree/phyloTree";
```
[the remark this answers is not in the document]
victorlin: It's worth noting that the type is often used to represent the object, so `PhyloTreeProps` would seem weird in contexts [such as](https://github.com/nextstrain/auspice/blob/12aff241c7fed96675ee691b445ffa6cd72552ea/src/components/tree/reactD3Interface/change.ts#L6-L11):

```ts
export const changePhyloTreeViaPropsComparison = (
  mainTree: boolean,
  phylotree: PhyloTreeProps,
  oldProps: TreeComponentProps,
  newProps: TreeComponentProps,
): {
```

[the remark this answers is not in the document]
jameshadfield: The uppercase pattern works nicely when you have a lowercase object (e.g. `thing`) and uppercase type (`Thing`). It breaks down here where we have an object to be used as a constructor which typically is uppercased (e.g. `PhyloTree`). 

How about `PhyloTree` + `PhyloTreeType`.

src/util/colorScale.ts
jameshadfield: Didn't know about `Math.trunc` but it looks perfect. I've used `parseInt(String(x), 10)` in the past to appease linters but in reality `parseInt` will cast to a string anyway.

src/components/tree/tree.tsx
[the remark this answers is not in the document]
victorlin: `PhyloTree` + `PhyloTreeType` sounds good. Though if we end up settling on the mapped types approach in https://github.com/nextstrain/auspice/pull/1864#discussion_r1828775085, I think it's fine to overload `PhyloTree` with the mapped `PhyloTree` type:

- Most references will use `PhyloTreeBase`
- References to the `PhyloTree` will be contextualized with the layout, for example `PhyloTree["rect"]`: https://github.com/nextstrain/auspice/blob/12aff241c7fed96675ee691b445ffa6cd72552ea/src/components/tree/phyloTree/layouts.ts#L59

I'll hold off on any rename until we decide whether to use that approach.

src/components/tree/tree.tsx
```diff
@@ -17,17 +18,102 @@ import { attemptUntangle } from "../../util/globals";
...
+
+export interface TreeComponentProps extends WithTranslation, TreeComponentPropsFromState {
+  dispatch: AppDispatch
+  height: number
+  width: number
+}
+
+// This is duplicated from RootState, but good to be explicit about what's
+// expected here.
+export interface TreeComponentPropsFromState {
```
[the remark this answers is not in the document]
jameshadfield: > I'm not sure about moving TreeComponentPropsFromState (the anchor point of this thread) and TreeComponentState. I think it makes sense to keep those near the component being described.

Agreed. I guess my general thinking is that the source of (most) state in Auspice is in the reducers, and the types should be co-located with that, and then state & types flow down from that. For props / react-component-state I agree it should be co-located with the component.

src/components/tree/phyloTree/types.ts
```diff
@@ -0,0 +1,260 @@
...
+}
+
+export interface ReactTreeNode {
+  existingFilterState?: "active" | "inactive" | null
+  idx?: number
+  isBranch: boolean
+  name?: string
+  node?: PhyloNode
+  treeId?: string
+}
```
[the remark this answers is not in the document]
jameshadfield: Thanks!

> This just shows how it's easy to write types that satisfy the compiler but accuracy isn't guaranteed!

This is probably my big takeaway from this PR. I'd say that accuracy _is_ being guaranteed by our use of so many optional parameters, but we're loosing a lot of the power types should give us this way.

src/components/tree/tree.tsx
```diff
@@ -110,11 +200,16 @@ class Tree extends React.Component {
     if (Object.keys(newState).length) this.setState(newState);
   }
 
-  componentWillUnmount() {
+  override componentWillUnmount() {
```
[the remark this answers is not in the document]
jameshadfield: Thanks! (No need to split into separate commit - up to you)

tsconfig.json
victorlin: `tsc-baseline` was worth the shot but I don't think it's the right move.

Added a post-merge task:

> - Post-merge: create issue to address `strictNullChecks` violations (started with 260ba0020e78d7ea4a071c290d32fc2a53e67732)

src/util/treeJsonProcessing.ts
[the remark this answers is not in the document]
jameshadfield: > This is not trivial. The non-optional property will need to be used everywhere ReduxNode is initialized

Oh yeah, the nested nature of these objects makes changing types as data flows through functions difficult. I also don't have enough experience to know how to do this effectively in practice. Let's continue thinking about these things post this PR tho.

> This wouldn't make any meaningful difference to the TypeScript compiler nor us. The docstring of this function makes it clear that fullTipCount is added here, and nothing is done after calling this function to assert that fullTipCount is indeed set.

But it should make a big difference to something that uses a `ReduxNode` object and wants to use `fullTipCount` - currently tsc is always going to say "this property may be undefined" whereas ideally we'd have a typesystem which guaranteed it's presence or absence.

src/components/tree/phyloTree/types.ts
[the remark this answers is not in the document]
jameshadfield: Oh cool! Shall we add those commits as a separate PR (I realise they're not part of this one) and we can continue discussion there? I think it'd be a good direction, but unsure of the difficulty / priority.

src/components/tree/phyloTree/types.ts
jameshadfield: Similar to https://github.com/nextstrain/auspice/pull/1864#discussion_r1830060286 let's continue this discussion in a separate PR. I do think reducing the number of optional properties is going to be very powerful, but let's keep this PR as-is.

src/components/tree/phyloTree/types.ts
```diff
@@ -0,0 +1,260 @@
...
+  x: any
+
+  // TODO: This should be `string | number`, conditional on layout
+  y: any
+}
+
+// Apply fixes from above
+export type SVGProperty = Extract<keyof CSSStyleDeclarationWithHyphens | keyof CustomCSSStyleDeclaration, string>
+
+export interface PhyloNode extends Partial<Omit<CSSStyleDeclarationWithHyphens, keyof CustomCSSStyleDeclaration>>, CustomCSSStyleDeclaration {
```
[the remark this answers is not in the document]
jameshadfield: Much better - thanks!

re:

```js
  // TODO: This should be `string | number`, conditional on layout
  x: any
```

Any reason not to juse use `x: string | number`? (I'm sure there is and I'm missing it)

src/components/tree/phyloTree/types.ts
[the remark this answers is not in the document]
victorlin: I have them in this PR currently (the commit links go stale upon force-push) but yeah I think it'd be better to consider separately. Added a post-merge task:

> - Post-merge: consider using mapped types in a new PR

src/components/tree/phyloTree/types.ts
victorlin: Added a post-merge task:

> - Post-merge: consider using mapped types in a new PR

src/components/tree/tree.tsx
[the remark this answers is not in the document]
victorlin: Deferring mapped types to a separate PR. I'll rename to `PhyloTree` + `PhyloTreeType`.

src/components/tree/phyloTree/types.ts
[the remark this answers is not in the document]
victorlin: Without narrowing down across all usages*, we get errors when one or the other is assumed. Example: https://github.com/nextstrain/auspice/blob/12aff241c7fed96675ee691b445ffa6cd72552ea/src/components/tree/phyloTree/layouts.ts#L361-L365

Error:

> Operator '<' cannot be applied to types 'string | number' and 'number'.

\* we could easily do so by adding type guards such as `typeof d.x === "number" && d.x < minX`, but there are 23 violations and I think the [mapped types approach](https://github.com/nextstrain/auspice/pull/1864#discussion_r1828775085) is the better solution.

src/components/tree/phyloTree/types.ts
victorlin: @genehack re: union

> defining a type for each validation combination and then combining them in a union type would be the way to go

This doesn't work. Example:

```ts
interface RadialPhyloNode {
  x: number
  y: number
  angle: number
  // …
}

interface ScatterPhyloNode {
  x: string
  y: string
  // ...
}

export interface PhyloNode extends RadialPhyloNode, ScatterPhyloNode
```

There is a type error:

> Interface 'PhyloNode' cannot simultaneously extend types 'RadialPhyloNode' and 'ScatterPhyloNode'.
> Named property 'x' of types 'RadialPhyloNode' and 'ScatterPhyloNode' are not identical.

Even if this did work, it's disconnected from the underlying conditions (layout) and prone to error. We'd only be using the combined type which makes everything available, and there is no validation that `angle` is only used for radial layout.

src/components/tree/tree.tsx
```diff
@@ -22,12 +26,71 @@ export const spaceBetweenTrees = 100;
...
+  treeToo: TreeTooState
+}
+
+export interface TreeComponentState {
+  hoveredNode?: {
+    node: PhyloNode
+    isBranch: boolean
+  } | null
+  tree?: PhyloTreeType | null
+  treeToo?: PhyloTreeType | null
```
victorlin, line 76: I've replaced e20a865b18d78461b20542d41e07df2cff6cd7bf with 7d6613315cbc0cca4610d3e0c83f818b9dbffa41. However, I'm wondering if something like bdc91da7ac2e4816b27738b1e86fa52495fa8063 is a better solution. I can't tell any difference from some basic testing.

src/components/tree/tree.tsx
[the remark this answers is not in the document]
victorlin: Done in [force-push](https://github.com/nextstrain/auspice/compare/12aff241c7fed96675ee691b445ffa6cd72552ea..9549dfbb63295ebd7b01efc6cd5c9a118d9f84b8) (among other changes)

src/components/tree/tree.tsx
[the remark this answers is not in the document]
victorlin: Split into c0daffa7.

While doing that, I also split some other commits off of the big typing commit: ecae75b0, 70c8f469, 6c18a0bc

src/util/treeJsonProcessing.ts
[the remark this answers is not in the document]
jameshadfield: @genehack and I discussed this general topic and one avenue we could explore (not in this PR!) is to have a type `A` with lots of optional properties that we can use as we build up the data structure then turn this into a type `B` once it arrives in redux and we know all/most properties have been defined. [Type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) may be useful here. Of everything in Auspice tree nodes are probably the worst to type due to the way we modify them in so many different places and do it on-the-fly rather than using immutable structures that redux suggests.

src/components/tree/phyloTree/types.ts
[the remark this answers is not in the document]
victorlin: Relatedly, I've added this snippet to the commit message of a1641c64:

> Note that there are many optional properties in the newly defined interfaces. This is not ideal and doesn't capture the fact that many properties are conditional on things such as layout. But it's a starting point to be improved over time.

src/util/treeJsonProcessing.ts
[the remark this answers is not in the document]
victorlin: I'm not quite sure if mapped types is the right approach for the `fullTipCount` example, but it seems related and I'll consider this in the same post-merge task

> - Post-merge: consider using mapped types in a new PR

src/components/tree/tree.tsx
jameshadfield: https://github.com/nextstrain/auspice/commit/7d6613315cbc0cca4610d3e0c83f818b9dbffa41 is identical to https://github.com/nextstrain/auspice/commit/bdc91da7ac2e4816b27738b1e86fa52495fa8063, no? What's the advantage of this approach, React's `setState` will [shallow merge the provided object](https://react.dev/reference/react/Component#setstate) anyway. 

As for https://github.com/nextstrain/auspice/commit/e20a865b18d78461b20542d41e07df2cff6cd7bf, using optional properties like `{tree?: PhyloTree, treeToo?: PhyloTree}` isn't ideal because if hovered node is set then it will always have `tree` on it (it's not optional). As for `hoveredNode?: ...` vs `hoveredNode: type | null` they seem pretty similar for me.

victorlin: Sorry, I copied the wrong SHA. 9549dfbb63295ebd7b01efc6cd5c9a118d9f84b8 vs bdc91da7ac2e4816b27738b1e86fa52495fa8063

victorlin: Oh, I didn't realize React does shallow merge. I think there's an issue with the type definition of `setState` then... there should be no need for either of the two changes.

jameshadfield: I'm still not sure I understand

> https://github.com/nextstrain/auspice/commit/bdc91da7ac2e4816b27738b1e86fa52495fa8063 update state instead of replacing with a partial

`setState` will never replace the state with the provided state it'll shallow merge it in, so we don't have to do that merge. I.e. `this.setState({...this.state, ...newState})` is equivalent to `this.setState(newState)` (but slower and less clear).

victorlin: There is a StackOverflow thread on this: [allow typescript compiler to call setState on only one react state property](https://stackoverflow.com/q/37300933)

The answer says:

> If you still have this error then you might want to consider updating your react definitions.

I just checked and we are using `@types/react` version 17.0.39 ([published in 2022](https://www.npmjs.com/package/@types/react/v/17.0.39)):

https://github.com/nextstrain/auspice/blob/9549dfbb63295ebd7b01efc6cd5c9a118d9f84b8/package-lock.json#L17434-L17435

Looks like the type has changed since that answer. It is now:

```ts
// We MUST keep setState() as a unified signature because it allows proper checking of the method return type.
// See: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18365#issuecomment-351013257
// Also, the ` | S` allows intellisense to not be dumbisense
setState<K extends keyof S>(
    state: ((prevState: Readonly<S>, props: Readonly<P>) => (Pick<S, K> | S | null)) | (Pick<S, K> | S | null),
    callback?: () => void
): void;
```

[This answer](https://stackoverflow.com/a/55824499) (less popular but more recent) recommends `setState<never>()`. It seems the most appropriate for our use case.

victorlin: Using `setState<never>()` in 7cefd6c545a2dff13e9c17536db11f9b37efb7d1

victorlin: > Edit: this message was written prior to reading the above response

This happened to me too. I think GitHub has real-time updates on issues but not PR comment threads. Unfortunate.

src/components/tree/phyloTree/types.ts
victorlin: Implemented a logical sort in the latest changes:

https://github.com/nextstrain/auspice/blob/7cefd6c545a2dff13e9c17536db11f9b37efb7d1/src/components/tree/phyloTree/types.ts#L13

https://github.com/nextstrain/auspice/blob/7cefd6c545a2dff13e9c17536db11f9b37efb7d1/src/components/tree/phyloTree/types.ts#L36

https://github.com/nextstrain/auspice/blob/7cefd6c545a2dff13e9c17536db11f9b37efb7d1/src/components/tree/phyloTree/types.ts#L50

https://github.com/nextstrain/auspice/blob/7cefd6c545a2dff13e9c17536db11f9b37efb7d1/src/components/tree/phyloTree/types.ts#L140

victorlin:
There's a few more small things I need to address in the checklist, but this should be ready for final review.

src/components/tree/phyloTree/change.ts
```diff
@@ -74,33 +80,42 @@ const svgSetters = {
...
-      });
-  }
-};
+function createUpdateCall(
+  treeElem: TreeElement,
+
+  /** e.g. ["visibility", "stroke-width"] */
+  properties: Set<SVGProperty>,
+):       (selection: Transition<SVGGElement, PhyloNode, SVGSVGElement, unknown>) => void {
+  return (selection: Transition<SVGGElement, PhyloNode, SVGSVGElement, unknown>) => {
```
jameshadfield, line 90: Wow I find TS hard to read sometimes!

Here we're typing the returned function signature twice, but (I think) we can drop the second duplicate signature.

```ts
function createUpdateCall(
  treeElem: TreeElement,
  properties: Set<SVGProperty>
): (selection: Transition<SVGGElement, PhyloNode, SVGSVGElement, unknown>) => void {
  return (selection) => {
    ...
```

Or even clearer (IMO):

```ts
type UpdateCall = (selection: Transition<SVGGElement, PhyloNode, SVGSVGElement, unknown>) => void;

function createUpdateCall(treeElem: TreeElement, properties: Set<SVGProperty>): UpdateCall {
  return (selection) => {
    ...
  }
}
```

src/components/tree/phyloTree/helpers.ts
```diff
@@ -156,15 +170,19 @@ export const formatDivergence = (divergence) => {
...
 /**
  * Are the provided nodes within some divergence / time of each other?
  * NOTE: `otherNode` is always closer to the root in the tree than `node`
  */
-function isWithinBranchTolerance(node, otherNode, distanceMeasure) {
+function isWithinBranchTolerance(
+  node: ReduxNode,
+  otherNode: ReduxNode,
+  distanceMeasure: Distance,
+) {
```
jameshadfield, line 185: [not a request for changes]

Here the return type is not specified, but tsc can infer it (`boolean`). The function below, `findFirstBranchWithAMutation`, specifies the return type (`ReduxNode`) but we can drop this and tsc will infer it. Any guidance on what style to follow?

src/util/colorScale.ts
```diff
@@ -109,7 +117,11 @@ export const calcColorScale = (colorBy, controls, tree, treeToo, metadata) => {
     };
   } catch (err) {
     /* Catch all errors to avoid app crashes */
-    console.error("Error creating color scales. Details:\n", err.message);
+    if (err instanceof Error) {
+      console.error("Error creating color scales. Details:\n", err.message);
+    } else {
+      console.error("Unknown error creating color scales. Details:\n", err);
+    }
```
jameshadfield, line 124: I've never seen non-Error-derived expressions thrown in Auspice, although I know it's technically possible. I'm not thrilled with this verbosity / conditional error messaging spreading throughout the code base so how about 
```ts
  } catch (err: any) {
    /* Catch all errors to avoid app crashes */
    console.error("Unknown error creating color scales. Details:\n", err?.message);
```

src/actions/tree.ts
```diff
@@ -61,16 +71,23 @@ export const applyInViewNodesToTree = (idx, tree) => {
...
+   * Change the in-view part of the tree.
+   *
+   * [0, 0]: reset. [undefined, undefined]: do nothing
+   */
+  root?: Root
+
+  cladeSelected?: string
+} = {}
+):       (dispatch: AppDispatch, getState: () => RootState) => void => {
+  return (dispatch: AppDispatch, getState: () => RootState) => {
```
jameshadfield, line 90: As commented elsewhere, these duplicate function types would be nice to avoid, unless you think they are somehow advantageous.

src/actions/tree.ts
```diff
@@ -200,48 +223,64 @@ export const changeDateFilter = ({newMin = false, newMax = false, quickdraw = fa
...
+    const d: AnyAction = {
       type: types.UPDATE_TIP_RADII, version: tree.tipRadiiVersion + 1
     };
     const tt = controls.showTreeToo;
     if (tipSelectedIdx) {
       d.data = calcTipRadii({tipSelectedIdx, colorScale, tree});
       if (tt) {
         const idx = strainNameToIdx(treeToo.nodes, tree.nodes[tipSelectedIdx].name);
-        d.dataToo = calcTipRadii({idx, colorScale, tree: treeToo});
+        d.dataToo = calcTipRadii({tipSelectedIdx: idx, colorScale, tree: treeToo});
```
jameshadfield, line 255: [note for posterity]

It's great we fixed this bug but note that we never call `updateTipRadii` with a `tipSelectedIdx` parameter, so this entire `if (tipSelectedIdx)` block is dead code.

src/reducers/tree.ts
```diff
@@ -0,0 +1,178 @@
...
+export interface ReduxNode {
+  /** the index of the node in the nodes array. set so that we can access visibility / nodeColors if needed */
+  arrayIdx?: number
+
+  branch_attrs?: {
+    mutations?: {
+      [gene: string]: string[]
+    }
+    labels?: Record<string, any>
+  }
```
jameshadfield, line 23: [not a request for changes]

This reminded me of Ivan's advice to type the underlying Auspice JSON and let types flow through from there. After doing this work do you think we've essentially typed the Auspice JSON by typing the structure it's processed into?

Also, `processBranchLabelsInPlace` turns this into `labels?: Record<string, string>` so something to consider when we start to distinguish between types in the data-constructing phase vs data-consuming phase.

jameshadfield (commented):
Read through the final code and made some minor comments.

For the upcoming work exploring reducing the optional parameters on types let's explore a few ideas at a small scale to better see which approach will be best for us!

src/util/colorScale.ts
victorlin: I think we should use `?` sparingly and am similarly not thrilled with its use spreading throughout the codebase.

In this case, `err?.message` would resolve to `undefined` instead of a potentially more meaningful value coming from `err` itself.

Would this still be too verbose?

```ts
} catch (err: unknown) {
  /* Catch all errors to avoid app crashes */
  const errorMessage = err instanceof Error ? err.message : String(err);
  console.error("Error creating color scales. Details:\n", errorMessage);
```

src/util/treeJsonProcessing.ts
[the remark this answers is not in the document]
victorlin: > turn this into a type `B` once it arrives in redux and we know all/most properties have been defined

This part isn't clear to me especially with the nested nature mentioned above. I'll create a separate issue to discuss further after this PR is merged.

src/reducers/tree.ts
jameshadfield: > For example, the bulk of [Metadata](https://github.com/nextstrain/auspice/blob/d5986de2d645958c68fc3f59fc3e1dd967eecb5e/src/metadata.ts#L3-L8) is still missing

I [started down this road a while ago](https://github.com/nextstrain/auspice/tree/james/ts), largely following Ivan's advice to start with the JSON and let types flow down. We/I should revisit!

src/util/colorScale.ts
victorlin: Done: https://github.com/nextstrain/auspice/blob/ad1f2cedbde4df6974412d37fd84119208902065/src/util/colorScale.ts#L118-L121

victorlin (commented):
Summarizing the force-push [above](https://github.com/nextstrain/auspice/compare/eb808a2450867c27174b9f10bbaa49a95300ab86..ad1f2cedbde4df6974412d37fd84119208902065):

- Rebased onto latest `master`
- Replaced `any`s with `unknown`
- Refactored placement and usage of [ChangeParams](https://github.com/nextstrain/auspice/blob/ad1f2cedbde4df6974412d37fd84119208902065/src/components/tree/phyloTree/types.ts#L187)
- Added "Set default value of focus to false" (302439fb)
- Added "Allow setState to take a partial state" (ad1f2ced)
- Added "Remove unused default value for performanceFlags" (9fecbee7)
- Updated "Properly handle unknown errors when creating color scales" (2af89645) for https://github.com/nextstrain/auspice/pull/1864#discussion_r1831736738

src/components/tree/phyloTree/change.ts
victorlin: After rebasing onto latest changes, I updated it to: https://github.com/nextstrain/auspice/blob/ad1f2cedbde4df6974412d37fd84119208902065/src/components/tree/phyloTree/change.ts#L77-L81

src/reducers/tree.ts
victorlin: Nice, I wasn't aware of that adventure! These look familiar... https://github.com/nextstrain/auspice/blob/ce33cbd65975e3891bb418b7fe6f3496db77e5cc/src/reducers/metadata.ts#L12-L31

src/components/tree/phyloTree/layouts.js
```diff
@@ -1,8 +1,7 @@
 /* eslint-disable no-multi-spaces */
 /* eslint-disable space-infix-ops */
 import { min, max } from "d3-array";
-import scaleLinear from "d3-scale/src/linear";
-import {point as scalePoint} from "d3-scale/src/band";
+import { scaleLinear, scalePoint } from "d3-scale";
```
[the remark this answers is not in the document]
victorlin: bundlesize [before](https://github.com/nextstrain/auspice/actions/runs/11713204361/job/32625506642) and [after](https://github.com/nextstrain/auspice/actions/runs/11714582605/job/32629505693?pr=1864) shows no significant difference.

image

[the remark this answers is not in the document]
jameshadfield: Nice! Thanks for double checking

src/reducers/tree.ts
jameshadfield: Yeah it was a bit of an adventure. Be fun to revisit it with the knowledge you've learned. I hadn't looked at it in months. This comment stood out](https://github.com/nextstrain/auspice/compare/master...james/ts#diff-c39c21eb00b72bc5c073924636b033832877770bc0b166a23654a99924be368fR7-R9) upon re-reading:

> Does it make sense to have two state types, a non-loaded one and a loaded one (discriminated type based on loaded: boolean) to give more assurances? Otherwise most properties are potential undefineds.

victorlin:
Summarizing the above [force-pushes](https://github.com/nextstrain/auspice/compare/83399300a046012abfde041cbb7ce14039009f9f..d6ecbc2beeaf8c769394aa559bf86627209bb3f5): I split `src/reducers/tree.js` into two files.

- `src/reducers/tree/index.ts`: the original file contents with types imported
- `src/reducers/tree/types.ts`: types live here now

Reasons:

- This now shows up as a rename for `tree.js`→ `index.ts`. Previously, the added types at the top of the file meant that the rename was not detected by default git rename thresholds.
- I think it's best practice to put types in a separate file if they are exported for external use. I'll add this to https://github.com/nextstrain/auspice/issues/1890 (will also need to re-align some existing code).

victorlin:
^ I had to fix some imports for the last force-push. While doing that I also converted `src/reducers/treeToo.ts` → `src/reducers/tree/treeToo.ts`.

src/components/tree/phyloTree/change.ts
```diff
@@ -297,6 +297,7 @@ export const change = function change(
     svgHasChangedDimensions = false,
     animationInProgress = false,
     changeNodeOrder = false,
+    focus = undefined,
```
joverlee521, line 300: > Set default value of focus to false 

```suggestion
    focus = false,
```

joverlee521 (commented):
Thank you for continuing to push on adding types! I mostly reviewed for my own learning purposes as I've been thinking about [adding types for the measurements panel](https://github.com/nextstrain/auspice/tree/measurements-types).

src/components/tree/tree.tsx
```diff
@@ -22,12 +26,72 @@ export const spaceBetweenTrees = 100;
...
+export interface TreeComponentOwnProps {
+  dispatch: AppDispatch
+  height: number
+  width: number
+}
+
+export interface TreeComponentProps extends WithTranslation, TreeComponentStateProps, TreeComponentOwnProps {}
+
+// This is duplicated from RootState, but good to be explicit about what's
+// expected here.
```
genehack, line 38: would defining RootState in terms of this interface reduce the duplication?

genehack (approved):
couple tiny non-blocking suggestions/questions.

src/components/tree/tree.tsx
victorlin: It's not trivial since thee same is different (e.g. `tree: state.tree` vs. `selectedNode: state.controls.selectedNode` vs. `colorings: state.metadata.colorings`): https://github.com/nextstrain/auspice/blob/3e23f8590764ce7863ad8e6a798ffc676008e0da/src/components/tree/index.ts#L6-L38

So it would need to be something like `type TreeComponentStateProps = PropsFromTreeState & PropsFromControlsState & PropsFromColoringsState …`.

I also think it's more logical to keep the RootState as the central source of truth.

victorlin: While looking into this, I realized that it makes more sense to put these types in a separate `types.ts` file so I've done that: 3e23f8590764ce7863ad8e6a798ffc676008e0da

victorlin:
I'm going to merge this. The types are largely in-progress with further follow-up to be done with https://github.com/nextstrain/auspice/issues/1888 and https://github.com/nextstrain/auspice/issues/1889, but we're at a good point now where this can be merged to allow new work to happen on top of these changes.

Issue #1709, opened 2023-10-09
Downstream PR preview creation and build slowness/timeouts


victorlin:
## Context

When working on UI changes, it's useful to previews the changes on nextstrain.org. Currently, [the process](https://github.com/nextstrain/auspice/blob/8377c4c47003468dab5b34cc02ab2d06eb952ca1/DEV_DOCS.md#test-on-downstream-repositories) is simple but it takes a long time, at least 20 minutes: >10m to make the PR, then >10m to build the preview version of nextstrain.org. The slowness seems to be mostly due to `npm install` being slow when resolving git references.

## Upstream issue

- https://github.com/npm/cli/issues/4896

## Past discussions

This has been disussed in https://github.com/nextstrain/auspice/pull/1567#issuecomment-1273892375 and https://github.com/nextstrain/auspice/pull/1661#issuecomment-1500667474. I thought it would be fixed with the upgrade to Node.js 16, but looks like it's still taking a while.

## Solutions

- For time to create the downstream PR: #1870
- For time to build the review app on Heroku: https://github.com/nextstrain/nextstrain.org/pull/1041

## Solutions that didn't work / weren't pursued

1. Accept the slowness and do things to fill the void while waiting, such as writing issues like this.
2. Try using `yarn`, maybe it's faster?
3. Do the `npm install` with `https://`
4. Replace `github:` with `https://` in package.json + package-lock.json
5. (in git config on both the workflow that creates the PR and the heroku build env) Use `url."ssh://git@".insteadOf git://` or `url."https://github".insteadOf "git://github"`

jameshadfield:
Recent changes to how nextstrain.org is built & deployed have improved this a lot. `npm` is still slower installing from GitHub cf. the npm registry, but I don't think it's worth switching to `yarn` or similar. Please re-open this issue if it's still frustrating for anyone!

victorlin:
Re-opening this because it is once again taking at least 20 minutes and even [timing out](https://github.com/nextstrain/nextstrain.org/pull/1025#issuecomment-2420582251) on Heroku builds.

victorlin:
Read through [the upstream issue](https://github.com/npm/cli/issues/4896) and added some more possible solutions which I'm trying in #1870.

victorlin:
In https://github.com/nextstrain/nextstrain.org/pull/1039 I tried some different workarounds mentioned in https://github.com/npm/cli/issues/4896 but none seemed to work:

[...]

PR #1870, opened 2024-10-18
Speed up review app installation

victorlin:

Reduce the time to create the preview PR from ~11m to ~3m.

> [!NOTE]
> I plan to drop e4adf4f5f2d348b447d44bad64644fb5443d1ae3 since it doesn't seem to have any effect: compare runs [before](https://github.com/nextstrain/auspice/actions/runs/11410864785/job/31754024500) and [after](https://github.com/nextstrain/auspice/actions/runs/11410911840/job/31754172608)
> EDIT: [keeping it](https://github.com/nextstrain/auspice/pull/1870#issuecomment-2426977204) as 2aba0cf9c320ca154484069435bb5ff7074ed2ee

## Related issue(s)

- #1709

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/preview_on_downstream_repo.yaml
+++ b/.github/workflows/preview_on_downstream_repo.yaml
@@ -42,6 +42,7 @@ jobs:
         shell: bash
         working-directory: ${{ inputs.directory }}
         run: |
+          npm config set ignore-scripts true
           npm ci
           npm install nextstrain/auspice#${{ github.sha }}
           git add package.json package-lock.json
--- a/.github/workflows/preview_on_downstream_repo.yaml
+++ b/.github/workflows/preview_on_downstream_repo.yaml
@@ -30,7 +30,7 @@ jobs:
     steps:
       - uses: actions/setup-node@v4
         with:
-          node-version: '16'
+          node-version: '20'
 
       - name: Checkout downstream repo
         uses: actions/checkout@v4
--- a/.github/workflows/preview_on_downstream_repo.yaml
+++ b/.github/workflows/preview_on_downstream_repo.yaml
@@ -41,6 +41,8 @@ jobs:
       - name: Install Auspice from PRs HEAD commit
         shell: bash
         working-directory: ${{ inputs.directory }}
+        env:
+          NODE_ENV: production
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/preview_on_downstream_repo.yaml
victorlin, line 1: Aside: Adding the label before opening the PR causes weird concurrency issues. See how these two were triggered at the same time:

image

Links:

- [canceled](https://github.com/nextstrain/auspice/actions/runs/11395018611)
- [running](https://github.com/nextstrain/auspice/actions/runs/11395018603)

victorlin: I thought they were different but they're actually the same. So this isn't really an issue and the concurrency rule is cancelling one as expected.

jameshadfield (approved):
Why drop https://github.com/nextstrain/auspice/commit/e4adf4f5f2d348b447d44bad64644fb5443d1ae3? Its commit message provides a compelling reason to keep it...

victorlin:
@jameshadfield I forgot to explain in detail – I added that commit thinking it would ignore installation of Auspice `devDependencies`, but it only applies to `devDependencies` in the current project, i.e. [nextstrain.org/auspice-client/package.json](https://github.com/nextstrain/nextstrain.org/blob/7617c384f3aa8fb3083c46a5c29f9f90db8c2558/auspice-client/package.json) which has no `devDependencies` at the moment. You have a good point though – if that every changes in the future, we would still want to ignore `devDependencies` during installation for the review app. I'll keep the commit with a revised message.

Side note: the reason that Auspice `devDependencies` were installed and causing slowness was due to having a [prepare script on a git install](https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-scripts). `NODE_ENV=production` has no effect in this case. It is addressed with c90a465715232c16fe81d71e61509825ad07fd90 and https://github.com/nextstrain/nextstrain.org/commit/c3d38f25911dfadbbb9161d9e98f907181ab1d5e elsewhere.

victorlin:
^ new preview PR is a bug. Will create a PR to fix.

Issue #1878, opened 2024-10-26
performance observations while interacting with a 23k-tip tree


jameshadfield:
A summary of observations while rendering & interacting with a huge tree. Ideally we'd like to have Auspice be performant on such big trees, but also this is a simple way to flag up slow parts of the code. The timings are just to indicate bottlenecks - while they are reasonably consistent for me they'll be different for you. (This issue is mainly for me to keep track of things in a public-facing way. If you'd like to work on anything please spin it out into a new issue.) 

### Setup: 

Dataset: `groups/blab-private/ncov/north-america/usa/washington/king-county/large?c=ns_kc&d=tree&p=full` (i.e. only tree panel rendering). 23k tips. The dataset is only 45MB, so lots of nodes but (relatively) few mutations (45MB minimised but uncompressed). 

Auspice 2.59.1 at master branch (35e4ca69) running in production mode with the following changes:
```diff
webpack.config.js
-    devtool: !devMode ? undefined : "eval-source-map",
+    devtool: "eval-source-map",
-      minimize: !devMode,
+      minimize: false,
```

Chrome 130.0 running in incognito mode on an Apple M3 Pro chip. GC triggered before each snapshot was taken.

## Test 1:

Filter to `f_region=South%20America` (n=259 tips selected). Start a snapshot then inactivate the filter, immediately hover over a tip so we can see when that listenerr will fire/render. This shares a lot of similarities to dragging the date range / animating but is much easier to reproduce consistently & isolate. `Trace-20241026T131320.json`

**Heap Size:** 230MB to 330Mb.
**Timings:**
1. 666ms JS "Event: click". This is our code & d3 code.
2. 240ms JS "Animation frame fired' (triggered from (1)). This is mostly d3 code.
3. 115ms recalculate style (triggered from (2))
4. 40ms (pre-paint, paint, layerize)  (no visual changes - looks identical)
5. 1170ms Commit 
   * During this time, partially-presented frames are rendered at 60fps
   * About ~100ms into the Commit cycle we get the tips all shown but not all the branches are correct. So it looks like partial d3 updates have been rendered?
   * All the other partially-presented frames look identical
6. <1ms Event: mouseover 
7. 3ms Recalculate style, layout, pre-paint
   * NOTE: no paint here
8. 75ms JS "Animation frame fired" (triggered from (1))
9. 80ms recalculate style (triggered from (6)) 
10. 45ms (pre-paint, paint, layerize)
[...]

PR #1879, opened 2024-10-28
Improve entropy recalculations

jameshadfield:
Improves the general speed of Auspice when the Entropy panel is not displayed (either toggled off or off-screen). The downside is a small amount of time where there's a blank entropy panel when it scrolls back into view after a state change which requires the entropy data to be recalculated. (The win is that we skip these recalculations so everything else is faster.)

Eyes  on the behaviour changes in this PR would be very welcome!

There's more we can do here, but I'll leave the following for subsequent PRs:
* I'd like to investigate shifting the entire recalculation off to a worker thread. The mutation data can be transferred once (it never changes) so I think the IO should be fast.
* I suspect the slowness is the myriad string slice operations - how much faster would it be to pre-process these and store as arrays. (And how much more memory would this take?)
* The debouncing could be smarter, perhaps utilising the `requestIdleCallback` API

Related to #1878


jameshadfield:
[Updated](https://github.com/nextstrain/auspice/compare/2c81abcc04c406e22db1591beb10583da4e68d96..4588d6ce850947e9aea1f9ac0b31e707f04630aa) to fix a bug when viewing narratives and also fix linting errors

src/components/entropy/index.js
```diff
@@ -193,16 +209,47 @@ class Entropy extends React.Component {
       select(this.d3entropy).selectAll(".handle--custom").style("visibility", "hidden");
     }
     this.setState({chart});
+
+    if (!props.bars) {
+      props.dispatch((dispatch, getState) => {
+        const { entropy, tree } = getState();
+        const [bars, maxYVal] = calcEntropyInView(tree.nodes, tree.visibility, entropy.selectedCds, entropy.showCounts);
+        dispatch({type: ENTROPY_DATA, data: bars, maxYVal});
+      });
```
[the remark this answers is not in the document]
jameshadfield: Oh, that's slipped through from an earlier version, probably from a rebase. I'll remove it, it's not needed. I'll similarly refactor the almost identical call within `visibilityOnScreenChange`. Thanks for looking!

src/components/entropy/index.js
```diff
@@ -193,16 +209,47 @@ class Entropy extends React.Component {
...
+      this.setUp(this.props); 
+      const observer = new IntersectionObserver(this.visibilityOnScreenChange.bind(this), {threshold: 0.0});
+      observer.observe(this.d3entropy)
     }
   }
   UNSAFE_componentWillReceiveProps(nextProps) {
     if (!nextProps.loaded) {
       this.setState({chart: false});
     }
+    if (!nextProps.bars) { // can happen on narrative slide changes
```
jameshadfield, line 250: There's a bug here. Fixing now...

src/components/entropy/index.js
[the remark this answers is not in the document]
jameshadfield: Block removed (it was essentially dead code, so no functional changes)

src/components/entropy/index.js
```diff
@@ -177,7 +178,21 @@ class Entropy extends React.Component {
       </div>
     );
   }
+  requestRecalculationOfBars() {
+    /* The <Entropy> component is now responsible for requesting a recalculation of the underlying entropy data
+    as this allows us to skip calculations when (e.g.) the panel is not rendered */
```
joverlee521, line 184: _non-blocking_

Hmm, this feels like an anti-pattern to me. I would expect the `TOGGLE_PANEL_DISPLAY` action to trigger the entropy re-calculation or clear the entropy data. Then there's a clear connection that the toggling of the panel display will trigger re-calculation.

src/components/entropy/index.js
```diff
@@ -209,9 +212,24 @@ class Entropy extends React.Component {
...
+    }
+    const onScreen = entries[0].isIntersecting;
+    if (onScreen===this.props.onScreen) return; // can happen when component initially rendered
+    // if gone off screen or come back on screen with the bars still valid then it's a simple toggle action
+    if (!onScreen || this.props.bars) {
+      return this.props.dispatch({type: ENTROPY_ONSCREEN_CHANGE, onScreen})
+    }
+    // else if back on screen and the bars are invalid then we want to regenerate them
+    this.requestRecalculationOfBars(onScreen);
+  }
```
joverlee521, line 227: _non-blocking_

Ditto, this feels like an anti-pattern. I would expect the one `ENTROPY_ONSCREEN_CHANGE` action and it would run the entropy re-calculation when appropriate.

src/components/entropy/index.js
```diff
@@ -178,13 +179,15 @@ class Entropy extends React.Component {
...
+  requestRecalculationOfBars(onScreen) {
     /* The <Entropy> component is now responsible for requesting a recalculation of the underlying entropy data
     as this allows us to skip calculations when (e.g.) the panel is not rendered */
     this.props.dispatch((dispatch, getState) => {
       const { entropy, tree } = getState();
       const [bars, maxYVal] = calcEntropyInView(tree.nodes, tree.visibility, entropy.selectedCds, entropy.showCounts);
-      dispatch({type: ENTROPY_DATA, data: bars, maxYVal});
+      const action = {type: ENTROPY_DATA, data: bars, maxYVal};
+      if (onScreen!==undefined) action.onScreen = onScreen;
+      dispatch(action);
```
joverlee521, line 190: _non-blocking_

> This results in slightly strange behaviour when the entropy panel will
be shown with no bars and they'll be drawn after a slight delay (while
the data's recalculated).

Would it be worth displaying a brief loading text (e.g. "Recalculating entropy") while waiting for the data?

joverlee521 (approved):
I played around with the `groups/blab/ncov-orf8ko/WA/20k` dataset and definitely see improvements in responsiveness! 

I only left non-blocking comments. In general, I find the split of management of entropy data between the Redux actions and Entropy component a bit weird. It feels like the Redux actions should trigger that entropy data updates appropriately and the Entropy component should just remain a consumer of that data.

src/components/entropy/index.js
jameshadfield: Agreed. I moved to this style both here and in the related comment. Thanks!

src/components/entropy/index.js
jameshadfield: Implemented in [3799a98](https://github.com/nextstrain/auspice/pull/1879/commits/3799a9873934684f71cc331e8844c607792dbc3c) - if people don't like it I'm happy to drop it, but I think it's an improvement in and of itself. Note that it's not entirely consistent -- when we change the filters and the tree & entropy panels are in view then we prioritize the tree update and defer entropy recalculations until that's done which leaves the old bars rendered until this time.

jameshadfield:
Rebased onto master which now includes [e3cfb27](https://github.com/nextstrain/auspice/pull/1879/commits/e3cfb275601e1d603596844a875f8f49f8f3f793) as part of #1883. Merging.

Issue #1857, opened 2024-09-12
Revisit Redux states for measurements


joverlee521:
Follow up to https://github.com/nextstrain/auspice/pull/1848

Redux state for measurements is a bit of a mess right now because of a few of reasons
1. deferred loading of the measurements JSON 
2. the ranking of state to use (url params > collection display defaults > app defaults)
3. narrative page changes with the same dataset but changes in URL params

Copying from @jameshadfield [comment](https://github.com/nextstrain/auspice/pull/1848#issuecomment-2346832018)

> Prior to this work, we load and parse the main JSON and read URL queries all in one redux step (largely `createStateFromQueryOrJSONs`). Sidecar JSONs are loaded via subsequent actions and there are no URL queries affecting these - `loadMeasurements` is the relevant function here. This prioritises rendering of the tree, with the downside that a measurements panel flashes in a second later ([example](https://next.nextstrain.org/community/blab/measurements-panel/flu/seasonal/h3n2/ha)). Narrative slide changes (query changes and/or dataset changes) operate the same way. 
> 
> This PR parses measurements panel URL queries within `createStateFromQueryOrJSONs` (although the JSON hasn't yet been parsed), and then reads them from redux state in `loadMeasurements`. Narrative slide changes where the dataset does change are akin to loading a new page. Narrative slide changes where the dataset doesn't change (but query does) apply these changes to the measurements data within `createStateFromQueryOrJSONs` (via `getCollectionDefaultControl`), and this is where it feels like the separation isn't as clean as it could be.
> 
> We discussed two alternative approaches:
> 1. the measurement URL queries are read within `loadMeasurements`. This is a nicer separation of concerns however narrative slide changes where the dataset doesn't change complicate this, as we never call `loadMeasurements`. We also want to process all changes within the same action to avoid out-of-sync rendering when a slide changes.
> 2. waiting for the measurements JSON to be fetched and loading everything within `createStateFromQueryOrJSONs`. The downside here is additional complexity in that function. The upsides are rendering all panels simultaneously and making it easier to one day implement measurements queries which modify the tree. 
>

joverlee521:
_Reminder for myself_

If the measurements JSON is fetched upfront and loaded in `createStateFromQueryOrJSONs`, then auspice.us [handling of files](https://github.com/nextstrain/auspice.us/blob/6aa71a694f98855dcae90939c2b78320dafae3ac/auspice_client_customisation/handleDroppedFiles.js#L200-L214) will need to updated as well.

PR #1848, opened 2024-09-09
Measurements URL query params

joverlee521:
## Description of proposed changes

Add URL query params for the measurements panel, see [docs for new query params](https://nextstrain--1848.org.readthedocs.build/projects/auspice/en/1848/advanced-functionality/view-settings.html#measurements-panel-url-query-options).
Examples using [community dataset](https://nextstrain-s-nextstrain-zlos8n.herokuapp.com/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17)

- [display collection that is not the default](https://nextstrain-s-nextstrain-zlos8n.herokuapp.com/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17?d=measurements&m_collection=egg_HI_raw)
- [display raw data](https://nextstrain-s-nextstrain-zlos8n.herokuapp.com/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17?d=measurements&m_display=raw)
- [group by serum](https://nextstrain-s-nextstrain-zlos8n.herokuapp.com/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17?d=measurements&m_groupBy=serum)
- [hide overall mean](https://nextstrain-s-nextstrain-zlos8n.herokuapp.com/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17?d=measurements&m_overallMean=hide)
- [hide threshold](https://nextstrain-s-nextstrain-zlos8n.herokuapp.com/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17?d=measurements&m_threshold=hide)
- [filter to single reference strain](https://nextstrain-s-nextstrain-zlos8n.herokuapp.com/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17?d=measurements&mf_reference_strain=A/Astrakhan/1/2011)
- [filter to two reference strains](https://nextstrain-s-nextstrain-zlos8n.herokuapp.com/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17?d=measurements&mf_reference_strain=A/Astrakhan/1/2011&mf_reference_strain=A/Bayern/69/2009)

[...]

Issue #1819, opened 2024-08-16
Color tree by measurements


huddlej:
## Description

Originally suggested by @trvrb [on Slack](https://bedfordlab.slack.com/archives/C03KWDET9/p1667408527847199?thread_ts=1667319132.502029&cid=C03KWDET9):

> General feature I've wanted for the Measurements panel is to be able to click on a row at the y-axis (maybe with a little gear or spotlight following the name) that would then color the tree by this row's measurements.

This feature is part of a class of features that would allow the measurements panel to control the display of other panels. This specific feature would function like a combination of 1) a filter to all tips in the tree that appear as test viruses in the selected row(s) and 2) a dynamically-defined coloring of tips (and the map and frequencies) by a quantitative color scale corresponding to the range of measurements in the selected row(s).

This functionality would be analogous to the nextflu implementation of gears that users can click on to display only tips that have measurements against that reference.

## Design questions

 - What color do we display for a tip with multiple measurements? How does the nextflu implementation deal with this?
 - Should we allow users to select multiple rows from the measurements panel?
 - How should we present the option to select a row to filter/color the tree (e.g., a "gear or spotlight"?)
 - How do we handle selection of a measurements row that is not grouped by reference strain (e.g., grouped by reference clade)? This relates to the earlier question about whether to support selection of multiple rows.
 - How should we highlight the reference strain(s) in the tree to indicate their distinct nature as a reference point for the filter and coloring?
 - How should we save the state of the selected row(s) in the URL? Thinking ahead to when we've implemented #1815.
 - How can we support display of imputed titers from models like nextflu produces? Some options include:
   - precompute these imputed titer values per reference strain (in a workflow) and store them in the measurements JSON as a separate collection with an option to group by reference strain
   - precompute the imputed titer values per reference strain (in a workflow) and store them as coloring in the main Auspice JSON
   - impute titer values on the fly like nextflu does using only the substitution model

huddlej:
I chatted with @tkanderson today at Options and he upvoted this feature request as high priority for their work. He is interested in the full nextflu functionality where the tree can display the raw measurements and imputed titer values per reference strain.

joverlee521:
We had a small group meeting today that went over the design questions ([meeting notes](https://docs.google.com/document/d/1nJYL-Y57PqUWNYW4uDWPmVRB4qNbFL1pLros1gnKJjk/edit?tab=t.0#heading=h.vv3d817fm16x)). 

Summary of v0:

[...]

PR #1881, opened 2024-10-30
Clean up measurements Redux states/actions

joverlee521:
_[Testing narrative](https://nextstrain-s-nextstrain-62bhxv.herokuapp.com/community/narratives/joverlee521/nextstrain-testing/test-multiple-dataset-measurements)_

Follow up to https://github.com/nextstrain/auspice/pull/1848, which left the measurements Redux state/actions in a bit of a mess as described in https://github.com/nextstrain/auspice/issues/1857. 
This PR cleans up the mess to make upcoming changes for https://github.com/nextstrain/auspice/issues/1819 simpler 🤞 

The main change is to fetch/load measurements JSON upfront instead of deferred loading with other sidecars JSONs. With this change, a lot of the Redux state complexities can be removed because the query params can be validated during the initial load of the page. 

## Related issue(s)

Resolves https://github.com/nextstrain/auspice/issues/1857

## Checklist

- [x] Checks pass

### Post-release

- [ ] merge https://github.com/nextstrain/auspice.us/pull/107


src/actions/recomputeReduxState.js
```diff
@@ -923,6 +923,11 @@ export const createStateFromQueryOrJSONs = ({
     measurements.collectionToDisplay = collectionToDisplay;
     controls = {...controls, ...collectionControls};
     query = updatedQuery;
+  } else {
+    // Hide measurements panel if loading failed
+    controls.panelsToDisplay = controls.panelsToDisplay.filter((panel) => panel !== "measurements");
+    controls.canTogglePanelLayout = hasMultipleGridPanels(controls.panelsToDisplay);
+    controls.panelLayout = controls.canTogglePanelLayout ? controls.panelLayout : "full";
```
jameshadfield, line 930: For completeness I'd also suggest removing any measurements queries. (There's no partial measurements state which is nice - it's either the loaded state or the default state.)

src/actions/measurements.js
```diff
@@ -263,34 +264,45 @@ export const parseMeasurementsJSON = (json) => {
...
+    collectionToDisplay
+  }
+};
+
+export const loadMeasurements = (measurementsData, dispatch) => {
+  let measurementState = getDefaultMeasurementsState();
+  let warningMessage = "";
+  if (measurementsData === undefined) {
+    // eslint-disable-next-line no-console
+    console.debug("No measurements JSON fetched");
```
jameshadfield, line 285: [aside]

We can strip these out of production builds if you want?

jameshadfield, line 285: For a dataset without measurements `measurementsData=false` so we're getting the warning message when we shouldn't. This happens because while the dispatch object sets `measurementsData: undefined` that won't override the default arg value of `createStateFromQueryOrJSONs({measurementsData = false, ...})`. This is a feature / bug of JS depending on your point of view:

```js
function test({aaa=false}={}) {console.log("inside", aaa)}
test() // inside false
test({aaa: undefined}) // inside false
```

You can use `null` if you want, but I'd suggest `false`

src/actions/measurements.js
```diff
@@ -170,7 +170,8 @@ const getCollectionDisplayControls = (controls, collection) => {
 };
 
 const parseMeasurementsJSON = (json) => {
-  const collections = json["collections"];
+  // Avoid editing the original json values since they are cached for narratives
+  const collections = cloneDeep(json["collections"]);
```
jameshadfield, line 174: [aside]

When we're not viewing a narrative we could avoid this and reduce the memory footprint. I don't actually want to do it because it makes testing harder and will invariably result in bugs somewhere, but something that came to mind as I read this.

jameshadfield (changes requested):
I think this is a lot nicer to reason with than the previous version - thanks for refactoring. Thanks also for the testing narrative - I was just about to create one!

I'd think this PR is blocking until the auspice.us PR is ready.

src/components/measurements/index.js
```diff
@@ -343,20 +343,19 @@ const Measurements = ({height, width, showLegend}) => {
...
+            {measurementsError}
+          </p>
+        </Flex> :
+          (measurementsLoaded &&
+            <MeasurementsPlot
+              height={height}
+              width={width}
+              showLegend={showLegend}
+              setPanelTitle={setTitle}
+            />)
```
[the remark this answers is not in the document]
joverlee521: > I'd suggest flipping the conditionals here: if loaded render the panel, if not show an error component with measurementsError || "unknown error".

I'll go with this!

> Update: https://github.com/nextstrain/auspice/commit/6c2308a5eef7871fe973186179ec2910340e5060 removes the entire panel if loaded: false, so these conditionals are never in play. We'll still get a warning notification "Failed to parse measurements collections" but no panel.

It loads the page without the panel, but the user can still toggle on the panel via the sidebar controls. I think it's helpful to display the error message if the user decides to toggle on the panel after the warning notification has disappeared.

src/actions/recomputeReduxState.js
joverlee521: Ah, good catch! Will do.

src/actions/measurements.js
joverlee521: Ah good catch, I'll go with `false`.

src/actions/measurements.js
joverlee521: Yeah, I thought about adding some conditionals around narratives, but like you said, it seemed like that would eventually result in bugs somewhere 🙈

src/components/measurements/index.js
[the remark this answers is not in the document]
joverlee521: Updated in https://github.com/nextstrain/auspice/pull/1881/commits/cab837e3103a99e4e58482a7b110366bbe2bc504

src/actions/measurements.js
joverlee521: Removed the console message as part of https://github.com/nextstrain/auspice/pull/1881/commits/0069e09eb66a8a3654b660d881bcb39dac765cd1

src/actions/recomputeReduxState.js
joverlee521: Added in https://github.com/nextstrain/auspice/pull/1881/commits/c03d336afb0c44969b7281047f5caa8fb54154f4

jameshadfield (approved):

PR #1884, opened 2024-11-06
Replace jest-puppeteer with Playwright

### LINKED ISSUE #1843: CI timing out during puppeteer browser download
A typical CI run calls `npm ci` 18 times. While this mostly just works, `npm ci` stochastically takes a _looong time_ ([couple](https://github.com/nextstrain/auspice/actions/runs/10515852239/job/29136804639) of [examples](https://github.com/nextstrain/auspice/actions/runs/10588576994/job/29341300964) from the past week where it hit the 6 hour time limit for an action) which means stochastic CI failures. Reducing the number of `npm ci` calls is an obvious way to make this less likely to affect ~a job~ the action.

Instead of running independent jobs for build, unit-test, smoke-test, type-check, bundlesize, check-lockfile we could collect them into a single job . Across a matrix of 4 node versions this would reduce the `npm ci` calls from 18 to 4.

This has the downside of not showing each step in the LHS sidebar's "Jobs summary", we'd just see 4 jobs of "test (node version)" or similar. Using `if: always()` for steps should allow them to run even if a preceeding step failed whilst still giving them the red check mark; `continue-on-error: true` for the job may do the same.

victorlin:

Advantages of the new Playwright setup:
 
 1. It auto-detects whether a server is already running (no need for two separate npm scripts)
 2. Browser installation is done in a separate command. Previously, it was done on installation of puppeteer which became difficult to debug when it timed out on remote build environments.
 3. It requires much fewer dependencies (see diff of package-lock.json)

## Related issue(s)

- Hopefully resolves #1843. Even if it doesn't, the separate browser installation command should make it easier to debug/configure.

## Checklist

- [x] Checks pass


victorlin:
Jobs using `npm ci` without `--ignore-scripts` (build, smoke-test) now run in [~1m30s](https://github.com/nextstrain/auspice/actions/runs/11711186448) compared to the previous [~6m](https://github.com/nextstrain/auspice/actions/runs/11710193771).

jameshadfield (approved):
Haven't tested thoroughly and it's been a while since I delved into puppeteer / smoke-testing in Auspice but this looks like a good move and CI looks good.

PR #1893, opened 2024-11-07
Autofix: When date range includes very-in-the-future dates, tree display is broken

lattaai3:
Modify the getDefaultControlsState function to set a wider default date range and update the changeDateFilter action to handle extreme future dates. 
> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.


tsibley:
Nope. (And the code is wrong, to boot.)

Issue #1888, opened 2024-11-07
TypeScript: Address `strictNullChecks` violations


victorlin:
_from https://github.com/nextstrain/auspice/pull/1864#discussion_r1821713176_

The TSConig option [strictNullChecks](https://www.typescriptlang.org/tsconfig/#strictNullChecks) will be disabled with #1864. We should re-enable it to catch existing bugs and prevent new bugs.

## Possible solutions

1. Ignore all current and future violations (status quo)
2. ⛔️ Ignore current violations only using a baseline approach
    - This seems good in theory but [does not work well](https://github.com/nextstrain/auspice/pull/1864#discussion_r1828591737).
3. Use [equality narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#equality-narrowing)
    - I started this in 260ba0020e78d7ea4a071c290d32fc2a53e67732 but it's tedious with the hundreds of violations.
4. Reduce the amount of optional properties (separate issue)

jameshadfield:
`strictNullChecks` as in [the handbook examples](https://www.typescriptlang.org/tsconfig/#strictNullChecks) will be really helpful to catch little bugs, but https://github.com/nextstrain/auspice/commit/260ba0020e78d7ea4a071c290d32fc2a53e67732 shows how intertwined it is with #1889. Using an abridged example from that commit:

```ts
const unrootedPlaceSubtree = (node: PhyloNode, totalLeafWeight: number) => {
  // FIXME: consider an UnrootedPhyloNode that guarantees presence of these? will need some casting...
  if (node.depth === undefined || ... ) {
    return;
  }
```

This looks safe (the function handles the optional properties) but if it's ever triggered it indicates a much larger bug in the architecture of Auspice, namely that this function should never be called if these values haven't been set. Returning or throwing essentially masks this deeper more structural issue. Ideally the type system would provide guarantees about the overall architecture and ensure that a function like `unrootedPlaceSubtree` can never be given an argument which doesn't have the necessary values set.

genehack:
> This looks safe (the function handles the optional properties) but if it's ever triggered it indicates a much larger bug in the architecture of Auspice, namely that this function should never be called if these values haven't been set. 

As we discussed yesterday, I think the best approach here is to have 2 types per "thing" — one where the properties are largely optional and a second where they're not (insert naming bikeshed convo here...) — and then use type guards to "up-convert" the optional form into the fully reified one.

victorlin:
[...]

PR #1897, opened 2024-11-12
Error boundaries for panels

jameshadfield:
We already use error boundaries in more tightly scoped parts of auspice (e.g. surrounding the legend), presumably in response to specific bugs/crashes observed therein. This adds a panel-specific error boundary which surrounds each panel.

Here's how it looks with an error in every panel:

image

---

image


corneliusroemer:
Oh this is great, thanks!

victorlin:
I can see this being very useful for #1888! I opened a PR to start using it in nextstrain.org too: https://github.com/nextstrain/nextstrain.org/issues/892

Issue #1890, opened 2024-11-07
TypeScript: Update DEV_DOCS with best practices


victorlin:
Now that TypeScript is being used more in this codebase, we should add some guidelines in the DEV_DOCS section.

Ideas:

- Usage of [`unknown` vs. `any`](https://github.com/nextstrain/auspice/pull/1864#discussion_r1831806967)
- [Typing function return calls](https://github.com/nextstrain/auspice/pull/1864#discussion_r1831776660)
    - should we just [enforce this with ESLint](https://typescript-eslint.io/rules/explicit-function-return-type/)?

PR #1915, opened 2024-11-21
Update entropy types

victorlin:

Updates TypeScript usage to be consistent with #1864, etc.

I thought of these changes after adding 6215591c49fc0c7b63065276bdc530b8bba5dd6a.

## Discussions

- [How to type unknown JSON inputs?](https://github.com/nextstrain/auspice/pull/1915#discussion_r1889284431)
- [Semicolons in interface definitions?](https://github.com/nextstrain/auspice/pull/1915#discussion_r1905857239)
- [Type cast value of `%3` operation `as Frame` or `as 0 | 1 | 2`?](https://github.com/nextstrain/auspice/pull/1915#discussion_r1905865269)

## Checklist

- [x] Checks pass

### DIFF
--- a/src/util/entropyCreateStateFromJsons.ts
+++ b/src/util/entropyCreateStateFromJsons.ts
@@ -2,8 +2,11 @@ import { genotypeColors } from "./globals";
 import { defaultEntropyState } from "../reducers/entropy";
 
 type JsonAnnotations = Record<string, JsonAnnotation>
+
 type Strand = '+' | '-' // other GFF-valid options are '.' and '?';
+
 type JsonSegmentRange = {start: number, end: number}; // Start is 1-based, End is 1-based closed (GFF)
+
 interface JsonAnnotation {
   /* Other properties are commonly set in the JSON structure, but the following are
   the only ones read by Auspice */
@@ -22,11 +25,13 @@ or defines the range of the genome (chromosome) itself.
 Start is always less than or equal to end. 
 Start is 1-based, End is 1-based closed. I.e. GFF. */
 type RangeGenome = [number, number];
+
 /* Same as RangeGenome but now relative to the nucleotides which make up the CDS
 (i.e. after slippage, splicing etc). The first CDS segment's RangeLocal will always
 start at 1, and the end value (of the last segment) corresponds to the number of nt in the CDS:
 range_segLast[1] - range_seg1[0] + 1 = 3 * number_of_amino_acids_in_translated_CDS */
 type RangeLocal = [number, number];
+
 type ChromosomeMetadata = {
   strandsObserved: Set<Strand>,
   posStrandStackHeight: number,
@@ -34,16 +39,19 @@ type ChromosomeMetadata = {
 }
 
 type GenomeAnnotation = Chromosome[];
+
 interface Chromosome {
   name: string;
   range: RangeGenome;
   genes: Gene[];
   metadata: ChromosomeMetadata
 }
+
 interface Gene {
   name: string;
   cds: CDS[];
 }
+
 interface CDS {
   length: number; /* length of the CDS in nucleotides. Will be a multiple of 3 */
   segments: CdsSegment[];
@@ -55,6 +63,7 @@ interface CDS {
   description?: string;
   stackPosition?: number;
 }
+
 interface CdsSegment {
   rangeLocal: RangeLocal;
   rangeGenome: RangeGenome;
--- a/src/util/entropyCreateStateFromJsons.ts
+++ b/src/util/entropyCreateStateFromJsons.ts
@@ -48,8 +48,6 @@ interface ChromosomeMetadata {
   negStrandStackHeight: number
 }
 
-type GenomeAnnotation = Chromosome[]
-
 interface Chromosome {
   name: string
   range: RangeGenome
@@ -112,7 +110,7 @@ interface CdsSegment {
  * ¹ The exception being a single CDS which wraps around the origin, which we are able
  * to split into two segments here.
  */
-export const genomeMap = (annotations: JsonAnnotations): GenomeAnnotation => {
+export const genomeMap = (annotations: JsonAnnotations): Chromosome[] => {
 
   const nucAnnotation = Object.entries(annotations)
     .filter(([name,]) => name==='nuc')
--- a/src/util/entropyCreateStateFromJsons.ts
+++ b/src/util/entropyCreateStateFromJsons.ts
@@ -5,13 +5,13 @@ type JsonAnnotations = Record<string, JsonAnnotation>
 
 type Strand = '+' | '-' // other GFF-valid options are '.' and '?';
 
-type JsonSegmentRange = {
+interface JsonSegmentRange {
   /** 1-based */
   start: number,
 
   /** 1-based closed (GFF) */
   end: number,
-};
+}
 
 interface JsonAnnotation {
   /* Other properties are commonly set in the JSON structure, but the following are
@@ -42,7 +42,7 @@ type RangeGenome = [number, number];
  */
 type RangeLocal = [number, number];
 
-type ChromosomeMetadata = {
+interface ChromosomeMetadata {
   strandsObserved: Set<Strand>,
   posStrandStackHeight: number,
   negStrandStackHeight: number,
--- a/src/util/entropyCreateStateFromJsons.ts
+++ b/src/util/entropyCreateStateFromJsons.ts
@@ -5,7 +5,13 @@ type JsonAnnotations = Record<string, JsonAnnotation>
 
 type Strand = '+' | '-' // other GFF-valid options are '.' and '?';
 
-type JsonSegmentRange = {start: number, end: number}; // Start is 1-based, End is 1-based closed (GFF)
+type JsonSegmentRange = {
+  /** 1-based */
+  start: number,
+
+  /** 1-based closed (GFF) */
+  end: number,
+};
 
 interface JsonAnnotation {
   /* Other properties are commonly set in the JSON structure, but the following are
@@ -20,16 +26,20 @@ interface JsonAnnotation {
   description?: string;
 }
 
-/* Specifies the range of the each segment's corresponding position in the genome,
-or defines the range of the genome (chromosome) itself.
-Start is always less than or equal to end. 
-Start is 1-based, End is 1-based closed. I.e. GFF. */
+/**
+ * Specifies the range of the each segment's corresponding position in the genome,
+ * or defines the range of the genome (chromosome) itself.
+ * Start is always less than or equal to end. 
+ * Start is 1-based, End is 1-based closed. I.e. GFF.
+ */
 type RangeGenome = [number, number];
 
-/* Same as RangeGenome but now relative to the nucleotides which make up the CDS
-(i.e. after slippage, splicing etc). The first CDS segment's RangeLocal will always
-start at 1, and the end value (of the last segment) corresponds to the number of nt in the CDS:
-range_segLast[1] - range_seg1[0] + 1 = 3 * number_of_amino_acids_in_translated_CDS */
+/**
+ * Same as RangeGenome but now relative to the nucleotides which make up the CDS
+ * (i.e. after slippage, splicing etc). The first CDS segment's RangeLocal will always
+ * start at 1, and the end value (of the last segment) corresponds to the number of nt in the CDS:
+ * range_segLast[1] - range_seg1[0] + 1 = 3 * number_of_amino_acids_in_translated_CDS
+ */
 type RangeLocal = [number, number];
 
 type ChromosomeMetadata = {
@@ -53,7 +63,8 @@ interface Gene {
 }
 
 interface CDS {
-  length: number; /* length of the CDS in nucleotides. Will be a multiple of 3 */
+  /** length of the CDS in nucleotides. Will be a multiple of 3 */
+  length: number;
   segments: CdsSegment[];
   strand: Strand;
   color: string;
@@ -64,12 +75,22 @@ interface CDS {
   stackPosition?: number;
 }
 
+type Phase = 0 | 1 | 2
+
+type Frame = 0 | 1 | 2
+
 interface CdsSegment {
   rangeLocal: RangeLocal;
   rangeGenome: RangeGenome;
-  segmentNumber: number; /* 1-based */
-  phase: number; /* 0, 1 or 2. Indicates where the next codon begins relative to the 5' end of this segment */
-  frame: number; /* 0, 1 or 2. The frame the codons are in, relative to the 5' end of the genome. It thus takes into account the phase */
+
+  /** 1-based */
+  segmentNumber: number;
+
+  /** Indicates where the next codon begins relative to the 5' end of this segment */
+  phase: Phase;
+
+  /** The frame the codons are in, relative to the 5' end of the genome. It thus takes into account the phase */
+  frame: Frame;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/util/entropyCreateStateFromJsons.ts
```diff
@@ -2,66 +2,96 @@ import { genotypeColors } from "./globals";
...
-  segments?: JsonSegmentRange[];
-  strand: Strand;
-  gene?: string;
-  color?: string;
-  display_name?: string;
-  description?: string;
+  end?: number
+  start?: number
+  segments?: JsonSegmentRange[]
+  strand?: Strand
```
genehack, line 23: making sure this needed to become optional to match some usage somewhere?

victorlin: In 2d3a331c the matched usage was `Object.prototype.hasOwnProperty.call(annotation, "strand")` which to me indicates that it is optional. It's hard to tell from other usage because the entire `JsonAnnotation` object is represented by `json.meta.genome_annotations` here:

https://github.com/nextstrain/auspice/blob/a7578d20f1c4db7b3755ae501816e9c11967a6d8/src/actions/recomputeReduxState.js#L819

https://github.com/nextstrain/auspice/blob/a7578d20f1c4db7b3755ae501816e9c11967a6d8/src/actions/recomputeReduxState.js#L870

@jameshadfield might have a better idea.

jameshadfield: It's optional for the `nuc` annotation. You could argue it should be required, but we relaxed this and just assert that it's not negative:

https://github.com/nextstrain/auspice/blob/c0ca42fed852bcd2709dfa853e134cd7ca6f2f46/src/util/entropyCreateStateFromJsons.ts#L93-L94

For CDSs to be considered valid it must be "+" or "-":

https://github.com/nextstrain/auspice/blob/c0ca42fed852bcd2709dfa853e134cd7ca6f2f46/src/util/entropyCreateStateFromJsons.ts#L187-L188

src/util/entropyCreateStateFromJsons.ts
```diff
@@ -20,7 +20,7 @@ interface JsonAnnotation {
   end?: number
   start?: number
   segments?: JsonSegmentRange[]
-  strand: Strand
+  strand?: Strand
```
jameshadfield, line 23: A broader comment / question about TS as it relates to the JSON dataset. Beyond being semantically valid JSON anything could be present in the JSON (we don't do formal schema validation within Auspice). So should any types relating to the dataset contents therefore allow any and every field to be optional, and also any values to be `any` type? 

This relates to a7578d20f1c4db7b3755ae501816e9c11967a6d8, which proposes

```diff
-  if (typeof annotation.display_name === 'string') {
+  if (annotation.display_name !== undefined) {
```

(`annotation` here is direct from the JSON dataset). According to our schema `display_name` will always be unset or a string, but in reality it could be anything.

src/util/entropyCreateStateFromJsons.ts
```diff
@@ -332,9 +332,24 @@ function _frame(
...
+function _mod3(n: number): 0 | 1 | 2 {
+  if (!Number.isInteger(n)) {
+    throw new Error(`${n} is not an integer.`);
+  }
+
+  /* TypeScript cannot infer the exact range of values from a modulo operation,
+   * so it is manually provided.
+   */
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  return (Math.abs(n) % 3) as 0 | 1 | 2
```
jameshadfield, line 352: For what it's worth, I find these changes kind of unfortunate (I have no doubt they are just to appease `tsc`). I'd prefer expanding out the ternary return into a longer function with more comments than additional small functions. I like the cast the best!

The lack of an integer type is pretty frustrating (JS, not TS, to be fair). If the numbers supplied to this function aren't integers we've got bigger problems in Auspice than a floating point frame!

jameshadfield (changes requested):

src/util/entropyCreateStateFromJsons.ts
victorlin: > So should any types relating to the dataset contents therefore allow any and every field to be optional, and also all values to be any type?

Technically yes, and it'd look like this:

```ts
interface InputJsonAnnotation {
  [key: string]: any
}
```

but that by itself is not very useful and defeats the purpose of having TypeScript at all. I think the proper way is to start with that loose type and use conditionals to narrow the types down to what we're expecting. That would be the schema validation that's currently missing.

The disconnect here and why this works now is that functions in `entropyCreateStateFromJsons.ts` operate under a reasonable assumption of a schema-conforming object. The type narrowing/validation should happen where the object is pulled from the JSON file, specifically in `recomputeReduxState.js` which is so far untyped and gets a free pass from tsc.

src/util/entropyCreateStateFromJsons.ts
victorlin: Reverted 14e93af0bcdb613edc861c295ad727496d586e29 and added the cast justification to df356ea1.

src/util/entropyCreateStateFromJsons.ts
```diff
@@ -2,66 +2,95 @@ import { genotypeColors } from "./globals";
...
+  start: number
+
+  /** 1-based closed (GFF) */
+  end: number
+}
+
 interface JsonAnnotation {
   /* Other properties are commonly set in the JSON structure, but the following are
   the only ones read by Auspice */
-  end?: number;
```
genehack, line 11: ...aren't these trailing semicolons required by Typescript? [The docs](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#interfaces) suggest they are.

src/util/entropyCreateStateFromJsons.ts
genehack: Just a meta-note, types like

```
 [key: string]: any
```

are (IMO) better written as 

```
 [key: string]: unknown
```

The `any` type disables typechecks, which means all the Typescript hoop-jumping adds zero value; the `unknown` type forces type disambiguation which can be a useful way to deal with dynamic values.

src/util/entropyCreateStateFromJsons.ts
```diff
@@ -280,17 +314,30 @@ function cdsFromAnnotation(cdsName: string, annotation: JsonAnnotation, rangeGen
...
+  positiveStrand: boolean,
+): Frame {
+  /* TypeScript cannot infer the exact range of values from a modulo operation,
+   * so it is manually provided.
+   */
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  return (positiveStrand ?
     (start+phase-1)%3 :
-    Math.abs((end-phase-genomeLength)%3);
+    Math.abs((end-phase-genomeLength)%3)) as 0 | 1 | 2;
```
genehack, line 340: I think this should work and it makes it more explicit what's going on...

```suggestion
    Math.abs((end-phase-genomeLength)%3)) as Frame;
```

src/util/entropyCreateStateFromJsons.ts
```diff
@@ -300,12 +347,13 @@ function _frame(start:number, end:number, phase: number, genomeLength:number, po
  * refers to this being a stacking problem.) The stack position starts at 1.
  * Returns the maximum position observed.
  */
-function calculateStackPosition(genes: Gene[], strand: (Strand|null) = null):number {
+function calculateStackPosition(
+  genes: Gene[],
+  strand: Strand,
+): number {
   /* List of CDSs, sorted by their earliest occurrence in the genome (for any segment) */
   let cdss = genes.reduce((acc: CDS[], gene) => [...acc, ...gene.cds], []);
```
genehack, line 355: This series of statements could be one stacked `.reduce().filter().sort()` operation, which would then let `cdss` be a `const` declaration instead of a `let`.

I don't think it would be that much harder to read, that way.

src/util/entropyCreateStateFromJsons.ts
victorlin: They're not required in JavaScript due to ASI. Whether to use them (or commas in this case) is mostly a stylistic choice – ESLint rule page [@stylistic/js/semi](https://eslint.style/rules/js/semi) has a good overview of pros and cons.

As mentioned in the commit message 7e6a7300795abf50b520be9defee387e781f7052, I removed them to improve consistency within the file and with other files (e.g. [phyloTree types](https://github.com/nextstrain/auspice/blob/8a6e56fa4b926de7230bf3bd35fa9e590d878fd5/src/components/tree/phyloTree/types.ts)).

src/util/entropyCreateStateFromJsons.ts
victorlin: Sure, but out of scope of this PR. I'll open another PR for that change.

src/util/entropyCreateStateFromJsons.ts
victorlin: My thinking is that `0 | 1 | 2` is logically the outcome of `%3` – if TypeScript were smart enough, it would know that without needing the cast. The return type of `Frame` in the function signature will ensure that `0 | 1 | 2` can be properly casted to `Frame`.

Does that make sense or do you still prefer `as Frame` here?

src/util/entropyCreateStateFromJsons.ts
victorlin: `entropyCreateState` doesn't return `JsonAnnotations` though, do you mean to propose another function like this?

```ts
function asJsonAnnotations(x: InputJsonAnnotation): JsonAnnotation {...}
```

We have something similar for measurements:

https://github.com/nextstrain/auspice/blob/8a6e56fa4b926de7230bf3bd35fa9e590d878fd5/src/reducers/measurements/types.ts#L59-L85

src/util/entropyCreateStateFromJsons.ts
genehack: Personally, I would write it `as Frame`; the comment makes it clear why the cast is there (Typescript can't infer the mod result, fine), and I don't have to go and look at what the type definition of `Frame` actually is. 🤷 

Non-blocking/weak opinion.

src/util/entropyCreateStateFromJsons.ts
victorlin: I'm only a slight lean towards no semicolons. I'll happily add semicolons to all the TypeScript code I've added if that's what the team desires. My stronger opinion is that we should either enforce stylistic consistency via ESLint rules and not nitpick comments on PRs.

Maybe a topic for dev chat?

src/util/entropyCreateStateFromJsons.ts
victorlin: Ok, I'll keep as-is unless there is another vote for `as Frame`.

src/util/entropyCreateStateFromJsons.ts
jameshadfield: At a conceptual level I'm trying to work out our best approach to parsing JSONs where in reality we don't know anything about how they're structured. The measurements approach looks a little more like what i'm imagining, although even `Partial<Measurement>` isn't correct as it assumes that if the `value` key is present it's a `number`. Is this an acceptable trade-off as it makes the code a lot easier? (and faster!) Or should we treat all JSON input as `[key: string]: unknown`?

To return to entropy, the better example which I should have used is:

```ts
type JsonAnnotations = Record<string, JsonAnnotation>
genomeMap = (annotations: JsonAnnotations): GenomeAnnotation
```

vs

```ts
type JsonAnnotations = [key: string]: unknown
genomeMap = (annotations: JsonAnnotations): GenomeAnnotation
```

victorlin: How about treating the input as unknown and using a function to validate + type narrow? Something like this (see eddb9b006de504b144c9f821184bf7ca584827d0 for full details):

https://github.com/nextstrain/auspice/blob/eddb9b006de504b144c9f821184bf7ca584827d0/src/util/entropyCreateStateFromJsons.ts#L192-L199

There will be duplication of type info between the validation function and the `interface`. I think that's inevitable due to the distinction of TypeScript being a compile-time check and the function being a run-time check, unless we generate a schema based on the `interface` (e.g. [ts-json-schema-generator](https://www.npmjs.com/package/ts-json-schema-generator)) and validate with that. Seems far fetched though.

src/util/entropyCreateStateFromJsons.ts
victorlin: Actually it seems fine to bundle with cad5eb63. Added as 28bd3ed9a4b1a03aeee9ddf9da6015d7679880e9.

src/util/entropyCreateStateFromJsons.ts
genehack: > My stronger opinion is that we should either enforce stylistic consistency via [ESLint rules](https://github.com/nextstrain/auspice/blob/8a6e56fa4b926de7230bf3bd35fa9e590d878fd5/.eslintrc.yaml#L40-L44) and not nitpick comments on PRs,

In my defense, I was not trying to "nitpick"; I was legitimately confused about why there had been semicolons and then they were removed in a revision to the same PR. (I had also, honestly, managed to forget "ASI" was a thing, and I was happier for it.) 

I also don't really see semicolon usage in Javascript as a "style" issue but that's probably down to having dealt with really hard to figure out bugs caused by semicolon insertion behaving in surprising ways.

victorlin: > I also don't really see semicolon usage in Javascript as a "style" issue but that's probably down to having dealt with really hard to figure out bugs caused by semicolon insertion behaving in surprising ways.

Fair, there are other "style" rules that we've adopted to prevent bugs, such as [eqeqeq](https://eslint.org/docs/latest/rules/eqeqeq). I'm down to add `@stylistic/js/semi: "always"` under the "Code quality rules" section.

src/util/entropyCreateStateFromJsons.ts
jameshadfield: I've been sceptical of schema validation because of how slow it is in augur. Rather than a `isValidGenomeAnnotations` function, I'm interested in seeing what the `genomeMap` function looks like if it's input was `UnknownJsonObject`? From my point of view this leverages typescript to help us build robust parsing functions. For instance,

image

Notably we already have runtime checks within `genomeMap` such as `if (!nucAnnotation) throw new Error("Genome annotation missing 'nuc' definition")` so this just seems to be expanding them to properly deal with unknown inputs.

src/util/entropyCreateStateFromJsons.ts
victorlin: I just [tried](https://github.com/nextstrain/auspice/compare/8a6e56fa4b926de7230bf3bd35fa9e590d878fd5...0536ee3a0cbc3b0438b03eb77ba4e904b5bfa357) `@stylistic/semi: error` and it doesn't enforce semicolons in this case (within an `interface` definition).

I'm going to close this out and keep as-is, but please re-open if you still think semicolons should be used here.

src/util/entropyCreateStateFromJsons.ts
victorlin: > this just seems to be expanding them to properly deal with unknown inputs.

After a bit of hacking, I also came to this conclusion. This is what I came up with, pushed to this PR: 0d9600b1ece01941a21ce341d0e92e9e0d0be456

The alternative is to use a library that does schema validation upfront (e.g. [superstruct](https://www.npmjs.com/package/superstruct), [zod](https://www.npmjs.com/package/zod)) so all the downstream code can avoid needing to check the inputs. However, there is arguably a benefit to keeping the validation close to where it is being used (better context and custom error messages).

jameshadfield (approved):
I ended up reading through the code changes all together as there was a fair bit of layering - it all looks good from what I can tell. If there's anything particular to look at please let me know & I can re-check.

victorlin:
Thanks for the reviews @genehack and @jameshadfield! Merging this now.

src/util/entropyCreateStateFromJsons.ts
victorlin: Noted this in #1890:

> [Don't assume types on variables derived from JSON input](https://github.com/nextstrain/auspice/pull/1915#discussion_r1909903115)

Issue #1822, opened 2024-08-16
SVG text splitting of y-axis labels. See also this comment by @huddlej


huddlej:

joverlee521:
The color by measurements crosshair icon is also affected by this issue, from https://github.com/nextstrain/auspice/pull/1924#discussion_r1902224814:

> Hmm, placement of the crosshair is not great when the grouping string is too long
![Screenshot 2025-01-03 at 2 50 12 PM](https://github.com/user-attachments/assets/ec17b506-1c55-4bb1-ab66-eba82ee80edd)

PR #1924, opened 2025-01-03
Color tree by measurements

joverlee521:

Clicking on a measurements group in the y-axis of the measurements panel will add the new coloring to the tree and a new color by to the drop-down. A crosshair icon is added to annotate the coloring group in the measurements panel and in the tree if there is a tip with a name that matches the coloring group. The measurements coloring uses the existing URL param for coloring, formatted as `c=m-<grouping-value>`. 

The coloring on the tree represents an average of the measurement values for the matching test strain within the selected measurements group. Filtering the measurements data (e.g. by source) will also filter the measurements values used to calculate the average value. Note that if the measurements filter removes all records within the selected color grouping, then the tree will be all grey. If the measurements coloring is no longer valid when changing the measurements collection or group by, then the tree will revert to its default coloring with a warning notification.

## Testing

- [community dataset](https://dev.nextstrain.org/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17)
- [community narrative (slides 11-13)](https://dev.nextstrain.org/community/narratives/joverlee521/nextstrain-testing/test-multiple-dataset-measurements?n=11)
- Try out datasets at https://dev.nextstrain.org/groups/nextflu-private (if you have access to the private Group)

## Related issue(s)

Resolves https://github.com/nextstrain/auspice/issues/1819

## Checklist

- [x] Checks pass


src/components/measurements/measurementsD3.js
```diff
@@ -492,3 +494,38 @@ export const addColorByAttrToGroupingLabel = (ref, treeStrainColors) => {
...
+        groupingLabel.append("svg")
+          .attr("id", colorGroupingCrosshairId)
+          .attr("stroke", "currentColor")
+          .attr("fill", "currentColor")
+          .attr("strokeWidth", "0")
+          .attr("viewBox", "0 0 256 256")
+          .attr("height", layout.yAxisColorByLineHeight * 2)
+          .attr("width", layout.yAxisColorByLineHeight * 2)
+          .attr("x", -width - (layout.yAxisColorByLineHeight * 2))
+          .attr("y", -layout.yAxisColorByLineHeight)
```
joverlee521, line 524: Hmm, placement of the crosshair is not great when the grouping string is too long
Screenshot 2025-01-03 at 2 50 12 PM

src/components/tree/phyloTree/change.ts
```diff
@@ -242,6 +248,11 @@ export const modifySVGInStages = function modifySVGInStages(
     this.updateTipLabels();
     this.drawTips();
     if (this.vaccines) this.drawVaccines();
+    if (this.measurementsColorGrouping) {
+      this.drawMeasurementsColoringCrosshair();
+    } else {
+      this.removeMeasurementsColoringCrosshair();
+    }
```
jameshadfield, line 255: For `modifySVGInStages` you'll want to remove the crosshair in step 1 and redraw it in step 3. Currently it stays in place while the tree tips move and then jumps to its new location

(Overall the PR looks really good)

j23414:
I can't comment on the implementation; but, I love the ability to click on the measurements panel to change the strain from which all [HI/FRA] measurements are compared against. Overall, I like the crosshair icon since we can still see the tip color on the tree.

src/actions/measurements.ts
```diff
@@ -58,6 +65,26 @@ interface Query extends MeasurementsQuery {
...
+  return colorBy.startsWith(measurementColoringPrefix);
+}
+export function encodeMeasurementColorBy(groupingValue: string): string {
+  return `${measurementColoringPrefix}${groupingValue}`;
+}
+export function decodeMeasurementColorBy(colorBy: string): string {
+  const prefixPattern = new RegExp(`^${measurementColoringPrefix}`);
+  return colorBy.replace(prefixPattern, '');
+}
+
```
genehack, line 87: i wonder if it's worth some some of comment here, explaining this data structure a bit (particularly the prefix bit and the why of the encode/decode stuff).

maybe that's already documented somewhere, in which case ignore me — i'm not at all familiar with this code and i'm just reacting to what's in the actual PR.

src/actions/measurements.ts
```diff
@@ -517,14 +577,176 @@ export const changeMeasurementsDisplay = (
...
+      dispatch
+    );
+  })
+}
+
+export function getActiveMeasurementFilters(
+  filters: MeasurementFilters
+): {string?: string[]} {
+  // Find active filters to filter measurements
+  const activeFilters: {string?: string[]} = {};
```
genehack, line 612: What does making the key part of this object optional accomplish? IOW, how is this different from `{string: string[]}` as a type declaration?

src/actions/measurements.ts
```diff
@@ -517,14 +577,176 @@ export const changeMeasurementsDisplay = (
...
+  Object.entries(filters).forEach(([field, valuesMap]) => {
+    activeFilters[field] = activeFilters[field] || [];
+    valuesMap.forEach(({active}, fieldValue) => {
+      // Save array of active values for the field filter
+      if (active) activeFilters[field].push(fieldValue);
```
genehack, line 617: i know it's widespread in the codebase but unbraced conditionals always make me twitch, just a little bit. 

just needed to get that off my chest in the new year.

huddlej (commented):
@joverlee521 This is such a cool set of new functionality and you clearly put a ton of thought and work into it! I've collected my thoughts below entirely based on my experience as a user interacting with the measurements panel. I'm leaving code review to other folks who are more qualified to offer feedback on that.

My main requests are for an option to filter tips from the tree that don’t have measurements and to have more hints in the user interface about the nature of the data being shown or how to interact with the new functionality. I also wouldn’t expect that any of these would be in the first prototype of the functionality. I only consider the filtering functionality as a potential blocker to releasing this as is.

**General likes and loves:** I love that I can [color recent H1 viruses by measurements to the current vaccine strain \(A/Wisconsin/67/2022\)](https://dev.nextstrain.org/groups/nextflu-private/flu/seasonal/2024-12-13/h1n1pdm/2y/titers/ha?c=m-A/Wisconsin/67/2022&d=tree,measurements&f_subclade=C.1.1,C.1.8,C.1.9,D.5&f_year_month=2024-06,2024-07,2024-08,2024-09,2024-10,2024-11,2024-12&label=clade:5a.2a&m_display=raw&mf_reference_strain=A/Wisconsin/67/2022&p=full) and get a sense of antigenic distance and clade size in a single view. I also like that I can color the tree by measurements and then scatterplot those measurements against other metadata fields in the tree. We didn’t previously have a way to do this kind of comparison inside the measurements panel. I like the target icon for the selected reference in the tree and the measurements panel. I like that when I change the grouping of the measurements panel from reference strain to reference subclade, I get a notification that “Measurement coloring is no longer valid”.

**Wishes and what ifs:** I wish that I could hide the gray tips that don’t have any measurements, so I could more clearly see the patterns among tips that have measurements. This would give a view like the current private site (minus the gears). This is what I would want for my default view, but I’ve also noticed it can be handy to have the gray tips shown when I’m comparing subclades with a specific genotype in a tree based on their coverage for a specific reference. For example, [this view](https://dev.nextstrain.org/groups/nextflu-private/flu/seasonal/2024-12-13/h3n2/2y/titers/ha?branchLabel=aa&c=m-A/Croatia/10136RV/2023&d=tree,measurements&gt=HA1.160K&label=Subclade:J.2&m_collection=cell_hi&mf_reference_strain=A/Croatia/10136/RV/2023&mf_reference_strain=A/Croatia/10136RV/2023&p=grid) shows J.2 descendants with the HA1:160K genotype colored by measurements against the new A/Croatia/10136RV/2023 vaccine strain. This view shows that one small clade of 160K is covered by measurements, but another sibling clade is not. This kind of analysis depends on the ability to show tips without filtering by the presence/absence of measurements. Maybe when we’ve colored by measurements, we could have a filter generated that hides the tips without measurements by default but which we could remove to show those tips as needed?

I wish there was a static visual cue to the user that they can click on a row of the measurements panel to color by those measurements and maybe an on-hover description of what would happen on click.

Along the same lines, there isn’t an obvious way to toggle the coloring back to what it was before clicking on the measurement row. You need to know that the action in the measurements panel should be undone by interacting with the coloring dropdown. What if clicking on the row again toggled back to the previously selected color?

When I color the tree by a specific reference’s measurements, the rest of the measurements panel updates to group measurements from other references by the new coloring scheme. This is the expected behavior for the measurements panel, but it took me a minute to wrap my head around what it actually means for the other measurements to be grouped in this way. I think this grouping allows me to look for mismatches like when test viruses have higher values against the selected reference than they do for the other references. I had originally wondered whether we should automatically filter measurements to just those in the selected row, but I think this view with all measurements will be useful in the long run.

Similarly, when coloring by one reference’s measurements, the other references get assigned a measurement value next to their names that seems to correspond to the color ramp bins. I think that value reflects the color bin that the other reference viruses map to based on their measurements against the selected reference and not some summary statistic of all measurements for the other references. Is that right?

Are the bins for the color ramp for the measurements scale fixed across all measurements in a given collection?

When I group measurements by subclade and color the tree by those measurements, I don’t have an indication of how the resulting color values are calculated. I would assume that the values are the mean or median for each test virus across all references in the measurements group. What if we reported that information somehow when coloring by measurements across more than one reference?

**Really cool unexpected features:** I love how the narrative examples effectively represent a tutorial on how to use the measurements panel! If the narrative source code was easily accessible, these slides could be a great learning tool.

One unexpected cool feature is how the frequency panel shows the timing of test viruses measured against a selected reference. This gives a sense of what time period in the tree is covered by a reference, what proportion of viruses in the tree are covered by the reference, and the change in antigenic distances between that reference and test viruses from different times. If we filtered tips without measurements from the tree and turned off normalization of the frequency panel, we’d get a kind of histogram of viruses measured against the selected reference by frequency pivot. This feature is especially helpful when grouping measurements by subclade, to see how many test viruses are covered by references from each subclade and how well they are covered. It’s also interesting to group measurements by source and see how the test viruses covered by different CCs varies through time.

Another unexpected implementation detail that seems like a feature is that the narrative view of the measurements panel allows the user to change the coloring in the tree by selecting a different row. For example, [this narrative slide](https://dev.nextstrain.org/community/narratives/joverlee521/nextstrain-testing/test-multiple-dataset-measurements?n=11) defaults to coloring by A/Astrakhan/1/2011, but I can click on A/Lviv/N6/2009 and change the coloring of the tree. This is unusual behavior, since the tree and other panels tend to be static in a narrative. I can zoom/pan the map, for instance, but I can’t change the coloring or filters. This ability to color by other references would be really helpful during live reports because it would let us stay in the context of the narrative without exiting to “explore the data yourself”.

Yet another cool feature is that we can group measurements by specific serum id and color by that serum’s measurements, showing the serum’s potency across all test viruses in a way we couldn’t easily visualize before.

src/actions/measurements.ts
joverlee521: The optional key allows `activeFilters` to be an empty object, which is needed since there are cases where all of the filters are "inactive".

genehack: ah, okay, thanks -- I guess in the past I've spelled this 

```
const foo: Record<string,string[]> = {};
```

which doesn't seem to have the same problem with an empty object. (Not asking for a change.)

src/actions/measurements.ts
joverlee521: Ah yeah...it doesn't bother me as much :face_with_peeking_eye:

src/actions/measurements.ts
```diff
@@ -517,14 +577,176 @@ export const changeMeasurementsDisplay = (
...
+
+  const nodeAttrs: MeasurementsNodeAttrs = {};
+  for (const [strain, measurements] of Object.entries(strainMeasurementValues)) {
+    const averageMeasurementValue = measurements.reduce((sum, value) => sum + value) / measurements.length;
+    nodeAttrs[strain] = {
+      [measurementColorBy]: {
+        value: averageMeasurementValue
+      }
+    };
+  }
```
joverlee521, line 663: > Maybe when we’ve colored by measurements, we could have a filter generated that hides the tips without measurements by default but which we could remove to show those tips as needed?

@huddlej thanks for bringing up the need for filtering to measurements tips! I think this can be accomplished similar to how we add a [boolean field for the drag-and-drop metadata](https://github.com/nextstrain/auspice/blob/8a6e56fa4b926de7230bf3bd35fa9e590d878fd5/src/actions/filesDropped/metadata.js#L143-L150). Here, I can add a new field (e.g. `_measurementsColors`) here with the `true` value so that it can be used to filter to those tips.

src/components/measurements/measurementsD3.js
```diff
@@ -264,6 +266,7 @@ export const drawMeasurementsSVG = (ref, xAxisRef, svgData) => {
         // but there're always limits of the available space so punting that for now.
         //    -Jover, 20 September 2022
         g.selectAll('text')
+          .attr("class", classes.groupingValue)
```
joverlee521, line 269: > I wish there was a static visual cue to the user that they can click on a row of the measurements panel to color by those measurements and maybe an on-hover description of what would happen on click.

I can add an [SVG title](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) here to add a basic tooltip to highlight the function.

src/components/measurements/index.tsx
```diff
@@ -131,39 +139,28 @@ const filterMeasurements = (
...
 
 const MeasurementsPlot = ({height, width, showLegend, setPanelTitle}) => {
+  const dispatch = useDispatch();
   // Use `lodash.isEqual` to deep compare object states to prevent unnecessary re-renderings of the component
   const { treeStrainVisibility, treeStrainColors } = useSelector((state: RootState) => treeStrainPropertySelector(state), isEqual);
-  const legendValues = useSelector((state: RootState) => state.controls.colorScale.legendValues, isEqual);
+  // Convert legendValues to string to ensure that subsequent attribute matches work as intended
+  const legendValues = useSelector((state: RootState) => state.controls.colorScale.legendValues.map((v): string => v.toString()), isEqual);
   const colorings = useSelector((state: RootState) => state.metadata.colorings);
   const colorBy = useSelector((state: RootState) => state.controls.colorBy);
```
joverlee521, line 162: > What if clicking on the row again toggled back to the previously selected color?

I _think_ if we save the `colorBy` as a local ref to reference the previously selected color then it can be used to toggle the color back. If this gets too complicated, I might just toggle to the tree's default coloring.

src/components/measurements/measurementsD3.js
```diff
@@ -466,7 +470,7 @@ export const addColorByAttrToGroupingLabel = (ref, treeStrainColors) => {
   svg.selectAll(`.${classes.yAxis}`).select(".tick")
     .each((_, i, elements) => {
       const groupingLabel = select(elements[i]);
-      const groupingValue = groupingLabel.text();
+      const groupingValue = groupingLabel.select(`.${classes.groupingValue}`).text();
       const groupingValueColorBy = treeStrainColors[groupingValue];
```
joverlee521, line 474: > I think that value reflects the color bin that the other reference viruses map to based on their measurements against the selected reference and not some summary statistic of all measurements for the other references. Is that right?

Yes, this is reflecting the color of the strain in the tree

src/actions/measurements.ts
```diff
@@ -517,14 +577,176 @@ export const changeMeasurementsDisplay = (
...
+  const sortedValues = collection.measurements
+    .map((m) => m.value)
+    .sort((a, b) => a - b);
+
+  // Matching the default coloring for continuous scales
+  const colorRange = colors[9];
+  const step = 1 / (colorRange.length - 1);
+  const measurementsColorScale: [number, string][] = colorRange.map((color, i) => {
+    return [quantile(sortedValues, (step * i)), color]
+  });
```
joverlee521, line 673: > Are the bins for the color ramp for the measurements scale fixed across all measurements in a given collection?

Yes, the color scale is created using all measurements in a given collection. I did this to have a stable coloring scale for each collection.

src/components/measurements/measurementsD3.js
joverlee521: > When I group measurements by subclade and color the tree by those measurements, I don’t have an indication of how the resulting color values are calculated. I would assume that the values are the mean or median for each test virus across all references in the measurements group. What if we reported that information somehow when coloring by measurements across more than one reference?

Reporting when coloring by more than one reference seems like a very flu titer data specific request. Maybe this can be generalized in the on-hover description? 

```
Click to color by an average of all measurement values for a test strain in this group
```

src/actions/measurements.ts
joverlee521: The main reason I've added the prefix bit is to lower the chances of the generated coloring from clashing with an existing coloring of the tree (similar to how Genotype coloring is prefixed by `gt-`). I'll add as comments here.

genehack: 🙌

src/actions/measurements.ts
genehack: 😆

tsibley: > i know it's widespread in the codebase but unbraced conditionals always make me twitch, just a little bit.

Would you prefer it if it was:

```js
activeFilters[field].push(fieldValue) if active;
```

;-)

genehack: I was literally considering tooting something, earlier today, about how the loss of postfix conditionals was one of the biggest losses in moving away from Perl. :lolsob:

src/actions/measurements.ts
joverlee521: Added in 972ed91ef8cd1c173b969c3978ff66f0af1ca22e

src/actions/measurements.ts
joverlee521: Added in 82e6f6d11db06cf23fc80f22c26666fd18c7cde5

src/components/measurements/measurementsD3.js
joverlee521: Added tooltip in 9f6b03260d2df285542ee5faa611dbca92353871

src/components/measurements/index.tsx
joverlee521: Added in 27cdbb07087c30e02184d3d4dc4e6d93dd8fb543

src/actions/measurements.ts
```diff
@@ -65,10 +65,18 @@ interface Query extends MeasurementsQuery {
   [key: string]: string | string[]
 }
 
+
+const hasMeasurementColorAttr = "_hasMeasurementColor";
```
jameshadfield, line 70: Probably beyond this PR, but this makes me think about some improvements to the filter badges - they haven't seen any development since they were first introduced and filtering has become a much bigger part of workflows since then.

It'd be nice to show the title not the key when hovering, and even nicer if we could somehow render this in the badge itself, as a badge simply showing "true" isn't very intuitive. We already do both of these in the sidebar
image

Would this (the `_hasMeasurementsColor` filter) be better as part of the separate measurements filtering section?

image

src/actions/measurements.ts
```diff
@@ -611,14 +612,23 @@ export const applyMeasurementsColorBy = (
...
-  dispatch({type: ADD_EXTRA_METADATA, newNodeAttrs, newColorings});
-  dispatch(changeColorBy(measurementColorBy));
+  /**
+   * Batching all dispatch actions together to prevent multiple renders
+   * This is also _required_ to prevent error in calcColorScale during extra renders:
+   * 1. REMOVE_METADATA removes current measurements coloring from metadata.colorings
+   * 2. This triggers the componentDidUpdate in controls/color-by, which dispatches changeColorBy.
+   * 3. calcColorScale throws error because the current coloring is no longer valid as it was removed by REMOVE_METADATA in step 1.
+   */
+  batch(() => {
```
jameshadfield, line 767: I didn't know about `batch` - was a good opportunity to catch up with some of the many changes in the redux/react world. [It seems that React 18 will do this automatically](https://react-redux.js.org/api/batch) (react 18 upgrade PR: https://github.com/nextstrain/auspice/pull/1659). I wondered why you didn't just add the new node info, change colour, then delete the old data... however since the `node.attrs` key names are reused I can imagine this causes some undesirable side-effects?

src/util/treeMiscHelpers.js
```diff
@@ -309,3 +309,20 @@ export const addNodeAttrs = (nodes, newAttrs) => {
...
+ * @param {Array} attrsToRemove
+ */
+export const removeNodeAttrs = (nodes, attrsToRemove) => {
+  nodes.forEach((node) => {
+    if(!node.node_attrs) return;
+
+    attrsToRemove.forEach((attrName) => {
+      if(attrName in node.node_attrs){
+        delete node.node_attrs[attrName];
+      }
```
jameshadfield, line 325: You can drop the conditional - `delete` doesn't throw if the `attrName` isn't a property

src/actions/measurements.ts
jameshadfield, line 767: I don't think it's worth changing what you've done (which includes a good docstring) but an alternative approach to consider for patterns like this if we use them again is to make "remove metadata" a thunk which checks that the attr being removed isn't currently used as a coloring, and if it is then thee thunk can do something about it.

src/actions/recomputeReduxState.js
```diff
@@ -917,10 +917,37 @@ export const createStateFromQueryOrJSONs = ({
...
+        return newColoringData.colorings[coloring].type !== "continuous"
+      });
+      tree.totalStateCounts = {
+        ...tree.totalStateCounts,
+        ...countTraitsAcrossTree(tree.nodes, nonContinuousColorings, false, true)
+      };
+      // Update treeToo if exists
+      if (treeToo && treeToo.loaded) {
+        addNodeAttrs(treeToo.nodes, newColoringData.nodeAttrs);
+      }
```
jameshadfield, line 976: Can we have two trees + measurements? I've come to the opinion that we should drop a lot of functionality when viewing two trees, so that if the second tree is displayed we don't allow measurements, frequencies, map to be displayed.

src/components/measurements/index.tsx
```diff
@@ -254,9 +256,25 @@ const MeasurementsPlot = ({height, width, showLegend, setPanelTitle}) => {
...
+      prevNonMeasurementColorBy.current = colorBy;
+    }
+  }, [colorBy]);
+
   const handleClickOnGrouping = useCallback((grouping: string): void => {
     if (grouping !== colorGrouping || !isMeasurementColorBy(colorBy)) {
       dispatch(applyMeasurementsColorBy(grouping));
+    } else if (grouping === colorGrouping && isMeasurementColorBy(colorBy)) {
+      // Clicking on the same grouping twice will toggle back to the previous non-measurements coloring
+      dispatch(changeColorBy(prevNonMeasurementColorBy.current));
```
jameshadfield, line 277: [Comment "Click on measurements grouping to toggle back to previous coloring"]

This'll leave the `_hasMeasurementsColor` filter enabled. Just a heads up - not a comment on whether that's the desired behaviour or not.

jameshadfield (approved):
Looks good @joverlee521. Read through the code and played around a little with a measurements dataset I had, although I didn't test out the narratives or change collections behaviour. I don't think any of my comments are blocking.

src/components/measurements/index.tsx
```diff
@@ -161,7 +161,8 @@ const filterMeasurements = (
 const MeasurementsPlot = ({height, width, showLegend, setPanelTitle}) => {
   // Use `lodash.isEqual` to deep compare object states to prevent unnecessary re-renderings of the component
   const { treeStrainVisibility, treeStrainColors } = useSelector((state: RootState) => treeStrainPropertySelector(state), isEqual);
-  const legendValues = useSelector((state: RootState) => state.controls.colorScale.legendValues, isEqual);
+  // Convert legendValues to string to ensure that subsequent attribute matches work as intended
+  const legendValues = useSelector((state: RootState) => state.controls.colorScale.legendValues.map((v): string => v.toString()), isEqual);
```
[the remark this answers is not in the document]
joverlee521: Ah good point about `toString` raising exceptions, I'll update to `String`. 

> Not being super familiar with measurements terminology, I didn't understand "attributes" in the commit message and had to dive into the code. I think it's node attributes / node values, right? 

These are referring to the coloring attributes, so yes these correspond to the node attributes / node values.

> Where is the Object.keys you mention - I followed the code a little bit but didn't find it.

These are referring to where the node attributes get compared to the legend values for ordering. These are in measurementsD3.js, within [jitterRawMeansByColorBy](https://github.com/nextstrain/auspice/blob/a6076d52e611dd5cc0d1f75e7592fa7cd5ed5694/src/components/measurements/measurementsD3.js#L343-L345) and [drawMeansForColorBy](https://github.com/nextstrain/auspice/blob/a6076d52e611dd5cc0d1f75e7592fa7cd5ed5694/src/components/measurements/measurementsD3.js#L390-L391).

src/actions/measurements.ts
joverlee521: > I wondered why you didn't just add the new node info, change colour, then delete the old data... however since the node.attrs key names are reused I can imagine this causes some undesirable side-effects?

Yeah, things can get complicated if the node.attrs key names are the same but the underlying values changed. A concrete example is changing measurements collections. They might have the same reference strain (which means they have the same node.attrs key), but completely different measurements so the data needs to be completely updated. It's cleaner to remove old data then apply the new data instead of trying to keep track of which nodes needs to be updated.

src/actions/recomputeReduxState.js
joverlee521: Two trees + measurements is definitely possible right now, where the measurements panel always uses the measurements JSON from the main tree. 

I'm not entirely sure how useful it is though...could it be useful for HA:NA + measurements?

src/components/measurements/index.tsx
joverlee521: Thanks for bringing this up! I left this in as a "feature", but not sure if others noticed.  

I thought it'd be useful to see the filtered tips in the other coloring option as you quickly toggle back and forth by clicking on the measurements grouping. I'm leaving it as-is unless I hear otherwise in the future.

src/util/treeMiscHelpers.js
joverlee521: Oops, Python bleeding into JavaScript...dropped conditional in aead1cf13c4000e7d76e090877932018268b9820.

src/actions/measurements.ts
jameshadfield: > I think it makes sense to keep the _hasMeasurementsColor filter in the overall filters since it is filtering the tree node attributes. The measurements filtering section is specific to filtering measurements records in the measurements panel.

Ahh, this is clarifying. Thanks!

src/actions/recomputeReduxState.js
jameshadfield: > I'm not entirely sure how useful it is though...could it be useful for HA:NA + measurements?

I can't answer that, but we should find the answer. Two trees complicates Auspice quite a lot, and I think it would be a worthwhile push to limit the cases in order to simplify the code.

jameshadfield (approved):

src/components/measurements/measurementsD3.js
joverlee521: Noted this in related issue https://github.com/nextstrain/auspice/issues/1822.

src/actions/recomputeReduxState.js
joverlee521: We can revisit simplifying code in the future, starting conversation in [Slack](https://bedfordlab.slack.com/archives/C03KWDET9/p1737491902430339).

trvrb:
Awesome to have this merged! This is a very cool feature. A couple post-merge points / desires (based on what's currently at https://dev.nextstrain.org/groups/nextflu-private/flu/seasonal/2024-12-13/h1n1pdm/2y/titers/ha).

1. The "filtered to true" is a bit strange and should be clearer.
 
Screenshot 2025-01-21 at 5 23 58 PM

The box in the side bar is properly clear:

Screenshot 2025-01-21 at 5 27 57 PM

2. I'd really like a mouse-over hovel panel when mousing over each row in the measurements panel that includes italic text at the bottom indicating that you can click on the row to color the tree by this measurement. This is the pattern we use elsewhere in Auspice to surface that you can click on things.

3. Very cool to be able to interface this with scatterplots like:

Screenshot 2025-01-21 at 5 32 07 PM

joverlee521:
> 2. I'd really like a mouse-over hovel panel when mousing over each row in the measurements panel that includes italic text at the bottom indicating that you can click on the row to color the tree by this measurement. This is the pattern we use elsewhere in Auspice to surface that you can click on things.

Ah, I didn't include it in the hover over each row because you have to click on the grouping in the y-axis. There is a small pop-up when you hover over group in the y-axis, although seems like it takes a brief moment for it to pop up:

https://github.com/user-attachments/assets/fe1ea255-0d57-474d-a7d4-145feb64cf9b

PR #1928, opened 2025-01-14
Show warnings in Info component

### LINKED ISSUE #1927: Add ability to show warning text from Auspice JSON
_[from Slack](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1736886133814889)_

Render a new JSON entry for warnings (e.g. `meta.warning`) along the lines of @trvrb's mock-up:

image

## Related issues

- https://github.com/nextstrain/augur/issues/1722

victorlin:
_preview on [auspice.us](https://auspice-us-nextstrain-b-9puakb.herokuapp.com), [nextstrain.org](https://nextstrain-s-nextstrain-eaqp9o.herokuapp.com/community/victorlin/nextstrain-test/ncov/warning)_

Inspired by code in byline.js (same parent component) and footer.js (also parses Markdown)

Markdown loads lazily, but I think it will be useful to support Markdown format instead of just plain text (which could be loaded earlier).

## Related issue(s)

Closes #1927

## Discussion threads

- [How to handle markdown parsing errors?](https://github.com/nextstrain/auspice/pull/1928#discussion_r1915849381)

## Checklist

- [ ] Checks pass

### DIFF
--- a/src/components/info/warning.js
+++ b/src/components/info/warning.js
@@ -1,0 +1,67 @@
+import React, { Suspense, lazy } from "react";
+import { connect } from "react-redux";
+import styled from 'styled-components';
+import { FaExclamationTriangle } from "react-icons/fa";
+import { withTranslation } from 'react-i18next';
+
+const MarkdownDisplay = lazy(() => import("../markdownDisplay"));
+
+const WarningContainer = styled.div`
+  background-color: #FFF1D1;
+  display: flex;
+  align-items: stretch;
+`;
+
+const WarningIconContainer = styled.div`
+  background-color: #FFC641;
+  padding: 0 7px;
+  display: flex;
+  align-items: center;
+`;
+
+const WarningIcon = styled.div`
+  padding-top: 3px;
+  color: white;
+  font-size: 24px;
+`;
+
+const WarningText = styled.div`
+  padding: 0 7px;
+  font-size: 15px;
+`;
+
+/**
+ * React component for the warning of the current dataset.
+ */
+@connect((state) => {
+  return {
+    warning: state.metadata.warning
+  };
+})
+class Warning extends React.Component {
+  render() {
+    const { warning } = this.props;
+
+    if (warning === undefined) return null;
+
+    return (
+      <Suspense fallback={null}>
+        <WarningContainer>
+          <WarningIconContainer>
+            <WarningIcon>
+              <FaExclamationTriangle />
+            </WarningIcon>
+          </WarningIconContainer>
+          <WarningText>
+            <MarkdownDisplay
+              mdstring={warning}
+              placeholder="This dataset contained a warning message to be displayed here, however it wasn't correctly formatted."/>
+          </WarningText>
+        </WarningContainer>
+      </Suspense>
+    );
+  }
+}
+
+const WithTranslation = withTranslation()(Warning);
+export default WithTranslation;
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -743,6 +743,9 @@ const createMetadataStateFromJSON = (json) => {
   if (json.meta.description) {
     metadata.description = json.meta.description;
   }
+  if (json.meta.warning) {
+    metadata.warning = json.meta.warning;
+  }
   if (json.version) {
     metadata.version = json.version;
   }
--- a/src/components/framework/fine-print.js
+++ b/src/components/framework/fine-print.js
@@ -161,7 +161,9 @@ export function getCustomFinePrint() {
   return (
     <Suspense fallback={<></>}>
       <Flex className='finePrint'>
-        <MarkdownDisplay mdstring={markdown} />
+        <MarkdownDisplay
+          mdstring={markdown}
+          placeholder="This dataset contained a fine print message to be displayed here, however it wasn't correctly formatted." />
       </Flex>
     </Suspense>
   );
--- a/src/components/framework/footer.js
+++ b/src/components/framework/footer.js
@@ -163,7 +163,10 @@ export const getAcknowledgments = (metadata, dispatch) => {
   if (metadata.description) {
     return (
       <Suspense fallback={<div />}>
-        <MarkdownDisplay className="acknowledgments" mdstring={metadata.description} />
+        <MarkdownDisplay
+          className="acknowledgments"
+          mdstring={metadata.description}
+          placeholder="This dataset contained acknowledgements to be displayed here, however it wasn't correctly formatted." />
       </Suspense>
     );
   }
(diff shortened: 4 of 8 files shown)


src/components/info/warning.js
```diff
@@ -0,0 +1,57 @@
+import React, { Suspense, lazy } from "react";
+import { connect } from "react-redux";
+import styled from 'styled-components';
+import { FaExclamationTriangle } from "react-icons/fa";
+import { withTranslation } from 'react-i18next';
+
+const MarkdownDisplay = lazy(() => import("../markdownDisplay"));
```
jameshadfield, line 7: An error parsing markdown results in the following rendered within the error box:

> There was an error parsing the Markdown. See the JS console.

I'd suggest customizing this to say something along the lines of "This dataset contained a warning message to be displayed here however the warning message itself wasn't correctly formatted!"

src/components/info/warning.js
```diff
@@ -0,0 +1,57 @@
...
+const WarningText = styled.div`
+  font-size: 15px;
+`;
+
+/**
+ * React component for the warning of the current dataset.
+ */
+@connect((state) => {
+  return {
+    metadata: state.metadata
```
jameshadfield, line 32: In general it's best to access state in the most fine-grained way possible to reduce unnecessary rendering. In this case it'd be `warning: state.metadata.warning`. Reality is a little more complex and depends on how the parent is rendered, but it's a good rule of thumb. 

(It's been a while since I was day-to-day wiring up React, but if you'd like me to track down specific docs let me know)

src/components/info/warning.js
victorlin: What do you think about using the raw text instead? e48af64e6bfc1afd3e4dda9e1e1713837184584a

src/components/info/warning.js
```diff
@@ -0,0 +1,57 @@
+import React, { Suspense, lazy } from "react";
+import { connect } from "react-redux";
+import styled from 'styled-components';
+import { FaExclamationTriangle } from "react-icons/fa";
+import { withTranslation } from 'react-i18next';
+
+const MarkdownDisplay = lazy(() => import("../markdownDisplay"));
+
+const WarningContainer = styled.div`
```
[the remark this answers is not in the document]
victorlin: Updated:

image

src/components/info/warning.js
victorlin: Good tip, much cleaner. Updated:

https://github.com/nextstrain/auspice/blob/866829a0777f53dc799feee4a243ca431d98824f/src/components/info/warning.js#L36-L40

src/components/info/warning.js
tsibley: Reading email, I saw James' comment but not Victor's, and came here to suggest falling back to the raw text. +1 for that. Markdown is intended to be readable.

genehack (approved):

src/components/info/warning.js
jameshadfield: Invalid markdown, as in JSON data which fails to be parsed by our parser such as a dict/object, looks much worse this way:

image

If you can find a JSON string which is invalid (I can't, although I'm sure it's possible) then won't your approach here open yourself up to script injection?

Secondly, this change is applied to any component using `<MarkdownDisplay>`. I'm guessing this is intentional?

I continue to think we want a component-specific error message displayed.

tsibley: Oh, jeez, I didn't actually read the commit Victor linked. James is correct that it absolutely opens up door for script injection and more. The Markdown needs to be run thru the HTML sanitizer, or better yet, avoid setting raw user input via `dangerouslySetInnerHTML` entirely when we don't need to, e.g.

```diff
diff --git a/src/components/markdownDisplay/index.tsx b/src/components/markdownDisplay/index.tsx
index 4b4abae7..8ddeab72 100644
--- a/src/components/markdownDisplay/index.tsx
+++ b/src/components/markdownDisplay/index.tsx
@@ -8,17 +8,19 @@ export default function MarkdownDisplay({
   mdstring: string,
   props: React.HTMLAttributes<HTMLDivElement>,
 }): JSX.Element {
-  let html: string;
   try {
-    html = parseMarkdown(mdstring);
+    return (
+      <div
+        {...props}
+        dangerouslySetInnerHTML={{ __html: parseMarkdown(mdstring) }}
+      />
+    );
   } catch (error) {
     console.error(`There was an error parsing the provided text as Markdown. Using the raw text as-is. Error: ${error}`);
-    html = `<p>${mdstring}</p>`;
+    return (
+      <div {...props}>
+        <p>${mdstring}</p>
+      </div>
+    );
   }
-  return (
-    <div
-      {...props}
-      dangerouslySetInnerHTML={{ __html: html }}
-    />
-  );
 }
```

victorlin:
I've dropped ff91cfd17a16063a6890b65d8a23c929297c3fc8 since it's blocked on https://github.com/nextstrain/nextstrain.org/pull/1110

src/components/info/warning.js
victorlin: Thanks for the suggestion @tsibley, I've incorporated it into 72ebf1611da762e9145840cc404b95a27950a721.

victorlin: @jameshadfield I've switched to component-specific placeholders in cb3d4e4e45d47614edb3c1c23c665797fcd9c4d4.

jameshadfield: Can you supply an example where the `if (typeof mdstring === 'string') {` clause is run?

I still don't think it's worth rendering the raw text, but if you must do it then please also include the placeholder in those cases.

victorlin: > an example where the `if (typeof mdstring === 'string') {` clause is run

It runs when there is a bug within parseMarkdown or the `marked` package.

> I still don't think it's worth rendering the raw text, but if you must do it then please also include the placeholder in those cases.

How about a generic placeholder in this case? Example:

image

jameshadfield: Ok, if we have

```
if internalAuspiceError -> render text as is
else -> render placeholder
```

And you're happy with the sanitisation / security details then that's fine with me.

victorlin: Updated to:

https://github.com/nextstrain/auspice/blob/f5552ae9c2dacf51a85e4b0464446bc424350ab7/src/components/markdownDisplay/index.js#L13-L21

It should be safe as long as we don't use `dangerouslySetInnerHTML`, so the text goes through React's [built-in sanitization](https://legacy.reactjs.org/docs/introducing-jsx.html#jsx-prevents-injection-attacks):

> By default, React DOM [escapes](https://stackoverflow.com/questions/7381974/which-characters-need-to-be-escaped-on-html) any values embedded in JSX before rendering them. Thus it ensures that you can never inject anything that’s not explicitly written in your application. Everything is converted to a string before being rendered. This helps prevent [XSS (cross-site-scripting)](https://en.wikipedia.org/wiki/Cross-site_scripting) attacks.

jameshadfield (approved):

Issue #1919, opened 2024-12-06
Unresponsiveness with 1Password extension


victorlin:
Recent versions of 1Password browser extension (8.10.56(.28), all browsers) causes Auspice views to be slow and in some cases seemingly unresponsive. The slowness scales with the number of elements rendered by Auspice. This means it is most noticeable for the following scenarios per-panel:

- Tree: large number of tips
- Map, Frequency: large number of distinct color-by values
- Entropy: large genome size

## Example for reproducing the issue

1. Revert #1932 and run a local dev server
2. Ensure the 1Password extension is active
3. Go to http://localhost:4000/tb/global?d=tree
4. Scroll down on the sidebar and click the toggle to the right of "Entropy"
5. Page will become unresponsive

## Workarounds

Ordered from least to most work involved. Instructions written for Chrome – may vary for other browsers.

1. ✅ Apply a workaround in Auspice
    - #1932
2. Tell affected users to disable the "Offer to save one-time passwords" toggle in 1Password extension settings > Autofill & save or use a different browser/profile without the extension installed.
3. Tell affected users to downgrade to an older version of 1Password.

    1. Download and unzip [1Password-8.10.52.25_0.zip](https://github.com/user-attachments/files/18084441/1Password-8.10.52.25_0.zip).
    2. Go to Chrome extension settings and enable **Developer mode** on the top right.
    3. Click **Load unpacked**.
    5. Select the unzipped directory.
    6. Check that 1Password extension version is `8.10.52.25`.

## Progress

- [x] Open [support ticket](https://support.nextstrain.org/agent/nextstrain/nextstrain/tickets/details/668036000008450227) with 1Password
- [x] Inform users: [blog](https://github.com/nextstrain/nextstrain.org/pull/1088), [Mastodon](https://mstdn.science/@nextstrain/113641321533662923)
- [x] Apply workaround in Auspice: #1932
- [x] Inform users: [blog](https://github.com/nextstrain/nextstrain.org/pull/1113), [Mastodon](https://mstdn.science/deck/@nextstrain/113874083547729384), [Bluesky](https://bsky.app/profile/nextstrain.org/post/3lgec3lbbvk24)

## Internal links

- Slack threads
    - [11/27 #nextstrain-general](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1732749045004729)
    - [12/4 #nextstrain-dev](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1733353970597669)
[...]

PR #1932, opened 2025-01-18
Workaround 1Password browser extension bug

victorlin:
preview on [auspice.us](https://auspice-us-nextstrain-b-3jy7rl.herokuapp.com/), [nextstrain.org](https://nextstrain-s-nextstrain-rbtee6.herokuapp.com/)

The bug pathologically slows down (and sometimes crashes) pages with thousands/millions of SVG elements in the DOM, which happens in reasonably sized datasets.¹

Short circuit the extension's calls to the very slow .innerText in its function:

```
  RM = (e) => {
    if (e == null) return null;
    let t = e.parentNode,
      n = 0;
    for (; n < 4 /*LM*/ && t !== null; ) {
      if (
        t instanceof HTMLElement &&
        t.innerText &&
        t.innerText.trim().length > 0
      )
        return t.innerText;
      (t = t.parentNode), n++;
    }
    return null;
  },
```

by nesting each panel's D3 elements deeper than extension's traversal limit of 4.  That is, we're moving the closest HTMLElement (e.g. \<div\>s containing \<svg\>s) further away from the leaf SVG elements.

For most panels, fewer \<g\>s would do, but adding four guards against future changes to D3 implementation.

The workaround was trivial for entropy and frequencies panels which already have a separation where size/styles are applied on a parent \<svg\> and D3 operations happen on a child \<g\>. It was still not too bad for map and measurements panels - those did not have a pre-existing separation, but adding the separation was trivial.

The tree panel was the least trivial - the code and type annotations assumed that the root SVG element for D3 operations was an \<svg\>, so switching it to \<g\> required updating those assumptions. Additionally, there is code that retrieves the width/height from the root SVG element, so it must be retained on the child element even when nested under an \<svg\>, which requires width and height to be set explicitly.

¹ <https://github.com/nextstrain/auspice/issues/1919>

## Related issue(s)

#1919

## Checklist

- [x] Checks pass (docs build failure is a linting error that can be ignored)
- [x] test entropy: tb [before](https://nextstrain.org/tb/global?d=entropy), [after](https://nextstrain-s-nextstrain-rbtee6.herokuapp.com/tb/global?d=entropy)
- [x] test frequencies: ncov color by pango lineage [before](https://nextstrain.org/ncov/gisaid/global/6m@2025-01-10?c=pango_lineage&d=frequencies), [after](https://nextstrain-s-nextstrain-rbtee6.herokuapp.com/ncov/gisaid/global/6m@2025-01-10?c=pango_lineage&d=frequencies)
    - The extension effects aren't as noticeable here, I had to use dev tools profiling to confirm that innerText is not excessively called
- [x] test map: ncov color by pango lineage [before](https://nextstrain.org/ncov/gisaid/global/6m@2025-01-10?c=pango_lineage&d=map), [after](https://nextstrain-s-nextstrain-rbtee6.herokuapp.com/ncov/gisaid/global/6m@2025-01-10?c=pango_lineage&d=map)
- [x] test tree: ncov [before](https://nextstrain.org/ncov/gisaid/global/6m@2025-01-10?d=tree), [after](https://nextstrain-s-nextstrain-rbtee6.herokuapp.com/ncov/gisaid/global/6m@2025-01-10?d=tree)
- [x] test measurements: tested locally with Auspice test dataset (no link)

### DIFF
--- a/src/components/entropy/index.js
+++ b/src/components/entropy/index.js
@@ -290,7 +290,12 @@ class Entropy extends React.Component {
           width={this.props.width}
           height={this.props.height}
         >
-          <g ref={(c) => { this.d3entropy = c; }} id="d3entropy"/>
+          {/* TODO: remove intermediate <g>s once the 1Password extension interference is resolved
+            * <https://github.com/nextstrain/auspice/issues/1919>
+            */}
+          <g><g><g><g>
+            <g ref={(c) => { this.d3entropy = c; }} id="d3entropy"/>
+          </g></g></g></g>
         </svg>
         {this.resetLayout(styles)}
         {this.entropyCountSwitch(styles)}
--- a/src/components/tree/tree.tsx
+++ b/src/components/tree/tree.tsx
@@ -27,8 +27,8 @@ const rhsTreeId = "RIGHT";
 export class TreeComponent extends React.Component<TreeComponentProps, TreeComponentState> {
 
   domRefs: {
-    mainTree: SVGSVGElement | null;
-    secondTree: SVGSVGElement | null;
+    mainTree: SVGGElement | null;
+    secondTree: SVGGElement | null;
   };
   tangleRef?: Tangle;
   clearSelectedNode: (node: SelectedNode) => void;
@@ -186,12 +186,23 @@ export class TreeComponent extends React.Component<TreeComponentProps, TreeCompo
   }) {
     return (
       <svg
-        id={mainTree ? "MainTree" : "SecondTree"}
+        id="d3treeParent"
         style={{pointerEvents: "auto", cursor: "default", userSelect: "none"}}
         width={width}
         height={height}
-        ref={(c) => {mainTree ? this.domRefs.mainTree = c : this.domRefs.secondTree = c;}}
-      />
+      >
+        {/* TODO: remove intermediate <g>s once the 1Password extension interference is resolved
+          * <https://github.com/nextstrain/auspice/issues/1919>
+          */}
+        <g><g><g><g>
+          <g
+            id={mainTree ? "MainTree" : "SecondTree"}
+            width={width}
+            height={height}
+            ref={(c) => {mainTree ? this.domRefs.mainTree = c : this.domRefs.secondTree = c;}}
+          />
+        </g></g></g></g>
+      </svg>
     );
   }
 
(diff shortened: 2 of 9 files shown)


src/components/tree/tree.tsx
```diff
@@ -186,12 +186,20 @@ export class TreeComponent extends React.Component<TreeComponentProps, TreeCompo
...
       <svg
-        id={mainTree ? "MainTree" : "SecondTree"}
+        id="d3treeParent"
         style={{pointerEvents: "auto", cursor: "default", userSelect: "none"}}
         width={width}
         height={height}
-        ref={(c) => {mainTree ? this.domRefs.mainTree = c : this.domRefs.secondTree = c;}}
-      />
+      >
+        <g><g><g>
```
genehack, line 194: previous guards used 4 `<g>`, this only has 3 — intentional?

src/components/entropy/index.js
```diff
@@ -290,7 +290,9 @@ class Entropy extends React.Component {
           width={this.props.width}
           height={this.props.height}
         >
-          <g ref={(c) => { this.d3entropy = c; }} id="d3entropy"/>
+          <g><g><g><g>
```
genehack, line 293: I wonder if these guards warrant an explanatory comment with the issue number …if nothing else, it will make it easier to excise them down the road, should we end up needing to.

src/components/tree/tree.tsx
victorlin: Copy/paste error, fixed:

https://github.com/nextstrain/auspice/blob/17b3f5c08af54c618de64d8673bd6adb5b16abcb/src/components/tree/tree.tsx#L197-L204

src/components/entropy/index.js
victorlin: Added comments:

https://github.com/nextstrain/auspice/blob/17b3f5c08af54c618de64d8673bd6adb5b16abcb/src/components/entropy/index.js#L293-L295

genehack (approved):

jameshadfield (approved):
I haven't tested this with the 1password extension, but gave it a quick test on its own and (as expected) it works fine. A minor annoyance for me when inspecting elements that there's more `<g>` elements to open in dev-tools, but overall a nice and simple fix for a big problem (assuming it does fix the problem).

PR #1934, opened 2025-01-22
Update changelog

victorlin:
Fixed formatting and added entries for internal changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 # Changelog
 
 * Added a new color tree by measurements feature.
-Clicking on a group in the measurements panel will add a new coloring to the tree,
-where the colors represent an average of the measurement values for the matching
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
victorlin, line 1: These changes render fine on the [GitHub Markdown preview](https://github.com/nextstrain/auspice/blob/0b6265e561cd7cd1db0767596fa7359564e3be55/CHANGELOG.md#version-2620---20250121) but not as great on the [GitHub Release page](https://github.com/nextstrain/auspice/releases/tag/v2.62.0), which seems to retain newlines from plain text:

image

We could probably update [extract-release-notes.js](https://github.com/nextstrain/auspice/blob/0b6265e561cd7cd1db0767596fa7359564e3be55/scripts/extract-release-notes.js) to address this, but 🤷

PR #1935, opened 2025-01-22
Retroactively add changelog entry

victorlin:
Accidentally skipped in the pull request 🤦

<https://github.com/nextstrain/auspice/pull/1932>


genehack (approved):

jameshadfield:
Thanks for [also adding it to the release notes!](https://github.com/nextstrain/auspice/releases/tag/v2.62.0)

victorlin:
Unfortunately this won't show up on the [versioned docs page](https://docs.nextstrain.org/projects/auspice/en/v2.62.0/releases/changelog.html), but I think that's something we can live with.

Issue #1441, opened 2021-12-21
Support build avatars sourced from non-GitHub URLs


huddlej:
**Context**  
Each Nextstrain build displays an avatar in the build's byline on nextstrain.org. [This avatar is sourced from the GitHub repository associated with the build](https://github.com/nextstrain/auspice/blob/a0e9080dcd61249f24f6c822dfcf6478e4e07ac2/src/components/info/byline.js#L36-L51) in the Auspice config JSON's `build_url` parameter. This approach allows Nextstrain Groups and community builds, like the Neher lab's European builds to include attribution for their group with the icon shown below:

<img width="808" alt="image" src="https://user-images.githubusercontent.com/85372/147007302-882f862b-b59e-45a0-b909-b63db2a7de0a.png">

**Description**   
Users (or groups of users) who deploy their data through Nextstrain Groups may not wish (or may not be allowed) to use GitHub to source their workflows. However, they do want a way to include their group logo in the byline of their Nextstrain builds.

Could we allow users to define a path to a custom avatar that is not hosted on GitHub?

**Examples**

For example, we could allow users to define their own avatar path in the Auspice JSON config like so:

```json
{
  "title": "Genomic epidemiology of novel coronavirus",
  "build_url": "https://nextstrain.org",
  "build_avatar": "https://path.to/avatar.png",
  "maintainers": [
    {"name": "Someone Great", "url": "https://nextstrain.org"}
  ],
```

**Possible solution**   

It seems like we could modify [the existing `renderAvatar` method](https://github.com/nextstrain/auspice/blob/a0e9080dcd61249f24f6c822dfcf6478e4e07ac2/src/components/info/byline.js#L36-L51) to look for an `avatarUrl` before it falls back to looking for a GitHub avatar associated with the build URL.

sagar-pathak:
I want to work on this problem. I will add this following code block inside renderAvatar function.

```
  const { buildAvatar, buildUrl } = metadata;
  // If buildAvatar is present, return the avatar image
  if (buildAvatar) {
    return <AvatarImg alt="avatar" width="28" src={buildAvatar} />;
  }
```

And also update the createMetadataStateFromJSON function (recomputeReduxState.js) so that buildAvatar will be available on the metadata state. 

```
  if (json.meta.build_avatar) {
    metadata.buildAvatar = json.meta.build_avatar;
  }
```

Let me know if this is how I should proceed. Also, I will update the schema information on augur.  Thanks

PR #1942, opened 2025-02-17
Support build avatars sourced from non-GitHub URLs #1441

sagar-pathak:
Users will now be able to define a path to a custom avatar that is not hosted on GitHub through the config file. 

## Related issue(s)
Related Issue: #1441 

- [ ] Checks pass

### DIFF
--- a/src/actions/recomputeReduxState.js
+++ b/src/actions/recomputeReduxState.js
@@ -755,6 +755,9 @@ const createMetadataStateFromJSON = (json) => {
   if (json.meta.build_url) {
     metadata.buildUrl = json.meta.build_url;
   }
+  if (json.meta.build_avatar) {
+    metadata.buildAvatar = json.meta.build_avatar;
+  }
   if (json.meta.data_provenance) {
     metadata.dataProvenance = json.meta.data_provenance;
   }
--- a/src/components/info/byline.js
+++ b/src/components/info/byline.js
@@ -39,9 +39,14 @@ const AvatarImg = styled.img`
  * which is a GitHub repo. The avatar image is fetched from GitHub (by the client).
  */
 function renderAvatar(t, metadata) {
-  const repo = metadata.buildUrl;
-  if (typeof repo === 'string') {
-    const match = repo.match(/(https?:\/\/)?(www\.)?github.com\/([^/]+)/);
+  const { buildAvatar, buildUrl } = metadata;
+  // If buildAvatar is present, return the avatar image
+  if (buildAvatar) {
+    return <AvatarImg alt="avatar" width="28" src={buildAvatar} />;
+  }
+
+  if (typeof buildUrl === 'string') {
+    const match = buildUrl.match(/(https?:\/\/)?(www\.)?github.com\/([^/]+)/);
     if (match) {
       return (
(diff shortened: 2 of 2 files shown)


jameshadfield (commented):
The changes look ok to me and work well on the datasets I've manually adjusted.

I think React's escaping of strings is enough to prevent XSS but would appreciate 👀  from @joverlee521 and @genehack on this. And also whether to allow base64 encoded images etc, however I think we already allow these in the markdown footer, right?

genehack:
> I think React's escaping of strings is enough to prevent XSS

From what I understand, with the way this is done (setting the `src` prop value via interpolation), React covers the XSS angle here. (To be clear, I did not try to inject something malicious, like a Base64-encoded hunk of JS, but I don't _think_ that will work in an `` element the way it would with a `<a href`. 

> And also whether to allow base64 encoded images etc, however I think we already allow these in the markdown footer, right?

I think supporting Base64-encoded images would be good; we may need to indicate something about how to do this in some docs somewhere, if we don't already?

I did leave one tweak suggestion on the PR that I think should be added, as a belt-and-suspenders check that somebody didn't stick an array or an object in the JSON instead of the expected string (the same way we check the `buildUrl` before the regex parsing is attempted).

sagar-pathak:
I added the check for the string and tested it with a base64 image, which works. However, I'm unsure how we will add validation for base64 images since they will be received as a string. @genehack @jameshadfield

tsibley:
This will allow visitor tracking by third-parties. Do we intend to allow that?

sagar-pathak:
> This will allow visitor tracking by third-parties. Do we intend to allow that?

Hi @tsibley, thank you for your response. Would you mind explaining more on this? Do you mean by injecting code through the build_avatar third-parties can track visitors?

jameshadfield:
> This will allow visitor tracking by third-parties. 

More so than allowing outbound `src` images in markdown footers etc?

tsibley:
> More so than allowing outbound `src` images in markdown footers etc?

Nope!

> Do you mean by injecting code through the build_avatar third-parties can track visitors?

Not code, HTTP requests.

sagar-pathak:
> Not code, HTTP requests.

Wouldn't it be the same in the URL field for maintainers? Sorry, I didn't quite understand. Could you please help me clarify?

corneliusroemer:
@sagar-pathak The URL isn't loaded by default, it requires a click IIUC.

The avatar would _always_ be loaded.

Do we have precedent for loading such resources? If so, this PR might not worsen things. However if this would be a first it would be significant.

It might even require adding a consent notice in the EU to allow the request (if it uses for tracking purposes - but we don't know the purpose so we'd sort of have to err on the side of caution)

@sagar-pathak I appreciate you suggesting this change, could you explain your use case? Maybe there are other, more privacy preserving ways to achieve the same goal.

sagar-pathak:
@corneliusroemer Thanks, that makes perfect sense. I will check what best practices are there to handle such situations.

jameshadfield (approved):
Thanks @sagar-pathak - much appreciated!

> [cornelius] Do we have precedent for loading such resources? 

Yes, we already allow (and load) images with outbound `src` in markdown footers & narratives.

PR #1947, opened 2025-03-28
Reset layout when nucleotide position is selected

### LINKED ISSUE #1774: Reset layout in entropy view has no effect when URL params are set: either disable button in that case, or make it override URL params
### Current Behavior

When one colors by a nucleotide, the entropy view panel does not respect clicks on "reset layout" - in contrast to at least my personal user experience.

### Expected behavior

I would expect one of the following things as a user instead:
a) If the button doesn't work in this case of coloring, disable it so it is greyed out and/or shows a 🚫 sign to notify the user
b) Override the auto-view setting, allow reset layout to uncouple from the initial zoom that is the result of coloring
c) add additional button to "unlock" the coupling

### How to reproduce

1. Go to entropy panel
2. Click on nucleotide
3. Click reset layout
4. Observe result

compressed

victorlin:

Previously, this did not work because the dispatch to reset layout was conditional on having a CDS selected.

## Related issue(s)

Closes #1774

## Checklist

- [ ] Checks pass

### DIFF
--- a/src/actions/entropy.js
+++ b/src/actions/entropy.js
@@ -69,9 +69,6 @@ export const changeEntropyCdsSelection = (arg) => (dispatch, getState) => {
   if (!entropy.loaded) return;
 
   if (arg === nucleotide_gene) {
-    if (entropy.selectedCds === nucleotide_gene) {
-      return
-    }
     action.selectedCds = arg;
     action.selectedPositions = [];
   } else if (typeof arg === 'string') {
--- a/src/components/entropy/index.js
+++ b/src/components/entropy/index.js
@@ -147,9 +147,8 @@ class Entropy extends React.Component {
                 zoomMin: this.state.chart.zoomBounds[0],
                 zoomMax: this.state.chart.zoomBounds[1],
               })
-            } else {
-              this.props.dispatch(changeEntropyCdsSelection(nucleotide_gene));
             }
+            this.props.dispatch(changeEntropyCdsSelection(nucleotide_gene));
           }}
         >
           <span style={styles.switchTitle}> {'RESET LAYOUT'} </span>


genehack (approved):

src/actions/entropy.js
```diff
@@ -69,9 +69,6 @@ export const changeEntropyCdsSelection = (arg) => (dispatch, getState) => {
   if (!entropy.loaded) return;
 
   if (arg === nucleotide_gene) {
-    if (entropy.selectedCds === nucleotide_gene) {
-      return
-    }
```
jameshadfield, line 74: If you've highlighted a nucleotide bar (in the entropy panel) and click "reset layout" the bar will no longer be highlighted because now `action.selectedPositions = [];` will be hit below.

src/components/entropy/index.js
```diff
@@ -147,9 +147,8 @@ class Entropy extends React.Component {
                 zoomMin: this.state.chart.zoomBounds[0],
                 zoomMax: this.state.chart.zoomBounds[1],
               })
-            } else {
-              this.props.dispatch(changeEntropyCdsSelection(nucleotide_gene));
             }
+            this.props.dispatch(changeEntropyCdsSelection(nucleotide_gene));
```
jameshadfield, line 151: With this change (and the earlier change), the following just above this line becomes unnecessary and can be removed:
```js
            if (viewingGenome) {
              this.state.chart.update({
                zoomMin: this.state.chart.zoomBounds[0],
                zoomMax: this.state.chart.zoomBounds[1],
              })
            }
```

This is because we end up running `update({selectedPositions: []})` which runs
```js
    if (selectedCds || selectedPositions !== undefined) {
      this._setZoomCoordinates(zoomMin, zoomMax, !!selectedCds);
    }
```

so we'll always reset the coordinates to show the entire nuc span.

jameshadfield (changes requested):

src/components/entropy/index.js
victorlin: It's still used for when no CDS or position is selected, when the min/max sliders are adjusted. I think the update can happen elsewhere though. Started with 1f40cd0d5caaecc1b7ae47e7bb2e707b2ed92452 but haven't gotten it fully working.

src/actions/entropy.js
victorlin: Simply dropping `action.selectedPositions = []` doesn't work, I think because zoom bounds are tightly coupled with selected positions here:

https://github.com/nextstrain/auspice/blob/1f40cd0d5caaecc1b7ae47e7bb2e707b2ed92452/src/components/entropy/entropyD3.js#L131-L145

PR #1948, opened 2025-03-28
Conditionally disable reset layout button

victorlin:

Render this button useless when the view is not zoomed.

## Related issue(s)

Follow-up to https://github.com/nextstrain/auspice/pull/1684#issuecomment-1681955133

## Checklist

- [ ] Checks pass
- [ ] Merge base PR #1947 

### DIFF
--- a/src/components/entropy/index.js
+++ b/src/components/entropy/index.js
@@ -72,11 +72,6 @@ const getStyles = (width) => {
     loaded: state.entropy.loaded,
     onScreen: state.entropy.onScreen,
     colorBy: state.controls.colorBy,
-    /**
-     * Note that zoomMin & zoomMax only represent the state when changed by a URL
-     * i.e. on dataset load or narrative page change. As such, they fall out-of-sync
-     * as soon as any user-zooming is performed.
-     */
     zoomMin: state.controls.zoomMin,
     zoomMax: state.controls.zoomMax,
     defaultColorBy: state.controls.defaults.colorBy,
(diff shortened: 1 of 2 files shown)


src/components/entropy/index.js
```diff
@@ -266,8 +229,65 @@ class Entropy extends React.Component {
...
+    return (
+      (this.props.selectedCds != nucleotide_gene) ||
+      (this.state.chart.zoomBounds && this.props.zoomMin != this.state.chart.zoomBounds[0]) ||
+      (this.state.chart.zoomBounds && this.props.zoomMax != this.state.chart.zoomBounds[1])
+    );
+  }
+
   render() {
-    const styles = getStyles(this.props.width);
+    const zoomed = (
```
jameshadfield, line 285: `zoomed` isn't used, which makes me think this PR is still a WIP?

src/reducers/controls.ts
```diff
@@ -512,15 +512,8 @@ const Controls = (state: ControlsState = getDefaultControlsState(), action): Con
...
-     * Currently the CHANGE_ZOOM action (entropy panel zoom changed) does not
-     * update the zoomMin/zoomMax, and as such they only represent the initially
-     * requested zoom range. The following commented out code will keep the
-     * state in sync, but corresponding changes will be  required to the entropy
-     * code.
-     */
-    // case types.CHANGE_ZOOM: // this is the entropy panel zoom
-    //   return {...state, zoomMin: action.zoomc[0], zoomMax: action.zoomc[1]};
+    case types.CHANGE_ZOOM: // this is the entropy panel zoom
+      return {...state, zoomMin: action.zoomc[0], zoomMax: action.zoomc[1]};
```
jameshadfield, line 516: I remember disabling this as it caused bugs, or maybe unnecessary renders, but I can't remember the specifics and 2a063a706b82e8b8938bdb6ceb9f5a3bde85b11c doesn't spell them out either. I'll do some more testing when the PR's ready to go!

jameshadfield: > or maybe unnecessary renders

This is it I think. With these changes we now trigger two calls to `update` rather than one. The first has no zoom information, the second sets min/max to the overall zoom bounds. (Sometimes three calls, but my suggested changes in #1947 will limit it to two calls.)

I don't think this is blocking, although would be worth spending a small amount of time to see if the logic can be improved so there's only one `update` call.

PR #1949, opened 2025-03-28
Add Auspice PR title to preview PR title

victorlin:

It's difficult to tell what the PR is for from a number alone. The title should provide better context.

One downside is that any title updates won't propagate, but I think we can live with that. The code could be tweaked if we really wanted to update the title.

## Checklist

- [x] Checks pass (docs failure unrelated)


genehack (approved):

jameshadfield:
Nice improvement - thanks!

victorlin:
Of course it's a good idea to not merge these preview PRs, but I don't think it's necessary to call out as all-caps text in the title. It could simply be noted in the description. If we really want to prevent human error, there are better ways to enforce* such as automatically keeping the PR in draft state.

* only goes so far without branch protection rules on the target branch

PR #1952, opened 2025-03-31
JSON-definable zoom label

jameshadfield:
First three commits tidy up previous code and add some more error handling  & improved behaviour in edge cases. 

Fourth commit allows us to specify a `display_default.label` in the JSON which behaves the same as the URL query `label` we have had for a long time. There are two improvements which in the interest of time I'd like to leave as future to-dos, but I can fix them here if people think they'll be encountered sooner rather than later:

1. One side effect with defining a display default label is that you can't always specify a URL which will load the tree at the tree root. If there's a branch label on the root node then you can almost get there by "zooming" to the root node which will then set a 'label' URL query.

2. Also note that when interacting with the tree, zooming to the same label as the display default will set a URL query 'label', which differs from our typical handling of defaults where we would not set any URL query here.

### DIFF
--- a/docs/advanced-functionality/view-settings.rst
+++ b/docs/advanced-functionality/view-settings.rst
@@ -51,6 +51,8 @@ These are exported as the (optional) property of the dataset JSON ``meta.display
 +---------------------------+-----------------------------------------------------------------------+----------------------------------------------------+
 | ``language``              | Language to display Auspice in                                        | "ja"                                               |
 +---------------------------+-----------------------------------------------------------------------+----------------------------------------------------+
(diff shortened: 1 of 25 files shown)


docs/advanced-functionality/view-settings.rst
victorlin, line 1: This may be nitpicky, but the "reset layout" button feels half-baked as-is. When there is a default set, clicking it will zoom out to the root. I see two options:

1. Update the button to reset to the default zoom.
2. Rename the button to something more descriptive of what it is doing, e.g. "zoom to root".

I don't have a strong opinion on which option to take.

PR #1959, opened 2025-05-14
Add Chinese localization support.

zhangmengjie-amy:

Add Chinese Localization Support

## Checklist

- [x] Checks pass


victorlin (approved):
Thank you @zhangmengjie-amy, this is great! I will edit the changelog and merge.

There are some bugs in the code that prevent some of these from showing, but that applies to all other languages. I will make a separate PR to fix that.

src/locales/zh/translation.json
```diff
@@ -0,0 +1,91 @@
+{
+  "__Header//Byline__": "########################################",
+  "Maintained by": "维护者",
+  "using data from": "数据来源",
+  "Built with": "构建工具",
```
victorlin, line 5: I noticed "构建工具[nextstrain/ncov](https://github.com/nextstrain/ncov)" reads weirdly as "build tool: [nextstrain/ncov](https://github.com/nextstrain/ncov)". My level of Chinese is not great, but I'm thinking something like "基于[nextstrain/ncov](https://github.com/nextstrain/ncov)构建" is a more accurate translation.

Right now it is not possible because the URL placement is not included in the translation, but we can change that. What do you think @zhangmengjie-amy?

zhangmengjie-amy: @victorlin Yes, that's a very nice propose, in Chinese, it is more standard and reasonable to translate it as "基于[nextstrain/ncov](https://github.com/nextstrain/ncov)构建". In this case, it is necessary to add a dynamic variable URL in the translation.

PR #1964, opened 2025-05-15
Include link in "built with …" translation key

victorlin:

For some languages, it does not make sense to place the link at the end of the sentence. The need to set placement per-language means it should be included in the translation key, similar to existing keys formatted as `{{...}}`.

This isn't possible with the `t()` function which only works with string interpolation, but can be accomplished with `<Trans>` and a more verbose translation key.

The new Chinese translation is improved as an example in this PR's changes. For other languages, the link has been included at the end to match existing behavior.

## Related issue(s)

Follow-up to https://github.com/nextstrain/auspice/pull/1959#discussion_r2090252016

## Checklist

- [x] Checks pass


jameshadfield (approved):
Tested the review app and read the code, but didn't look into the `<Trans>` API. Thanks!

Issue #1855, opened 2024-09-12
Normalise frequencies error when changing datasets via sidebar


jameshadfield:
When changing between two datasets via the sidebar dropdown UI, each with a tip frequencies sidecar, we saw someone get a warning banner indicating a failure to normalise frequencies and the sidebar option is not available. The relevant code for the sidebar is:

https://github.com/nextstrain/auspice/blob/de94b72e155e7d386f1c82509727b422f0c6f99d/src/components/controls/frequency-normalization.js#L22-L32

and `checkIfNormalizableFromRawData` seems to always return `true` (!)

https://github.com/nextstrain/auspice/blob/de94b72e155e7d386f1c82509727b422f0c6f99d/src/util/processFrequencies.js#L62-L63

implying a race condition around `this.props.frequencies.loaded && this.props.tree.loaded`. 

Reloading the page works as expected.

I don't have any reproducible datasets (have asked for them), and note that changing https://nextstrain.org/seasonal-flu/h3n2/ha/3y to 2y (via sidebar) works as expected. Perhaps it's dataset specific (hard to understand after skimming the relevant code) or using an old version of auspice (normalisation checking was removed in https://github.com/nextstrain/auspice/commit/7b9495b107517008ef5d1dca7466e83f3db2984a in 2001)

joverlee521:
I can reproduce locally by changing the dataset via one of the dropdowns that is _not_ the last one, e.g. changing from http://localhost:4000/flu/seasonal/h1n1pdm/ha/12y to na via sidebar triggers the warning.

jameshadfield:
This bug is only observed when using the Auspice server (i.e. not within nextstrain.org). Steps to reproduce:

1. provision appropriate datasets
```
nextstrain remote download nextstrain.org/seasonal-flu/h3n2/ha/2y
nextstrain remote download nextstrain.org/seasonal-flu/h3n2/na/2y
nextstrain remote download nextstrain.org/seasonal-flu/h3n2/ha/3y
```
2. Within auspice, load seasonal-flu/h3n2/ha/2y
3. Use the sidebar to change "ha" to "na". 
4. The error is shown and the entire frequencies panel is missing

jameshadfield:
Notes from a briefly looking into this (have to jump to other things now):

In the above example, we use an incomplete URL to fetch the tip frequencies:
```
this.apiCalls.tipFrequencies = /charon/getDataset?prefix=seasonal-flu/h3n2/na&type=tip-frequencies
[...]

Issue #1528, opened 2022-06-20
Unconventional camel case for CLI args rather than GNU convention kebab-case


corneliusroemer:
### Context
I was surprised to find that the (long) CLI args of Auspice are camel case rather than kebab case, as is conventional for CLIs

### Description
Switch to using kebab-case by default, camel case should be kept for backwards compatibility only, but help should show kebab caes.

### Examples
Current arguments
```
❯ npx auspice view --help
usage: auspice view [-h] [--verbose] [--handlers JS] [--datasetDir PATH]
                    [--narrativeDir PATH]
                    

Launch a local server to view locally available datasets & narratives. The 
handlers for (auspice) client requests can be overridden here (see 
documentation for more details). If you want to serve a customised auspice 
client then you must have run "auspice build" in the same directory as you 
run "auspice view" from.

Optional arguments:
  -h, --help           Show this help message and exit.
  --verbose            Print more verbose logging messages.
  --handlers JS        Overwrite the provided server handlers for client 
                       requests. See documentation for more details.
  --datasetDir PATH    Directory where datasets (JSONs) are sourced. This is 
                       ignored if you define custom handlers.
  --narrativeDir PATH  Directory where narratives (Markdown files) are 
                       sourced. This is ignored if you define custom handlers.
```

jameshadfield:
Good to be thinking about this. If we change these I'd be more inclined to accept any number of paths at the end of the command which are scanned for markdown / json files. This would allow you to use recursive expansion if you store files in a hierarchy etc. 
```
auspice [arguments] <directory> [<directory> ...]
```

corneliusroemer:
Agree, since the directory is mandatory it makes a lot of sense to make it a positional argument

PR #1967, opened 2025-05-26
[v3] [cli] Allow one or more positional path args

jameshadfield:
The previous interface was slightly cumbersome - I often found myself wanting to source both datasets & narratives from some directory X and having to type in the rather long `--datasetDir X --narrativeDir X`, and often I'd misspell / pluralise the argument names; using a simple positional argument `X` is much nicer. The ability to use multiple directories is also functionality I've long wanted.

The previous CLI interface is preserved in its entirety for backwards compatibility and is mutually exclusive with the new positional-args style.


jameshadfield:
Will close #1528

jameshadfield:
Closes #1855 (in addition to an already-merged PR)

PR #1981, opened 2025-05-29
Adjust types

victorlin:

Add missing types, reuse existing definitions, and add newlines for readability.

## Related issue(s)

Follow-up to #1902

## Checklist

- [ ] Checks pass

### DIFF
--- a/src/actions/treeStreams.ts
+++ b/src/actions/treeStreams.ts
@@ -3,8 +3,9 @@ import { processStreams, labelStreamMembership, isNodeWithinAnotherStream } from
 import { getParentStream } from "../components/tree/phyloTree/helpers";
 import { updateVisibleTipsAndBranchThicknesses } from "./tree";
 import { warningNotification } from "./notifications";
+import { ThunkFunction } from "../store";
 
-export function toggleStreamTree() {
+export function toggleStreamTree(): ThunkFunction {
   return function(dispatch, getState) {
     const {controls, tree} = getState();
     const showStreamTrees = !controls.showStreamTrees; // new state
@@ -44,7 +45,7 @@ export function toggleStreamTree() {
 } 
 
 
-export function changeStreamTreeBranchLabel(newLabel) {
+export function changeStreamTreeBranchLabel(newLabel: string): ThunkFunction {
   return function(dispatch, getState) {
     const {controls, tree} = getState();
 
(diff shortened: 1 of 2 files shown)


src/actions/treeStreams.ts
```diff
@@ -3,8 +3,9 @@ import { processStreams, labelStreamMembership, isNodeWithinAnotherStream } from
 import { getParentStream } from "../components/tree/phyloTree/helpers";
 import { updateVisibleTipsAndBranchThicknesses } from "./tree";
 import { warningNotification } from "./notifications";
+import { ThunkFunction } from "../store";
 
-export function toggleStreamTree() {
+export function toggleStreamTree(): ThunkFunction {
```
jameshadfield, line 8: Can we use linting to enforce return types?

src/actions/treeStreams.ts
```diff
@@ -44,7 +45,7 @@ export function toggleStreamTree() {
 } 
 
 
-export function changeStreamTreeBranchLabel(newLabel) {
+export function changeStreamTreeBranchLabel(newLabel: string): ThunkFunction {
```
jameshadfield, line 48: Ditto - any reason not to use linting to enforce parameter types?

jameshadfield (commented):
Stepping back a bit, I'd like us to use linters to enforce things like this and thus avoid having to make these fix-up PR. (And similarly for the periodic PRs across different repos which remove trailing whitespace etc)

Previous discussions about linting have (I think?) made the decision to not enforce stylistic aspects. But if people feel strongly enough about stylistic aspects that they'll go to the effort of making a PR then I think that should be part of linting!

Going through these, using types for function args / return values seems like something we should do. (Discussed above.)

Readability is inherently subjective, for instance I'm ambivalent about `false|Node` vs `false | Node` so if you prefer the latter I'm happy to have the linter guide my hand. For other things I find

```js
function labelStreamMembership(tree: ReduxNode, branchLabelKey): Streams {
```

much more readable than
```js
function labelStreamMembership(
  tree: ReduxNode,
  branchLabelKey: string,
): Streams {
```

Maybe others disagree? I'd prefer these to be left up to the original author in most cases, or enforce a line length for function defs or something.

victorlin:
Absolutely, sorry I took it too far with the stylistic aspects. I'll open a separate PR with just the lint-able parts and linter enforcement enabled.

PR #1988, opened 2025-06-05
Unit tests for createStateFromQueryOrJSONs

jameshadfield:
This was primarily motivated by the bug identified in <https://github.com/nextstrain/auspice.us/issues/122> and fixed via a one-line patch in <https://github.com/nextstrain/auspice/pull/1986>. I confirmed that the tests added here would have flagged this bug (via reverting https://github.com/nextstrain/auspice/commit/ece625f57106e56c4e59c4516f4b04dbe07355cc).

There's a lot of future directions we can take tests such as these, this is only the starting point.

### DIFF
--- a/test/data/generic.json
+++ b/test/data/generic.json
@@ -1,0 +1,91 @@
+{
+    "version": "v2",
+    "meta": {
+      "title": "Generic (small) feature-rich dataset for testing purposes",
+      "updated": "2023-03-28",
+      "build_url": "https://github.com/nextstrain/auspice",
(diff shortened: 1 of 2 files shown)


genehack (approved):

jameshadfield:
This PR originally polyfilled `structuredClone`, which required a core-js bump, which then increased the bundlesizes over our (completely arbritrary) limits. I went down a deep rabbit hole of bundle-related stuff, but for the purposes of this PR I've just dropped the polyfill and used the good old JSON parse/stringify round trip.

PR #1993, opened 2025-06-10
Focal zoom mode acts on both axis

trvrb:
This causes date updates and other filters to dynamically update the XY bounds, ensuring that the tree fills the page. This works well for zooming to early samples and give a generally more dynamic feeling. It combines very well with "focus on selected".

I believe that this could hacked on to combine with "focus on selected", but the intent here is just to produce a review app for playing around with.

### DIFF
--- a/src/components/tree/treeButtons.tsx
+++ b/src/components/tree/treeButtons.tsx
@@ -21,7 +21,8 @@ export function TreeButtons(props: Props): ReactElement {
   const tree = props.mainTree ? props.tree : props.treeToo;
   const filtered = !!tree.idxOfFilteredRoot &&
     tree.idxOfInViewRootNode !== tree.idxOfFilteredRoot;
-  const activeZoomButton = filtered;
+  const showZoomButton = !focusMode;
+  const activeZoomButton = filtered; // only applicable if `showZoomButton` is true
   const treeZoomed = tree.idxOfInViewRootNode !== 0;
 
   const containerStyles: React.CSSProperties = {zIndex: 100, position: "absolute", right: props.offsetPx, top: 0};
@@ -36,9 +37,11 @@ export function TreeButtons(props: Props): ReactElement {
         <FaSearchMinus/>
       </button>
 
-      <button style={activeZoomButton ? selectedButtonStyles : unselectedButtonStyles} onClick={zoomToSelected}>
-        {props.t("Zoom to Selected")}
-      </button>
+      {showZoomButton && (
+        <button style={activeZoomButton ? selectedButtonStyles : unselectedButtonStyles} onClick={zoomToSelected}>
+          {props.t("Zoom to Selected")}
+        </button>
+      )}
 
       <button style={treeZoomed ? selectedButtonStyles : unselectedButtonStyles} onClick={redrawTree}>
         { /* focus mode can't zoom back to the root, so change our language to suit */


src/components/tree/tree.tsx
```diff
@@ -287,31 +207,78 @@ export class TreeComponent extends React.Component<TreeComponentProps, TreeCompo
...
+        <TreeButtons {...this.props} />
       </Card>
     );
   }
 }
 
 export default withTranslation()(TreeComponent);
+
+
+function TreeButtons(props: TreeComponentProps): ReactElement {
```
jameshadfield, line 219: There is subtlety around how the tree buttons ("zoom to selected" etc) behave with this functionality!

**Situation 1** ZIKV, filtered to USVI & focus on Selected:

image 3

> NOTE: the green "root" shown in focus on selected is not tracked in redux state - it only looks like a root due to a viewport shift. Let's call this the "viewport root". We can figure it out with a tree traversal to find the first visible node.

What does clicking on the viewport root and each of the buttons do in this context? My thoughts:

* "-" doesn't make sense and should be inactivated. (Same as clicking on the viewport root branch)
* "zoom to selected" doesn't make sense, as we're continually keeping things focused on selected!
* "zoom to root" - should be shown inactivated (see situation 3)

**Situation 2** 
Things get more interesting when we shift the min date slider, as we are presented with a situation Auspice has so far completely avoided:

image 4

We now have three viewport roots. In this context:

* clicking on any of the viewport roots zooms into that subtree (current behaviour)
* "-" doesn't make sense and should be inactivated
* "zoom to selected" doesn't make sense and should be hidden
* "zoom to root" - should be shown inactivated (see situation 3)

**Situation 3** when zoomed into a USVI subtree, e.g:

image 5

* clicking on the viewport root does make sense and should zoom out (taking us towards the screenshots in situation 1 or 2, depending on the min-date slider value)
* "-" should do the same
* "zoom to selected" doesn't make sense and should be hidden
* "zoom to root" does make sense, and we should zoom out to the screenshots in situation 1 or 2, depending on the min-date slider value. This button should be renamed to something like "show all selected" or something.

jameshadfield: cc @trvrb does this line up with your expectations?

src/components/tree/phyloTree/layouts.ts
```diff
@@ -360,10 +361,32 @@ export const mapToScreen = function mapToScreen(this: PhyloTreeType): void {
...
-  let nodesInDomain = this.nodes.filter((d) => d.inView && d.y!==undefined && d.x!==undefined);
-  // scatterplots further restrict nodes used for domain calcs - if not rendering branches,
-  // then we don't consider internal nodes for the domain calc
-  if (this.layout==="scatter" && this.scatterVariables.showBranches===false) {
+  let nodesInDomain;
+  if (this.focus) {
+    /* ------------------------------------------------------------------
+    * New: limit the axis domains to nodes Auspice will actually render,
+    * Note: nodes marked as `inView` may be off-screen in this mode
+    * ------------------------------------------------------------------ */
```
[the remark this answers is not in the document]
jameshadfield: Worth noting that radial displays already behave as if this patch was applied, e.g. they look like:

image

victorlin:
Thanks for showing that this is possible!

My main concern is the conflation of "zoom" and "focus" as defined in https://github.com/nextstrain/auspice/issues/1368#issuecomment-2354019338.

I'd suggest putting this new feature behind its own toggle, at least for testing purposes. Example: 82c3e355cd6918e0e04a8890cafa436cc4ae25be
This enables views such as "zoom only" below.

zoom-vs-focus

It might make more sense to use this feature to replace the existing "Zoom to selected" button, rather than adding it to the "Focus on selected" toggle.

src/components/controls/toggle-focus.tsx
```diff
@@ -47,14 +49,28 @@ export function ToggleFocus(): JSX.Element {
...
+        dispatch(setFocus(valueAfterToggling))
       }}
       label={label}
       style={{ marginBottom: 8 }}
     />
   );
 }
+
+
+function setFocus(newToggleValue: null|"selected"): ThunkFunction {
```
jameshadfield: > [@victorlin] My main concern is the conflation of "zoom" and "focus" as defined in https://github.com/nextstrain/auspice/issues/1368#issuecomment-2354019338.

For testing purposes, perhaps a review app of https://github.com/nextstrain/auspice/commit/82c3e355cd6918e0e04a8890cafa436cc4ae25be would be good enough? Conceptually this PR can split them out, but there would be a lot of lines of changed code.

There's always a tension between more controls & complexity vs simplicity but less control. The conversations I had with Trevor led us to the view that exposing a simpler UI here was preferred. I'd like to hear what others think about this...

P.S. Your definitions of "zoom" vs "focus" are really helpful for internal discussions, but I don't think there's a problem calling everything "focus" in the Auspice UI if we do choose to use a single toggle.

src/components/tree/treeButtons.tsx
```diff
@@ -35,12 +37,16 @@ export function TreeButtons(props: Props): ReactElement {
         <FaSearchMinus/>
       </button>
 
-      <button style={activeZoomButton ? selectedButtonStyles : unselectedButtonStyles} onClick={zoomToSelected}>
-        {props.t("Zoom to Selected")}
-      </button>
+      {showZoomButton && (
```
jameshadfield: > [@victorlin] It might make more sense to use this feature to replace the existing "Zoom to selected" button, rather than adding it to the "Focus on selected" toggle.

I think it's a toggle state not a button action, as I like the dynamic behaviour as you change the date slider / filters. But you're right that they are similar, which is why I got rid of the "zoom to selected" button when in focus mode!

joverlee521: Ah I see, thanks for the examples in the review app. I was mixing up your idea of "dynamic zoom" with the existing "zoom to selected" button.

trvrb:
I was just reaching for this to try to look at https://nextstrain.org/measles/genome?c=division&dmin=2025-01-01&f_country=USA&focus=selected. It would be nice to have this merged (I didn't actually realize it wasn't). I do think that a single toggle to "Focus on selected" that applies both to y values and to x values makes sense semantically to me. Note with this measles example I can't click "Zoom to selected" in the tree panel.

PR #2003, opened 2025-08-11
Export URLs in TSVs via additional column

jameshadfield:
PR #1998 (merged but unreleased) exported URLs associated with metadata by appending to the value, i.e. `<value> :<url>`. This commit switches to using a separate column `<column_name>__url` for the URL. This change is motivated by a desire to encode URLs consistently across the nextstrain ecosystem, including in pathogen repo ingest pipelines¹ and `augur export`.

This commit essentially reverts commits 7315a9b4fddadf499bb9f85f77745442a72c9342, d7c85058caf61bfb25f42e000ed1bdd4c1ae50ad and 68c0fa6095ad4e11de790e39dece8b65a3b0d58b (all part of #1998).

¹ https://github.com/nextstrain/rsv/pull/94#discussion_r2214163028

- [ ] CHANGELOG

### DIFF
--- a/src/components/download/helperFunctions.js
+++ b/src/components/download/helperFunctions.js
@@ -250,7 +250,7 @@ export const strainTSV = (dispatch, filePrefix, nodes, nodeVisibilities) => {
     const traits = Object.keys(node.node_attrs).filter((k) => !nodeAttrsToIgnore.includes(k));
     for (const trait of traits) {
       const value = getTraitFromNode(node, trait);
-      if (value) {
+      if (value !== undefined) {
         headerInsert(headerFields, null, trait);
         if (typeof value === 'string') {
           tipTraitValues[node.name][trait] = value;
--- a/src/components/download/helperFunctions.js
+++ b/src/components/download/helperFunctions.js
@@ -235,7 +235,7 @@ export const strainTSV = (dispatch, filePrefix, nodes, nodeVisibilities) => {
     const numDate = getTraitFromNode(node, "num_date");
     if (numDate) {
       const traitName = "date"; // matches use in augur metadata.tsv
-      if (!headerFields.includes(traitName)) headerFields.push(traitName);
+      headerInsert(headerFields, null, traitName)
       const numDateConfidence = getTraitFromNode(node, "num_date", {confidence: true});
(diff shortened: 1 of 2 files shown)


src/components/download/helperFunctions.js
```diff
@@ -249,14 +249,15 @@ export const strainTSV = (dispatch, filePrefix, nodes, nodeVisibilities) => {
     const nodeAttrsToIgnore = ["author", "div", "num_date", "vaccine", "accession"];
     const traits = Object.keys(node.node_attrs).filter((k) => !nodeAttrsToIgnore.includes(k));
     for (const trait of traits) {
-      if (!headerFields.includes(trait)) headerFields.push(trait);
       const value = getTraitFromNode(node, trait);
       if (value) {
+        headerInsert(headerFields, null, trait);
```
joverlee521, line 254: _completely separate tangent_

I just realized this means any `0` value will not be included in metadata downloads. Clearly it hasn't been an issue since I don't see any related bug reports, but probably would be good to fix:

```suggestion
      if (value !== undefined) {
        headerInsert(headerFields, null, trait);
```

src/components/download/helperFunctions.js
```diff
@@ -268,27 +269,32 @@ export const strainTSV = (dispatch, filePrefix, nodes, nodeVisibilities) => {
...
+        tipTraitValues[node.name][urlify(traitName)] = fullAuthorInfo.paper_url;
       }
     }
 
     /* handle `vaccine` specially */
     const vaccine = getVaccineFromNode(node);
     if (vaccine && vaccine.selection_date) {
       const traitName = "vaccine_selection_date";
-      if (!headerFields.includes(traitName)) headerFields.push(traitName);
+      headerInsert(headerFields, null, traitName);
```
joverlee521, line 284: _non-blocking nit_

There are 3 other instances of 
```js
if (!headerFields.includes(traitName)) headerFields.push(traitName);
``` 
in this file that can be replaced with the new `headerInsert` function ;)

joverlee521 (approved):
Changes LGTM based on the exported accession and author URLs in the test datasets.

src/components/download/helperFunctions.js
jameshadfield: Good find! Fixed

PR #2012, opened 2025-09-17
Declare support for Node.js 24

victorlin:

I did some local testing with Node.js 24 and it worked fine without any code changes.

## Related issue(s)

Follow-up to https://github.com/nextstrain/auspice/pull/1975#discussion_r2110811691

## Checklist

- [x] Checks pass
- [x] ~(to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].~ N/A, downstream repos use Node.js 20

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [20, 22]
+        node: [20, 22, 24]
     steps:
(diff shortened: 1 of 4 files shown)


jameshadfield (approved):
[Node 24 allows us to write TS](https://nodejs.org/api/typescript.html#type-stripping) on the (Auspice) server which is nice. It won't be coming anytime soon here since v20 (and perhaps v22?) don't support it, but for the nextstrain.org server this would be a big win.

Issue #2011, opened 2025-09-15
When "focus selected", exiting single click-focused sequence with Esc keeps focus on single sequence


corneliusroemer:
When the "Focus on selected" toggle is ON, and one clicks on a single sequence to focus it and then hits the Escape key to defocus the sequence, the modal disappears but the focus remains on the single (previously focused) sequence in disagreement with actual filters

### How to reproduce

1. Toggle on "Focus on selected"
2. Click on a single sequence/tip with mouse
3. Hit Escape key
4. Observe results

### Current behavior

Focus is as if one had filtered for the sequence one had clicked before 

### Expected behavior

Focus is back to how it was before clicking on the single sequence/tip

### Your environment: if running Nextstrain locally
 - Operating system: macOS
 - Browser: Chrome
 - Version (e.g. `auspice 2.7.0`): Auspice v2.66.0

![Image](https://github.com/user-attachments/assets/c829152f-2b41-4ff0-87df-c379c275e951)

victorlin:
Interesting, this seems to be a timing issue. It doesn't happen for me when clicking outside of the modal to dismiss it, unless I add a delay here:

```diff
--- a/src/components/tree/phyloTree/change.ts
+++ b/src/components/tree/phyloTree/change.ts
@@ -438,7 +438,9 @@ export const change = function change(
 
   /** display order calculations */
   if (newDistance || updateLayout || changeNodeOrder || streamDefinitionChange) {
+    setTimeout(() => {
     setDisplayOrder({nodes: this.nodes, focus, streams: this.params.showStreamTrees && this.streams});
+    }, 100);
   } else if (this.params.showStreamTrees && (changeColorBy || changeVisibility)) {
     // rippleDisplayOrders are typically called by setDisplayOrder however for ∆{colorBy,visibility} we don't want to pay
     // the price of recomputing the display orders for the entire tree, we just need to recompute the
```

victorlin:
A minimal patch addressing the focus bug:

```diff
--- a/src/components/tree/phyloTree/change.ts
+++ b/src/components/tree/phyloTree/change.ts
@@ -437,8 +437,10 @@ export const change = function change(
    */
 
   /** display order calculations */
-  if (newDistance || updateLayout || changeNodeOrder || streamDefinitionChange) {
+  if (newDistance || updateLayout || changeNodeOrder || streamDefinitionChange || changeVisibility) {
+    queueMicrotask(() => {
[...]

PR #2013, opened 2025-09-17
Move escape key handling to modal component

victorlin:

This co-locates handling of the escape key within the component that it is used for, where clearSelectedNode is already used as an onClick callback to close the modal.

## Related issue(s)

Thought of this while investigating #2011. While it doesn't fix the bug, it helps answer the first question that came to mind: "how different are the esc and click handling code paths?"

## Checklist

- [x] Checks pass

### DIFF
--- a/src/components/tree/infoPanels/click.js
+++ b/src/components/tree/infoPanels/click.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from "react";
+import React, { useEffect } from "react";
 import { isValueValid, strainSymbol } from "../../../util/globals";
 import { infoPanelStyles } from "../../../globalStyles";
 import { numericToCalendar } from "../../../util/dateHelpers";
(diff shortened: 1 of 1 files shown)


src/components/tree/infoPanels/click.js
```diff
@@ -248,6 +248,18 @@ const Trait = ({node, trait, colorings, isTerminal}) => {
...
+      if (event.key==="Escape" && selectedNode) {
+        clearSelectedNode(selectedNode);
+      }
+    }
+    document.addEventListener('keyup', handleEscapeKey);
+    return function cleanup() {
+      document.removeEventListener('keyup', handleEscapeKey);
+    };
+  }, [selectedNode, clearSelectedNode]);
+
```
jameshadfield: One salient difference is that this will add/remove listeners every time a modal appears/disappears as opposed to a single listener set up when the tree panel mounts. In reality this won't matter because opening a modal is a one-off action, but it's something to be aware of when designing listeners.

An improvement would be to return early to avoid adding an unnecessary listener when we _close_ the modal:
```js
  useEffect(() => {
    if (!selectedNode) return
    ...
```

jameshadfield (approved):

jameshadfield:
P.S. "The Effect Hook is a [direct replacement](https://legacy.reactjs.org/docs/hooks-effect.html#effects-with-cleanup) for componentDidMount and componentWillUnmount" just isn't true. Perhaps `useEffect` with an explicit & empty dependency array is the same, although I haven't checked the exact timings of when code runs relative to DOM mutations.

src/components/tree/infoPanels/click.js
victorlin: Thanks for noting the difference. There is a way to retain the existing behavior using a ref, which I've incorporated into the main commit (c136e2588239fe54e6e82e5043c584b852ba9480). I added another commit to switch to the plain `useEffect` pattern with your proposed improvement above (eae3fbad08f44e60f85639fa94f81a1a6ef73a45), which we can drop if you think the existing behavior is better.

jameshadfield: I need to dive back into the world of hooks one day, although I wasn't very enamored with them. Perhaps [the rest of the group feel differently!](https://github.com/nextstrain/auspice/pull/2018#discussion_r2474387773) Is adding an additional `useRef` & `useEffect` the recommended way? I find it somewhat convoluted to read, although I can figure out how it works. Up to you whether this is your preferred method vs. a single but slightly more costly `useEffect` - either works for me.

Issue #2014, opened 2025-09-25
Unexpected behavior for binned continuous metadata


kimandrews:
### 1. Frequencies panel shows bins based on `"bounds"` rather than`"display"`

[Here is an example tb dataset](https://nextstrain.org/staging/tb/aws/test/n1000c) generated using bins defined for the `Number of drug resistant variants` in the auspice-config.json like this:

```
   {
    "key": "num_dr_variants",
    "title": "Number of drug resistance variants",
    "type": "continuous",
    "legend": [
      { "value": 0,    "display": "0",     "bounds": [0, 0] },
      { "value": 1.5,  "display": "1-2",   "bounds": [0.5, 2.5] },
      { "value": 3.5,  "display": "3-4",   "bounds": [2.5, 4.5] },
      { "value": 5.5,  "display": "5-6",   "bounds": [4.5, 6.5] },
      { "value": 7.5,  "display": "7-8",   "bounds": [6.5, 8.5] },
      { "value": 9.5,  "display": "9-10",  "bounds": [8.5, 10.5] },
      { "value": 11.5, "display": "11+",   "bounds": [10.5, 100] }
    ]
  },
```

In the example dataset, the auspice frequencies panel shows the upper and lower values of `"bounds"` for each bin, which means that samples with 1-2 mutations are shown as `0.50-2.50`, whereas it would be preferable to show the `"display"` value of `1-2`. Changing the `"bounds"` to `[1, 2]` does not solve the problem because samples with 1 mutation do not fit into those bounds.

<img width="1179" height="767" alt="Image" src="https://github.com/user-attachments/assets/3e1b39b8-ae32-4683-b852-462be83e3cd5" />

### 2. Coloring behavior seems to vary depending on whether samples are present within each bin

Here are 3 example datasets that differ in the number of samples, and whether the auspice_config.json has a `"scale"` section for the `Percent reads mapped` (note that all datasets have `Number of drug resistant variants` bins as described above, with no `"scale"` section):
* 8 samples, with `"scale"` section: https://nextstrain.org/staging/tb/n7/test1
* 8 samples, without `"scale"` section: https://nextstrain.org/staging/tb/n7/test2
* ~1000 samples, with `"scale"` section: https://nextstrain.org/staging/tb/aws/test/n1000c

Here is the `'Percent reads mapped'` block WITH the `"scale"` section in the auspice_config.json:
```
{
    "key": "pct_reads_mapped",
    "title": "Percent reads mapped",
[...]

PR #2015, opened 2025-09-29
Improve legend & frequencies display of continous color scales

jameshadfield:
A number of small (but nice) improvements to how we display information associated with a legend entry (bin) for continuous scales. See commit messages for full details as well as screenshots in this PR. Motivated by bug no. 1 in #2014 which this fixes.

**on-hover legend displays the bounds for continuous scales, including date-formatting**
image

**frequencies panel uses JSON-defined "display", where available**
image

### DIFF
--- a/src/util/colorScale.ts
+++ b/src/util/colorScale.ts
@@ -608,7 +608,10 @@ function parseUserProvidedLegendData(
  */
 export function formatBounds(bounds: [number, number], temporal: boolean):string {
   if (temporal) {
-    return `(${numericToCalendar(bounds[0])}, ${numericToCalendar(bounds[0])}]`;
+    // lower/uppermost bounds are often infinity, which doesn't go to a calendar date nicely!
+    const lower = bounds[0] === -Infinity ? '-∞' : numericToCalendar(bounds[0]);
+    const upper = bounds[1] === Infinity ? '∞' : numericToCalendar(bounds[1]);
+    return `(${lower}, ${upper}]`;
   } 
   return `(${bounds[0].toFixed(2)}, ${bounds[1].toFixed(2)}]`;
 }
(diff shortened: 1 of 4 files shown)


src/util/colorScale.ts
```diff
@@ -583,9 +579,36 @@ function parseUserProvidedLegendData(
       });
     if (boundArrays.length===legendValues.length) {
       legendValues.forEach((v, i) => {legendBounds[v]=boundArrays[i];});
+      userProvidedLegendBounds = true;
     } else {
       legendBounds = createLegendBounds(legendValues);
+      userProvidedLegendBounds = false;
```
joverlee521, line 585: Noticed this in an expected boundary display in a [testing dataset](https://auspice-james-display-n-mpd108.herokuapp.com/flu/seasonal/h1n1pdm/ha/2y?c=num_date) that does not define bounds

Screenshot 2025-09-29 at 1 37 51 PM

I'm not sure this is catching _all_ cases where the user does not provide legend bounds? At beginning of the function, it immediately returns `false` if the provided legend is not an array

https://github.com/nextstrain/auspice/blob/26d2f6091376f44775f4a0976cb923ad4a5e1ff2/src/util/colorScale.ts#L547

jameshadfield: Thanks for the catch! It wasn't the early return it was [a typo elsewhere](https://github.com/nextstrain/auspice/compare/26d2f6091376f44775f4a0976cb923ad4a5e1ff2..3da26eb7221287462dcf7d764a488318e67d68a3). Now looks like this:

image

joverlee521: Ah, nice! I'm was confused by this bit in the commit message

> I considered displaying the bounds for _all_ continuous scales but when
the bounds are chosen by Auspice it's less ideal. (Still, maybe we
should show them in a pop-up or similar.)

Now I'm seeing this is referring to the legend _label_ and not the hover text 🤦‍♀️

Issue #1766, opened 2024-04-11
Entropy panel: shift click to select multiple positions


joverlee521:
This was briefly brought up in today's office hours by @huddlej as a potential alternative to using the color-by input to select multiple positions via the entropy panel.

jameshadfield:
I remember thinking this _was_ the behaviour and being surprised it wasn't! We should do this.

With the redesigned entropy panel you can only ever view AAs for one CDS at a time, which matches our genotype color-by restriction nicely. But there's still a few out-of-sync situations to catch.

PR #2018, opened 2025-10-28
[info panel] Clickable mutations

jameshadfield:
Mutations in a node-info modal (via tip click or shift+branch click) are now clickable. Clicking on a mutation will colour the tree by the genotypes at that position. Shift+clicking will append that position to the current genotype colouring (if possible), or remove it if it's already part of it. Command clicking (with or without shift) will behave similarly but filter the tree via the specific mutated state instead of changing the colouring.

Closes #2017 

This _should_ make it straightforward to shift+click on bars in the entropy panel and add them to the current colouring, but that's left for a future PR. _UPDATE: This was super easy - it's #2019_

- [x] Checks pass
- [x] If making user-facing changes, add a message in [CHANGELOG.md](https://github.com/nextstrain/auspice/blob/HEAD/CHANGELOG.md) summarizing the changes in this PR
- [ ] (to be done by a Nextstrain team member) [Create preview PRs on downstream repositories][1].

[1]: https://github.com/nextstrain/auspice/blob/-/DEV_DOCS.md#test-on-downstream-repositories

PR #2019, opened 2025-10-29
Shift click entropy bars

jameshadfield:
This builds on the ability to shift-click additional mutations in a tree's info modal <https://github.com/nextstrain/auspice/pull/2018> and have them added/removed from the current color-by, as applicable.

Closes https://github.com/nextstrain/auspice/issues/1766


src/components/entropy/index.js
joverlee521: Small request to document the shift+click ability, maybe adding to the tooltip for each bar

https://github.com/nextstrain/auspice/blob/22124f8eacec2c46f39806aa95472581f65fc25a/src/components/entropy/entropyD3.js#L1095-L1097

jameshadfield: Done - thanks!

Issue #2033, opened 2026-01-19
[dev] allow auspice client - nextstrain.org dev server comms


jameshadfield:
Testing Auspice client changes against a locally running nextstrain.org server is very useful.

We have a little code [here](https://github.com/nextstrain/auspice/blob/master/src/components/narrativeEditor/updateApiCalls.js) which can be configured to do this in a certain context (narratives in the narrative editor); and it works by running Auspice (server/client) and the nextstrain.org dev server on different (localhost) ports.

We should generalise it. There are two main configurations I want to reach for, including:

1. All charon requests handled by nextstrain.org
2. Narratives sourced from Auspice, but their datasets from nextstrain.org

Modifying the addresses below using the code above gets us ~half of the way to (1) and maybe all the way to (2)?

https://github.com/nextstrain/auspice/blob/6370cc5a682824b607dcd0314c1821e99bd636f7/src/actions/loadData.js#L266-L270

P.S. CORS isn't an issue because the nextstrain.org server [allows cross-origin](https://github.com/nextstrain/nextstrain.org/blob/655c6c5206725440ccbe475903d9983aa7e98729/src/middleware.js#L19-L28) for non-credentialed GET/HEAD requests.

P.P.S. Another approach would be to use the Auspice server as a reverse proxy and have it do the fetch to the nextstrain.org server. This is probably nicer as it keeps complexity out of the frontend. 🤔

PR #2035, opened 2026-01-20
[dev] reverse proxy

jameshadfield:
Running with `auspice view --handlers cli/dev/reverse-proxy.js` will proxy all requests to localhost:5000 by default, allowing you to run a local nextstrain.org server and test against a new Auspice client. Specifying the env variable `PROXY="https://nextstrain.org"` easily allows you to use the live nextstrain.org host as the charon API, allowing easy testing of nextstrain.org functionality such as groups, community etc.

This essentially closes
https://github.com/nextstrain/auspice/issues/2033. There's one (niche!) use-case where you may want to source narratives locally (i.e. default Auspice server behaviour) but their datasets through a proxy. If I ever need to reach for that particular behaviour I'll revisit this.


victorlin:
Tested locally, works great!

Small low-priority suggestion, feel free to take or leave: ce0e70b38ec07174ef28fb258d072253478b3410

PR #2039, opened 2026-03-16
update how new metadata is added

### LINKED ISSUE #2031: drag-and-drop metadata: Support coloring merging
Allow drag-and-drop metadata to merge with existing metadata for fields that are used as coloring

https://github.com/nextstrain/auspice/blob/6370cc5a682824b607dcd0314c1821e99bd636f7/src/actions/filesDropped/metadata.js#L228-L240

jameshadfield:
~The code needs polish, but it's working well enough~ Code is now ready for review. The new functionality closes CoE 4.4 / Issue #2031. It also adds the ability to drag-and-drop node data JSONs.


src/actions/filesDropped/constants.js
```diff
@@ -14,8 +14,11 @@ const isWindowsTsv = (file) => file.type === "" && file.name.endsWith('.tsv');
 // Handle Excel exported .csv files
 const isExcelCsv = (file) => file.type === "application/vnd.ms-excel" && file.name.endsWith('.csv');
 
-const isAcceptedFileType = (file) => acceptedFileTypes.includes(file.type) || isWindowsTsv(file) || isExcelCsv(file);
```
jameshadfield, line 17: Auspice.us uses
```js
import { isAcceptedFileType as isAuspiceAcceptedFileType } from "@auspice/actions/filesDropped/constants";
```

So we should keep this around, maybe as:

```js
function isAcceptedFileType(file) {
  return fileTypeIsCSVLike(file) || fileTypeIsJson(file)
}
```

src/actions/updateMetadata/updateMetadata.ts
```diff
@@ -0,0 +1,108 @@
...
+import { UPDATE_METADATA } from "../types";
+import { NewMetadata } from "./updateMetadata.types";
+import { changeColorBy } from "../colors";
+/**
+ * A redux thunk action to update node attributes and related data. The newMetadata
+ * has not been cross-referenced with Redux state and so this thunk does that work,
+ * dispatching notifications as needed. The resulting dispatched action contains
+ * validated data which the reducers can simply merge into state.
+ */
+export const updateMetadata = (newMetadata: NewMetadata) => {
```
joverlee521, line 12: Is the thought that `updateMetadata` will be used for the editable dataset feature as well? I'd think they would be separate since this is focused on updating sample metadata in the tree whereas the editable dataset feature is updating the dataset metadata. 

Maybe I'm splitting hairs, but I could see this dispatching one action to update the tree node_attrs, then a second action to update the dataset metadata (i.e. color and geographic). This second action could also be used for the editable dataset feature.

jameshadfield: My current plan is to allow extra properties in the action structure to facilitate new metadata. While actions can be split into multiple actions I've encountered race-condition bugs in them before and thus I try to put things into a single action where possible. Are there advantages to splitting them up in your eyes?

joverlee521: The main advantage is to keep things simple with reusable actions that don't have too many conditionals, e.g. when to expect n>0 strains in the payload. 

If we split this into separate actions: 

1. Update tree metadata. This must _always_ include strains to merge new metadata with existing metadata and update the tree nodes.
2. Update dataset metadata. This should _not_ include strains so changes are verified against the existing tree in redux store but does not update the tree nodes. 

I'd think it'd be safe to dispatch [1], get the updated state, then dispatch [2] in the same thunk without any race conditions. Then the editable dataset feature can use [2] without worrying about how to work around/avoid tree updates.

jameshadfield: This strikes me as a question of where to put the complexity, not a way to get rid of it. With multiple actions the complexity gets sprinkled around the codebase, with each dispatch point having to think about this and understand the redux code. With a single action the thunk gets more complex (and the reducers a little bit too), but the call-points (UI elements) become very easy to reason with. This is a general pattern I've liked, and have tried to implement in `applyFilter`, `changePage` etc. Where we haven't done it, and multiple actions need to be fired to achieve the desired change, it feels brittle.

joverlee521: Just want to clarify, I'm not suggesting the call-points to dispatch multiple actions and the complexity would still be hidden in the thunks and reducers. But point taken that these are just different ways of organizing the complexity. 

I'm mainly confused by how the strains/node_attrs are included in the colorings. I can see using this for editable datasets if there's shift in the payload type, so it's easy to see if there are no `strains` defined:

```diff
diff --git a/src/actions/updateMetadata/updateMetadata.types.ts b/src/actions/updateMetadata/updateMetadata.types.ts
index f5595cdf7c..905fd8bb73 100644
--- a/src/actions/updateMetadata/updateMetadata.types.ts
+++ b/src/actions/updateMetadata/updateMetadata.types.ts
@@ -6,7 +6,13 @@
  * covering use cases from drag-and-drop metadata TSVs to in-app editing of attr colours.
  */
 export interface NewMetadata {
-  attributes?: Record<string, ColoringInfo>;
+  strains?:{
+    [strain: string ]: {
+      [node_attr: string]: NodeAttr
+    }
+  }
+
+  colorings?: ColoringInfo[];

   geographic?: {
     /** trait name */
```

If we're using `updateMetadata` as the central thunk, then `filterStrains`, `filterAttrs`, and the `successNotification` feel like they should be moved out (probably to `handleFilesDropped`?).

jameshadfield: I worked on this a little more today - you may be interested in [a4f52e4](https://github.com/nextstrain/auspice/pull/2039/commits/a4f52e4bcf2c7efe1c622999407a847125f4e9aa) which hopefully makes what I'm going for a little clearer. Most of the complexity is being shifted into the `updateMetadata` thunk, with the reducers themselves as simple as they can be. I'm still playing around with ideas so the code is messy!

src/actions/updateMetadata/updateMetadata.ts
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  if (replace) {
+    attrDetails.colors
+    return {
+      title: attrDetails.name,
+      type: attrDetails.scaleType,
+      ...(attrDetails.colors?.length && { scale: attrDetails.colors }),
+    }
+  }
```
joverlee521, line 208: Confirming my understanding here for the downstream editable datasets. If a user wants to edit a single color of a continuous coloring (e.g. dates), this should replace the whole coloring object?

Concretely, the `newMetadata` look like 

```ts
const newMetadata: NewMetadata = {
  attributes: {
    [coloring.key]: {
      key: coloring.key,
      name: coloring.title,
      scaleType: "categorical", // previously "continuous"
      strains: [],
      colors: updatedColors,
    }
  }
}

```

src/actions/filesDropped/constants.js
```diff
@@ -14,8 +14,17 @@ const isWindowsTsv = (file) => file.type === "" && file.name.endsWith('.tsv');
...
+const fileTypeIsCSVLike = (file) => acceptedFileTypes.includes(file.type) || isWindowsTsv(file) || isExcelCsv(file);
+
+const fileTypeIsJson = (file) => file.type === "application/json" || file.name.endsWith('.json');
+
+/** Used by auspice.us */
+function isAcceptedFileType(file) {
+  return fileTypeIsCSVLike(file) || fileTypeIsJson(file);
```
joverlee521, line 23: Not wanting to block the work here, just wondering how we would differentiate Auspice JSONs vs node data JSONs in auspice.us?

src/actions/updateMetadata/updateMetadata.ts
joverlee521: Thanks for the latest work here, added [specific question](https://github.com/nextstrain/auspice/pull/2039#discussion_r3228886489) about how coloring merging works.

src/actions/filesDropped/constants.js
jameshadfield: We won't be able to, so you can't drag on dataset + node-data JSONs at the same time, but you should be able to drag on and load a dataset, and then drag on a node-data JSON.

(I need to check this expected behaviour is actually the behaviour!)

(And obviously we could add heuristics to check what type a JSON is, but I don't think it's worth it)

src/actions/updateMetadata/updateMetadata.ts
joverlee521: Ah, just tried this locally and it's not as straightforward because the tree doesn't have `totalStateCounts` for continuous colorings so converting it to categorical is not ideal. 

Maybe just loosen to 

```ts
if (!replace && state.type !== attrDetails.scaleType) {
  console.warn(`Merging scale colors is only possible same scale type`)
  replace = true;
}
```

jameshadfield: Up front: I fully expect some code tweaks etc to be needed as we expand the functionality with your editable-UI work. How we'll actually change a continuous scale also seems up for grabs -- do we modify a single stop, like I think you're implying, or use a UI more like 👇 which will (for action purposes) replace all the stops, or something else entirely?

image

Two related thoughts I had while thinking through this:

1. Should we leave continuous scale modifications out of the v1 editable UI? 

2. We could (I think!) enable an auspice-config JSON download, which would let people use GUIs to choose configuration options and then have them stamped across all future builds...

joverlee521: > Up front: I fully expect some code tweaks etc to be needed as we expand the functionality with your editable-UI work.

Yes, totally get that, just wanted to talk through expectations before I make the changes. 

> do we modify a single stop, like I think you're implying,

Yes, this is the v1 of the editable UI. I think we can add the feature in the future to make the continuous scale editing more user friendly. 

For now, I am able to get the single stop editing working locally with some minor tweaks on your changes here. I'll clean it up and actually rebase https://github.com/nextstrain/auspice/pull/2046 on top of this PR to hook it all up tomorrow.

joverlee521: See https://github.com/nextstrain/auspice/pull/2046#discussion_r3236732426

src/actions/filesDropped/constants.js
jameshadfield: I tested this and it was messy (auspice.us would sometimes attempt to load the node-data JSON as the main dataset file). It turned out [to be trivial](https://github.com/nextstrain/auspice.us/pull/148) to check if a JSON has top-level `meta` & `tree` keys, which then allows us to drag on node-data files onto auspice.us' splash page and have things Just Work.

src/actions/updateMetadata/updateMetadata.types.ts
```diff
@@ -0,0 +1,79 @@
...
+  colorings?: {
+    [coloringKey: string]: ColoringInfo;
+  };
+  geoResolutions?: {
+    demes: {
+      [demeName: string]: LatLong;
+    };
+    key: string;
+  }[];
+}
```
joverlee521, line 43: _non-blocking_

(Asking to know where I should add more types for editing datasets)

How does this fit with the existing [Metadata type](https://github.com/nextstrain/auspice/blob/43006efa77f49c4beb6c5c849f8d71270c3e9aae/src/metadata.ts#L3-L10)? Should `ActionMetadata` just be `Partial<Metadata>`?

jameshadfield: Hmmm. Yes, maybe. Perhaps not `Partial<>`, but cherry-picking the `Metadata` properties the action supports would be preferable. 

Aside: The metadata state is not typed, and this has led to a range of half-complete types around the codebase. I like the approach of https://github.com/nextstrain/auspice/commit/448e97da1876261d57f118bffb6a1228a111bbf0 where the JSON type is `any` which we parse into a concrete type for `redux.metadata.sharing` which downstream components can easily reason with.

joverlee521: > Aside: The metadata state is not typed, and this has led to a range of half-complete types around the codebase.

Hah, I might be adding to the problem with [993e013](https://github.com/nextstrain/auspice/pull/2046/commits/993e0134ba98f73a37be960dc4eb69641de0128a)

jameshadfield: I wrote this up in https://github.com/nextstrain/auspice/issues/2053 rather than fixing it here!

PR #2046, opened 2026-05-08
Editable dataset

### LINKED ISSUE #140: Support editing of datasets
See [design GDoc](https://docs.google.com/document/d/1C51A6b2F94QQkxevCH5kzgTXFDoUNPyWtz7cnKfjx1c/edit?tab=t.0)

Allow users to edit datasets 
- metadata fields (title, authors, description etc)
- color palettes for variables
- legend position - blocked by https://github.com/nextstrain/auspice/issues/2042

Cannot save edits until https://github.com/nextstrain/auspice/issues/2000 is implemented.

### LINKED ISSUE #2042: Support legend placement customization
Prompted by https://github.com/nextstrain/auspice.us/issues/140

In order to allow users to edit legend placement, Auspice first has to support legend placement customization via the dataset JSON. The legend placement is currently [hardcoded in the legend component](https://github.com/nextstrain/auspice/blob/3f9d1f8b0bd59097b61a6c8389738ca827cbfcdf/src/components/tree/legend/legend.js#L191-L201), with minor variation to display on the [left side in the tree panel](https://github.com/nextstrain/auspice/blob/3f9d1f8b0bd59097b61a6c8389738ca827cbfcdf/src/components/tree/tree.tsx#L154) and on the [right side in the map panel](https://github.com/nextstrain/auspice/blob/3f9d1f8b0bd59097b61a6c8389738ca827cbfcdf/src/components/map/map.js#L623). 

This will require changes in augur to update the [Auspice JSON schema](https://github.com/nextstrain/augur/blob/ce66ccf076adc6f28d649f433b2ed2efd73c53e1/augur/data/schema-export-v2.json) and in auspice to parse and use the new legend placement field(s).

joverlee521:

Add modals for editing dataset info and for updating colors.
1. Click on "Edit" button near title to open dataset editing modal
Screenshot 2026-05-08 at 2 47 33 PM

2. Shift-click on legend swatch to open color editing modal
Screenshot 2026-05-08 at 2 22 47 PM

3. Hit submit and see updated info and coloring. Both should also trigger a warning notification.
Screenshot 2026-05-08 at 2 24 27 PM

## Related issue(s)

Resolves https://github.com/nextstrain/auspice/issues/2042
Resolves https://github.com/nextstrain/auspice.us/issues/140
Resolves https://github.com/nextstrain/private/issues/200

## Checklist

- [x] Checks pass
- [x] Add docs for editing datasets


src/components/datasetEditor/colorByEditor.tsx
joverlee521, line 1: Rebased on https://github.com/nextstrain/auspice/pull/2039 to start propagating edits through to Redux state. 

As of a4b9af311b2acab0dc9efdd19d207d364b17be2d, color by editing should work for categorical and continuous coloring. I just realized that Genotype color editing does not work as expected...will look into it.

jameshadfield: > Genotype color editing does not work as expected

Urgh. Genotypes are an oddity in Auspice as their values aren't stored in `node_attrs`, and their colorings data is similarly special-cased.

joverlee521: Hrrrm. Maybe we can talk through this in dev chat.

joverlee521:
All editing should propagate through Redux state as of [a92e46f](https://github.com/nextstrain/auspice/pull/2046/commits/a92e46f19f3596164ab7e7172c174b70305502b9) (with the exception of [updating coloring of Genotype](https://github.com/nextstrain/auspice/pull/2046#discussion_r3236732426)).

src/components/datasetEditor/colorByEditor.tsx
joverlee521: Reference for previous work on keeping aa/nuc colors consistent: https://github.com/nextstrain/auspice/pull/739.

I'm going to punt on editing Genotype coloring for this particular PR, but we can revisit in the future.

joverlee521: Disabled genotype color edits in a97b6848c83c56ffd6fd90f42202123d67273adb.

joverlee521:
Rebased to build on top of the latest Redux types and the download function. This should now be completely hooked up to edit -> update Redux state -> download JSON with updates.

joverlee521:
Did some brief testing in the [auspice.us test app](https://auspice-us-nextstrain-b-yecndc.herokuapp.com/) to confirm this works for 
- [x] dataset JSONs
- [x] dataset JSONs + additional drag-and-drop metadata
- [x] nwk 
- [x] nwk + additioanl drag-and-drop metadata

src/actions/updateMetadata/updateMetadata.ts
```diff
@@ -18,18 +18,23 @@ export const SUCCESS = "SUCCESS";
 export const updateMetadata = (
   newMetadata: NewMetadata,
   /** Replace redux state where possible, rather than merge */
-  replace = false
+  replace = false,
+  /** Expect tree updates, only returns SUCCESS when tree is updated */
+  treeUpdates = true,
```
jameshadfield, line 23: > This required minor updates to the `updateMetadata` thunk to skip tree
updates for this feature. All fields are just passed through without any
transformations. 

This stood out to me. All things being equal, I'd prefer to reduce the needed arguments, or put another way, reduce the cognitive overhead for code paths which are calling `updateMetadata`.

Could we instead make `_reduxTree` track if anything was actually changed, and conditionally inject the `tree` into the dispatched payload accordingly? Or just dispatch it anyway and pay the potential cost of an extra re-render?

I think this runs into problems because of the early `return "No matching nodes in tree!";`? If so, I can revisit why I did that and see if we can drop it or make it more intelligent.

src/components/datasetEditor/datasetEditor.tsx
```diff
@@ -0,0 +1,252 @@
...
+      maintainers: convertToNameAndUrl(maintainers),
+      dataProvenance: convertToNameAndUrl(dataProvenances),
+    }, true, false));
+
+    if (result===SUCCESS) {
+      dispatch(warningNotification({
+        message: t("This dataset has been edited!"),
+        details: t("Edited dataset warning detail"),
+        autoClose: false,
+      }));
```
jameshadfield: If I make back-to-back[-to-back] edits I get a stack of warnings. How hard would it be to limit it to one?

Image

src/components/datasetEditor/datasetEditor.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+
+  return (
+    <DatasetEditorForm onSubmit={handleSubmit}>
+      <div className="row">
+        <div className="col-sm-6">
+          <div className="row">
+            <div className="col-lg-6">
+              <label>
+                {t("Dataset title")}:<br/>
```
jameshadfield, line 90: [minor] some of these headings (Maintainers, Data provenances, Legend placement) are underlined, and others are not. Is there a reason for this or can we make them all underlined or not?

src/components/datasetEditor/colorByEditor.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { LegendValues } from "../../reducers/controls";
+import Flex from "../framework/flex";
+import { ColorScale } from "../../reducers/controls";
+import { updateMetadata, SUCCESS } from "../../actions/updateMetadata/updateMetadata";
+import { NewMetadata } from "../../actions/updateMetadata/updateMetadata.types";
+import { warningNotification } from "../../actions/notifications";
+import { getBrighterColor, getDarkerColorHex } from "../../util/colorHelpers";
+import { ColorByEditorForm } from "./styles";
+
+export default function ColorByEditor({
```
jameshadfield, line 15: [not sure where exactly this comes from]

While editing colour scales the dev console gets spammed with warnings like 👇 which would be nice to fix

```
The specified value "rgb(81, 131, 255)" does not conform to the required format.  The format is "#rrggbb" where rr, gg, bb are two-digit hexadecimal numbers.
```

src/components/tree/legend/legend.js
```diff
@@ -149,6 +150,8 @@ class Legend extends React.Component {
   legendItems() {
     const values = this.props.colorScale.visibleLegendValues;
     const maxNumPerColumn = Math.ceil(values.length/2); // hardcoded to 2 columns
+    // We do not support editing Genotype colors because we do not keep nuc/aa colors in Redux state.
+    const supportColorEdit = !isColorByGenotype(this.props.colorBy);
```
jameshadfield: It'd be nice when shift-clicking on a genotype color swatch to dispatch a notification saying something like "genotype color editing is not supported at the moment"

src/components/datasetEditor/colorByEditor.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const maxNumPerColumn = Math.ceil(legendValues.length/2);
+  const values = legendValues.slice(0, maxNumPerColumn);
+  const values2 = legendValues.slice(maxNumPerColumn);
+
+  /**
+   * The color input value is set to the internally calculated brighter
+   * color to match the legend swatch. The border color is set to the
+   * darker color and saved as a data attribute so that it can be saved as
+   * the main color in the Redux state.
+   */
```
jameshadfield, line 116: I regret doing this, but there's no going back now!

src/components/tree/legend/item.js
```diff
@@ -22,6 +24,11 @@ const LegendItem = ({
     onMouseLeave={() => {
       dispatch(updateTipRadii());
     }}
+    onClick={(e) => {
+      if (supportColorEdit && e.shiftKey) {
+        dispatch({ type: SET_MODAL, modal: "colorByEditor" });
+      }
+    }}
```
jameshadfield: When I shift-click on a specific swatch I get the color-by-editor model and then have to re-click the swatch I want to edit. It'd be slick to have that swatch's color-picker thing open up automagically, but I think this would be hard to wire up and so I'm pretty happy leaving this out. Did you look into this?

src/components/datasetEditor/datasetEditor.tsx
```diff
@@ -0,0 +1,252 @@
...
+}
+
+function convertToNameAndUrl(state: NameAndUrlState[]): NameAndUrl[] {
+  return state.map((item): NameAndUrl => {
+    delete item.id;
+    return item;
+  })
+}
+
+export default function DatasetEditor({
```
jameshadfield, line 30: General comment about this modal and the color-picker one:

We (hardcode?) the behaviour where clicking outside the modal dismisses it, but when we're making edits accidentally clicking outside the modal gets rid of our proposed edits. Frustrating! It would be nicer to stop this behaviour and have two buttons at the bottom of the form: "apply changes" & "discard changes" (with the enter & escape keys bound to these as well).

src/components/datasetEditor/datasetEditor.tsx
```diff
@@ -0,0 +1,252 @@
...
+            {t("Description")}:<br/>
+            <textarea
+              rows={10}
+              name="description"
+              defaultValue={metadata.description || ""}/>
+          </label>
+        </div>
+      </div>
+      <Flex justifyContent="center">
+        <button type="submit">{t("Submit")}</button>
```
jameshadfield: Thoughts on "apply changes" instead? Submit makes me think somethings going across the wire

src/components/datasetEditor/datasetEditor.tsx
```diff
@@ -134,6 +141,9 @@ export default function DatasetEditor({
               name="description"
               defaultValue={metadata.description || ""}/>
           </label>
+          <LegendPlacementInputs
```
jameshadfield, line 152: Did you consider having the legend positioning interface as part of the color-changer modal? I haven't prototyped this but wonder if it'd be nicer. I am imagining a future where the capabilities of the "color changer" modal are greater -- editing the title, changing the color type, etc. 

It would also remove a (very slight!) oddity: currently the dataset editor shows has the legend placement selector for the map even when the map is toggled off or when the tree/map are shown side-by-side and thus there's no legend shown on the map.

src/components/datasetEditor/editButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+  height: 1em;
+`;
+
+export default function EditButton(): JSX.Element {
+  const { t } = useTranslation();
+  const dispatch = useDispatch();
+  const title="EditButton";
+  return (
+    <>
+      <Button data-tip data-for={title}
```
jameshadfield, line 26: [Big UI comment 1/2]

I'm not sold on the edit button as currently designed. I don't want dataset editing to be too hidden, but at the moment it's (IMO!) too prominent. Pinning it to the RHS looks better to me:

Image

src/components/datasetEditor/datasetEditor.tsx
```diff
@@ -0,0 +1,252 @@
...
+      buildUrl: e.currentTarget.buildUrl.value,
+      buildAvatar: e.currentTarget.avatarUrl.value.trim() || undefined,
+      warning: e.currentTarget.warning.value.trim() || undefined,
+      description: e.currentTarget.description.value,
+      maintainers: convertToNameAndUrl(maintainers),
+      dataProvenance: convertToNameAndUrl(dataProvenances),
+    }, true, false));
+
+    if (result===SUCCESS) {
+      dispatch(warningNotification({
```
jameshadfield: [Big UI comment 2/2]

Did you consider a different UI to convey this? I like the consistency (of reusing the notifications) but I'm tempted to use something bolder, e.g

Image

jameshadfield (commented):
Very cool! The behaviour is great. Mostly minor (optional!) suggestions / requests, with the UI suggestions being the bigger ones which we should perhaps discuss a little before changing anything.

src/components/datasetEditor/datasetEditor.tsx
```diff
@@ -0,0 +1,346 @@
...
+}
+
+function convertToNameAndUrl(state: NameAndUrlState[]): NameAndUrl[] {
+  return state.map((item): NameAndUrl => {
+    delete item.id;
+    return item;
+  })
+}
+
+export default function DatasetEditor({
```
jameshadfield, line 30: Oh - Also - I think we should drop editing ability from mobile

src/actions/updateMetadata/updateMetadata.ts
joverlee521: Yes, this was definitely added because of the early return of "No matching nodes". I had considered passing in the existing tree, but wasn't sure if that would cause any unexpected changes/renderings.

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Ah, yeah, I can see that being annoying...Will look into limiting to one.

src/components/datasetEditor/datasetEditor.tsx
joverlee521: I underlined the fields that have multiple inputs as "headers". I can remove them if this is unclear.

src/components/tree/legend/item.js
joverlee521: I thought about this briefly. It would require dispatching `SET_MODAL` with an additional parameter and adding the selected swatch as a central Redux state...I wasn't sure it would be worth it.

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Yeah, I'm also getting frustrated by accidentally clicking outside. I'll toggle off the behavior and add the separate buttons.

src/components/datasetEditor/editButton.tsx
joverlee521: I'm not tied to the current positioning, RHS looks good to me!

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Hmm, yeah. Actually this would probably be easier to do than trying to [limit to one warning notification](https://github.com/nextstrain/auspice/pull/2046#discussion_r3432911033). I'll go this route.

src/components/tree/legend/item.js
jameshadfield: Not worth it!

src/components/datasetEditor/datasetEditor.tsx
joverlee521: This is no longer relevant since we are changing to the [warning banner](https://github.com/nextstrain/auspice/pull/2046#discussion_r3433096833)

src/components/datasetEditor/colorByEditor.tsx
joverlee521: Thanks for flagging, resolved with [fbcd560](https://github.com/nextstrain/auspice/pull/2046/commits/fbcd560f507223610be78cea30713530616393cd)

src/components/tree/legend/legend.js
joverlee521: Added in [639e848](https://github.com/nextstrain/auspice/pull/2046/commits/639e848ff895350a2a9d00bb6da93c66b99727fc)

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Updated behavior and added separate buttons in [0a668d9](https://github.com/nextstrain/auspice/pull/2046/commits/0a668d9cca198a077c3805a7a6df10911118d943). 

I did not venture into the key bindings yet since I wasn't sure if these should be added to the [existing bindings in the modal](https://github.com/nextstrain/auspice/blob/3bbf410bb90af9df80c35149bf11b7405742b5d1/src/components/modal/Modal.jsx#L22-L34).

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Updated in [0a668d9](https://github.com/nextstrain/auspice/pull/2046/commits/0a668d9cca198a077c3805a7a6df10911118d943).

src/components/datasetEditor/editButton.tsx
joverlee521: Done in [1ec2ee1](https://github.com/nextstrain/auspice/pull/2046/commits/1ec2ee12716f9bb47aa920b1ad75a61462b6d4bf).

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Since we are prompting download here, this made me wonder how editing should interact with the `metadata.sharing.dataset_json`. If the original author does not allow sharing, then this should probably not allow downloading either (otherwise this would be a pretty bad loophole). 

If user can't download the dataset, should they still be allowed to edit it?

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Updated in [14420a4](https://github.com/nextstrain/auspice/pull/2046/commits/14420a4ccd75a252fce30fc4bcc49e5872b2bb2b)

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Switched from notification to permanent banner in [7754f8c](https://github.com/nextstrain/auspice/pull/2046/commits/7754f8c4522a510b659e395c363decaa2aaeef86).

src/components/datasetEditor/datasetEditor.tsx
jameshadfield: I didn't get the association with multiple inputs, but it's ok as is I think

src/components/datasetEditor/datasetEditor.tsx
jameshadfield: > I did not venture into the key bindings yet since I wasn't sure if these should be added to the [existing bindings in the modal](https://github.com/nextstrain/auspice/blob/3bbf410bb90af9df80c35149bf11b7405742b5d1/src/components/modal/Modal.jsx#L22-L34).

I started (very briefly) centralising them in https://github.com/nextstrain/auspice/pull/1999 but that needs more work so isn't helpful here. Just adding the bindings to the `<DatasetEditor>` is what I'd do:

```diff
diff --git a/src/components/datasetEditor/datasetEditor.tsx b/src/components/datasetEditor/datasetEditor.tsx
index 996a75fbc..b373e8321 100644
--- a/src/components/datasetEditor/datasetEditor.tsx
+++ b/src/components/datasetEditor/datasetEditor.tsx
@@ -1,4 +1,5 @@
-import React, { useRef, useState, Dispatch } from "react";
+import React, { useRef, useState, useEffect, Dispatch } from "react";
+import Mousetrap from "mousetrap";
 import { useDispatch } from "react-redux"
 import { useTranslation } from "react-i18next";
 import { useSelector } from "react-redux";
@@ -38,6 +39,20 @@ export default function DatasetEditor({
   const { legendPlacements } = metadata;
   const [ maintainers, setMaintainers ] = useState<NameAndUrlState[]>(convertToNameAndUrlState(metadata.maintainers || []));
   const [ dataProvenances, setDataProvenances ] = useState<NameAndUrlState[]>(convertToNameAndUrlState(metadata.dataProvenance || []));
+  const formRef = useRef<HTMLFormElement>(null);
+
+  useEffect(() => {
+    Mousetrap.bind('enter', () => {
+      formRef.current?.requestSubmit();
+    });
+    Mousetrap.bind('escape', () => {
+      dismissModal();
+    });
+    return () => {
+      Mousetrap.unbind('enter');
+      Mousetrap.unbind('escape');
+    };
+  }, [dismissModal]);
 
   function handleSubmit(e: React.FormEvent<HTMLFormElement>): void {
     e.preventDefault();
@@ -65,7 +80,7 @@ export default function DatasetEditor({
 
 
   return (
-    <DatasetEditorForm onSubmit={handleSubmit}>
+    <DatasetEditorForm ref={formRef} onSubmit={handleSubmit}>
       <div className="row">
         <div className="col-sm-6">
           <div className="row">
```

src/actions/updateMetadata/updateMetadata.ts
jameshadfield: I think the code is fine as-is, although I'll try to improve `updateMetadata` so it's not needed. That shouldn't block this PR however.

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Thinking on this briefly, it would require propagating _which_ legend is currently being edited through the Redux state and I don't think that's worth it here. 

Maybe revisit if we ever redesign the modal to [render with a portal](https://react.dev/reference/react-dom/createPortal#rendering-a-modal-dialog-with-a-portal)?

src/components/datasetEditor/datasetEditor.tsx
joverlee521: Thanks! Added to both the dataset editor and color editor modals in [c0cd9ce](https://github.com/nextstrain/auspice/pull/2046/commits/c0cd9cec0a578b5472a1cddafc568b2789c6d004)

joverlee521:
Tested briefly in a [auspice.us review app](https://auspice-us-nextstrain-b-7d4rzf.herokuapp.com). I'm happy with where this is now and I think it'd be fine to merge and release as v2.73.0 instead of waiting for the V3 release.

joverlee521:
Final bits of testing:

1. auspice.us: Pushed [commit in test PR](https://github.com/nextstrain/auspice.us/pull/146/commits/d49fed6bc5f91c9ef194ce9312ecaa2b5f374e65) to enable the dataset editor and confirmed works as expected in [test app](https://auspice-us-nextstrain-b-12bww4.herokuapp.com). 
2. nextstrain.org: Confirmed the [test app](https://nextstrain-s-nextstrain-2qrskb.herokuapp.com) does _not_ have the editing features.

joverlee521:
I'll merge and release this tomorrow morning.

PR #2047, opened 2026-05-12
Allow JSONs to control downloading ability

jameshadfield:
Allows a new JSON.metadata.sharing object/dict to provide fine grained control over the ability to download each asset. There is no change in current behaviour for datasets which don't define the 'sharing' data structure.

This commit also refactors things a little to guarantee the redux metadata state is valid up-front (for the 'sharing' key), thus making the rendering components simpler.

[Related slack thread](https://bedfordlab.slack.com/archives/C0K3GS3J8/p1778484044141669)

Needs to be coordinated with updated docs and updated JSON schemas (in Augur)

### DIFF
--- a/src/components/download/downloadButtons.js
+++ b/src/components/download/downloadButtons.js
@@ -25,12 +25,12 @@ export const DownloadButtons = ({dispatch, t, tree, entropy, metadata, colorBy, 
   const filePrefix = getFilePrefix();
   const temporal = distanceMeasure === "num_date";
 
-  /* set gisaidProvenance based on supplied JSON metadata */
-  const gisaidProvenance = (metadata?.dataProvenance || []).filter((el) => el.name.toUpperCase()==='GISAID').length>0;
-  /* Verify that we can parse dataset name from URL to support Auspice JSON download */
+  /**
+   * Currently our ability to download JSONs relies on parsing the URL and re-fetching the
+   * JSON. Check that we can parse datasets from the URL before exposing the button here.
+   * This check will be removed as part of <https://github.com/nextstrain/auspice/issues/2000>
+   */
   const datasetNames = getDatasetNamesFromUrl(window.location.pathname);
-  const supportAuspiceJsonDownload = !gisaidProvenance && datasetNames.some(Boolean);
-
   const entropyBar = entropy?.selectedCds===nucleotide_gene ? "nucleotide" : "codon";
 
   return (
@@ -44,7 +44,7 @@ export const DownloadButtons = ({dispatch, t, tree, entropy, metadata, colorBy, 
         <p/>
         {partialData ? `Currently ${selectedTipsCount}/${totalTipCount} tips are displayed and will be downloaded.` : `Currently the entire dataset (${totalTipCount} tips) will be downloaded.`}
       </div>
-      {supportAuspiceJsonDownload && (
+      {metadata.sharing.dataset_json && datasetNames.some(Boolean) && (
         <Button
           name="Auspice (Nextstrain) JSON"
           description={`The main Auspice dataset JSON(s) for the current view`}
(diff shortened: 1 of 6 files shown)


src/metadata.ts
```diff
@@ -34,3 +36,18 @@ export type Legend = {
...
+ * applicable assets
+ */
+export interface MetadataSharing {
+  dataset_json: boolean;
+  metadata_tsv: boolean;
+  authors: boolean;
+  trees: boolean;
+  entropy: boolean;
+  gisaid_acknowledgments: boolean;
+  screenshot: boolean;
```
joverlee521, line 53: Just confirming, is the plan to add these exact keys under a top level `sharing` key in the auspice config schema in Augur?

src/metadata.ts
```diff
@@ -34,3 +36,18 @@ export type Legend = {
...
+ * assets of the dataset. The defaults are to allow downloading of all
+ * applicable assets
+ */
+export interface MetadataSharing {
+  dataset_json: boolean;
+  metadata_tsv: boolean;
+  authors: boolean;
+  trees: boolean;
+  entropy: boolean;
+  gisaid_acknowledgments: boolean;
```
joverlee521: Thoughts on changing `gisaid_acknowledgments` to `acknowledgments`? We might want to extend this interface to allow users to define which node attrs to include in the acknowledgements download (e.g. change the hardcoded `gisaid_epi_isl` to `ppx_accession`)

src/util/metadataJsonParsing.ts
```diff
@@ -0,0 +1,57 @@
...
+    sharing[key] = value;
+  }
+
+  /* Hardcode overrides for GISAID datasets */
+  const gisaid = (metadataState?.dataProvenance || []).filter((el) => el.name.toUpperCase() === 'GISAID').length > 0;
+  if (gisaid) {
+    sharing.dataset_json = false;
+    sharing.metadata_tsv = false;
+    sharing.gisaid_acknowledgments = true;
+  }
```
joverlee521, line 29: Should this _always_ override the user's settings? E.g. I can see wanting to share downloads in a private Group dataset.

src/components/download/downloadButtons.js
joverlee521, line 1: Makes sense that each individual button depends on the associated `sharing` key. Should this have an easy override for auspice.us to allow downloads across the board or should auspice.us stay true to the dataset's settings?

joverlee521 (commented):
Nice! I didn't test, but this looks good from reading through the code. I left some small questions, but the main comment is naming (of course).

src/metadata.ts
jameshadfield: everything except `gisaid_acknowledgments`

src/metadata.ts
jameshadfield: My plan is for `gisaid_acknowledgments` to be an auspice-code option only, not a JSON-exposed one. If we implement some general acknowledgements download we can absolutely go in the direction you propose.

src/util/metadataJsonParsing.ts
jameshadfield: Yeah, I debated this, but I felt the potential for misuse and collateral damage was too great at the moment. We can relax it in the future :)

src/components/download/downloadButtons.js
jameshadfield: That's a good question. At the moment I think we enforce what the JSON specifies -- a little odd, but hopefully not terrible. (And, like everything here, easily side-stepped.)

src/metadata.ts
joverlee521: Gotcha, that makes sense. 

Small note, if `gisaid_acknowledgments` is an auspice-code option only, then the default should always be `true` to match existing behavior.

jameshadfield: > Small note, if gisaid_acknowledgments is an auspice-code option only, then the default should always be true to match existing behavior.

The rendering logic is a little different -- simpler! -- and now uses these flags to enable/disable the toggles. So this needs to be `false` to not expose this option for non-gisaid datasets

jameshadfield: I've updated things a little, so `gisaid_acknowledgments` is not present in defaults (it's an optional key in the type now), and gets set (to `true`) only when it's a gisaid dataset

joverlee521: > So this needs to be false to not expose this option for non-gisaid datasets

Oh hah, yeah I got turned around, that makes sense!

Issue #2000, opened 2025-07-22
Download dataset JSON(s) from redux state


jameshadfield:
Currently we expose the ability to download the underlying dataset JSON:

https://github.com/nextstrain/auspice/blob/b2965188e6cf137be95108c313d0006f02dbcf2f/src/components/download/helperFunctions.js#L627-L630

We enable this functionality if (a) it meets certain data-use conditions and (b) it's possible to make a HTTP request for the JSON:

https://github.com/nextstrain/auspice/blob/b2965188e6cf137be95108c313d0006f02dbcf2f/src/components/download/downloadButtons.js#L30-L32

specifically this prevents auspice.us from attempting to download JSONs, which isn't possible using the above approach as the data was drag-and-dropped. Additionally we can't update the dataset JSON so that it better reflects the current UI settings - e.g. changing the display defaults color-by to that currently used, or exporting additional metadata which has been dragged on. 

Instead we should create a JSON from the current redux-state and download that. This allows the functionality to work in auspice.us and allows for future work to better match the downloaded dataset to the current view settings.

### Implementation details
On data load we process the dataset JSON and store (parts of) it in various reducers, e.g. [here's us processing the top-level "meta" key](https://github.com/nextstrain/auspice/blob/master/src/actions/recomputeReduxState.js#L818-L897). The idea would be to allow such functions to work in reverse. This doesn't mean we'd create an identical dataset as "errors" (e.g. invalid default color-by) wouldn't be exported. 

### Bigger picture

We've discussed ways to edit various parts of datasets in-browser - for instance, changing the title, color schemes, re-rooting etc. The ability to export the dataset will be essential for this.

joverlee521:
_Prompted by [question re: MicrobeTrace](https://github.com/nextstrain/nextstrain.org/issues/1188#issuecomment-3120222815)_

As part of "Bigger picture", consider generating a dataset JSON for a subtree view of large trees.

jameshadfield:
Notes before I start with this. Work in progress on branch [james/download-json-from-redux-state](https://github.com/nextstrain/auspice/tree/james/download-json-from-redux-state)

#### Simplest use-case (MVP)

Take the redux-state objects and recreate the JSON. Ignore any filters and related state. This by itself will revel a lot of information about potential difficulties. Things that come to mind:
* `meta.extensions` (which Nextclade uses) isn't stored in redux state. What else isn't?!
* We create a dummy root node in the tree which needs to be pruned.
[...]

PR #2054, opened 2026-05-25
JSON.meta parsed in TS

### LINKED ISSUE #2053: Fixup Metadata types
TS adoption in Auspice is (necessarily) incomplete & piecemeal, however the types for the `metadata` redux state need some attention. This issue is specifically motivated by https://github.com/nextstrain/auspice/pull/2039#discussion_r3237935166

* The redux state is untyped (the other "big" reducers - tree & controls - are now in TS).
  * #1887 was an experiment to do this, but I don't think it's quite the right approach now. I like the approach of https://github.com/nextstrain/auspice/commit/448e97da1876261d57f118bffb6a1228a111bbf0 where the JSON type is `any` but the redux state is comprehensively typed. 
* There are some scattered types for parts of the state / actions:
  * `src/metadata.ts` types part of the state. We should move this to `src/reducers/metadata.types.ts` I think.
  * #2039 added [src/actions/updateMetadata/updateMetadata.types.ts](https://github.com/nextstrain/auspice/pull/2039/files#diff-28154e2e767bb464179fd17487ec9c640a34ed80fc1a8fadd827a011c5b462f8), which is most probably duplicating types and would be better to re-use existing types
  * #2046 may also be doing the same

jameshadfield:
I did this as part of the [james/download-json-from-redux-state](https://github.com/nextstrain/auspice/tree/james/download-json-from-redux-state) branch (which addresses #2000), but it's big enough to be a PR especially as other work such as #2046 may want to use it

~Partially closes #2053, but not completely as the redux store is still untyped (and so is the majority of code which consumes the metadata state!)~

Closes #2053


src/actions/updateMetadata/updateMetadata.ts
```diff
@@ -199,18 +199,29 @@ function _updateColoring(
...
+    throw new Error('Unreachable: merge path only handles categorical scales');
+  }
+
   const updatedScale: [string, string][] = Object.entries({
     // existing scale pairs, less any values which no longer exist on the tree
     ...Object.fromEntries(
-      (state.scale || []) // restrict to strings as we only consider categorical scales
+      (state.scale || [])
         .filter(([value,]) => typeof value === 'string' && stateCounts?.get(value) > 0)
     ),
```
victorlin, line 226: The `typeof` check is redundant now that `state.scale` is typed

```suggestion
      (state.scale || [])
        .filter(([value,]) => stateCounts?.get(value) > 0)
    ),
```

victorlin (commented):
+1 for this direction! I didn't test the functional changes but I'm guessing most should be reasonable with the added validation.

src/reducers/metadata.types.ts
```diff
@@ -5,41 +5,83 @@ import { ScaleType } from "./controls"
...
+    colorBy?: string;
+    distanceMeasure?: "num_date" | "div";
+    layout?: "rect" | "radial" | "unrooted" | "clock";
+    selectedBranchLabel?: string;
+    label?: string;
+    tipLabelKey?: string;
+    streamLabel?: string;
+    showTransmissionLines?: boolean;
+    language?: string;
+    sidebar?: "open" | "closed";
```
joverlee521: Similar to `PANEL_VALUES`/`Panel`, should `distanceMearsure`, `layout`, and `sidebar` options be pulled out as their own types to be referenced by consumers (e.g. [parseJsonMetaBlock](https://github.com/nextstrain/auspice/blob/38d2543ffabafcc1b6e5465f7c0b6cb37d291b5c/src/util/metadataJsonParsing.ts#L78-L87))?

src/reducers/metadata.ts
```diff
@@ -1,19 +1,26 @@
...
+    entropy: true,
+    screenshot: true,
+  },
   rootSequence: undefined,
   identicalGenomeMapAcrossBothTrees: false,
   rootSequenceSecondTree: undefined,
-  colorOptions // this can't be removed as the colorScale currently runs before it should
-}, action) => {
+  colorOptions, // this can't be removed as the colorScale currently runs before it should
+};
```
joverlee521, line 21: Should this `initialState` be exported to be used in `metadataJsonParsing`, i.e. use `initialState` instead of redefining the default state 

https://github.com/nextstrain/auspice/blob/db4372be9e3a5cc010d0572dbc9b8f75486da7a4/src/util/metadataJsonParsing.ts#L9-L17

joverlee521 (approved):
Left some minor comments, but overall looks good. Thanks for pushing through this!

src/reducers/metadata.ts
jameshadfield: Right -- `computeMetadataSharing` duplicates this. Not good.

Conceptually I'm not sure what a "complete / loaded" `Metadata` type should actually refer to here - is it the `parseJsonMetaBlock` return value? Is it the state the reducer guarantees? Is it a state the reducer guarantees conditional on `metadata.loaded = true`? I _think_ it should be the last, although I haven't implemented it that way.

src/reducers/metadata.types.ts
jameshadfield: Yes, probably!

src/reducers/metadata.ts
joverlee521: Good question! I agree that it should be a state the reducer guarantees conditional on `metadata.loaded = true`. I think that should represent whatever minimal state is required for the app to work when `metadata.loaded = true`.

src/reducers/metadata.types.ts
jameshadfield: Fixed up - thanks for the suggestion

src/reducers/metadata.ts
jameshadfield: I'm going to resolve this in a separate PR as I'll make some extra changes which need some more testing...

jameshadfield:
Merging. One step in CI is failing (playwright download stalled out), but everything else passes.

src/reducers/metadata.ts
jameshadfield: https://github.com/nextstrain/auspice/pull/2058

PR #2055, opened 2026-05-27
Download json from redux state

### LINKED ISSUE #2000: Download dataset JSON(s) from redux state

jameshadfield:
Downloads JSON from redux state (rather than the previous approach of re-fetching the data).

Additionally, many UI states are now exported in the JSON. This means you can change the color-by, tree layout, tree distance measure, selected branch & tip labels, stream tree label, map geo resolution, map transmission lines, panels, language and (in some cases) tree zoom and have that state reflected in the JSON.

Closes https://github.com/nextstrain/private/issues/201
Closes #2000

### DIFF
--- a/src/components/download/downloadButtons.js
+++ b/src/components/download/downloadButtons.js
@@ -8,7 +8,6 @@ import { materialButton } from "../../globalStyles";
 import * as helpers from "./helperFunctions";
 import { getFullAuthorInfoFromNode } from "../../util/treeMiscHelpers";
 import { getNumSelectedTips } from "../../util/treeVisibilityHelpers";
-import { getDatasetNamesFromUrl } from "../../actions/loadData";
 
 const RectangularTreeIcon = withTheme(icons.RectangularTree);
 const PanelsGridIcon = withTheme(icons.PanelsGrid);
@@ -30,13 +29,6 @@ export const DownloadButtons = ({ relevantPublications }) => {
   const partialData = selectedTipsCount !== totalTipCount;
   const filePrefix = getFilePrefix();
   const temporal = state.controls.distanceMeasure === "num_date";
-
-  /**
-   * Currently our ability to download JSONs relies on parsing the URL and re-fetching the
-   * JSON. Check that we can parse datasets from the URL before exposing the button here.
-   * This check will be removed as part of <https://github.com/nextstrain/auspice/issues/2000>
-   */
-  const datasetNames = getDatasetNamesFromUrl(window.location.pathname);
   const entropyBar = state.entropy?.selectedCds===nucleotide_gene ? "nucleotide" : "codon";
 
   return (
@@ -51,12 +43,12 @@ export const DownloadButtons = ({ relevantPublications }) => {
         {partialData ? `Currently ${selectedTipsCount}/${totalTipCount} tips are displayed and will be downloaded.` : `Currently the entire dataset (${totalTipCount} tips) will be downloaded.`}
       </div>
 
-      {state.metadata.sharing.dataset_json && datasetNames.some(Boolean) && (
+      {state.metadata.sharing.dataset_json && (
         <Button
           name="Auspice (Nextstrain) JSON"
-          description={`The main Auspice dataset JSON(s) for the current view`}
+          description={`The main Auspice dataset JSON with the current view as the display default.`}
           icon={<DatasetIcon width={iconWidth} selected />}
-          onClick={() => helpers.auspiceJSON(dispatch, datasetNames)}
+          onClick={() => helpers.auspiceJSON(dispatch, state, filePrefix)}
         />
       )}
       {state.metadata.sharing.trees && (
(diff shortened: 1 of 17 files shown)


src/util/constructDatasetJson.ts
```diff
@@ -0,0 +1,76 @@
...
+  const reduxState = getState();
+  const rootSequence = handleRootSequences(reduxState.metadata);
+  const json = {
+    version: "v2" as const,
+    meta: metadataStateToJson(reduxState),
+    tree: treeStateToJson(reduxState.tree),
+    ...(rootSequence && { root_sequence: rootSequence}),
+  };
+  return json
+}
```
joverlee521, line 35: _non-blocking_

For future debugging purposes, it'd be helpful to differentiate when a JSON is created by Auspice vs Augur and if created by Auspice, what version was used to create it. I think this would be a `generated_by` key similar to what gets added in [augur_write_json](https://github.com/nextstrain/augur/blob/024292af6daf1f8aae7e8d690bd2f944e1c7322d/augur/utils.py#L135). 

(Although, as I'm writing this, I'm realizing that `generated_by` is not included in the JSONs created by `augur export`...)

src/util/metadataJsonParsing.ts
```diff
@@ -249,3 +251,86 @@ function _validateLegendArray(
...
+}
+
+
+function _computeJsonSharing(sharing: Metadata['sharing']): Record<string, false> {
+  return Object.fromEntries(
+    Object.entries(sharing)
+      .filter(([key, _value]) => key!=='gisaid_acknowledgments')
+      .filter(([_key, value]) => value===false)
+  )
+}
```
joverlee521, line 325: Hmm, I'm wary of only including the `false` values. If we change Auspice's internal defaults in a future version, then the downloaded JSON's sharing will no longer work as expected.

src/components/download/downloadButtons.js
```diff
@@ -44,76 +42,77 @@ export const DownloadButtons = ({dispatch, t, tree, entropy, metadata, colorBy,
         <p/>
         {partialData ? `Currently ${selectedTipsCount}/${totalTipCount} tips are displayed and will be downloaded.` : `Currently the entire dataset (${totalTipCount} tips) will be downloaded.`}
       </div>
-      {metadata.sharing.dataset_json && datasetNames.some(Boolean) && (
+
+      {state.metadata.sharing.dataset_json && (
         <Button
           name="Auspice (Nextstrain) JSON"
           description={`The main Auspice dataset JSON(s) for the current view`}
```
joverlee521: Small suggestion to update the download description to match the new behavior.

```suggestion
          description={`The main Auspice dataset JSON with the current view as the display default.`}
```

joverlee521 (approved):
Work as expected in local testing. I'll rebase my editable UI work on this to connect everything up.

src/util/constructDatasetJson.ts
jameshadfield: Good suggestion -- added via `meta.extensions.generated_by`

src/util/metadataJsonParsing.ts
jameshadfield: I hear ya, but this concern is applicable to every dataset JSON out there! If there was a way to track the `true` values which originated from the original JSON then maybe we could do this, but my expected usage of `sharing` is that 99% of the uses will be to define `false` values (like ANDV)

jameshadfield:
Merging, as all tests passing except for CI on node 24 where it stalls out at `npx playwright install chromium` (which we've seen elsewhere too)

PR #2058, opened 2026-05-31
Improve metadata types

jameshadfield:
Follow ups to https://github.com/nextstrain/auspice/pull/2054

### DIFF
--- a/src/actions/updateMetadata/updateMetadata.types.ts
+++ b/src/actions/updateMetadata/updateMetadata.types.ts
@@ -1,6 +1,6 @@
 import { TreeState, NodeAttr } from "../../reducers/tree/types";
 import type { ScaleType, ControlsState } from "../../reducers/controls";
-import type { ColoringInfo } from "../../reducers/metadata.types";
+import type { Colorings, GeoResolutions } from "../../reducers/metadata.types";
 
 /**
  * Struct containing new metadata information for merging with the current redux state.
@@ -9,16 +9,7 @@ import type { ColoringInfo } from "../../reducers/metadata.types";
  */
 export interface NewMetadata {
   attributes?: Record<string, AttrDetails>;
-
-  geographic?: {
-    /** trait name */
-    key: string;
-
-    /** map of trait value (the deme name) to lat/long */
-    demes: {
-      [deme: string]: LatLong;
-    }
-    }[];
+  geographic?: GeoResolutions[];
 }
 
 export interface UpdateMetadataAction {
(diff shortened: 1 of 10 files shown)


src/reducers/metadata.ts
```diff
@@ -65,4 +69,12 @@ function getBuildUrlFromGetAvailableJson(availableData: AvailableDataset[] | und
   return false;
 }
 
+function assertLoaded(state: MetadataReduxState, actionType: string): state is (Metadata & { loaded: true }) {
```
joverlee521: Since `Metadata.loaded = true`, can this be simplified to

```suggestion
function assertLoaded(state: MetadataReduxState, actionType: string): state is Metadata {
```

src/actions/updateMetadata/updateMetadata.types.ts
```diff
@@ -29,20 +20,10 @@ export interface UpdateMetadataAction {
...
-    };
-    key: string;
-  }[];
-}
-
-
+type ActionMetadata = Partial<{
+  colorings: Colorings;
+  geoResolutions: GeoResolutions[];
+}>;
```
joverlee521, line 26: In this file, I'm seeing two patterns of directly using the custom types (e.g. `Colorings`) vs the state type (e.g. `TreeState['totalStateCounts']`). I think using the state type makes sense to tie these action types closely with the redux state:

```suggestion
type ActionMetadata = Partial<Pick<Metadata, "colorings" | "geoResolutions">>;
```

joverlee521 (approved):
Thanks for continuing to push on type improvements. I only left minor comments.

src/reducers/metadata.ts
jameshadfield: Ahhh yes -- it can. This is left over from the v1 I had where `loaded` wasn't part of `Metadata` (it becomes a lot simpler if it is)

src/actions/updateMetadata/updateMetadata.types.ts
joverlee521: Made this change in https://github.com/nextstrain/auspice/pull/2046 (https://github.com/nextstrain/auspice/pull/2046/commits/bd476d84ed5dbe61099cdae1bf861b1bab8c87e7)

src/reducers/metadata.ts
jameshadfield: [fixed up](https://github.com/nextstrain/auspice/compare/95950685db7c006893a59110a658ae8f068e4d66..7c32a64a789d467ffe838c456c0abc6340efd2a7)

PR #2060, opened 2026-06-01
Map: use vector tiles

### LINKED ISSUE #2057: Switch map tiles
_Originally brought up by Trevor on [Slack](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1779768455902969)_

Looking at [MapBox pricing](https://www.mapbox.com/pricing), we can cut costs in half by switching from the [Static Tiles API](https://docs.mapbox.com/api/maps/static-tiles) ($0.50 / 1,000) to the [Vector Tiles API](https://docs.mapbox.com/api/maps/vector-tiles/) ($0.25 / 1,000).

jameshadfield:
_This is a breaking change to Auspice, i.e. will need a v3 version release, so it's a good time to think about the best direction to take this_

## Proposed changes

We swap to mapbox vector tiles -- see added comments in `DEV_DOCS.md` & `api.rst` for the interaction with libraries and API calls, but the high level overview is:

* Switch to MapLibre GL JS as the rendering engine, which uses webworkers to render vector tiles, applying any custom styles client-side.
* Change our customisation interface to expect a MapLibre Style Spec JSON for vector tiles (and other things)
* Hardcode our (custom!) Mapbox styles JSON into source code, having rewritten `mapbox://` URIs into MapLibre-compatible HTTPS equivalents.

image

_This PR (LHS) vs current nextstrain.org (RHS)_

image

_Vector images (LHS) noticibly sharper. (The GPS coords are slightly shifted, but this was a dataset issue not an Auspice one.)_

## Customised tiles (Auspice extensions)

Added an example customization JSON which specifies an [openfreemap](https://openfreemap.org/) [MapLibre style document](https://maplibre.org/maplibre-style-spec/) using the [positron theme](https://github.com/openmaptiles/positron-gl-style)

image


jameshadfield:
I revisited this today in the context of [June 4th's discussion](https://docs.google.com/document/d/1aYP3nv5UhvQNi-Tr-4uh8Mbc7JyFJLNjBhzkkMW71p4/edit?tab=t.lu17x849feca#heading=h.3mduq76za5jb) and #827

I'm now using OpenFreeMap tiles for everything with our own custom MapLibre styles JSON to emulate our custom MapBox styles. Here's comparison screenshots between existing MapBox (LHS) and new OpenFreeMap (RHS) - the JSON provides a huge amount of customisation to how layers are painted so we can tweak things more, either in this PR or later on.

image

image

image

### Improvements in offline usage

Using these tiles allows us to pursue ideas from #827, as well as building on the service worker introduced in #2062:

1. Allow service worker to serve tiles which exist in the browser cache - this allows offline access for areas of the map you've recently explored
2. Active pre-fetching of tiles from low zoom levels, perhaps coupled with `navigator.storage.persist()`, to improve offline access
3. For auspice installs (e.g. `nextstrain view`) we could fetch map tiles from low zoom levels & bundle them with the app so that entirely offline usage works for basic use. You can get to zoom level 5 for a total of ~15MB, or zoom level 7 for a total of ~150Mb, which might be acceptable increase for our docker image. (The screenshot of the caribbean is level 4, for reference.)

### Self hosting tiles

We could self-host the PBF tiles, glyphs and sprites on our (cloudfronted) S3 buckets to be less reliant on OpenFreeMap if we wanted to. There would be no auspice code changes beyond changing the addresses in the style JSON.

PR #2062, opened 2026-06-04
Add option to enable service worker

victorlin:
Use Workbox's webpack plugin to install a simple service worker that enables offline use of the index page.

This is added primarily for auspice.us, but should work for any Auspice server where assets are served at the root path. Gated behind a build-time env var setting to prevent unintended service worker installation (e.g. on nextstrain.org).

## Review threads

- [x] [Add self-unregistering service worker?](https://github.com/nextstrain/auspice/pull/2062#discussion_r3495735819) yes
- [x] [Add docs?](https://github.com/nextstrain/auspice/pull/2062#discussion_r3495753531) yes
- [x] [How do updates work?](https://github.com/nextstrain/auspice/pull/2062#discussion_r3501298278) immediately across all open tabs
- [x] [Which paths to serve?](https://github.com/nextstrain/auspice/pull/2062#discussion_r3501302161) just root
- [x] [How to enable feature?](https://github.com/nextstrain/auspice/pull/2062#discussion_r3502354568) env var

## Checklist

- [x] [auspice.us review app](https://auspice-us-nextstrain-b-hkzrnq.herokuapp.com) with `AUSPICE_ENABLE_SERVICE_WORKER=true` set in Heroku installs a service worker and [works offline](https://github.com/nextstrain/auspice/pull/2062#issuecomment-4836134836)
- [x] [nextstrain.org review app](https://nextstrain-s-nextstrain-oa1hkv.herokuapp.com) does not install a service worker when viewing a dataset
- [x] [Docs preview](https://nextstrain--2062.org.readthedocs.build/projects/auspice/en/2062/server/overview.html#offline-use)
- [x] Checks pass
- [x] Rebase onto latest changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+* Added support for offline use through the `AUSPICE_ENABLE_SERVICE_WORKER` environment variable read at build-time. This is added primarily for auspice.us, but should work for any Auspice server where assets are served at the root path.
+  ([#2062](https://github.com/nextstrain/auspice/pull/2062))
+
 ## version 2.72.0 - 2026/06/30
 
 
--- a/src/index.js
+++ b/src/index.js
@@ -32,6 +32,16 @@ if (!window.NEXTSTRAIN) {window.NEXTSTRAIN = {};}
 /* google analytics */
 initialiseGoogleAnalyticsIfRequired();
 
+/* service worker */
+if (process.env.ENABLE_SERVICE_WORKER && 'serviceWorker' in navigator) {
+  window.addEventListener('load', () => {
+    navigator.serviceWorker.register('/service-worker.js')
+      // Check for updates on each network-connected page load.
+      .then((registration) => registration.update())
+      .catch((error) => console.warn("Failed to register service worker", error)); // eslint-disable-line no-console
+  });
+}
+
 i18n
   .use(initReactI18next)
   .init({
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -94,7 +94,8 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
   });
   const pluginHtml = new HtmlWebpackPlugin({
     filename: 'index.html',
-    template: './src/index.html'
+    template: './src/index.html',
+    scriptLoading: 'blocking'
   });
   const cleanWebpackPlugin = new CleanWebpackPlugin({
     cleanStaleWebpackAssets: true
@@ -123,23 +124,6 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
       // Serve the index page for offline navigation requests to the root path.
       navigateFallback: "/dist/index.html",
       navigateFallbackAllowlist: [/^\/(\?.*)?$/],
-
-      // Add the index page to precache, updating when the webpack build output changes.
-      // TODO: upgrade to html-webpack-plugin to v5 and remove this block. The
-      // newer version would emit index.html earlier and allow Workbox to
-      // discover it natively and precache with a content-based revision.
-      manifestTransforms: [
-        (manifestEntries, compilation) => {
-          return {
-            manifest: [
-              ...manifestEntries,
-              // NOTE: compilation.hash captures changes unrelated to this file,
-              // but it's better than no revision.
-              {url: "/dist/index.html", revision: compilation.hash}
-            ]
-          };
-        }
-      ],
 
       // A separate runtime would be imported from root, but Auspice serves assets from /dist/.
       inlineWorkboxRuntime: true,
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -62,6 +62,7 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
     }
     // console.log("extensionData", extensionData);
   }
+  const enableServiceWorker = !devMode && process.env.AUSPICE_ENABLE_SERVICE_WORKER === "true";
 
   /* plugins */
   /* inject strings into the client-accessible process.env */
@@ -69,6 +70,7 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
     "process.env": {
       NODE_ENV: devMode ? JSON.stringify("development") : JSON.stringify("production"),
       SKIP_REDUX_CHECKS: JSON.stringify(process.env.SKIP_REDUX_CHECKS),
+      ENABLE_SERVICE_WORKER: JSON.stringify(enableServiceWorker),
       EXTENSION_DATA: JSON.stringify(extensionData)
     }
   });
@@ -111,6 +113,46 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
     pluginHtml,
     cleanWebpackPlugin
   ];
+
+  if (enableServiceWorker) {
+    const WorkboxPlugin = require('workbox-webpack-plugin');
+    plugins.push(new WorkboxPlugin.GenerateSW({
+      // The asset name of the service worker file created by this plugin.
+      swDest: "service-worker.js",
+
+      // Serve the index page for offline navigation requests to the root path.
+      navigateFallback: "/dist/index.html",
+      navigateFallbackAllowlist: [/^\/(\?.*)?$/],
+
+      // Add the index page to precache, updating when the webpack build output changes.
+      // TODO: upgrade to html-webpack-plugin to v5 and remove this block. The
+      // newer version would emit index.html earlier and allow Workbox to
+      // discover it natively and precache with a content-based revision.
+      manifestTransforms: [
+        (manifestEntries, compilation) => {
+          return {
+            manifest: [
+              ...manifestEntries,
+              // NOTE: compilation.hash captures changes unrelated to this file,
+              // but it's better than no revision.
+              {url: "/dist/index.html", revision: compilation.hash}
+            ]
+          };
+        }
+      ],
+
+      // A separate runtime would be imported from root, but Auspice serves assets from /dist/.
+      inlineWorkboxRuntime: true,
+
+      // Control all clients immediately (no reload required).
+      clientsClaim: true,
+      skipWaiting: true,
+
+      // Rarely needed, but shouldn't hurt.
+      // <https://github.com/GoogleChrome/workbox/issues/1407#issuecomment-463687081>
+      cleanupOutdatedCaches: true,
+    }));
+  }
 
   if (analyzeBundle) {
     const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
--- a/DEV_DOCS.md
+++ b/DEV_DOCS.md
@@ -110,6 +110,38 @@ If embarking on this journey, consider using Playwright since it is already used
 Linting via eslint: `npm run lint`, typechecking via `npm type-check`
 
 
+## Offline use
+
+Auspice can register a [service worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) for use without a network connection.
+This is opt-in and primarily intended for [auspice.us](https://auspice.us); it is enabled by setting the environment variable `AUSPICE_ENABLE_SERVICE_WORKER=true` at build time.
+It is **disabled by default** and is **always disabled in dev mode** (a caching service worker would interfere with hot-reloading).
+
+When enabled, the service worker is registered at the site root with a scope of the whole origin.
+
+What it intercepts:
+
+* The build's static assets – `index.html` and referenced JS/CSS bundles.
+* [Navigation requests](https://web.dev/articles/handling-navigation-requests) to the root path `/`.
+
+What it does **not** intercept:
+
+* **Charon API requests**
+* **Map tiles**
+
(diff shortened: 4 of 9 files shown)


victorlin:
auspice.us review app works:

1. Open the review app to "install" the service worker
2. Disconnect from internet
3. Open the review app in a new tab/window
4. Drag/drop an Auspice JSON

victorlin:
Demo using [auspice.us](https://auspice.us) (0:00-0:25) vs. [review app](https://auspice-us-nextstrain-b-hkzrnq.herokuapp.com/) (0:25-0:53):

https://github.com/user-attachments/assets/06a85992-43b8-45f8-8e35-aa96352078be

webpack.config.js
```diff
@@ -111,6 +143,9 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
     pluginHtml,
     cleanWebpackPlugin
   ];
+  if (enableServiceWorker) {
+    plugins.push(pluginServiceWorker);
+  }
```
jameshadfield, line 139: If we enable it, then disable it, the service worker will remain on the client (forever?). This is how (via claude) we can remove it if needed:

For production (e.g. if auspice.us ever wanted to turn it off for real users) the same trap applies at scale: every user who ever loaded the SW keeps it until they manually clear it. The standard fix is to deploy a self-destroying service worker — a `service-worker.js` whose entire body is roughly:
```js
self.addEventListener('install', () => self.skipWaiting());
self.addEventListener('activate', async () => {
  await self.registration.unregister();
  await Promise.all((await caches.keys()).map(k => caches.delete(k)));
  const clients = await self.clients.matchAll();
  clients.forEach(c => c.navigate(c.url));   // reload now-uncontrolled pages
});
```
You'd serve that at the same `/service-worker.js` path (keeping the no-cache header so browsers actually fetch it). The next update cycle replaces the Workbox worker with this one, which unregisters itself and clears the precache on activation.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 This allows additional dropped-on metadata to be exported, as well as allowing the functionality to work in auspice.us.
 Additionally, the current state of the UI (such as the color-by, which panels are toggled on etc) are saved into the downloaded JSON where possible.
 ([#2055](https://github.com/nextstrain/auspice/pull/2055))
+* The Auspice server has a new build-time customisation, `enableServiceWorker`, to generate and register a service worker for offline use. This is added primarily for auspice.us, but should work for any Auspice server where assets are served at the root path.
```
jameshadfield: We should add to the dev docs to describe how to (a) test service workers locally and (b) remove them after we've finished testing. To remove it, i've used the first of these suggestions (from claude):

1. Manual, via DevTools (simplest):
- Application → Service Workers → Unregister (for the localhost:4000 origin).
- Application → Storage → Clear site data (this wipes Cache Storage, where the ~54 precached entries live, plus everything else for the origin).
- Close and reopen the tab. Unregistering stops it controlling future loads, but an already-open tab stays controlled until you close it.

2. Programmatic, paste into the console (handy if you're scripting or doing this repeatedly):
navigator.serviceWorker.getRegistrations().then(rs => rs.forEach(r => r.unregister()));
caches.keys().then(ks => ks.forEach(k => caches.delete(k)));
Then reload. Same effect as the DevTools steps.

jameshadfield (commented):
I reviewed this with claude as I'm not an expert on service workers. It seems to work well (localhost & [the auspice.us review app](https://auspice-us-nextstrain-b-hkzrnq.herokuapp.com/). I've read stories which make me wary of them due to the difficulty in undoing mistakes that have gone out, but I do see the advantage for auspice.us. Some general comments on the PR

1. We need more dev docs - especially around what requests will be intercepted (bundles defined in index.html and all navigation requests in the domain) and which won't (e.g. map tiles, charon), how to test them, how to remove them etc.
2.  If a user has an auspice.us tab open (with the service worker running), then refreshing the tab won't get an updated auspice.us version as far as I can tell. Something to be aware of, but probably not a deal breaker for auspice.us
3. Is there any anticipated use-case beyond auspice.us? The worker won't handle the charon API, so at best it's giving an auspice error page. Should we use  `navigateFallbackAllowlist: [/^\/(\?.*)?$/],` so that it only works for the root URL?

src/index.js
```diff
@@ -32,6 +32,16 @@ if (!window.NEXTSTRAIN) {window.NEXTSTRAIN = {};}
 /* google analytics */
 initialiseGoogleAnalyticsIfRequired();
 
+/* service worker */
+if (process.env.ENABLE_SERVICE_WORKER && 'serviceWorker' in navigator) {
+  window.addEventListener('load', () => {
+    navigator.serviceWorker.register('/service-worker.js')
+      // Check for updates on each network-connected page load.
+      .then((registration) => registration.update())
```
victorlin, line 40: > If a user has an auspice.us tab open (with the service worker running), then refreshing the tab won't get an updated auspice.us version as far as I can tell.

It should! This is controlled by the precache, which contains all JS bundles and assets. You can see the full list in the compiled service-worker.js file which looks like:

```js
function(e) {
    C().precache(e)
}([{
    url: "/dist/09172b1910dff418491f.woff2",
    revision: null
}, {
// ...
    url: "/dist/auspice.chunk.163.bundle.f2dc78be857ccb8ba92e.js",
    revision: null
}, {
// ...
    url: "/dist/index.html",
    revision: "4cb1bfc61bc2a453d2d5"
}]),
```

and is checked on each network-connected page load:

https://github.com/nextstrain/auspice/blob/5a56fc0d150c155a1122e7546e85336c23b4df21/src/index.js#L38-L40

webpack.config.js
```diff
@@ -94,6 +97,35 @@ const generateConfig = ({extensionPath, devMode=false, customOutputPath, analyze
     filename: 'index.html',
     template: './src/index.html'
   });
+  const pluginServiceWorker = new WorkboxPlugin.GenerateSW({
+    // The asset name of the service worker file created by this plugin.
+    swDest: "service-worker.js",
+
+    // Serve the index page for offline navigation requests.
+    navigateFallback: "/dist/index.html",
```
victorlin: > Is there any anticipated use-case beyond auspice.us? The worker won't handle the charon API, so at best it's giving an auspice error page. Should we use navigateFallbackAllowlist: [/^\/(\?.*)?$/], so that it only works for the root URL?

`navigateFallbackAllowlist` doesn't apply to charon API requests from the app because those are not considered navigation requests. If we add that, it will show the browser's "no internet" page for URLs like https://auspice.us/foo. I think better to fall back to the index page instead, but maybe that only makes sense for auspice.us?

src/index.js
jameshadfield: My understanding is that while `registration.update()` runs on a page refresh, if it updates the service worker (from A to B) then B becomes ready in a "waiting" state but the tab continues to use A.  I believe `index.html` will also stay version A, because service worker A just does `cache.match()` without any revalidation. I don't have enough experience with service workers to give concrete advice here. Maybe this is all just fine?

webpack.config.js
jameshadfield: P.S. That https://auspice.us/foo works right now isn't intentional, it's a side-effect of the SPA design used for local Auspice / nextstrain.org. Ideally those URLs should never be used, and if we put `auspice.us` into static hosting (which we should do one day!) then those URLs will cease to function.

victorlin: These are the options:

1. Serve a custom 404 page
2. Serve the root page
3. Generic "not found" page (or "no internet" when offline)

(1) is ideal but requires extra work as you mention. We're not getting there in this PR. Do you prefer 2 or 3?

jameshadfield: Don't know! (2) is more consistent with auspice.us as it works right now, my thinking for (3) is that it reduces the blast radius for things going wrong with service workers. I'll leave it with you, I don't have strong intuitions or experience here.

docs/customise-client/api.rst
```diff
@@ -31,6 +31,7 @@ The following are definable as top-level keys of the JSON file. A useful referen
 -  ``googleAnalyticsKey`` You can specify a Google Analytics key to enable (some) analytics functionality. This is deprecated and will be removed from an upcoming release.
 -  ``serverAddress`` Specify the address / prefix which the auspice client uses for API requests.
 -  ``mapTiles`` Specify the address (and other information) for the tiles used to render the map.
+-  ``enableServiceWorker`` Set to ``true`` to generate and register a service worker for offline use. Disabled by default.
```
victorlin: I'm changing this to an environment variable `AUSPICE_ENABLE_SERVICE_WORKER=true`.

I realized that putting the feature behind `enableServiceWorker` means it's all-or-nothing for auspice.us and its review apps. While it's nice to have the service worker on the current review app for testing, after merge it should be turned off for future review apps, otherwise each ephemeral review app would register its own service worker that lives beyond the review app's lifetime!

The environment variable can be easily added on the Heroku side for only the production auspice.us app.

webpack.config.js
victorlin: Added a modified version: [cli/server/cleanupServiceWorker.js](https://github.com/nextstrain/auspice/blob/56e3f031dfe44dc826e69a298b6d7791009a449a/cli/server/cleanupServiceWorker.js)

CHANGELOG.md
victorlin: Added an "offline use" section to:

- [Server docs](https://nextstrain--2062.org.readthedocs.build/projects/auspice/en/2062/server/overview.html#offline-use)
- [Dev docs](https://github.com/nextstrain/auspice/blob/56e3f031dfe44dc826e69a298b6d7791009a449a/DEV_DOCS.md#offline-use)

Service worker removal is much easier with your suggestion above.

src/index.js
victorlin: I've added `skipWaiting: true` so that B will immediately take control in all existing tabs.

https://github.com/nextstrain/auspice/blob/56e3f031dfe44dc826e69a298b6d7791009a449a/webpack.config.js#L131-L133

webpack.config.js
victorlin: That's fair, easier to start with 3 and switch to 2 later, than the other way around. Done:

https://github.com/nextstrain/auspice/blob/56e3f031dfe44dc826e69a298b6d7791009a449a/webpack.config.js#L124-L127

docs/customise-client/api.rst
victorlin: Done:

https://github.com/nextstrain/auspice/blob/56e3f031dfe44dc826e69a298b6d7791009a449a/webpack.config.js#L65

jameshadfield: I like this approach!

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 This allows additional dropped-on metadata to be exported, as well as allowing the functionality to work in auspice.us.
 Additionally, the current state of the UI (such as the color-by, which panels are toggled on etc) are saved into the downloaded JSON where possible.
 ([#2055](https://github.com/nextstrain/auspice/pull/2055))
+* Added support for offline use through the `AUSPICE_ENABLE_SERVICE_WORKER` environment variable read at build-time. This is added primarily for auspice.us, but should work for any Auspice server where assets are served at the root path.
```
joverlee521: Reminder to rebase/fixup changelog before merge since v2.72.0 was released.

joverlee521 (approved):
Will use my own time to understand service workers, but the user facing changes, both the offline behavior and the envvar to enable service workers, look good to me.

victorlin:
I'll hold off merging until at least after dev chat tomorrow.

Pushed some tests for the functionality: 8f58dd8a28040a68c80977188571554505e6ab44

src/index.js
jameshadfield: Nice. For completeness, this is what can go wrong with `skipWaiting: true` but I don't think it applies to us as requests will fallthrough to the network if not handled by the service worker

> If version-A HTML lazy-loads a hashed chunk after SW B has activated — say a dynamic import, or an image the A page requests on interaction — that request goes to SW B, which no longer has version A's assets in precache. If your precache-only strategy returns nothing for an unknown URL (and the network copy is also gone because you shipped B), that fetch fails. This is the classic "`skipWaiting` breaks in-flight pages" hazard, and it's specifically a problem for cache-first/precache setups with content-hashed assets.

victorlin:
Tests in 8f58dd8a28040a68c80977188571554505e6ab44 significantly changed CI run time: +1m of smoke testing and +2m of unit testing ([before](https://github.com/nextstrain/auspice/actions/runs/28483197265?pr=2062) and [after](https://github.com/nextstrain/auspice/actions/runs/28556957233?pr=2062)).

I've dropped that commit in the rebase and will open a separate PR to explore ideas for testing.

src/index.js
victorlin: Actually, this does apply to us:

> the network copy is also gone because you shipped B

though it's very specifically for when someone has an old tab open, opens a new tab, and tries to use the old tab again without refreshing*.

I guess this is why `clientsClaim` and `skipWaiting`  default to `false`. There is a flip side though, because the browser can only have one active service worker per origin: for the user who has an old tab open, they'll be stuck with the old version on the new tab too, until all tabs are closed to end the waiting phase.

In our case, I think it's better that new tabs use the latest version while breaking old tabs, than to have all tabs stuck on the old version. I'll keep both options set to `true`.

\* could be mitigated by auto-refresh on ChunkLoadError?

Issue #1558, opened 2022-10-04
Auspice testing data (get-data)


jameshadfield:
## Background

This repo doesn't contain any datasets beyond some [very minimal examples to explain the dataset format](https://github.com/nextstrain/auspice/tree/master/examples). Instead we rely on the [get-data script](https://github.com/nextstrain/auspice/blob/master/scripts/get-data.sh) which downloads a slew of (nextstrain core) datasets. In times gone by, this was an accurate listing of all of our core datasets (and other sources didn't exist -- groups, community etc). This script is often run manually (e.g. `npm run get-data`) so you can have some data to play with, and heroku runs this during setup (`npm run heroku-postbuild`) which results in usable datasets in review apps.

## Shortcomings

A lot of auspice's functionality cannot be tested with the data here and two PRs in the last couple of weeks have highlighted this: https://github.com/nextstrain/auspice/issues/1463 and https://github.com/nextstrain/auspice/pull/1552. This means the heroku-review apps are not useful, and people have to manually checkout the auspice branch and obtain an appropriate dataset for testing.

## Proposal

We should make the `get-data` script obtain a useful set of testing datasets, preferably using timestamped datasets so we can ensure reproducibility. For PRs which need additional datasets to test, these should be added to the get-data script as part of the PR.

victorlin:
Note that the PR review apps allow us to test on live nextstrain.org data

jameshadfield:
> Note that the PR review apps allow us to test on live nextstrain.org data

That's true (and a big positive) but it doesn't help with local development and, some time in the future, actual tests within auspice.

I'll take a crack at this issue today - it seems simple on the surface!

jameshadfield:
Resurrecting this issue, we now have:

* two (git-committed) datasets in `test/data`, both used for unit-tests. Ideally bugfixes would include a unit-test to prevent regressions, although I'm not sure how feasible this is.

* some very basic smoketests which currently depend on the `data/zika.json` dataset being provisioned.

* 8 "testing" narratives which are git-committed, but aren't leveraged for automated testing. They were added by me to help development of narratives but have largely fallen out of date because the reference datasets (in `data/`) which have been updated since the narrative was added.

* A `get-data.sh` script which downloads a mixture of date-stamped datasets and continuously updated datasets. 

I'm considering the following plan:

* Update all testing narratives (`narratives/`) to use nextstrain.org's `@YYYY-MM-DD` version ID or point to a specific git commit for community URLs. 
[...]

PR #2067, opened 2026-06-09
[v3] Default data changes

jameshadfield:
A set of commits which move away from provisioning data within the auspice repo / install directory.

There's a little polish / thinking to be done around how `npm run view` / `npm run dev` should work...

Heroku changes also needed, although not time-critical.


jameshadfield:
Closes #1664

Todo: Update issue #1558 after merging this PR

PR #2070, opened 2026-06-09
Add status code to splash page errors

### LINKED ISSUE #2068: Improve splash page errors
### Background

When Auspice fails to load a dataset, the error message is the same regardless of the type of error:

https://github.com/nextstrain/auspice/blob/9342561080e5f507f887b3a0ce2c1245ef57fb7c/src/components/splash/splash.js#L52-L67

This can be customized by the client (see [nextstrain.org](https://github.com/nextstrain/nextstrain.org/blob/71699b2edc5409c8c0d127cbb4ede346d04d5d21/auspice-client/customisations/splash.js#L93-L115)), but it's still limited by the props taken by the splash page React component ([src](https://github.com/nextstrain/auspice/blob/9342561080e5f507f887b3a0ce2c1245ef57fb7c/src/components/splash/index.js#L41-L49), [doc](https://docs.nextstrain.org/projects/auspice/en/stable/customise-client/api.html#splash-component)).

### Example

An ideal error message for https://github.com/nextstrain/nextstrain.org/issues/1347 would be something like:

> Nextstrain couldn’t fetch this dataset from genome.ucsc.edu. The dataset host returned HTTP 503, which may mean it is temporarily unavailable or blocking requests from Nextstrain’s server. Try again later or contact the host.

### Possible solution

The message above would need another prop that includes the error status code 503. Example: 9e052fdd36aba4e327f706bb51080bf13bdc56c0

victorlin:
Closes #2068

## Checklist

- [ ] Checks pass


jameshadfield (commented):
I think these are valuable improvements. 

There are additional places `FetchError`s are handled, such as `loadSidecars`, and it'd be nice to use the extra status-code information for them.

PR #2071, opened 2026-06-10
Remove Bootstrap v3.3.7 CSS dependency

jameshadfield:
The splash page styling looks fine for me testing locally.

Will create nextstrain.org & auspice.us review apps to double check they weren't (accidentally?) using any bootstrap CSS classes in their customisations, as the CSS was available site-wide.


src/components/splash/splash.js
```diff
@@ -96,7 +96,7 @@ const SplashContent = ({available, browserDimensions, errorMessage}) => {
   return (
     <>
       <NavBar sidebar={false}/>
-      <div className="static container">
+      <div className="static" style={{marginRight: "auto", marginLeft: "auto", paddingLeft: "15px", paddingRight: "15px", maxWidth: "1150px"}}>
```
joverlee521, line 99: This change reminds me of discussion in https://github.com/nextstrain/nextstrain.org/issues/963. Even though that was aimed at nextstrain.org, I think we can use the same guidelines in Auspice. 

Even if we don't use the bootstrap.css, I think something like the `container` class can be a global.css.

joverlee521 (commented):
Just FYI, I'm using the Bootstrap grid system for styling the form for editable datasets in https://github.com/nextstrain/auspice/pull/2046, but I can move those classes to the component specific styles.

src/components/splash/splash.js
jameshadfield: Let's chat about this when we next meet. The CSS is only ~2.7kb gzipped...

joverlee521:
Discussed in dev chat today and decided to just keep the bootstrap.css since it's a pretty small file and useful for generic styling.

PR #2073, opened 2026-06-11
Remove hardcoded footers

### LINKED ISSUE #980: Remove hardcoded footer descriptions
I believe most of the datasets now supply the footer description via the `meta.description` key, and thus we can clear out the hardcoded ones in `footer-descriptions.js`.

Nextstrain.org datasets to check:

- [ ] ebola
- [ ] zika
- [ ] mumps
- [ ] lassa
- [ ] WNV/NA
- [ ] h7n9 (takes precidence over `flu`)
- [ ] flu
- [ ] inrb-drc

jameshadfield:
Closes https://github.com/nextstrain/auspice/issues/980 (see that issue for more context)

~Blocked on https://github.com/inrb-drc/ebola-nord-kivu/pull/9~ _merged_

I re-added the (dataset-defined) description to the download modal which matches the intention of [the original implementation in 2017](https://github.com/nextstrain/auspice/commit/c283b256a000dfa1e0ba6c38732069b979b007e2) which had been unintentionally removed over the years.


src/components/download/downloadModal.js
```diff
@@ -111,6 +107,15 @@ class DownloadModalContents extends React.Component {
...
         </div>
+
+        {metadata.description && (
+          <>
+            <div style={infoPanelStyles.modalSubheading}>
+              {t("Dataset description")}:
+            </div>
+            <FooterDescription md={metadata.description} />
+          </>
+        )}
```
joverlee521, line 118: This is essentially hidden at the bottom of the modal, I had to scroll past everything to even see it. 
No harm having it here, but is it even worth surfacing? You have to scroll past the description to get the download button.

joverlee521 (approved):
Nice, thanks for checking all of the relevant datasets!

src/components/download/downloadModal.js
jameshadfield: It shows up on big screens so I'm going to leave it -- as you say, it's not a big deal if it goes unnoticed sometimes / most of the time.

Issue #1904, opened 2024-11-16
Tree svg scales incorrectly when window is resized (e.g. by opening dev console)


corneliusroemer:
On next.nextstrain.org (v2.61.1) I noticed that the tree panel scales incorrectly and ends up in a wrongly scaled state when resizing the window, e.g. by opening the dev console. See GIF for details:

![2024-11-16 17 14 38](https://github.com/user-attachments/assets/bfb006e9-9dde-427e-9674-cb6363680692)

jameshadfield:
Bisecting points to

```console
3a18d773adedafe6f41fce26fa7d814ceeec395f is the first bad commit
commit 3a18d773adedafe6f41fce26fa7d814ceeec395f
Author: james hadfield <hadfield.james@gmail.com>
Date:   Tue Oct 29 10:04:38 2024 +1300

    Skip (some) tree animations for big trees
```

jameshadfield:
Debugging one situation where this happens on a <4k tip tree:

* Resizing the window triggers a number of calls to [update the tree](https://github.com/nextstrain/auspice/blob/6e047d378bd6c4f7b7affa2500d41838ae8dc723/src/components/tree/phyloTree/change.ts#L287) in quick succession (aside: this is poor, we should debounce it and only trigger on the trailing edge).
* The `transitionTime` is set to `0` which means "no animation" due [to the frequency of resize actions](https://github.com/nextstrain/auspice/blob/6e047d378bd6c4f7b7affa2500d41838ae8dc723/src/components/tree/phyloTree/change.ts#L325-L330).
* We are correctly calculating the position of elements (wrt to the SVG size)
* The d3 update call, when console.logged indicates we're changing the position (attrs) of elements (e.g. tips) to their correct location. This call doesn't have a d3 transition set on it (see above) nor a timer flush (shouldn't be needed if there's no transition)
* The attr on the underlying element isn't being changed (e.g. when inspected the position hasn't updated on the DOM element)
* We can restore correct behaviour by adding back in a `.transition().duration(transitionTime);` + `timerFlush();`

This may be the same underlying bug as #1903

jameshadfield:
This situation (an update with a 500ms duration "interrupted" by another action) is occurring a lot - window resizes, date slider changes, mousing over legend swatches, even quick successive changes to color-by. I'm fairly confident now that it's behind #1903. 

[...]

PR #2077, opened 2026-07-01
Fix zoom freezing tips whose radius changed mid-transition

trvrb:
I discovered this bug randomly. To reproduce, go to nextstrain.org/ebola/ebov-2013, then without doing anything else, click on branch NODE_0001155 highlighted here:

auspice-bug-1

After the zoom transition you'll see that all the "Western Urban" tips haven't moved.

auspice-bug-2

Eventually we tracked this down to the fact that after clicking on the branch, the legend briefly opens, the cursor is over the legend swatch for "Western Urban" and this triggers a separate transition to resize Western Urban tip radii. This cancels the original transition and the Western Urban tips are left in place. The below fix seems to work and should catch other edge cases.

_Below via Claude_

## Summary

Zooming into a clade could leave a subset of tips frozen at their pre-zoom positions while the rest of the tree zoomed. Reproduced reliably on `/ebola`: hover a division in the legend (e.g. *Western Urban*), then click a branch to zoom — every tip of that division stays put.

## Root cause

A zoom triggers **two** `change()` calls in quick succession:

1. **The zoom** (`changeVisibility` + `zoom`) — a 500ms d3 transition animating every tip's `cx`/`cy` to its new position.
2. **A tip-radius update** — dispatched by the legend item's `onMouseEnter` (`updateTipRadii({selectedLegendItem})`, `src/components/tree/legend/item.js`). Landing <1s after the zoom, its `transitionTime` is `0`, so it started a **zero-duration transition** on exactly the tips whose radius changed.

In d3, starting a transition **interrupts** any running (default-named) transition on the same elements. So the zero-duration radius transition cancelled those tips' in-flight `cx`/`cy` animation, and the subsequent `timerFlush()` completed everyone else — leaving the hovered division's tips stranded.

This also explains the reported quirks: it only reproduced on *fresh load → zoom* (the hovered `selectedLegendItem` persisted; opening/closing the legend cleared it), and it was keyed on the legend value rather than tree structure. The node data and bindings were correct — `mapToScreen` computed the right new positions and `d.update` was set — the DOM writes were simply cancelled mid-flight.

It's a general issue: any tip-radius change (legend hover, or map deme hover) during any tree transition (zoom, filter, layout change) would freeze the matching tips.

## Fix

`genericSelectAndModify` now applies `transitionTime === 0` updates **directly** (`selection.call(updateCall)`) rather than through a zero-duration transition, so they update the attrs/styles immediately without interrupting other running transitions on those elements.

## Test plan

- [x] `/ebola` → zoom `NODE_0001155`: previously 144 tips (all *Western Urban*) stayed frozen; now **0 stuck**, all 1493 tips reposition.
- [x] No animation regression — verified identical visible-tip behavior against a stashed baseline build.
- [x] Date filtering still hides tips correctly (1493 → 195 with `dmax=2014-08-01`); no console errors.
- [x] `tsc` and `eslint` clean.
- [x] Manual check in a browser: hover a legend swatch, zoom a clade, confirm all tips move; confirm animation/date-slider/filter still smooth.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jameshadfield:
This is essentially a reversion of https://github.com/nextstrain/auspice/commit/5a66b778acbcfd5325910aa4a7e1425273467b8e, which is itself a partial reversion of https://github.com/nextstrain/auspice/commit/3a18d773adedafe6f41fce26fa7d814ceeec395f

So these changes fix the bug described here (confirmed for me too), they bring back bugs described in https://github.com/nextstrain/auspice/issues/1904 -- for instance see here how changing the browser dimensions doesn't trigger the tree to correctly reposition. I'll try and debug some more and see if we can get a better solution...

https://github.com/user-attachments/assets/ddb101a6-c547-40e8-b28c-d4fa79eb8be5

PR #2088, opened 2026-07-29
Transpile CLI TS code to JS

jameshadfield:
I recently converted (most of) the `cli/*` code to TS -- something I'd wanted to do
for a long time and which was now usable in a developer-friendly way with node 24+'s
type-stripping. However I didn't realise node won't strip types if Auspice is installed
in node_modules, which it almost always is for users! The fix is to transpile Auspice
at publish time so that JS files are used instead. See DEV_DOCS for full details.


jameshadfield:
I've tested this PR manually against our projects which consume them (both locally and via review apps from the PRs below). Each consuming project needed a few changes to work with Auspice v3, which is why our normal PRs-created-by-actions don't work (the PR is successfully created, but the repo doesn't build as the consumer project needs code changes). I'm going to merge this PR tomorrow given the successful testing.

* Nextstrain.org: https://github.com/nextstrain/nextstrain.org/pull/1365
* Auspice.us https://github.com/nextstrain/auspice.us/pull/161

PR #2089, opened 2026-07-30
Release Auspice via GitHub Actions

jameshadfield:
This approach follows other projects in the Nextstrain ecosystem. In addition to being more ergonomic and easier to use for the wider team, we also now have a straightforward way to make prereleases.

See additions to DEV_DOCS.md for instructions on how to use.

The action has a dry-run mode, but we can't test it until it's in master. I've reviewed the code (and made edits, although mainly to DEV_DOCS and comments) -- @victorlin and/or @joverlee521 would you mind taking a look and if you're happy we can merge and then run a dry-run action?

Note: CI is required to pass on the commit which we run the action from, however the action adds a subsequent commit to bump the version; this version-bump commit won't get its own CI run since pushes made with GITHUB_TOKEN don't trigger workflows. I think this is OK.
  
  
**How this relates to v3**

I'd like to:

1. Merge this PR
2. Rebase v3 branch onto master (to get the action YAML)
3. Release v3 branch as a prerelease (3.0.0.beta.0)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,275 @@
...
+        run: npm version "${{ needs.prepare-and-tag.outputs.version }}" --no-git-tag-version
+
+      # Don't use `--ignore-scripts` -- the "prepare" life cycle script builds what we publish
+      - run: npm ci --loglevel verbose
+
+      # --tag is always passed explicitly, because `npm publish` would otherwise point
+      # `latest` at a prerelease and serve it to every plain `npm install auspice`.
+      - run: npm publish --tag "$DIST_TAG" ${{ inputs.dry-run && '--dry-run' || '' }}
+        env:
+          DIST_TAG: ${{ needs.prepare-and-tag.outputs.is-prerelease == 'true' && 'next' || 'latest' }}
```
victorlin: The dist tag values `next`/`latest` are defined both here and in the markdown summary, prone to getting out of sync.

Suggestion: define it once in compute-release-version.js.

```suggestion
          DIST_TAG: ${{ needs.prepare-and-tag.outputs.dist-tag }}
```

A logical follow-up would be to remove the `is-prerelease` output, as prerelease is implied by a dist tag of `next`. Full changes, untested: d7b0721e

.github/workflows/release.yaml
```diff
@@ -0,0 +1,275 @@
...
+    outputs:
+      version: ${{ steps.compute.outputs.version }}
+      tag: ${{ steps.compute.outputs.tag }}
+      is-prerelease: ${{ steps.compute.outputs.is-prerelease }}
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          # Tags and origin/master are both needed below, by the existing-tag check and
+          # the drift warning respectively.
+          fetch-depth: 0
```
victorlin, line 48: I think `fetch-depth: 0` is fine here. Noting that `ref: master` + `fetch-tags: true` should work in theory, but there's a bug: https://github.com/actions/checkout/issues/1471

DEV_DOCS.md
```diff
@@ -228,20 +228,82 @@ If a translation of a particular string is not yet available, then auspice will
...
    If new entries are needed, commit these and push them to the remote (GitHub).
-1. Ensure you are on the master branch and up to date with the remote, have `node` and `npm` in your environment, and are logged into the GitHub CLI (`gh auth status`).
-1. Run the `./releaseNewVersion.sh` script.
-    This will prompt you for the version number increase and push changes to both the master and release branches.
-    * Auspice's [CI Action](https://github.com/nextstrain/auspice/actions/workflows/ci.yaml) will publish the new version to npm. 
-    It takes some time for the new version to appear in the npm registry.
+   Put them at the top of the file, directly below the `# Changelog` heading; the workflow will add the version heading above them.
+1. Go to the [release workflow](https://github.com/nextstrain/auspice/actions/workflows/release.yaml) and click "Run workflow". Then:
+    * Leave **"Use workflow from"** set to `master`. This selects the branch being released, and Auspice's normal releases always come from `master`.
+    * Set **bump** to `major` (e.g. 2.73.0 → 3.0.0), `feat` for a feature release (→ 2.74.0), or `minor` for a fix (→ 2.73.1).
```
victorlin: Describe `promote` too?

DEV_DOCS.md
```diff
@@ -228,20 +228,82 @@ If a translation of a particular string is not yet available, then auspice will
...
+    * It takes some time for the new version to appear in the npm registry.
     To check if it has been released run `npm view auspice versions --json | tail`; you may wish to wait for the email to the slack channel `#nextstrain-admin` but it's unclear if there is still a lag after the email arrives before it's available to `npm` APIs.
     None of the following numbered steps will work until the new version is available on npm. 
-    * The release script will automatically create an entry on [github.com/nextstrain/auspice/releases](https://github.com/nextstrain/auspice/releases/) containing information from the changelog.
     * Ensure the [docker-base CI action triggered by nextstrain-bot](https://github.com/nextstrain/docker-base/actions/workflows/ci.yml?query=branch%3Amaster+actor%3Anextstrain-bot) runs successfully.
-    (This action in nextstrain/docker-base is triggered by Auspice's CI action and installs Auspice from the release branch such that it's not dependant on availability in npm.)
+    (This action in nextstrain/docker-base is triggered by the release action and installs Auspice from the release branch such that it's not dependant on availability in npm.)
```
victorlin: Small wording nit: reflect the switch from `ci.yaml / release` (action) to `release.yaml` (workflow).

```suggestion
    (This action in nextstrain/docker-base is triggered by the release workflow and installs Auspice from the release branch such that it's not dependant on availability in npm.)
```

Similarly, docker-base's `ci.yml` is a workflow, not action.

victorlin (approved):
+1 for direction, just some very minor suggestions/comments.

.github/workflows/release.yaml
jameshadfield: Good background. Will revisit if we see failures!

DEV_DOCS.md
jameshadfield: Thanks - I don't think I appreciated the subtlety here

DEV_DOCS.md
jameshadfield: Oof, yeah that was a bad omission. Added.

.github/workflows/release.yaml
jameshadfield: I implemented this -- it simplified things nicely which I appreciate as there's a whole bunch of complexity involved here already.

jameshadfield:
Updated npm.js so the  "Trusted Publisher" is now `release.yaml` not `ci.yaml`

